PR #53, opened 2024-09-11
Ifu 20240613

ipanfilo:
IFU includes changes up to 905d94 (June13)


tests/pytorch/test_numerics.py
```diff
@@ -78,22 +85,48 @@ def get_causal_attn_mask(sq: int) -> torch.Tensor:
...
+    if dtype == torch.float32:
+        return dict(rtol=1.3e-6, atol=1e-5)
+    if dtype == torch.float16:
+        return dict(rtol=1e-3, atol=1e-5)
+    if dtype == torch.bfloat16:
+        return dict(rtol=1.6e-2, atol=1e-5)
+    raise ValueError(f"Unsuppored dtype ({dtype})")
+
+
+def rocm_attn_tols() -> Dict[str, float]:
```
wangye805, line 103: Based on the name rocm_attn_tols, we can skip the if IS_HIP_EXTENSION

tests/pytorch/test_numerics.py
```diff
@@ -78,22 +85,48 @@ def get_causal_attn_mask(sq: int) -> torch.Tensor:
...
+    if IS_HIP_EXTENSION:
+        use_fused_attn = int(os.getenv("NVTE_FUSED_ATTN", "1"))
+        use_fused_attn_ck = int(os.getenv("NVTE_FUSED_ATTN_CK", "1"))
+        # TODO: wait for the ck branch supporting determinism
+        if use_fused_attn and (not use_fused_attn_ck):
+            # AOTriton backend and non-fused attn are deterministic
+            # TODO(PIV) should clear tols?
+            pass
```
wangye805, line 111: Why we do not provide a tol value pair?

wangye805 (commented):
Do you have a complete unit test success/skip/failure list for all different configuration (pytorch (rocblas or hipblaslt) and jax) before and after the IFU? 

Overall LGTM

transformer_engine/common/common.h
```diff
@@ -19,13 +19,8 @@
 
 #include <cuda_runtime_api.h>
 #include <cuda_fp16.h>
-#ifndef __HIP_PLATFORM_AMD__
 #include <cuda_bf16.h>
 #include <cuda_fp8.h>
-#else
-#include <hip/hip_bfloat16.h>
-#include "amd_detail/hip_float8.h"
```
wenchenvincent, line 27: Why do we remove this?

ipanfilo: The same includes are result of hipifying so we don't need explicitly add them with #if

wangye805: hip/hip_bfloat16.h should be able to be generated by cuda_bf16.h, but I doubt "amd_detail/hip_float8.h" can be obtained from hipifying.

ipanfilo: cuda_fp8.h converted to it by hipify_custom_map.json rule

PR #60, opened 2024-09-23
Fixed wrong test methods arguments in some cases

ipanfilo:
When test with non default env variables some tests pass invalid arguments to underlying methods causing exception and test failure


wangye805 (commented):
So this is hipblaslt related but not CK/AOTriton related?

ipanfilo:
One fix is AOTriton related, the other is HIPBLASLT related, not actual GEMM lib but env

wangye805 (approved):
LGTM

PR #61, opened 2024-09-24
Update supported tests in README.rst

ipanfilo:
## Type of change

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -70,13 +70,15 @@ Note that some of operator unit tests fail in hipBLASLt config due to limited in
 Pytorch framework integration tests
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-All Pytorch integration pytests under tests/pytorch/ and tests/pytorch/fused_attn/ are supported 
+Pytorch integration pytests under tests/pytorch/ and tests/pytorch/fused_attn/ are supported 
 Except the following tests that are not supported in rocBLAS configuration
 
 * tests/pytorch/test_cuda_graph.py 
 * tests/pytorch/test_sanity.py::test_gpt_guda_graph
 
 Env `ROCBLAS_STREAM_ORDER_ALLOC=1` should be used when run tests in pytorch-rocblas configuration. 
+
+Also test_onnx_export.py does not support FP8 dues to absence of custom QDQ operatrs library
 
 Jax framework integration tests
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/README.rst
+++ b/README.rst
@@ -36,7 +36,7 @@ Execute the following commands to install ROCm Transformer Engine from source on
   git clone --recursive https://github.com/ROCmSoftwarePlatform/TransformerEngine-private.git
   
   cd TransformerEngine-private
-  export NVTE_FRAMEWORK=pytorch #optionally set framework, currently only support pytorch and jax
+  export NVTE_FRAMEWORK=pytorch,jax #optionally set framework, currently only support pytorch and jax; if not set will try to detect installed frameworks
   export PYTORCH_ROCM_ARCH=gfx942 # CK fused attn only support MI200 and MI300 and fp8 features are only supported on MI300
   pip install .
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -36,7 +36,7 @@ Execute the following commands to install ROCm Transformer Engine from source on
   git clone --recursive https://github.com/ROCmSoftwarePlatform/TransformerEngine-private.git
   
   cd TransformerEngine-private
-  export NVTE_FRAMEWORK=pytorch #optionally set framework, currently only support pytorch and jax
+  export NVTE_FRAMEWORK=pytorch,jax #optionally set framework, currently only support pytorch and jax; if not set will try to detect installed frameworks
```
wangye805, line 39: Have you tested whether it can detect installed frameworks? I never tested before

README.rst
```diff
@@ -65,45 +65,23 @@ After a successful Transformer Engine installation via `pip install`, execute th
...
-  tests/pytorch/test_float8tensor.py
-  tests/pytorch/test_jit.py
-  tests/pytorch/test_torch_save_load.py
-  tests/pytorch/test_fused_rope.py
-  tests/pytorch/test_deferred_init.py
-  tests/pytorch/fused_attn/test_fused_attn.py
-  tests/pytorch/fused_attn/test_fused_attn_with_cp.py
-
-Env `ROCBLAS_STREAM_ORDER_ALLOC=1` can be used when run tests in pytorch-rocblas configuration. 
-It can be dropped when the hipGraph feature is fully supported in Pytorch on AMDGPUs.
```
wangye805, line 89: Emm, do we support test_cuda_graph.py in hipblaslt configuration?

wangye805: BTW, how about test_convergence.py and the .so generation in test_onnx_export.py?

wangye805 (changes requested):

README.rst
ipanfilo: Yes, I did

README.rst
ipanfilo: We support test_cuda_graph with hipblaslt. 
About test_convergence you were right - it does nothing w/o data. QA section contains test script that downloads Megathron-LM and uses test_convergence but it does rely on existing data too. So I cannot confirm whether it works for us or not.
FP8 custom q/dq ops for ONNX - I did not find source of it so nothing to port to us. Making it from scratch requires better knowledge of ONNX

wangye805: Thanks for the confirmation.

I recall test_cuda_graph tested mha and dpa, which may call our rocm fused attn backends. But CK and AOTriton have not supported cuda graph yet. Do you mean it passes with NVTE_FUSED_ATTN=0?

We probably need to mention those tricky points in our README

README.rst
wangye805: Oh, great to know!

wangye805 (changes requested):
Let's mention those tricky scenarios (test_cuda_graph, test_onnx_export, and test_convergence)

README.rst
ipanfilo: Test_cuda_graph indeed tests dpa and uses fused attn_backend. It however works with both CK and AOTriton. So they rather partially support cuda_graph than don't support at all.

wangye805: Thanks for the confirmation. Please help confirm whether the dropout probability is 0 or not?

wangye805: Okay, I saw the dropout prob are all 0 in test_cuda_graph.py: https://github.com/ROCm/TransformerEngine/blob/285163a24a5472641cffc1ebb91a56b22fe9c3e2/tests/pytorch/test_cuda_graphs.py#L177-L190. Therefore, it's expected that this test can pass with our current rocm fused attn backends.

How about test_convergence.py and test_onnx_export.py? Let's mention the tricky points we discussed?

wangye805 (approved):
Overall LGTM. 
Minor issues: to mention tricky points in test_convergence.py and test_onnx_runtime.py

README.rst
ipanfilo: Confirmed from test code and by debugging attention that dropout is non-zero in test_numeric and 0 in test_cuda_graph. 
Also those tests use QKV memory layout.

But what about test_convergence? It is not in the list of supported tests.

wangye805: Oh I see, it's in tests/pytorch/distributed. Nevermind

PR #65, opened 2024-10-04
Add simultaneous support of hipBlasLt and rocBlas

ipanfilo:
# Description

Allow build TE with simultaneous support of hipBLASLt and rocBLAS GEMM backends. 
Fix building on docker w/o rocBLAS installed

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Changes

- Guard rocBlas and hipBlasLt codepaths with separate #ifdef rather than one #ifdef/#else
- Make both rocBlas and hipBlasLt built-in by default instead of rocBlas only
- Add new env variable NVTE_USE_ROCBLAS to build with rocBlas only support
- Make runtime rocBlas/hipBlasLt selection besed on env variables
- Make hipBlasLt a default backend if no evn variables is set


transformer_engine/common/CMakeLists.txt
```diff
@@ -17,6 +18,11 @@ else()
   set(USE_CUDA TRUE)
 endif()
 
+if (NOT USE_HIPBLASLT AND NOT USE_ROCBLAS)
```
wangye805: Put this under use_rocm true? Otherwise it will fail CUDA build

transformer_engine/common/gemm/rocm_gemm.hip
```diff
@@ -1261,5 +1263,54 @@ void cublas_gemm(const Tensor *inputA,
...
+                 int math_sm_count, int m_split, int n_split, bool gemm_producer,
+                 const Tensor *inputCounter, hipStream_t stream)
+{
+/*If no backend is specified with env variable use HIPBLASLT unless it is disabled
+  If HIPBLASLT backend is enabled and requested, use it despite ROCBLAS status
+  Otherwise use ROCBLAS 
+*/
+
+#if !defined(USE_HIPBLASLT) && !defined(USE_ROCBLAS)
+#error GEMM backend is not specified
```
wangye805, line 1295: NVTE_ERROR()?

README.rst
```diff
@@ -40,13 +40,14 @@ Execute the following commands to install ROCm Transformer Engine from source on
   export PYTORCH_ROCM_ARCH=gfx942 # CK fused attn only support MI200 and MI300 and fp8 features are only supported on MI300
   pip install .
 
-The default installation above will use rocblas in GEMM computation. The hipBlasLt alternative can be selected by setting the environment variable `NVTE_USE_HIPBLASLT` before the `pip install` as:
+The default installation above supports both rocBlas and hipBlasLt in GEMM computation. Building with single backend support can be done by setting `NVTE_USE_HIPBLASLT` or `NVTE_USE_ROCBLAS` environment variable before `pip install` as:
```
wangye805, line 43: Emm, so NVTE_USE_HIPBLASLT and NVTE_USE_ROCBLAS will be both used during compilation and application? What if our customer select different values for them during installation and application? Will we error out?

Or maybe we can introduce NVTE_INSTALL_HIPBLASLT and NVTE_INSTALL_ROCBLAS to control the compilation and NVTE_USE_XXX to control the application

wangye805 (changes requested):
Overall LGTM but just concerned that our customers may set different values for NVTE_USE_XXXBLAS during compilation and application

transformer_engine/common/gemm/rocm_gemm.hip
ipanfilo: NVTE_ERROR is runtime, here we catch compile time error.

README.rst
ipanfilo: By default both are compiled in and customer uses env runtime to specify which one to use.
If different values are specified during compilation and runtime we have two options: ignore runtime variables since only one GEMM library is available or check them and throw error if env does not match.
Current implementation uses the former approach, if you think the latter one is better for customers, it can be implemented as well

transformer_engine/common/gemm/rocm_gemm.hip
wangye805: Okay. I see. Thanks

README.rst
wangye805: Emm, I think your current way of ignoring runtime variables should be good. But can we put some warning messages to tell our customers under such scenarios?

wangye805 (approved):

PR #70, opened 2024-10-12
fix: rocm transformer engine install instructions

functionstackx:


wenchenvincent (approved):
Good catch. Thanks!

PR #80, opened 2024-10-17
Fix undesired nv_fusion disabling after reentrant_activation_recompute

ipanfilo:
# Description

Fix permanent disabling of bias_gelu_nv_fusion after using non-reentrant activation recompute.

Fixes # (issue)

https://github.com/ROCm/frameworks-internal/issues/9444

## Type of change

## Changes

- Re-enable reentrant mode in checkpoint routine after completing non-reentrant pass
- Do not disable bias_gelu_nv_fusion when non-reentrant activation recompute is in progress, it currently does not cause issues for ROCm 
- Fixed tensor compare in rocBLAS path of test_numerics.py::test_transformer_layer_hidden_states_format

### DIFF
--- a/tests/pytorch/test_numerics.py
+++ b/tests/pytorch/test_numerics.py
@@ -23,6 +23,7 @@ from transformer_engine.pytorch.utils import (
     is_bf16_compatible,
 )
 if IS_HIP_EXTENSION:
+    from functools import lru_cache
     from transformer_engine.pytorch.utils import is_mi200
 
 from transformer_engine.pytorch import (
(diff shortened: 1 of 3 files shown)


tests/pytorch/test_numerics.py
```diff
@@ -53,6 +54,11 @@
 _cpu_rng_state = torch.get_rng_state()
 _cuda_rng_state = torch.cuda.get_rng_state()
 
+if IS_HIP_EXTENSION:
+    @lru_cache(maxsize=1)
+    def use_hipblaslt() -> bool:
+        return (os.getenv("NVTE_USE_HIPBLASLT") is not None
```
wenchenvincent, line 60: I think `NVTE_USE_HIPBLASLT` and `NVTE_USE_ROCBLAS` could be either set to 1 or 0. So it seems not sufficient to just check if they are None or not.

wenchenvincent: Let's clean up that piece of logic  in another PR as it will involve a lot of changes.

wenchenvincent (approved):
LGTM.

PR #84, opened 2024-10-24
Match ROCm platform detection in CMake and setup.py. Issue #9690

ipanfilo:
# Description

Update TE package setup.py ROCm platform detection to match logic from TE core CMake

Fixes # (issue)

#9690

### DIFF
--- a/build_tools/utils.py
+++ b/build_tools/utils.py
@@ -166,16 +166,16 @@ def found_pybind11() -> bool:
 @functools.lru_cache(maxsize=None)
 def rocm_build() -> bool:
     """ ROCm build should be performed if:
-    - It is configured with NVTE_USE_ROCM env
+    - It is configured with NVTE_USE_ROCM=1 env
       OR:
-    - HIP compiler is found and CUDA on is not
+    - HIP compiler is found and CUDA one is not
     """
-    if os.getenv("NVTE_USE_ROCM"):
+    if bool(int(os.getenv("NVTE_USE_ROCM", "0"))):
         return True
 
     try:
         cuda_path()
-        return True
+        return False
     except FileNotFoundError:
         pass
 
--- a/build_tools/utils.py
+++ b/build_tools/utils.py
@@ -164,6 +164,26 @@ def found_pybind11() -> bool:
 
 
 @functools.lru_cache(maxsize=None)
+def rocm_build() -> bool:
+    """ ROCm build should be performed if:
+    - It is configured with NVTE_USE_ROCM env
+      OR:
+    - HIP compiler is found and CUDA on is not
+    """
+    if os.getenv("NVTE_USE_ROCM"):
+        return True
+
+    try:
+        cuda_path()
+        return True
+    except FileNotFoundError:
+        pass
+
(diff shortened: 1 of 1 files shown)


build_tools/utils.py
```diff
@@ -163,6 +163,26 @@ def found_pybind11() -> bool:
...
+    - It is configured with NVTE_USE_ROCM env
+      OR:
+    - HIP compiler is found and CUDA on is not
+    """
+    if os.getenv("NVTE_USE_ROCM"):
+        return True
+
+    try:
+        cuda_path()
+        return True
```
wangye805: If cuda_path() is found, should we return False?

build_tools/utils.py
```diff
@@ -238,6 +258,9 @@ def get_frameworks() -> List[str]:
             _frameworks.extend(arg.replace("--framework=", "").split(","))
             sys.argv.remove(arg)
 
+    if rocm_build():
+        _requested_frameworks = [framework.lower() for framework in _frameworks]
```
wangye805, line 262: Duplicated with line 292

build_tools/utils.py
```diff
@@ -271,6 +294,31 @@ def get_frameworks() -> List[str]:
         if framework not in supported_frameworks:
             raise ValueError(f"Transformer Engine does not support framework={framework}")
 
+    if rocm_build():
+        _unsupported_frameworks = []
+        if "pytorch" in _frameworks:
+            try:
+                from torch.utils.cpp_extension import IS_HIP_EXTENSION
```
wangye805, line 301: During docker image building, will this "from torch.utils.cpp_extension import" works

wangye805 (changes requested):
Thanks for working on this Ilya! 

I left some comments. 

In addition, there are some rocm/cuda determination needed in transformer_engine/common dir, for example, https://github.com/ROCm/TransformerEngine/blob/3ce0f87acd7442b1a9059e236139d016b189643f/transformer_engine/common/__init__.py#L105

Can you help address them as well? Or we can leave it for the next PR

build_tools/utils.py
ipanfilo: Yes, I mixed up cuda)is_found with use rocm

build_tools/utils.py
ipanfilo: It is similar but assigns different variable. Indeed, line 292 can be moved here but I try to minimize changes from upstream

build_tools/utils.py
ipanfilo: It was used in original ROCm detection path, so yes

ipanfilo:
> https://github.com/ROCm/TransformerEngine/blob/3ce0f87acd7442b1a9059e236139d016b189643f/transformer_engine/common/__init__.py#L105
> 
> Can you help address them as well? Or we can leave it for the next PR

Since this is runtime, not build time. We cannot rely on hipcc for ROCm detection. Should we use other logic like env variable but if it is not set try/except to load  CUDA or ROCm libraries ?

wenchenvincent (approved):

PR #89, opened 2024-11-04
add new branch heyi_cast_transpose and update cast_transpose optimizations

eliotwang:
# Description

Optimize the cast_transpose kernel by:
1. refactoring the kernel to support more flexible parameter tuning methods; 
2. use assembly instructions to optimize the conversion for fp8 output types; 
3. apply performance analysis on different input shapes and provide empirical parameter configuration methods, ensuring that the HIP kernel's performance exceeds Triton's performance in all tests;

# Steps to run:
docker pull rocm/pytorch:latest

docker run -it --network=host -v /home/yigex/heyi:/workspace --device=/dev/kfd --device=/dev/dri --group-add video --security-opt seccomp=unconfined --ipc=host --name heyi_te_wx rocm/pytorch:latest

mkdir heyi && cd heyi

git clone -b heyi_cast_transpose --recursive https://github.com/eliotwang/TransformerEngine.git

cd TransformerEngine

export NVTE_FRAMEWORK=pytorch
export NVTE_ROCM_ARCH=gfx942

pip install .

mkdir tests/cpp/build && cd tests/cpp/build

cmake .. && make

./operator/test_operator

[ RUN      ] OperatorTest/CTTestSuite.TestCastTranspose/bfloat16Xfloat8e5m2X2048X12288
GPU execution time: 54.7456 us
[       OK ] OperatorTest/CTTestSuite.TestCastTranspose/bfloat16Xfloat8e5m2X2048X12288 (2854 ms)
[ RUN      ] OperatorTest/CTTestSuite.TestCastTranspose/bfloat16Xfloat8e5m2X256X65536
GPU execution time: 39.4748 us
[       OK ] OperatorTest/CTTestSuite.TestCastTranspose/bfloat16Xfloat8e5m2X256X65536 (1690 ms)

# Performance:
image


tests/cpp/operator/test_cast_transpose.cu
```diff
@@ -40,29 +41,53 @@ void compute_ref(const InputType *data, OutputType *output_c, OutputType *output
...
   std::unique_ptr<OutputType[]> ref_output_c = std::make_unique<OutputType[]>(N * H);
   std::unique_ptr<OutputType[]> ref_output_t = std::make_unique<OutputType[]>(N * H);
 
   fillUniform(&input);
   setRandomScale(&output_c);
   output_t.shareFP8Meta(output_c);
-
-  nvte_cast_transpose(input.data(), output_c.data(), output_t.data(), 0);
+ 
+  int warm_iter = 3;
```
wenchenvincent: I suggest that we keep this test file as is. This file was just for unit testing. We can measure performance when running it with rocprof.

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -271,62 +271,120 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *cast_output
...
-            add_config(1, 1);
-            const auto &kernel_config =
-                *std::min_element(kernel_configs.begin(), kernel_configs.end());
-            NVTE_CHECK(kernel_config.valid, "invalid kernel config");
-            const size_t load_size = kernel_config.load_size;
-            const size_t store_size = kernel_config.store_size;
-            const size_t num_blocks = kernel_config.num_blocks;
-
+            size_t load_size;
+            size_t store_size;
```
wenchenvincent: We would like to keep the original NVTE CUDA code while adding the new code for ROCm. You can use `__HIP_PLATFORM_AMD__` to guard the code. Here is an example: https://github.com/ROCm/TransformerEngine/blob/dev/transformer_engine/common/fused_softmax/scaled_masked_softmax.cu#L72-L80

transformer_engine/common/transpose/rtc/cast_transpose.cu
```diff
@@ -17,10 +17,44 @@ using OType = __OTYPE__;
 constexpr size_t load_size = __LOAD_SIZE__;
 constexpr size_t store_size = __STORE_SIZE__;
 constexpr size_t warps_per_tile = __WARPS_PER_TILE__;
+constexpr size_t iter_size = __ITER_SIZE__;
 constexpr size_t block_size = __BLOCK_SIZE__;
 
 }  // namespace
 
+__device__ OType convert_from_fp32(float v) {
```
wenchenvincent: What's the key difference between this function and https://github.com/ROCm/TransformerEngine/blob/dev/transformer_engine/common/amd_detail/hip_float8.h?

eliotwang: This function appears to be more complete. When compiling with the __gfx940__ macro enabled, it results in an error related to function overloading

eliotwang:
1. Revert the tests/cpp/operator/test_cast_transpose.cu to its original version;
2. Add __HIP_PLATFORM_AMD__ to guard modified code in transformer_engine/common/transpose/cast_transpose.cu;
3. Modify hip_f8 funcs in hip_float8.h to ensure the code runs correctly on MI308X;
4. Adopt hip_f8 funcs in hip_float8.h to do precision conversion and remove custom conversion functions;

eliotwang:
fix cast_transpose issue where the width and height of the tensor cannot be divided by the tile size. pass pytest including test_float8tensor.py，test_numerics.py，fused_attn/test_fused_attn.py and test_sanity.py

transformer_engine/common/transpose/cast_transpose.cu
wangye805: +1

Line 280 to 286 should also be guarded by #ifdef __HIP_PLATFORM_AMD__. 

Basically, after removing the rocm/amd specific ifdefs, we would like to see our repo are exactly the same as the upstream NVTE

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -276,56 +277,115 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *cast_output
...
+            size_t wpt_size = warps_per_tile;
+            size_t iter_size = THREADS_PER_WARP / wpt_size;
+            int num_blocks;
+            size_t rtc_block_size;
+            bool do_general_config = true;
+
+#ifdef __HIP_PLATFORM_AMD__            
+            if((std::is_same<OutputType, fp8e5m2>::value) || (std::is_same<OutputType, fp8e4m3>::value)){
```
wangye805: can use if constexpr since OutputType is determined at compile time

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -276,56 +277,115 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *cast_output
...
+               // Estimate number of SMs
+              // Note: H100 has 132 SMs, A100 has 108 SMs.
+              // Note: Directly querying number of SMs with cudaGetDeviceProperties is
+              // slow (>1 ms). Consider querying once and caching.
+              const int n_sms = 128;
+              // Helper functions to get kernel configuration
+              auto get_n_tiles = [=] (size_t load_size, size_t store_size) -> int {
+                constexpr size_t threads_per_warp = static_cast<size_t>(THREADS_PER_WARP);
+                size_t nvec_in = load_size / sizeof(InputType);
+                size_t nvec_out = store_size / sizeof(OutputType);
```
wangye805: Actually the sizeof(OutputType) is already fixed under this if condition

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -276,56 +277,115 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *cast_output
...
+                size_t n_tiles = DIVUP(row_length, nvec_in * threads_per_warp) *
+                                DIVUP(num_rows, nvec_out * threads_per_warp);
+              return n_tiles;
```
wangye805: Should be the same indent as line 301

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -276,56 +277,115 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *cast_output
...
+              };
+              auto get_n_blocks = [=] (size_t n_tiles, size_t cast_transpose_num_threads, size_t wpt_size) -> int {
+                size_t n_warps_per_block = cast_transpose_num_threads / THREADS_PER_WARP;
+                size_t n_blocks = DIVUP(n_tiles * wpt_size, n_warps_per_block);
+              return n_blocks;
```
wangye805: indent issue

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -276,56 +277,115 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *cast_output
...
+                load_size = 4,
+                store_size = std::is_same<InputType, float>::value ? 4 : 2;
+              }
+              
+              const size_t row_tile_elements = load_size * THREADS_PER_WARP / itype_size;
+              const size_t col_tile_elements = store_size * iter_size * wpt_size / otype_size;
+              // Number of CUDA blocks
+              num_blocks = (row_length / row_tile_elements) * (num_rows / col_tile_elements);
+              rtc_block_size = THREADS_PER_WARP * wpt_size;  
+              do_general_config =!(row_length % row_tile_elements == 0 && num_rows % col_tile_elements == 0);
```
wangye805, line 331: do_general_config is just not aligned()?

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -276,56 +277,115 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *cast_output
...
-            const size_t store_size = kernel_config.store_size;
-            const size_t num_blocks = kernel_config.num_blocks;
+              std::vector<KernelConfig> kernel_configs;
+              kernel_configs.reserve(16);
+              const size_t sm_count = static_cast<size_t>(cuda::sm_count());
+              auto add_config = [&](size_t load_size, size_t store_size) {
+                kernel_configs.emplace_back(row_length, num_rows, itype_size, otype_size, load_size,
+                                            store_size, sm_count);
+              };
+              add_config(8, 8);
```
wangye805: Regarding the big picture, in my opinion, you are trying to introduce a wpt_size and a iter_size so that wpt_size * iter_size <= THREADS_PER_WARP, in contrast to warps_per_tile x num_interations = THREADS_PER_WARP. So that each thread can save some registers and overall shared mems. Is my understanding correct? 

If so, I would expect add_config() will add wpt_size and iter_size into the cost model. But here we don't see additional parameters for add_config. Why?

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -276,56 +277,115 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *cast_output
...
+              return n_tiles;
+              };
+              auto get_n_blocks = [=] (size_t n_tiles, size_t cast_transpose_num_threads, size_t wpt_size) -> int {
+                size_t n_warps_per_block = cast_transpose_num_threads / THREADS_PER_WARP;
+                size_t n_blocks = DIVUP(n_tiles * wpt_size, n_warps_per_block);
+              return n_blocks;
+              }; 
+
+              wpt_size = 8;
+              iter_size = THREADS_PER_WARP / wpt_size;
```
wangye805: if iter_size x wpt_size = THREADS_PER_WARP here, why do we need to add ITER_SIZE into the subsequent RTC params?

wangye805 (changes requested):
Sorry for the late review. 

I was only able to review the cast_transpose part, not the cast_transpose_fusion part. I gave some comments on the coding style. But I'm still quite confused about the big picture as how wpt_size, iter_size works.

transformer_engine/common/transpose/cast_transpose.cu
zora-my: This is an added parameter for optimization, aimed at testing the impact of different wpt_size and iter_size configurations on performance.

transformer_engine/common/transpose/cast_transpose.cu
zora-my: Under the AMD framework, if the output type is FP8 and the current tile_size meets certain conditions, the optimized configuration will be used; otherwise, it will fall under the do_general_config case.

transformer_engine/common/transpose/cast_transpose.cu
zora-my: Yes, when wpt_size * iter_size <= THREADS_PER_WARP, it can save some registers and overall shared memory, thus improving the occupancy of the cu. How can we reflect the impact of wpt_size and iter_size on performance improvement in the cost model?

wangye805: Emm, we probably first need to understand how latency is affected by wpt_size*iter_size before changing the cost model. 

By the way, from https://github.com/eliotwang/TransformerEngine/blob/b9359d65666bbcbd6734376e180c8fbe513c50ee/transformer_engine/common/transpose/cast_transpose.cu#L312, it seems that wpt_size*iter_size is still tied to THREADS_PER_WARP, which is 32?

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -275,66 +275,149 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *cast_output
...
           const bool aligned =
               (row_length % THREADS_PER_WARP == 0 && num_rows % THREADS_PER_WARP == 0);
-          if (aligned && rtc::is_enabled()) {  // Runtime-compiled tuned kernel
+          if(aligned&&rtc::is_enabled()){
+            bool do_general_config = true;
+#ifdef __HIP_PLATFORM_AMD__            
+            const char* env_var = std::getenv("HEYI_CAST_TRANSPOSE");
+            if(((std::is_same<OutputType, fp8e5m2>::value) || (std::is_same<OutputType, fp8e4m3>::value))
+              &&(env_var != nullptr && env_var[0] == '1')){
+              do_general_config = false;
```
wangye805: This do_general_config assignment will be overwritten by line 329. Let's remove it?

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -275,66 +275,149 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *cast_output
...
+                size_t n_warps_per_block = cast_transpose_num_threads / THREADS_PER_WARP;
+                size_t n_blocks = DIVUP(n_tiles * wpt_size, n_warps_per_block);
+              return n_blocks;
+              }; 
+
+              size_t  wpt_size = 8;
+              size_t iter_size = THREADS_PER_WARP / wpt_size;
+              const size_t estimated_n_tiles = get_n_tiles(16, 8);
+              unsigned int cast_transpose_num_threads = wpt_size * THREADS_PER_WARP;
+              const size_t estimated_n_blocks = get_n_blocks(estimated_n_tiles, cast_transpose_num_threads, wpt_size);
```
wangye805: After digesting these three lines, I found that n_blocks is just n_tiles since 
n_blocks = n_tiles * wpt_size/ (wpt_size*threads_per_warp/threads_per_warp) = n_tiles

So can we help clean those up?

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -275,66 +275,149 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *cast_output
...
+                                reinterpret_cast<const CType *>(noop.data.dptr),
+                                static_cast<OutputType *>(cast_output.data.dptr),
+                                static_cast<OutputType *>(transposed_output.data.dptr),
+                                static_cast<const CType *>(cast_output.scale.dptr),
+                                static_cast<CType *>(cast_output.amax.dptr),
+                                static_cast<CType *>(cast_output.scale_inv.dptr), row_length,
+                                num_rows);
+            }          
+#endif         
+            if(do_general_config){
```
wangye805, line 365: No matter what do_general_config is assigned to in line 329, we will launch a rtc kernel for this case. What if do_general_config is true? Then it will launch another general tuning kernel. Does that mean we will run two kernels for the one case?

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -275,66 +275,149 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *cast_output
...
+                code = regex_replace(code, "__OTYPE__", otype_name);
+                code = regex_replace(code, "__LOAD_SIZE__", load_size);
+                code = regex_replace(code, "__STORE_SIZE__", store_size);
+                code = regex_replace(code, "__WARPS_PER_TILE__", wpt_size);
+                //code = regex_replace(code, "__ITER_SIZE__", iter_size);
+                code = regex_replace(code, "__BLOCK_SIZE__", rtc_block_size);
+                rtc_manager.compile(kernel_label, "cast_transpose_optimized_kernel", code,
+                                    "transformer_engine/common/transpose/rtc/cast_transpose.cu");
+              }
+              rtc_manager.launch(kernel_label, num_blocks, rtc_block_size, 0, stream,
```
wangye805: This fails TE pytorch pytests  tests/pytorch/test_numerics.py::test_grouped_linear_accuracy[True-True-126m-1-6-dtype0]

To reproduce, we can run the following command: 
HEYI_CAST_TRANSPOSE=1 pytest -s -v tests/pytorch/test_numerics.py::test_grouped_linear_accuracy[True-True-126m-1-6-dtype0] 

I dived a little into this failure, it seems that for the case that row_length: 768, num_rows: 96,  this rtc manager will try to launch a kernel with num_blocks = 0

transformer_engine/common/transpose/cast_transpose.cu
zora-my: Yes, we have cleaned it up in the new PR.

transformer_engine/common/transpose/cast_transpose.cu
zora-my: Thank you for your feedback. The current logic here is incorrect. The correct behavior should be: only when do_general_config=false will the tuned kernel be invoked; otherwise, the original kernel will be executed.

transformer_engine/common/transpose/cast_transpose.cu
zora-my: This is due to the logical error with do_general_config earlier. When the shape is [768, 96], it should call the original RTC kernel, but it is currently calling the tuning kernel, which causes the execution to fail.

wangye805 (approved):
LGTM

PR #94, opened 2024-11-20
Fix to use HW conversion instructions for fp8 casting on gfx941 and gfx942

wenchenvincent:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

### DIFF
--- a/transformer_engine/common/amd_detail/hip_f8_impl.h
+++ b/transformer_engine/common/amd_detail/hip_f8_impl.h
@@ -126,8 +126,14 @@ uint8_t cast_to_f8(T _x, bool stoch, uint32_t rng) {
      but after shift right by 4 bits, it would look like midpoint.
   */
 
-  if (exponent_diff>0)
-    mantissa >>= exponent_diff;
+  if (exponent_diff>0) {
+    //mantissa >>= exponent_diff;
+    // Use a loop to work around the compiler bug that produced the wrong result for the right shift
+    // TODO: Change back after the compiler bug is resolved.
+    int shift_amount = exponent_diff > 32 ? 32 : exponent_diff;
+    for (int i=0;i<shift_amount;++i)
+      mantissa >>= 1;
+  }
   else if (exponent_diff == -1)
     mantissa <<= -exponent_diff;
   bool implicit_one = mantissa & (1 << mfmt); 
--- a/transformer_engine/common/amd_detail/hip_f8_impl.h
+++ b/transformer_engine/common/amd_detail/hip_f8_impl.h
@@ -128,11 +128,9 @@ uint8_t cast_to_f8(T _x, bool stoch, uint32_t rng) {
 
   if (exponent_diff>0) {
     //mantissa >>= exponent_diff;
-    // Use a loop to work around the compiler bug that produced the wrong result for the right shift
+    // Work around the compiler bug that produced the wrong result for the right shift
     // TODO: Change back after the compiler bug is resolved.
-    int shift_amount = exponent_diff > 32 ? 32 : exponent_diff;
-    for (int i=0;i<shift_amount;++i)
-      mantissa >>= 1;
+    mantissa = exponent_diff >= 32 ? 0 : (mantissa >> exponent_diff);
   }
   else if (exponent_diff == -1)
     mantissa <<= -exponent_diff;
--- a/transformer_engine/common/amd_detail/hip_f8_impl.h
+++ b/transformer_engine/common/amd_detail/hip_f8_impl.h
@@ -23,9 +23,6 @@ uint8_t cast_to_f8(T _x, bool stoch, uint32_t rng) {
   static_assert(wm+we==7, "wm+we==7");
   static_assert(is_half || is_float, "Only half and float can be cast to f8");
 
-  //if(sizeof(T)==2 && we==5 && !negative_zero_nan)
-    //return cast_to_f8_no_range_reduce<2, 5, __half>(_x, stoch, rng);
-
   const int mfmt = (sizeof(T)==4) ? 23 : 10;
   uint32_t x;
   if(sizeof(T)==4)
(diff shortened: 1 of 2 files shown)


wangye805 (commented):
Quick question: why this change is needed with the cast_transpose PR? Our dev CI is not failing without this change.

wenchenvincent:
> Quick question: why this change is needed with the cast_transpose PR? Our dev CI is not failing without this change.

This is for performance. They found that the current fp8 header was using the SW conversion on gfx942 and thus wrote their own conversion function. We should fix this in the fp8 header and let them use the fp8 header.

wangye805:
> > Quick question: why this change is needed with the cast_transpose PR? Our dev CI is not failing without this change.
> 
> This is for performance. They found that the current fp8 header was using the SW conversion on gfx942 and thus wrote their own conversion function. We should fix this in the fp8 header and let them use the fp8 header.

If you take a look at the most recent CI: http://ml-ci-internal.amd.com:8080/job/TransformerEngine/job/TransformerEngine/job/PR-89/7/pipeline-console/?selected-node=356

wenchenvincent:
> > > Quick question: why this change is needed with the cast_transpose PR? Our dev CI is not failing without this change.
> > 
> > 
> > This is for performance. They found that the current fp8 header was using the SW conversion on gfx942 and thus wrote their own conversion function. We should fix this in the fp8 header and let them use the fp8 header.
> 
> If you take a look at the most recent CI: http://ml-ci-internal.amd.com:8080/job/TransformerEngine/job/TransformerEngine/job/PR-89/7/pipeline-console/?selected-node=356

What's the issue?

wangye805:
> > > > Quick question: why this change is needed with the cast_transpose PR? Our dev CI is not failing without this change.
> > > 
> > > 
> > > This is for performance. They found that the current fp8 header was using the SW conversion on gfx942 and thus wrote their own conversion function. We should fix this in the fp8 header and let them use the fp8 header.
> > 
> > 
> > If you take a look at the most recent CI: http://ml-ci-internal.amd.com:8080/job/TransformerEngine/job/TransformerEngine/job/PR-89/7/pipeline-console/?selected-node=356
> 
> What's the issue?

Those act test suites are failing: 
image

wenchenvincent:
Seems it failed for this PR as well. Let me dig into it.

transformer_engine/common/amd_detail/hip_f8_impl.h
```diff
@@ -126,8 +126,14 @@ uint8_t cast_to_f8(T _x, bool stoch, uint32_t rng) {
      but after shift right by 4 bits, it would look like midpoint.
   */
 
-  if (exponent_diff>0)
-    mantissa >>= exponent_diff;
+  if (exponent_diff>0) {
+    //mantissa >>= exponent_diff;
+    // Use a loop to work around the compiler bug that produced the wrong result for the right shift
+    // TODO: Change back after the compiler bug is resolved.
+    int shift_amount = exponent_diff > 32 ? 32 : exponent_diff;
```
ipanfilo: Is this limit sufficient by itself, w/o loop?

wenchenvincent: I tried to make it w/o loop and it worked. Code updated.

ipanfilo (approved):

transformer_engine/common/amd_detail/hip_f8_impl.h
```diff
@@ -126,8 +126,12 @@ uint8_t cast_to_f8(T _x, bool stoch, uint32_t rng) {
      but after shift right by 4 bits, it would look like midpoint.
   */
 
-  if (exponent_diff>0)
-    mantissa >>= exponent_diff;
+  if (exponent_diff>0) {
+    //mantissa >>= exponent_diff;
+    // Work around the compiler bug that produced the wrong result for the right shift
+    // TODO: Change back after the compiler bug is resolved.
+    mantissa = exponent_diff >= 32 ? 0 : (mantissa >> exponent_diff);
```
draganmladjenovic: Driveby comment. Not sure if there is a compiler bug here. Out of range shifts are UB.

wenchenvincent: Thanks for the comment. Is there a reference?

ipanfilo: Indeed. C++ standard: 'If the value of rhs is negative or is not less than the number of bits in lhs, the behavior is undefined. '

wenchenvincent: Thanks. I updated the code and comments.

transformer_engine/common/amd_detail/hip_f8_impl.h
```diff
@@ -118,16 +115,23 @@ uint8_t cast_to_f8(T _x, bool stoch, uint32_t rng) {
...
   /* This part is a bit tricky. The judgment of whether it is a tie needs to be done before we shift right 
      as shift right could rip off some residual part and make something not midpoint look like midpoint. 
      For example, the fp16 number 0x1002 (0 00100 0000000010), it is larger than midpoint, 
      but after shift right by 4 bits, it would look like midpoint.
   */
 
   if (exponent_diff>0)
-    mantissa >>= exponent_diff;
+    // Clip the exponent_diff as if the right shift when exponent_diff > 31 is undefined behavior
+    mantissa >>= (exponent_diff > 31 ? 31 : exponent_diff);
```
ipanfilo, line 134: It works because mantissa is actually shorter than 32 bits but strictly speaking there is no correct cutting of exponent_diff: 0x80000000 >> 31 = 1 while we would expect result of 0 for any shift value >31.

ipanfilo (approved):

transformer_engine/common/amd_detail/hip_f8_impl.h
wenchenvincent: Yeah, you're right :)

PR #96, opened 2024-12-04
Ipanfilo/fa support

ipanfilo:
# Description

This is work-in-progress review for ROCm FlashAttn2 support

Fixes #[9693](https://github.com/ROCm/frameworks-internal/issues/9693)

## Type of change

## Changes

- Enable FA specific code on ROCm platforms
- Cherry-pick upstream TE changes needed for proper FA working
- Modify test scripts to accommodate new FA


tests/pytorch/test_numerics.py
```diff
@@ -55,7 +55,7 @@
 _cuda_rng_state = torch.cuda.get_rng_state()
 
 if IS_HIP_EXTENSION:
-    @lru_cache(maxsize=1)
+    @cache
```
wangye805: What's the difference between lru_cache and cache? Why do we need to change it from lru_cache to cache?

ci/_utils.sh
```diff
@@ -55,17 +55,24 @@ configure_gemm_env() {
...
             unset NVTE_FUSED_ATTN NVTE_FUSED_ATTN_AOTRITON
         ;;
         "ck")
+            export NVTE_FLASH_ATTN=0
             export NVTE_FUSED_ATTN_AOTRITON=0
             unset NVTE_FUSED_ATTN NVTE_FUSED_ATTN_CK
         ;;
+        "flash")
+            export NVTE_FUSED_ATTN=0 NVTE_FUSED_ATTN_CK=0 NVTE_FUSED_ATTN_AOTRITON=0
+            unset NVTE_FLASH_ATTN
```
wenchenvincent, line 72: I think we want to `export NVTE_FLASH_ATTN=1` here?

tests/pytorch/fused_attn/test_fused_attn_with_cp.py
```diff
@@ -12,7 +12,7 @@
 from test_fused_attn import ModelConfig
 from transformer_engine.pytorch.attention import (
     _flash_attn_2_plus,
-    _flash_attn_2_3_plus,
+    _flash_attn_2_6_plus,
```
wenchenvincent: It seems that this is not used in this file?

ci/_utils.sh
ipanfilo: All NVTE_FLASH_ATTN and NVTE_FUSED_ATTN* are "1" by default so unset is equivalent to set 1

tests/pytorch/fused_attn/test_fused_attn_with_cp.py
ipanfilo: Oh, yes. It left from pre IFU 1.11 commit where THD was only skipped with 2.6+

wenchenvincent: Oh, also, I think we need to merge the updates from the dev branch (now with IFU 1.11) to make sure there is no breakage from that.

ipanfilo: It has those changes

README.rst
```diff
@@ -21,7 +21,7 @@ Feature Support Status
 
 * Activation, cast, fused softmax, layernorm, rmsnorm, transpose, fused rope, fp8 recipe, HipRTC: fully supported
 * GEMM: partially supported with following input/output types: (fp32/fp32), (fp16/fp16), (bf16/bf16), (fp8, bf8/fp16, bf16, fp32)
-* Attention (Flash Attention, Fused Multihead Attention): partially supported: Fused Attention with AOTriton and CK backends
+* Attention (Flash Attention, Fused Multihead Attention): partially supported: Fused Attention with AOTriton and CK backends, FlashAttention-2 with fixed length sequences
```
wangye805: nit: usually we don't say fixed length sequences in flash-attn/fused-attn. We can change to FlashAttention-2 without variable sequence length feature

transformer_engine/pytorch/attention.py
```diff
@@ -635,13 +629,6 @@ def get_attention_backend(
                     "with causal mask, no dropout, and qkv_format = bshd/sbhd"
                 )
                 use_fused_attention = False
-            # ROCm TE can support generic sliding window with dropout
```
wangye805, line 638: This PR is about flash-attn. Why do we change use_fused_attntion?

transformer_engine/pytorch/attention.py
```diff
@@ -860,7 +847,7 @@ def get_attention_backend(
 
     # Select FusedAttention for performance
     if (
-        use_flash_attention
+        use_flash_attention and (not IS_HIP_EXTENSION)
```
wangye805, line 850: Do we have similar logic to prefer fused attn or flash attn?

wangye805 (changes requested):
By the way, how about the dropout seed offset issue in flash-attn package?

tests/pytorch/test_numerics.py
ipanfilo: @cache is python 3.9 feature, which should be lighter than lru_cache: https://stackoverflow.com/questions/70301475/difference-between-functools-cache-and-lru-cache
It is not critical. The rest of test_numerics does not use caches so I eventually removed it too not to import something just for HIP

transformer_engine/pytorch/attention.py
ipanfilo: I think it was discussed some time ago. This code duplicates code on line 451 that is more generic. If you like, I can make it separate PR

transformer_engine/pytorch/attention.py
ipanfilo: We use generic logic which prefers flash attention over fused attention

tests/pytorch/test_numerics.py
wangye805: Got it. Thanks

transformer_engine/pytorch/attention.py
wangye805: Oh, I see line 451. Yeah, you are right. Let's include your change in this PR.

transformer_engine/pytorch/attention.py
wangye805: Do you mean https://github.com/ROCm/TransformerEngine/blob/e877c03bf662b4889458ba09c73a01e23100dad3/transformer_engine/pytorch/attention.py#L878? 

I saw in line https://github.com/ROCm/TransformerEngine/blob/e877c03bf662b4889458ba09c73a01e23100dad3/transformer_engine/pytorch/attention.py#L850, NV prefer cudnn fused attn "f16_arbitrary_seqlen" over flash attn for performance reason. Maybe later we can try to benchmark which one is better for us in ROCm

ipanfilo: They do it for specific workload and GPU. Yes, we could do the same but currently we don't even have a list of critical workloads and what is model GPU

wangye805 (approved):
LGTM

PR #99, opened 2024-12-10
JAX: runtime ROCm detection is the same as compile time. Removed dupl…

ipanfilo:
…icated code

# Description

Updated how JAX ROCm support is detected runtime to be similar to compile time.
Removed duplicated ROCm specific code

Fixes 

CI failures on Torch+JAX docker. Original way of ROCm JAX detection with devices enumeration modified default ROCm device on multi-GPU systems. Because it is called on library initialization it affected Torch tests. Moreover, it is time consuming

## Type of change

- [ x] Bug fix (non-breaking change which fixes an issue)
- [x ] Code refractor

## Changes

- Removed duplicated is_hip_extension and FP8 data type definition in JAX plugin
- is_hip_extension checks ROCm support in JAX similar to build-time instead of enumerating devices


wenchenvincent (approved):
LGTM!

PR #103, opened 2025-01-02
Disable unsupported test configs

ipanfilo:
# Description

Disable unsupported test configs

Fixes # (issue)

ONNX crash on CI server

## Type of change

## Changes

-Disable unsupported test configs and remove filtering from CI scripts
-Disable multiple test workers usage because some tests are not compatible with it
-Fix ONNX test crash
-Add config test environment config logging
-Don't run tests with multiple backends if those backends are not used bu tests

# Checklist:

### DIFF
--- a/ci/pytorch.sh
+++ b/ci/pytorch.sh
@@ -59,7 +59,7 @@ run_test_config(){
     run_default_fa 1 test_recipe.py
     run 1 test_sanity.py
     run_default_fa 1 test_torch_save_load.py
-    run_default_fa 1 fused_attn/test_fused_attn.py
+    run_default_fa 1 fused_attn/test_fused_attn.py # Backend selection is controlled by the test
 }
 
 run_test_config_mgpu(){
--- a/ci/pytorch.sh
+++ b/ci/pytorch.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-# Copyright (c) 2024, Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (c) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
 #
 # See LICENSE for license information.
 
@@ -9,7 +9,7 @@ DIR=`dirname $0`
 
 TEST_DIR=${TE_PATH}tests/pytorch
 
-: ${TEST_WORKERS:=4}
+#: ${TEST_WORKERS:=4}
 
 install_prerequisites() {
     pip install 'numpy>=1.22.4,<2.0' onnx onnxruntime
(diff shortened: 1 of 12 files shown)


ci/pytorch.sh
```diff
@@ -26,51 +26,49 @@ run() {
...
+    run_default_fa 1 test_gqa.py
+    run_default_fa 1 test_jit.py
+    run_default_fa 1 test_multi_tensor.py
+    run 1 test_numerics.py
+    run_default_fa 3 test_onnx_export.py # All FA are disabled in ONNX export mode
+    run_default_fa 1 test_permutation.py
+    run_default_fa 1 test_recipe.py
+    run 1 test_sanity.py
+    run_default_fa 1 test_torch_save_load.py
+    run_default_fa 1 fused_attn/test_fused_attn.py
```
wangye805: For tests/pytorch/fused_attn/test_fused_attn.py, we recall we are able to run through both CK and AOTriton backends. Why do we only test default_fa here?

ci/pytorch.sh
```diff
@@ -26,51 +26,49 @@ run() {
...
+    echo ==== Run with GEMM backend: $_gemm and Fused attention backend: $_fus_attn ====
+    #_WORKERS_COUNT=$TEST_WORKERS
+    #run_default_fa 1 test_cast_transpose_triton.py
+    run 1 test_cuda_graphs.py
+    run_default_fa 1 test_deferred_init.py
+    run_default_fa 1 test_float8tensor.py
+    run_default_fa 1 test_fused_rope.py
+    run_default_fa 1 test_fusible_ops.py
+    test $_gemm = "hipblaslt" && run_default_fa 3 test_gemm_autotune.py
+    run_default_fa 1 test_gqa.py
```
wangye805: GQA is also related to fused attn, so let's run all backends in gqa

wangye805 (changes requested):

ci/pytorch.sh
ipanfilo: Done

ci/pytorch.sh
ipanfilo: The test controls backend selection itself so setting it from outside interferes with the test and doesn't work properly

wangye805: Emm, regarding the backend selection logic, are you referring to this section of code: https://github.com/ROCm/TransformerEngine/blob/b279c5a130a138464d5b5977b6dfee4093055659/tests/pytorch/fused_attn/test_fused_attn.py#L196-L204
This NVTE_FUSED_ATTN_BACKEND env is used in NV side (transformer_engine/common/fused_attn: https://github.com/ROCm/TransformerEngine/blob/b279c5a130a138464d5b5977b6dfee4093055659/transformer_engine/common/fused_attn/fused_attn.cpp#L199) but not our side. 

Our current backend selection logic using NVTE_FUSED_ATTN_CK and NVTE_FUSED_ATTN_AOTRITON is not compatible with this section of code in tests/pytorch/fused_attn/test_fused_attn.py. 

Perhaps you can try to add NVTE_FUSED_ATTN_BACKEND env to our rocm side to be consistent with NVTE?

ci/pytorch.sh
```diff
@@ -26,51 +26,49 @@ run() {
...
-    run 1 test_torch_save_load.py
-    test $_fus_attn != "unfused" && run 1 fused_attn/test_fused_attn.py
+    run_default_fa 1 test_multi_tensor.py
+    run 1 test_numerics.py
+    run_default_fa 3 test_onnx_export.py # All FA are disabled in ONNX export mode
+    run_default_fa 1 test_permutation.py
+    run_default_fa 1 test_recipe.py
+    run 1 test_sanity.py
+    run_default_fa 1 test_torch_save_load.py
+    run_default_fa 1 fused_attn/test_fused_attn.py # Backend selection is controlled by the test
```
wangye805, line 62: As I replied in our conversation, we probably need to apply env NVTE_FUSED_ATTN_BACKEND into our https://github.com/ROCm/TransformerEngine/blob/b279c5a130a138464d5b5977b6dfee4093055659/transformer_engine/common/fused_attn_rocm/fused_attn.cpp#L104 in order to use the backend selection from the test

wangye805 (changes requested):

ci/pytorch.sh
ipanfilo: My bad, I added it to different branch. Now fixed.
I thought about using NVTE_FUSED_ATTN_BACKEND but its current meaning is recommended backend when multiple are supported which is different from our logic. Sure, we can redefine the logic on ROCm or use it as recommended flag in parallel with NVTE_FUSED_ATTN_* envs but I decided to put is aside of this PR

wangye805 (approved):
LGTM. Let's wait for the manual CI results and then merge if no failures.

PR #109, opened 2025-01-15
Fix crash on hipblaslt algo selection. Add test precision workaround …

ipanfilo:
…on MI308. Add pytorch 2.5 support

# Description

-Fix crash in corner cases with hipBlasLt algo selection
-Guard selected hipblaslt algo logging with NVTE_LOG_GEMM_CONFIG
-Add workaround for numerical test error on MI308 with ROCm6.3
-Add torch.compiler reset to avoid test failure on pytorch 2.5 due to cache limit reaching

Fixes # (issue)

https://github.com/ROCm/frameworks-internal/issues/10435
https://github.com/ROCm/frameworks-internal/issues/10561
Crash on debugging https://github.com/ROCm/frameworks-internal/issues/9693

## Type of change

### DIFF
--- a/tests/pytorch/test_numerics.py
+++ b/tests/pytorch/test_numerics.py
@@ -24,7 +24,7 @@ from transformer_engine.pytorch.utils import (
     is_bf16_compatible,
 )
 if IS_HIP_EXTENSION:
-    from transformer_engine.pytorch.utils import is_mi200
+    from transformer_engine.pytorch.utils import is_mi200, is_mi308
 
 from transformer_engine.pytorch import (
     DotProductAttention,
@@ -1864,7 +1864,11 @@ def test_transformer_layer_hidden_states_format(dtype, bs, model):
     # TODO: wait for the full determinism fix from hipblaslt
     if IS_HIP_EXTENSION:
         if use_hipblaslt():
-            torch.testing.assert_close(y_bshd, y_sbhd.transpose(0, 1).contiguous())
+            tols = dtype_tols(dtype)
+            if dtype in (torch.float16, torch.bfloat16) and is_mi308():
+                # mi308 hipblaslt precision issue
+                tols["atol"] = 5e-3
+            torch.testing.assert_close(y_bshd, y_sbhd.transpose(0, 1).contiguous(), **tols)
         else:
             assert torch.equal(y_bshd, y_sbhd.transpose(0, 1).contiguous()), "Tensors are not equal"
     else:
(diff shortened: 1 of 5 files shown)


transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1307,6 +1304,11 @@ void hipblaslt_gemm(const Tensor *inputA,
       cached_algo.ws_size_min = algoArr[bestAlgo].workspaceSize;
       cached_algo.ws_size_max = workspaceSize;
 
+      if (const char* env_p = std::getenv("NVTE_LOG_GEMM_CONFIG") ) {
```
wenchenvincent: `NVTE_LOG_GEMM_CONFIG` is used to log the gemm configs. And this is for log the solution information in autotuning. Let's use a different env var like `NVTE_LOG_HIPBLASLT_TUNING`

transformer_engine/pytorch/distributed.py
```diff
@@ -242,14 +242,14 @@ def _get_active_autocast_contexts():
 
     gpu_autocast_enabled = torch.is_autocast_enabled()
     gpu_autocast_dtype = torch.get_autocast_gpu_dtype()
-    gpu_autocast_ctx = torch.cuda.amp.autocast(
-        gpu_autocast_enabled, gpu_autocast_dtype, autocast_cached
+    gpu_autocast_ctx = torch.amp.autocast('cuda',
```
wenchenvincent, line 245: Is this change for pytorch 2.5?

tests/pytorch/test_numerics.py
```diff
@@ -670,6 +670,7 @@ def test_gpt_full_activation_recompute(dtype, bs, model, fp8, fp8_model_params,
         pytest.skip(reason_for_no_fp8)
 
     config = model_configs[model]
+    torch.compiler.reset() # avoid cache size limit overflow
```
wenchenvincent, line 673: Is this the workaround for the inductor issue? If so, might need to put a TODO here to remove it later when the issue is fixed in pytorch.

transformer_engine/pytorch/distributed.py
ipanfilo: Yes, old API is deprecated. It is not removed yet, just throws warnings

tests/pytorch/test_numerics.py
ipanfilo: It is workaround for new JIT cache overflow handling like described in https://github.com/ROCm/frameworks-internal/issues/10561

wangye805 (approved):
LGTM

PR #110, opened 2025-01-15
Bump to AOTriton 0.8.2b, and related build system changes

xinyazhang:
# Description

This bumps the AOTriton used by TE from 0.5b to 0.8b. However, no newly supported features are enabled right now.

Prerequisite to Fixes of SWDEV-490644

## Type of change

- [x] Infra/Build change
- [ ] Code refractor

## Changes

- Bump AOTriton to 0.8b
  + Use new API, notable changes are using Tensor for Philox PRNG arguments
  + Compile AOTriton with _TEprivate suffix, avoding name conflicts at both source code level (`namespace aotriton_TEprivate`) and binary level (`libaotriton_TEprivate_v2.so`)
  + Download AOTriton GPU kernels from pre-compiled package, to avoid re-compiling Triton GPU kernels

# Checklist:


xinyazhang:
The code compiles, AOTriton GPU kernels and libraries can be found in the final wheel file. `ldd` shows the install `libtransformer_engine.so` can correctly locate `libaotriton_TEprivate_v2.so.0.8.0` under the pip wheel install location

image

johnnynunez:
is it working with gx1100? I went from 364 compilation errors with rocm6.1 to 20 with rocm6.2 to 7 with rocm6.3.

xinyazhang:
> is it working with gx1100? I went from 364 compilation errors with rocm6.1 to 20 with rocm6.2 to 7 with rocm6.3.

AOTriton 0.8b supports gfx1100, but by default this fork of TE only builds for gfx942.
To support gfx1100 you need to edit `transformer_engine/common/CMakeLists.txt` and change  `-DTARGET_GPUS=MI300X` to `-DTARGET_GPUS=MI300X;Navi31`

johnnynunez:
> > is it working with gx1100? I went from 364 compilation errors with rocm6.1 to 20 with rocm6.2 to 7 with rocm6.3.
> 
> AOTriton 0.8b supports gfx1100, but by default this fork of TE only builds for gfx942. To support gfx1100 you need to edit `transformer_engine/common/CMakeLists.txt` and change `-DTARGET_GPUS=MI300X` to `-DTARGET_GPUS=MI300X;Navi31`

Uoh thank you so much. I will try

xinyazhang:
```
[2025-01-24T23:19:57.589Z] =========================== short test summary info ============================
[2025-01-24T23:19:57.589Z] FAILED ../tests/pytorch/test_numerics.py::test_transformer_layer_hidden_states_format[126m-2-dtype1] - AssertionError: Tensor-likes are not close!
[2025-01-24T23:19:57.589Z] 
[2025-01-24T23:19:57.589Z] Mismatched elements: 5810 / 3145728 (0.2%)
[2025-01-24T23:19:57.589Z] Greatest absolute difference: 0.001953125 at index (0, 1032, 484) (up to 1e-05 allowed)
[2025-01-24T23:19:57.589Z] Greatest relative difference: inf at index (1, 2003, 92) (up to 0.001 allowed)
[2025-01-24T23:19:57.589Z] FAILED ../tests/pytorch/test_numerics.py::test_transformer_layer_hidden_states_format[126m-2-dtype2] - AssertionError: Tensor-likes are not close!
[2025-01-24T23:19:57.589Z] 
[2025-01-24T23:19:57.589Z] Mismatched elements: 944 / 3145728 (0.0%)
[2025-01-24T23:19:57.589Z] Greatest absolute difference: 0.001953125 at index (0, 384, 642) (up to 1e-05 allowed)
[2025-01-24T23:19:57.589Z] Greatest relative difference: inf at index (0, 186, 472) (up to 0.016 allowed)
[2025-01-24T23:19:57.589Z] =========== 2 failed, 590 passed, 765 warnings in 229.58s (0:03:49) ============
```
Probably needs some tolerance adjustment but I cannot precisely locate it. @wangye805 @wenchenvincent

ipanfilo:
> Probably needs some tolerance adjustment but I cannot precisely locate it. @wangye805 @wenchenvincent

It is already done in dev. You might want to merge it to the branch or just ignore those errors

transformer_engine/common/CMakeLists.txt
```diff
@@ -222,19 +223,64 @@ if (USE_CUDA)
   target_compile_definitions(transformer_engine PUBLIC NV_CUDNN_FRONTEND_USE_DYNAMIC_LOADING)
 else()
   if(USE_FUSED_ATTN_AOTRITON)
+    set(__AOTRITON_VER "0.8.2b")
```
ipanfilo: It is better to have just file name. And there should be ROCm version detection

build_tools/utils.py
```diff
@@ -9,6 +9,7 @@
 import functools
 import glob
 import importlib
+import importlib.metadata
```
ipanfilo, line 12: Why is it needed?

transformer_engine/common/CMakeLists.txt
```diff
@@ -222,19 +223,64 @@ if (USE_CUDA)
...
+      if(USE_FUSED_ATTN_AOTRITON_BUILD_GPU_KERNELS)
+        set(AOTRITON_NOIMAGE_MODE OFF)
+      else()
+        set(AOTRITON_NOIMAGE_MODE ON)
+      endif()
+      include(ExternalProject)
+      ExternalProject_Add(aotriton_external
+        URL ../../3rdparty/aotriton 
+        CMAKE_ARGS -DCMAKE_INSTALL_PREFIX:PATH=${__AOTRITON_INSTALL_DIR}
+        -DTARGET_GPUS=MI300X
```
ipanfilo: This option used to be configurable

ipanfilo (commented):
Copyright dates should also be updated

build_tools/utils.py
xinyazhang: `importlib.metadata` is needed on certain version: https://discuss.python.org/t/module-importlib-has-no-attribute-metadata/20947

transformer_engine/common/CMakeLists.txt
xinyazhang: With ROCM version detection the logic will be slight more complicated and contradicting to first goal ("just file name")

transformer_engine/common/CMakeLists.txt
xinyazhang: I saw `gfx942` is hardcoded at somewhere else, can you point me the preferred cmake option in the build system? Is `CMAKE_HIP_ARCHITECTURES` okay?

build_tools/utils.py
ipanfilo: Oh, yes. I saw it and added to my PR before but later could not reproduce the problem and it was removed.

transformer_engine/common/CMakeLists.txt
ipanfilo: It does not, there should be multiple filenames for different ROCm versions. Each file i.e. URL has associated SHA. If ROCm version is not supported, buld has to failed or fallback to kernel build path.

transformer_engine/common/CMakeLists.txt
ipanfilo: It is hardcoded but like an option which can be overridden. Like MI300X used to be an option which could be changed when call cmake if needed

transformer_engine/common/CMakeLists.txt
```diff
@@ -218,34 +218,79 @@ if (USE_CUDA)
...
+        set(gfx1100_key Navi31)
+
+        foreach(X IN LISTS CMAKE_HIP_ARCHITECTURES)
+          set(key ${X})
+          string(APPEND key "_key")
+          string(APPEND target_gpus ${${key}})
+          string(APPEND target_gpus "|")
+        endforeach()
+      endmacro()
+      get_target_gpus_from_pytorch(aotriton_target_gpus)
```
ipanfilo: The name is confusing. TE builds w/o pytorch.

transformer_engine/common/CMakeLists.txt
```diff
@@ -218,34 +218,79 @@ if (USE_CUDA)
...
+      # GPU Targets
+      macro(get_target_gpus_from_pytorch target_gpus)
+        set(gfx90a_key MI200)
+        set(gfx942_key MI300X)
+        set(gfx1100_key Navi31)
+
+        foreach(X IN LISTS CMAKE_HIP_ARCHITECTURES)
+          set(key ${X})
+          string(APPEND key "_key")
+          string(APPEND target_gpus ${${key}})
```
ipanfilo: What is behavior if target architecture is not in the list? And just in case - does AOTriton distinct MI300 and MI308 for example? It is nice feature but might be hard to make it robust

transformer_engine/common/CMakeLists.txt
```diff
@@ -217,24 +218,114 @@ if (USE_CUDA)
...
+        # len("rocm") == 4
+        string(SUBSTRING ${AOTRITON_ROCM_BUILD_STR} 4 -1 AOTRITON_ROCM_BUILD)
+        # Find the last build that <= system's ROCM
+        # Assume the list is from lower to higher
+        if(AOTRITON_ROCM_BUILD VERSION_GREATER __AOTRITON_SYSTEM_ROCM)
+          break()
+        endif()
+        set(__AOTRITON_ROCM ${AOTRITON_ROCM_BUILD})
+      endforeach()
+      list(FIND __AOTRITON_ROCM_LIST "rocm${__AOTRITON_ROCM}" __AOTRITON_ROCM_INDEX)
```
ipanfilo: What is behaviour if build for different ROCm version, say 6.4?

xinyazhang: 6.4 will use largest available ROCM build. This is an example of the code above: https://github.com/pytorch/pytorch/pull/137443#issuecomment-2405822324

Overall AOTriton is supposed to compiled against a newer ROCM version only when there were ABI breaks.

transformer_engine/common/CMakeLists.txt
xinyazhang: > What is behavior if target architecture is not in the list?

Then the list will be empty and the build is likely to fail. However on that situation I think it is more suitable to set `-DUSE_FUSED_ATTN_AOTRITON=OFF` since there is no point to compile AOTriton if `CMAKE_HIP_ARCHITECTURES` doesn't contain MI200/300/Navi31.

> does AOTriton distinct MI300 and MI308 for example

Planned in 0.9 but now deferred to 0.10. Need a new dispatcher design to fit all demands shown below:
* Same arch (e.g., gfx942), different tuning info (MI325X vs MI300X)
* Same arch, same tuning info, but different brand name (MI300X vs MI300A)
* Different arch, same tuning info (due to similarity), different brand name (gfx1100/Navi31 vs gfx1101/Navi32)

(BTW since we are not supposed to expose internal codenames so in the final version Navi31 will become `RX7900XTX`)

transformer_engine/common/CMakeLists.txt
xinyazhang: Renamed to translate_arch_to_gpu_names with a new commit.

transformer_engine/common/CMakeLists.txt
xinyazhang: Oh wait a sec. I forgot the reason I didn't add multiple ROCM support.

Okay here it is. Unlike PyTorch. TE doesn't need to pick the the matching rocm version -- even if there are ABI breaks in the runtime.

The changes to the build system  will download the AOTriton pre-compiled binary, and only use it GPU kernels (This is to avoid build GPU kernel from source which is a pretty fragile process due to the ~1GiB LLVM package).
The runtime will be re-built (with `_TEprivate` suffix) on whatever ROCM that's being used.

(To clarify: all released packages of AOTriton contains the same GPU kernels. The only difference is the C++ dispatcher/shim library)

I'll remove the ROCM version detection since it is unnecessary.

transformer_engine/common/CMakeLists.txt
ipanfilo: So, if ROCm detection is no needed like stated in another comment, it should definitely be just file name rather than multiple components.

xinyazhang: Done

transformer_engine/common/CMakeLists.txt
```diff
@@ -222,19 +223,76 @@ if (USE_CUDA)
...
+          set(key ${X})
+          string(APPEND key "_key")
+          string(APPEND target_gpus ${${key}})
+          string(APPEND target_gpus "|")
+        endforeach()
+      endmacro()
+      translate_arch_to_gpu_names(aotriton_target_gpus)
+      include(ExternalProject)
+      ExternalProject_Add(aotriton_external
+        URL ../../3rdparty/aotriton
```
ipanfilo: Do I understand right that this instructs CMake to copy content of ../../3rdparty/aotriton to {BINDIR}/aotriton_external and build there?

xinyazhang: `aotriton_external` is the target name. The precise build directory is `./build/cmake/aotriton_external-prefix/src/aotriton_external-build/`.
It will be installed to `${CMAKE_CURRENT_BINARY_DIR}/aotriton`. To be specific it is `./build/cmake/aotriton`

The final installation to `./transformer_engine/lib/` happens somewhere else.

ipanfilo: Why not build it from ../../3rdparty/aotriton then?

xinyazhang: > Why not build it from ../../3rdparty/aotriton then?

I'm confused. What the point of doing an in-source build?

ipanfilo: Sorry, I mean build from original source. According to doc, now it copies sources to  ./build/cmake/aotriton_external-prefix/src/aotriton_external and then builds to ./build/cmake/aotriton_external-prefix/src/aotriton_external-build. Why is this copy needed?

xinyazhang: oops I should use `SOURCE_DIR` instead of `URL`

ipanfilo (approved):

transformer_engine/common/CMakeLists.txt
```diff
@@ -222,19 +223,76 @@ if (USE_CUDA)
...
+                            URL_HASH SHA256=${__AOTRITON_SHA256}
+                            SOURCE_DIR ${CMAKE_CURRENT_BINARY_DIR}/aotriton_tarball
+                            CONFIGURE_COMMAND ""
+                            BUILD_COMMAND ""
+                            INSTALL_COMMAND ${CMAKE_COMMAND} -E copy_directory
+                            "${CMAKE_CURRENT_BINARY_DIR}/aotriton_tarball/lib/aotriton.images"
+                            "${__AOTRITON_INSTALL_DIR}/lib/aotriton.images")
+        add_dependencies(aotriton aotriton_images)
+      endif()
+      install(DIRECTORY
```
ipanfilo, line 288: I've tried to use those changes and found minor issue:
It installs two copies of libaotriton library - *.so and .*so.0.8.2. The build make the former a symlink to the latter but wheel packager doesn't support symlinks so it end up with two copies. So it would be nice to exclude symlink from installation.
Can you add such parameters to install command?
              PATTERN "cmake" EXCLUDE
              PATTERN "libaotriton${__AOTRITON_SUFFIX}_v2.so" EXCLUDE)
 
It also installs all kernels for all supported architectures despite target GPUs which is probably OK supposing the libary will load only needed ones

xinyazhang: Actually I have more concern about the failed PyTorch Integration unittests. Did you figure out why? It seems the GPU images are not copied to correct place.

ipanfilo (approved):
LGTM.  The failing tests are fixed in dev. Examples failed due to server connection error

PR #116, opened 2025-01-24
Added NVTE_PLATFORM: For robust ROCm and CUDA platform detection and …

VeeraRajasekhar:
# Description

Our customers may have a docker image contain both /opt/rocm and /bin/nvcc. Our current platform selection logic will not work then.  This PR introduces changes to robustly detect and select the appropriate platform.

Fixes # (issue)

## Type of change

- [X] Infra/Build change
- [ ] Code refractor

## Changes

- Enhanced platform detection logic to prioritize NVTE_PLATFORM if set.
- Improved fallback detection for both ROCm and CUDA environments.
- Raised ValueError for invalid or ambiguous platform configurations.
- Ensured ROCm is prioritized when both /opt/rocm and /bin/nvcc exist.

# Checklist:


build_tools/utils.py
```diff
@@ -1,6 +1,6 @@
 # This file was modified for portability to AMDGPU
-# Copyright (c) 2024, Advanced Micro Devices, Inc. All rights reserved.
-# Copyright (c) 2022-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
```
wangye805: 2024-2025, this file started to exist for us in 2024

build_tools/utils.py
```diff
@@ -1,6 +1,6 @@
...
+# Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
wangye805: No need to change the NV side year

build_tools/utils.py
```diff
@@ -165,23 +165,59 @@ def found_pybind11() -> bool:
...
 @functools.lru_cache(maxsize=None)
 def rocm_build() -> bool:
-    """ ROCm build should be performed if:
-    - It is configured with NVTE_USE_ROCM=1 env
-      OR:
-    - HIP compiler is found and CUDA one is not
+    """
+    ROCm build should be performed if:
+    1. It is configured with NVTE_USE_ROCM=1 env
+    2. If `NVTE_PLATFORM` is set:
```
wangye805: I have different opinions about the introduction of "NVTE_PLATFORM" here. 

I think one of NVTE_USE_ROCM/NVTE_PLATFORM is sufficient. We don't need both. 

Recall our motivation that when both nvcc and hipcc exist, we want one additional variable to distinguish the platform. We can keep the previous logic for rocm_build() but add a default value 1 for NVTE_USE_ROCM

transformer_engine/common/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 # This file was modified for portability to AMDGPU
-# Copyright (c) 2022-2024, Advanced Micro Devices, Inc. All rights reserved.
-# Copyright (c) 2022-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2022-2025, Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
wangye805: I don't think we need to update NV side. We only update them when we cherry-pick commits or IFU from upstream

transformer_engine/common/CMakeLists.txt
```diff
@@ -12,10 +12,32 @@ option(USE_ROCBLAS "Use ROCBLAS" ON)
 option(USE_FUSED_ATTN_AOTRITON "Use aotriton backend" ON)
 option(USE_FUSED_ATTN_CK "Use ck backend" ON)
 
-if(NOT USE_ROCM)
-  if(((EXISTS "/opt/rocm/") OR (EXISTS $ENV{ROCM_PATH})) AND NOT (EXISTS "/bin/nvcc"))
+# Check for the environment variable to decide between ROCm and CUDA
+if(DEFINED ENV{NVTE_PLATFORM})
```
wangye805: Same here. I don't think NVTE_PLATFORM is needed. Besides, we would prefer flags passed from setup.py, not reading directly from env

wangye805 (changes requested):

build_tools/utils.py
ipanfilo: Agree. We don't need two flags. NVTE_USE_ROCM seems easier to handle and it is already here. The logic would be little more complicated than just set default to 1 though - it should distinct "True/False/Unset" and in the latter case use ROCm if it is installed

ipanfilo (changes requested):
In addition, build logic for tests/cpp should also be updated

build_tools/utils.py
VeeraRajasekhar: I agree that having two flags might be unnecessary. My concern was that the name "NVTE_USE_ROCM" might already be interpreted differently by users. Currently, it's used to explicitly specify ROCm, even if CUDA is available, but setting it to zero doesn't necessarily imply that CUDA should be used. The name of the variable itself isn't very intuitive.

To avoid breaking existing behavior, I introduced a new variable. However, if you'd prefer to stick with NVTE_USE_ROCM, I can modify its behavior to define 1 as using ROCm, 0 as using CUDA, and default to ROCm if it's unset. Essentially, replacing the NVTE_PLATFORM variable with NVTE_USE_ROCM will achieve the same outcome in the current changes.

build_tools/utils.py
VeeraRajasekhar: Updated

transformer_engine/common/CMakeLists.txt
VeeraRajasekhar: Updated

build_tools/utils.py
VeeraRajasekhar: Updated

build_tools/utils.py
ipanfilo: The effect of this env will indeed change. For users who used to correctly set it though effect will remain the same.

VeeraRajasekhar: Updated the codebase to use NVTE_USE_ROCM

transformer_engine/common/CMakeLists.txt
VeeraRajasekhar: Added flags and used them here.

build_tools/utils.py
```diff
@@ -165,23 +165,58 @@ def found_pybind11() -> bool:
 
 @functools.lru_cache(maxsize=None)
 def rocm_build() -> bool:
-    """ ROCm build should be performed if:
-    - It is configured with NVTE_USE_ROCM=1 env
-      OR:
-    - HIP compiler is found and CUDA one is not
     """
-    if bool(int(os.getenv("NVTE_USE_ROCM", "0"))):
+    Determines which build platform to use:
```
ipanfilo, line 169: Update docstring

build_tools/utils.py
```diff
@@ -165,23 +165,58 @@ def found_pybind11() -> bool:
...
+    
+    Returns:
+        bool: `True` for ROCm, `False` for CUDA.
+
+    Raises:
+        ValueError: For invalid `NVTE_USE_ROCM` or if neither ROCm nor CUDA is detected.
+    """
+    nvte_use_rocm = os.getenv("NVTE_USE_ROCM")
+    if nvte_use_rocm:
+        nvte_use_rocm = int(nvte_use_rocm)
```
ipanfilo: Let's make it simple !=0 / ==0. No big value in preventing used from using other numbers except 0 or 1

build_tools/utils.py
```diff
@@ -165,23 +165,58 @@ def found_pybind11() -> bool:
...
         pass
 
-    _, hipcc_bin = rocm_path()
-    return hipcc_bin.is_file()
-
+    # If neither ROCm nor CUDA is detected, raise an error
+    raise ValueError(
+        "Could not detect ROCm or CUDA platform. "
+        "Please ensure one of them is installed and accessible."
+    )
```
ipanfilo: The logic is inconsistent. Sometimes it throws ValueError sometimes FileNotFoundError for the same conditions.

transformer_engine/common/CMakeLists.txt
```diff
@@ -1,31 +1,27 @@
...
+# Copyright (c) 2022-2025, Advanced Micro Devices, Inc. All rights reserved.
 # Copyright (c) 2022-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # See LICENSE for license information.
 
 cmake_minimum_required(VERSION 3.21)
 
-option(USE_ROCM "Use ROCm" OFF)
+option(USE_ROCM "Use ROCM" ON)
+option(USE_CUDA "Use CUDA" FALSE)
```
ipanfilo: USE_CUDA is not an option but variable. No need to add it here. It rather should be removed from tests/cpp

transformer_engine/common/CMakeLists.txt
```diff
@@ -1,31 +1,27 @@
...
+if (NOT USE_ROCM)
+  set(USE_CUDA TRUE)
 endif()
 
 if (USE_ROCM)
   if (NOT USE_HIPBLASLT AND NOT USE_ROCBLAS)
     message(FATAL_ERROR "Need specify at least one GEMM library to use: HIPBLASLT or ROCBLAS")
   endif()
   unset(USE_CUDA)
-else()
```
ipanfilo, line 27: Why is it removed from here?

tests/cpp/CMakeLists.txt
```diff
@@ -6,14 +6,46 @@
 
 cmake_minimum_required(VERSION 3.18)
 
-option(USE_CUDA "Use CUDA" ON)
-option(USE_ROCM "Use ROCm" OFF)
+option(USE_CUDA "Use CUDA" OFF)
+option(USE_ROCM "Use ROCm" ON)
+
+# Check for the environment variable to decide between ROCm and CUDA
+if(DEFINED ENV{NVTE_USE_ROCM})
```
ipanfilo: Like was discussed, it should rely on USE_ROCM option rather than NVTE_USE_ROCM env

setup.py
ipanfilo, line 1: Copyright date

ipanfilo (commented):
Please update building section of README

tests/cpp/CMakeLists.txt
wangye805: This is in tests/cpp/CMakeslists, we don't have a way to pass USE_ROCM from setup.py to here

build_tools/utils.py
VeeraRajasekhar: Updated

build_tools/utils.py
VeeraRajasekhar: Changed the logic and updated.

build_tools/utils.py
VeeraRajasekhar: Made all the exceptions consistent.

transformer_engine/common/CMakeLists.txt
VeeraRajasekhar: Removed USE_CUDA.

tests/cpp/CMakeLists.txt
VeeraRajasekhar: This file cannot access the labels created from setup.py. As we will run this separately, I read the env variable and populated the logic here.

setup.py
VeeraRajasekhar: Updated.

transformer_engine/common/CMakeLists.txt
VeeraRajasekhar: Fixed this.

VeeraRajasekhar:
> Please update building section of README

Updated the README.

build_tools/utils.py
```diff
@@ -165,23 +165,53 @@ def found_pybind11() -> bool:
...
+        - Non-zero value: Use ROCm, if hipcc is detected.
+        - Zero value: Use CUDA, if nvcc is detected.
+    - If `NVTE_USE_ROCM` is not set:
+        - Attempt to auto-detect: Check for ROCm first, then CUDA.
+
+    Returns:
+        bool: `True` for ROCm, `False` for CUDA.
+
+    Raises:
+        FileNotFoundError: If required tools (hipcc or nvcc) are not found.
```
ipanfilo, line 182: It also raises ValueError

tests/cpp/CMakeLists.txt
```diff
@@ -1,19 +1,56 @@
...
+# Copyright (c) 2022-2025, Advanced Micro Devices, Inc. All rights reserved.
 # Copyright (c) 2022-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # See LICENSE for license information.
 
 cmake_minimum_required(VERSION 3.18)
 
-option(USE_CUDA "Use CUDA" ON)
-option(USE_ROCM "Use ROCm" OFF)
+set(USE_ROCM ON)
```
ipanfilo: USE_ROCM is an option

tests/cpp/CMakeLists.txt
ipanfilo: Let's make both CMake behave similar way.

build_tools/utils.py
```diff
@@ -165,23 +165,53 @@ def found_pybind11() -> bool:
...
+    if nvte_use_rocm:
+        try:
+            nvte_use_rocm = bool(int(nvte_use_rocm))
+        except ValueError:
+            raise ValueError(
+                f"Invalid value for NVTE_USE_ROCM: '{nvte_use_rocm}'.")
+
+        if nvte_use_rocm:
+            _, hipcc_bin = rocm_path()
+            if hipcc_bin.is_file():
```
wangye805, line 194: Looks like duplicated with line 202-203. 

Okay to keep as it is. Or perhaps we can make the code more concise by:

nvte_use_rocm = os.getenv("NVTE_USE_ROCM")
if nvte_use_rocm is None or bool(int(nvte_use_rocm)):
  #try to detect rocm
   _, hipcc_bin = rocm_path()
   if hipcc_bin.is_file():
      return true
   else:
      raise FileNotFoundError
else:
   #try to detect cuda 
   #blablabla

build_tools/utils.py
```diff
@@ -165,23 +165,53 @@ def found_pybind11() -> bool:
...
+                return True
+            else:
+                 raise FileNotFoundError(f"Could not find hipcc at {hipcc_bin}")
```
wangye805: raise should be with same indent as line 194 return

tests/cpp/CMakeLists.txt
```diff
@@ -1,19 +1,56 @@
...
+      set(USE_ROCM OFF)
+      set(USE_CUDA ON)
+      message("NVIDIA GPU detected.")
+    else()
+      message(FATAL_ERROR "FileNotFoundError: Could not find NVCC at '/bin/nvcc'")
+    endif()
+  endif()
+
+else()
+# if NVTE_PLATFORM is not set, check for rocm first then cuda.
```
wangye805: We do not have NVTE_PLATFORM anymore, right?

tests/cpp/CMakeLists.txt
```diff
@@ -1,19 +1,56 @@
...
+    set(USE_ROCM ON)
+  elseif (EXISTS "/bin/nvcc")
+    set(USE_ROCM OFF)
+    set(USE_CUDA ON)
+    message("NVIDIA GPU detected.")
+  else()
+    message(FATAL_ERROR "Could not detect ROCm or CUDA platform. Please install")
+  endif()
+
+endif()
```
wangye805, line 28: We can make the same logic concise here, similar to my comments in python rocm_build() function

if ENV not set or ENV int value >0:
   #try to detect hipcc
else:
   #try to detect nvcc

wangye805 (changes requested):

build_tools/utils.py
ipanfilo: I think with this approach 'else: raise FileNonFoundError' should rather be 'elif nvte_use_rocm is not None: raise FileNonFoundError'
And consequently 'else: #try do detect cuda' becomes unconditional.

VeeraRajasekhar: hey @wangye805, @ipanfilo, I can make these changes, but it will just save us 2 lines of repetition and it will be little difficult to understand on the first glance. What do you guys suggest. The current version will be little cleaner and easier to understand.

tests/cpp/CMakeLists.txt
wenchenvincent: Since whether using ROCm or CUDA is already determined in setup.py and passed to the CMakeList.txt by the macro USE_ROCM. I think we don't need to do detection in CMakeList.txt again.

build_tools/utils.py
```diff
@@ -165,23 +165,45 @@ def found_pybind11() -> bool:
...
+            raise FileNotFoundError(f"Could not find hipcc at {hipcc_bin}")
+        else:
+            try:
+                cuda_path()
+                return False
+            except FileNotFoundError:
+                # If neither ROCm nor CUDA is detected, raise an error
+                raise FileNotFoundError("Could not detect ROCm or CUDA platform. Please install.")
+    else:
+        cuda_path()
```
wangye805: Sorry I didn't think the logic through last time. My previous suggestion was actually worse as we still need two cuda_path() runs. 

I think your previous implementation is better as it follows your comments closely

tests/cpp/CMakeLists.txt
```diff
@@ -1,19 +1,55 @@
...
+  #Check for invalid values.
+  if(NOT NVTE_USE_ROCM MATCHES "^[0-9]+$")
+	  message(FATAL_ERROR "ValueError: Invalid value for NVTE_USE_ROCM: '${NVTE_USE_ROCM}'")
+  endif()
+  
+  math(EXPR NVTE_USE_ROCM_INT "${NVTE_USE_ROCM}")
+endif()
+
+#If NVTE_USE_ROCM>=1 or None, check for ROCm
+if (NOT DEFINED ENV{NVTE_USE_ROCM} OR "${NVTE_USE_ROCM_INT}" GREATER_EQUAL 1)
```
wangye805: Same as the rocm_build() function, let's revert it back to your previous logic which closely follow the comments

wangye805 (changes requested):

wangye805 (approved):
LGTM

Before merge, let's make sure we tested both with NVTE_USE_ROCM and without NVTE_USE_ROCM

build_tools/utils.py
```diff
@@ -165,23 +165,54 @@ def found_pybind11() -> bool:
...
         cuda_path()
         return False
     except FileNotFoundError:
         pass
 
-    _, hipcc_bin = rocm_path()
-    return hipcc_bin.is_file()
-
+    # If neither ROCm nor CUDA is detected, raise an error
+    raise FileNotFoundError("Could not detect ROCm or CUDA platform. Please install.")
```
ipanfilo: You can raise it on line 212 where catching FileNotFoundError from cuda. Also, please remove "Please install"

VeeraRajasekhar:
Ran the build, with and without NVTE_USE_ROCM.

build_tools/utils.py
```diff
@@ -165,23 +165,52 @@ def found_pybind11() -> bool:
...
     try:
         cuda_path()
         return False
     except FileNotFoundError:
-        pass
-
-    _, hipcc_bin = rocm_path()
-    return hipcc_bin.is_file()
-
+    # If neither ROCm nor CUDA is detected, raise an error
```
ipanfilo: Comment line should be aligned with the code below

tests/cpp/CMakeLists.txt
VeeraRajasekhar: Done.

tests/cpp/CMakeLists.txt
```diff
@@ -1,19 +1,34 @@
...
+set(USE CUDA OFF)
 
-if(((EXISTS "/opt/rocm/") OR (EXISTS $ENV{ROCM_PATH})) AND NOT (EXISTS "/bin/nvcc"))
-  message("AMD GPU detected.")
-  set(USE_ROCM ON)
-  set(USE_CUDA OFF)
+if (USE_ROCM)
+  #Check if ROCm exists
+  if( ((EXISTS "/opt/rocm/") OR (EXISTS $ENV{ROCM_PATH})) )
+    set(USE_ROCM ON)
```
ipanfilo: USE_ROCM is already set at this point. And please update message to AMD platform because we're not actually detecting GPU here

tests/cpp/CMakeLists.txt
```diff
@@ -1,19 +1,34 @@
...
 # Copyright (c) 2022-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # See LICENSE for license information.
 
 cmake_minimum_required(VERSION 3.18)
 
-option(USE_CUDA "Use CUDA" ON)
-option(USE_ROCM "Use ROCm" OFF)
+option(USE_ROCM "Use ROCm" ON)
+set(USE CUDA OFF)
```
ipanfilo: Please make USE_CUDA initialization matching common CMakeLists.txt

tests/cpp/CMakeLists.txt
```diff
@@ -1,19 +1,34 @@
...
-  message("AMD GPU detected.")
-  set(USE_ROCM ON)
-  set(USE_CUDA OFF)
+if (USE_ROCM)
+  #Check if ROCm exists
+  if( ((EXISTS "/opt/rocm/") OR (EXISTS $ENV{ROCM_PATH})) )
+    set(USE_ROCM ON)
+    message("AMD GPU detected")
+  else()
+    message(FATAL_ERROR "FileNotFoundError: Could not find HIPCC at '/opt/rocm'")
```
ipanfilo: FileNotFoundError is a name of python exception. No reason to use it in CMake. Just say that you cannot find ROCm installation

ipanfilo (approved):

PR #118, opened 2025-02-03
Issue#9690 - restore TE CUDA libary loading on init

ipanfilo:
# Description

Try to load CUDA libraries needed for TE work on NV platform but not fail initialization if those libraries are not found

Fixes #9690

## Type of change


wangye805 (approved):
LGTM but let's re-launch the CI to make sure it does not fail our pytests

PR #120, opened 2025-02-04
Fixed the hipify_torch error when run with 'setup.py install'

VeeraRajasekhar:
# Description

Fixed the hipify_torch error when run with 'setup.py install'.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

## Changes

- Changed build_tools/utils.py: Added the path to the library hipify_torch to sys.path.

# Checklist:

### DIFF
--- a/build_tools/utils.py
+++ b/build_tools/utils.py
@@ -417,11 +417,11 @@ def uninstall_te_wheel_packages():
     )
 
 def hipify(base_dir, src_dir, sources, include_dirs):
-    hipify_path = base_dir / "3rdparty" / "hipify_torch"
+    hipify_path = str(base_dir / "3rdparty" / "hipify_torch")
     cwd = os.getcwd()
-    os.chdir(hipify_path)
+    if hipify_path not in sys.path:
+        sys.path.insert(1, hipify_path)
     from hipify_torch.hipify_python import hipify as do_hipify
-    os.chdir(cwd)
 
     hipify_result = do_hipify(
         project_directory=src_dir,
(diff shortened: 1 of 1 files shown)


build_tools/utils.py
```diff
@@ -417,11 +417,11 @@ def uninstall_te_wheel_packages():
     )
 
 def hipify(base_dir, src_dir, sources, include_dirs):
-    hipify_path = base_dir / "3rdparty" / "hipify_torch"
+    hipify_path = str(base_dir / "3rdparty" / "hipify_torch")
     cwd = os.getcwd()
```
wenchenvincent, line 420: I think this line can also be removed.

build_tools/utils.py
```diff
@@ -417,11 +417,11 @@ def uninstall_te_wheel_packages():
...
-    os.chdir(hipify_path)
+    if hipify_path not in sys.path:
+        sys.path.insert(1, hipify_path)
```
wenchenvincent: Why insert at index 1?

build_tools/utils.py
VeeraRajasekhar: This cwd is used later in the code in #450 line.

build_tools/utils.py
VeeraRajasekhar: No major reason, just to give priority to the **current working directory** in case of pip install & **/root/TransformerEngine** in case of python setup.py install.

build_tools/utils.py
ipanfilo: But why need add base_dir/3rdparty/hipify_torch to serach path if base_dir is there. You should be able import 3rdparty.hipify_torch.hipify_torch.hipify_python w/o changing directory

VeeraRajasekhar: Yes, we can do that. Will change this and update

ipanfilo (approved):

PR #125, opened 2025-02-13
[PyTorch] Drop FA as an installation requirement (#1226)

ipanfilo:
# Description

Import changes that make standalone FA optional

Fixes # 
https://github.com/ROCm/frameworks-internal/issues/11110

## Type of change

## Changes

Cherry-pick upstream 161b1d9 and necessary perquisites 
Partial import of e762592 to simplify 161b1d9 merging to current baseline

# Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -117,7 +117,7 @@ def setup_requirements() -> Tuple[List[str], List[str], List[str]]:
     # Framework-specific requirements
     if (not rocm_build()) and (not bool(int(os.getenv("NVTE_RELEASE_BUILD", "0")))):
         if "pytorch" in frameworks:
-            install_reqs.extend(["torch", "flash-attn>=2.0.6,<=2.6.3,!=2.0.9,!=2.1.0"])
+            install_reqs.extend(["torch"])
             test_reqs.extend(["numpy", "onnxruntime", "torchvision", "prettytable"])
         if "jax" in frameworks:
             install_reqs.extend(["jax", "flax>=0.7.1"])
--- a/tests/pytorch/fused_attn/test_fused_attn.py
+++ b/tests/pytorch/fused_attn/test_fused_attn.py
@@ -23,9 +23,8 @@ from transformer_engine.pytorch.attention import (
     MultiheadAttention,
     RotaryPositionEmbedding,
     get_attention_backend,
-    _flash_attn_2_plus,
     _flash_attn_2_3_plus,
-    _flash_attn_3_plus,
+    _flash_attn_3_is_installed,
     check_set_window_size,
     AttentionParams,
     _attention_backends,
@@ -1380,7 +1379,7 @@ def test_mha_fp8_vs_f16(dtype, model, qkv_format, input_layernorm, fp8_dpa_bwd, 
     os.environ["NVTE_FP8_DPA_BWD"] = "1" if fp8_dpa_bwd else "0"
     config = model_configs_fp8_vs_f16[model]
 
-    if _flash_attn_3_plus and not is_training:
+    if _flash_attn_3_is_installed and not is_training:
         if RoPE:
             pytest.skip("Flash Attention doesn't support FP8 MHA with RoPE.")
         os.environ["NVTE_FLASH_ATTN"] = "1"
@@ -1408,7 +1407,7 @@ def test_mha_fp8_vs_f16(dtype, model, qkv_format, input_layernorm, fp8_dpa_bwd, 
     rtol = 5e-1
     rmse_tol = 0.15
     logging.debug("========== {:^25s} ==========".format("forward output"))
-    if _flash_attn_3_plus and not is_training:
+    if _flash_attn_3_is_installed and not is_training:
         _error(
             flash_attn_fwd_fp8,
             fused_attn_fwd_f16,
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -117,7 +117,7 @@ def setup_requirements() -> Tuple[List[str], List[str], List[str]]:
     # Framework-specific requirements
     if (not rocm_build()) and (not bool(int(os.getenv("NVTE_RELEASE_BUILD", "0")))):
         if "pytorch" in frameworks:
-            install_reqs.extend(["torch", "flash-attn>=2.0.6,<=2.6.3,!=2.0.9,!=2.1.0"])
```
wenchenvincent, line 120: If we don't make this change and don't have flash-attn pre-installed, then it would try to install the upstream repo of flash attention, right?

ipanfilo: It does not apply to us because of not rocm_build() at line 118. So this change is rather for consistency

wenchenvincent: Oh, does this apply to the other place at pytorch/setup.py?

tests/pytorch/fused_attn/test_fused_attn.py
```diff
@@ -23,9 +23,8 @@
     MultiheadAttention,
     RotaryPositionEmbedding,
     get_attention_backend,
-    _flash_attn_2_plus,
```
wenchenvincent, line 26: Why remove this one?

tests/pytorch/fused_attn/test_fused_attn.py
```diff
@@ -23,9 +23,8 @@
...
     _flash_attn_2_3_plus,
-    _flash_attn_3_plus,
+    _flash_attn_3_is_installed,
```
wenchenvincent, line 27: Is it necessary to change the name?

transformer_engine/pytorch/attention.py
```diff
@@ -89,67 +90,119 @@
...
+_flash_attn_2_4_1_plus = False
+_flash_attn_2_5_7_plus = False
+_flash_attn_2_6_plus = False
+_flash_attn_2_7_plus = False
+try:
+    _flash_attn_version = PkgVersion(get_pkg_version("flash-attn"))
+except PackageNotFoundError:
+    if (IS_HIP_EXTENSION or get_device_compute_capability() >= (8, 0)) and _NVTE_FLASH_ATTN:
+        fa_logger.debug(
+            "flash-attn v2 is not installed. To use, please install it by"
```
wenchenvincent, line 136: Are we able to install FA rocm by `pip install flash-attn`?

transformer_engine/pytorch/attention.py
```diff
@@ -89,67 +90,119 @@
...
+            """ "pip install flash-attn".""",
+        )
+else:
+    if _flash_attn_version_required <= _flash_attn_version <= _flash_attn_max_version:
+        from flash_attn.flash_attn_interface import flash_attn_func, flash_attn_varlen_func
+        from flash_attn.flash_attn_interface import (
+            _flash_attn_varlen_forward as _flash_attn_forward,
```
wenchenvincent, line 143: Why import _flash_attn_varlen_forward as _flash_attn_forward? Do we only use the varseq len backend?

transformer_engine/pytorch/setup.py
```diff
@@ -58,7 +58,7 @@
         description="Transformer acceleration library - Torch Lib",
         ext_modules=ext_modules,
         cmdclass={"build_ext": CMakeBuildExtension},
-        install_requires=[] if rocm_build() else ["torch", "flash-attn>=2.0.6,<=2.6.3,!=2.0.9,!=2.1.0"],
```
wenchenvincent, line 61: If we don't make this change and don't have flash-attn pre-installed, then it would try to install the upstream repo of flash attention, right?

setup.py
ipanfilo: All prerequisites here, in pytorc/setup.py and jax/setup.py are currently disabled for ROCm path because pip would install CUDA versions of them.  
They are, however, maintained to match CUDA TE build and to be re-enabled once ROCm versions are upsreamed

tests/pytorch/fused_attn/test_fused_attn.py
ipanfilo: It is not used.

tests/pytorch/fused_attn/test_fused_attn.py
ipanfilo: I attempted to make those changes to be compatible with what is in IFU. So to avoid conflicts in attention.py I picked some changes that are not strictly necessary for the purpose of PR but to match IFU. Name change is one of them

transformer_engine/pytorch/attention.py
ipanfilo: Good point. It tries to download src package and build ROCm version but fails due to missing CK. I think this is FA packaging error so probably the message should remain AS-IS and the error should be reported to FA

transformer_engine/pytorch/attention.py
ipanfilo: Yes, those methods are only used for CP and they are used despite actual varlen usage.

transformer_engine/pytorch/setup.py
ipanfilo: For CUDA build only

transformer_engine/pytorch/attention.py
wenchenvincent: Thanks for checking. Could you report this issue to FA?

wenchenvincent (approved):
LGTM.

PR #135, opened 2025-02-27
[ROCm] integrate ck varlen feature with THD layout into TE

wangye805:
# Description

Integrate the CK THD layout without padding between sequence into ROCM fused attn. 

Currently only the pytorch side can run since JAX THD layout requires padding between sequences

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

## Changes
Including softmax lse shape conversion kernels

# Checklist:

### DIFF
--- a/tests/pytorch/fused_attn/test_fused_attn.py
+++ b/tests/pytorch/fused_attn/test_fused_attn.py
@@ -657,25 +657,48 @@ model_configs_layout_thd = {
 }
 
 
+# ROCm fused attn does not use cudnn, return high numbers to avoid tests filtering out
 @pytest.mark.skipif(get_cudnn_version() < (9, 0, 0), reason="cuDNN 9.0.0+ is required.")
 @pytest.mark.skipif(
+    (not IS_HIP_EXTENSION) and
     get_device_compute_capability() < (9, 0), reason="THD is only supported on Hopper+."
 )
 @pytest.mark.parametrize("dtype", param_types_lean)
 @pytest.mark.parametrize("model_configs", [model_configs_layout_thd])
 @pytest.mark.parametrize("model", model_configs_layout_thd.keys())
 @pytest.mark.parametrize("qkv_layout", qkv_layouts_thd)
-def test_dpa_qkv_layout_thd(dtype, model_configs, model, qkv_layout):
+@pytest.mark.parametrize("pad_between_seqs", [False, True])
+def test_dpa_qkv_layout_thd(dtype, model_configs, model, qkv_layout, pad_between_seqs):
     """Test DotProductAttention module with different QKV layouts"""
-    pad_between_seqs = True
+    if (pad_between_seqs==False and get_cudnn_version() < (9, 3, 0)):
+        pytest.skip("cuDNN 9.3.0+ is required to run pad_between_seqs = False");
     test_dot_product_attention(
         dtype, model_configs, model, False, True, qkv_layout, False, pad_between_seqs
     )
-    if get_cudnn_version() >= (9, 3, 0):
-        # cuDNN 9.3.0+ is required to run pad_between_seqs = False/True in the same run
-        pad_between_seqs = False
+
+# ROCm specific pytests, not in upstream NVTE repo
+@pytest.mark.parametrize("dtype", param_types_lean)
+@pytest.mark.parametrize("model_configs", [model_configs_layout_thd])
+@pytest.mark.parametrize("model", model_configs_layout_thd.keys())
+@pytest.mark.parametrize("qkv_layout", qkv_layouts_thd)
+def test_dpa_qkv_layout_thd_mqa_gqa(dtype, model_configs, model, qkv_layout):
+    if (not IS_HIP_EXTENSION):
+        pytest.skip("ROCm specific pytests.");
+
+    def find_factors(x):
+        f = []
+        for i in range(2, x + 1):
+            if x % i == 0:
+                f.append(i)
+        return f
+
+    config = model_configs[model]
+    num_querys_per_gqa_group = find_factors(config.num_heads)
+
+    for num_q_per_gqa_group in num_querys_per_gqa_group:
+        config.num_gqa_groups = config.num_heads // num_q_per_gqa_group
         test_dot_product_attention(
-            dtype, model_configs, model, False, True, qkv_layout, False, pad_between_seqs
+            dtype, model_configs, model, False, True, qkv_layout, False, False
         )
 
 
--- a/tests/pytorch/fused_attn/test_fused_attn.py
+++ b/tests/pytorch/fused_attn/test_fused_attn.py
@@ -676,15 +676,12 @@ def test_dpa_qkv_layout_thd(dtype, model_configs, model, qkv_layout, pad_between
         dtype, model_configs, model, False, True, qkv_layout, False, pad_between_seqs
     )
 
-# ROCm specific pytests, not in upstream NVTE repo
+@pytest.mark.skipif(not IS_HIP_EXTENSION, reason="ROCm TE specific pytests.")
 @pytest.mark.parametrize("dtype", param_types_lean)
 @pytest.mark.parametrize("model_configs", [model_configs_layout_thd])
 @pytest.mark.parametrize("model", model_configs_layout_thd.keys())
 @pytest.mark.parametrize("qkv_layout", qkv_layouts_thd)
 def test_dpa_qkv_layout_thd_mqa_gqa(dtype, model_configs, model, qkv_layout):
-    if (not IS_HIP_EXTENSION):
-        pytest.skip("ROCm specific pytests.");
-
     def find_factors(x):
         f = []
         for i in range(2, x + 1):
--- a/tests/pytorch/test_numerics.py
+++ b/tests/pytorch/test_numerics.py
@@ -1824,36 +1824,29 @@ def test_transformer_layer_hidden_states_format(dtype, bs, model):
         attn_input_format="bshd",
     )
 
-    #TODO: release after rocm fused attn support var seq len features
-    if not IS_HIP_EXTENSION:
-        torch.manual_seed(0)
-        block_thd = TransformerLayer(
-            config.hidden_size,
-            4 * config.hidden_size,
-            config.num_attention_heads,
-            layernorm_epsilon=config.eps,
-            init_method=init_method,
-            output_layer_init_method=output_layer_init_method,
-            hidden_dropout=0,
-            attention_dropout=0,
-            kv_channels=config.embed,
-            params_dtype=dtype,
-            apply_residual_connection_post_layernorm=False,
-            output_layernorm=False,
-            device="cuda",
-            attn_input_format="thd",
-            self_attn_mask_type="padding_causal",
-        )
+    torch.manual_seed(0)
+    block_thd = TransformerLayer(
+        config.hidden_size,
+        4 * config.hidden_size,
+        config.num_attention_heads,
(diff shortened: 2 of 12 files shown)


tests/pytorch/fused_attn/test_fused_attn.py
```diff
@@ -657,25 +657,48 @@ def test_dpa_qkv_layout(dtype, model_configs, model, qkv_layout):
...
+    if (pad_between_seqs==False and get_cudnn_version() < (9, 3, 0)):
+        pytest.skip("cuDNN 9.3.0+ is required to run pad_between_seqs = False");
     test_dot_product_attention(
         dtype, model_configs, model, False, True, qkv_layout, False, pad_between_seqs
     )
-    if get_cudnn_version() >= (9, 3, 0):
-        # cuDNN 9.3.0+ is required to run pad_between_seqs = False/True in the same run
-        pad_between_seqs = False
+
+# ROCm specific pytests, not in upstream NVTE repo
```
ipanfilo: Better put skipif here. And remove wording about upstream - on the one hand there is a lot of other code not in upstream, on the other hand, when eventually the changes merged to upstream this wording will be misleading

tests/pytorch/fused_attn/test_fused_attn.py
```diff
@@ -657,25 +657,48 @@ def test_dpa_qkv_layout(dtype, model_configs, model, qkv_layout):
...
 @pytest.mark.skipif(
+    (not IS_HIP_EXTENSION) and
     get_device_compute_capability() < (9, 0), reason="THD is only supported on Hopper+."
 )
 @pytest.mark.parametrize("dtype", param_types_lean)
 @pytest.mark.parametrize("model_configs", [model_configs_layout_thd])
 @pytest.mark.parametrize("model", model_configs_layout_thd.keys())
 @pytest.mark.parametrize("qkv_layout", qkv_layouts_thd)
-def test_dpa_qkv_layout_thd(dtype, model_configs, model, qkv_layout):
+@pytest.mark.parametrize("pad_between_seqs", [False, True])
```
ipanfilo, line 670: What is difference in behaviour - both old and new code run test_dot_product_attention() with pad_between_seq False and True

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -114,6 +114,9 @@ def is_fused_attn_kernel_available(self):
 
     def get_fused_attn_backend(self):
         """Get the fused attention kernel backend"""
+        # temporary hack to disable thd in rocm fused attn as no pad_between_sequence
```
ipanfilo, line 117: Doesn't this call go down to nvte_get_fused_attn_backend()? Can it be filtered there instead of hack here?

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -34,37 +35,76 @@ bool is_ck_backend_supported(
...
+    if(nvte_log_ck_config){
+      std::cout<<"CK fused attn does not support pre_scale bias"<<std::endl;
+    }
+    return false;
+  }
+
+  const int device_id = cuda::current_device();
+  const std::string sm_arch_name_ = cuda::sm_arch_name(device_id);
+  //only MI300X supported
+  if(!(sm_arch_name_.find("gfx942")!=std::string::npos)){
```
ipanfilo, line 85: Can cuda::sm_arch(device_id)==94 be used instead?

tests/pytorch/test_numerics.py
```diff
@@ -1824,36 +1824,29 @@ def test_transformer_layer_hidden_states_format(dtype, bs, model):
...
+        output_layer_init_method=output_layer_init_method,
+        hidden_dropout=0,
+        attention_dropout=0,
+        kv_channels=config.embed,
+        params_dtype=dtype,
+        apply_residual_connection_post_layernorm=False,
+        output_layernorm=False,
+        device="cuda",
+        attn_input_format="thd",
+        self_attn_mask_type="padding_causal",
```
wenchenvincent, line 1843: Do we support "padding_causal" with this PR now?

transformer_engine/common/ck_fused_attn/include/ck_fused_attn/ck_fused_attn.hpp
```diff
@@ -15,22 +15,22 @@
 namespace ck_fused_attn{
 
 // input qkv dtypes
-enum DType {
+enum class DType {
```
wenchenvincent, line 18: Why do we need to change to class?

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_fwd.cpp
```diff
@@ -17,6 +17,102 @@
...
+  float* lse){
+
+  int32_t total_seqlen_q = cu_seqlen_q[b];
+  
+  for(uint64_t bh_idx = blockIdx.x*blockDim.x + threadIdx.x; bh_idx < b*h; bh_idx += blockDim.x * gridDim.x){
+    uint64_t b_idx = bh_idx/h;
+    uint64_t h_idx = bh_idx%h;
+    // loop over a given batch and head idx
+    for(uint64_t s_idx = cu_seqlen_q[b_idx]; s_idx < cu_seqlen_q[b_idx + 1]; s_idx++){
+      lse[bh_idx * s_q + s_idx - cu_seqlen_q[b_idx]] = lse_thd[h_idx * total_seqlen_q + s_idx];
```
wenchenvincent, line 34: For those padding positions in lse, do they need to be a specific value? Is the buffer initialized somewhere?

tests/pytorch/fused_attn/test_fused_attn.py
wangye805: I include this parameter "pad_between_seqs" on purpose, to split the previous one testcase with both values to two testcases. 

With the previous code, we first test when pad_between_seqs=True. In https://github.com/ROCm/TransformerEngine/blob/fc4e9c9683fd30a2ad895657b1483437edc1a8df/tests/pytorch/fused_attn/test_fused_attn.py#L288, if we found less than two backens, we will skip the entire pytest testcase, including the following one when pad_between_seq = False. 

With my code change, this will not happen

tests/pytorch/fused_attn/test_fused_attn.py
wangye805: Done. Thanks

tests/pytorch/test_numerics.py
wangye805: Yes, padding and causal are independent in CK

transformer_engine/common/ck_fused_attn/include/ck_fused_attn/ck_fused_attn.hpp
wangye805: enum class can support forward declaration but enum does not. I used those forward declaration in the corresponding header: https://github.com/ROCm/TransformerEngine/blob/3060d5c59012762f0d4230a6e06a912d3a2652c0/transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.hpp#L45-L50

transformer_engine/jax/cpp_extensions/attention.py
wangye805: The nvte_get_fused_attn_backend() does not contain param pad_between_seq. In this PR, I only support the true "THD" qkv layout with padding mask, which corresponds to the scenario that "pad_between_seq=False". 

In pytorch, we have a get_attention_backend() function which has this param. However, in jax, currently qkv tensors are always BSHD, which means that they are always "pad_between_seq=true". And people still pass in qkv layout as THD. So I have to do this hack

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
wangye805: Not sure whether gfx940 or gfx941 can be supported. Previously CK told me they only support gfx942

ipanfilo: GFX940/941 are MI300 too. I do not know if they are really used to bother about them. In any case please change wording from MI300X to MI300 - because gfx942 (or GFX9.4) is whole MI300 family not only MI300X.

tests/pytorch/test_numerics.py
```diff
@@ -1896,25 +1889,26 @@ def test_transformer_layer_hidden_states_format(dtype, bs, model):
...
-            torch.manual_seed(0)
-            y_thd = block_thd(
-                x_thd,
-                cu_seqlens_q=x_thd_cumsum,
-                cu_seqlens_kv=x_thd_cumsum,
-                max_seqlen_q=config.seq_len,
-                max_seqlen_kv=config.seq_len,
-            )
+    # in ROCm TE, THD only supported with ck backend
+    if IS_HIP_EXTENSION and os.getenv("NVTE_FUSED_ATTN_CK", "1") == "0":
```
ipanfilo: Use rocm_attn_backend() instead of direct env checking

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_fwd.cpp
wangye805: Those padded locations are not used in lse. But from math point of view, we should initialize them -inf because log sum(exp(-inf)) is still -inf

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
wenchenvincent: The MI300 program asked us to remove gfx940 and gfx941 targets as we will only have gfx942 in production now. Yes, gfx942 is for the whole MI300 family.

wenchenvincent: > Not sure whether gfx940 or gfx941 can be supported. Previously CK told me they only support gfx942

Do they support gfx90a?

tests/pytorch/test_numerics.py
wenchenvincent: But CK kernels do not support "padding" or "padding_causal", right? Why do we enable it in this PR?

tests/pytorch/test_numerics.py
wangye805: Done. Thanks

tests/pytorch/test_numerics.py
wangye805: In fact, all variable sequence length related in TE must be with "padding". Even the traditional thd without padding still requires "padding" mask.

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
wangye805: I suppose they should support gfx90a as CK example support it. But I didn't compile it under gfx90a

ipanfilo: So GFX 9.4 is effectively the same as "gfx942". But I see that you didn't add this piece of code but moved it here from below. What is a reason to move it here?

wangye805: Oh, I found those filter conditions are not organized well previously. So this time I would like to re-organize them. Single conditions/filters come first, then joint conditions/filters.

ipanfilo (approved):

wenchenvincent (approved):
LGTM.

PR #149, opened 2025-03-17
[ROCm] integrate rmsnorm_bwd from Xiaohu's PR and add the same cpp unit tests

brunomazzottiamd:
# Description

This PR adds RMSNorm backward Triton kernels (developed by Xiaohu) to Transformer Engine.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

## Changes

- Change A
- Change B

# Checklist:


wangye805 (approved):
LGTM

tests/pytorch/triton_kernels/test_rmsnorm_triton.py
```diff
@@ -43,58 +39,155 @@ def get_tolerances(in_dtype):
...
     ln_out_triton, rsigma_triton = te_rmsnorm_fwd_fp8_noalloc_triton(input_tensor, gamma_tensor, epsilon, ln_out_triton, out_dtype, zero_centered_gamma)
-    # run the reference hipified kernel path
+
+    # run the fwd reference hipified kernel path
+    # dummy fp8 meta
     scale_tensor = torch.empty(0, dtype=torch.float32, device='cuda')
     amax_tensor = torch.zeros(0, dtype=torch.float32, device='cuda')
     scale_inv_tensor = torch.empty(0, dtype=torch.float32, device='cuda')
-
     fwd_ln_sm_margin = int(os.getenv("NVTE_FWD_LAYERNORM_SM_MARGIN", "0"))
```
brunomazzottiamd: @wangye805, I have a question bout this environment variable:

Is `NVTE_FWD_LAYERNORM_SM_MARGIN` the correct name? Should we use `NVTE_FWD_RMSNORM_SM_MARGIN` instead of `NVTE_FWD_LAYERNORM_SM_MARGIN`. I'm not sure what's the best option since I'm not used to TE codebase.

Thanks!

wangye805: Emm, those envs are named by NV. It's tricky that they use the same envs for layernorm and rmsnorm. We would better to keep the same name for compatibility. 

But I did make a mistake here. We have both NVTE_FWD_LAYERNORM_SM_MARGIN and NVTE_BWD_LAYERNORM_SM_MARGIN:
https://github.com/ROCm/TransformerEngine/blob/b8b92dcc1e4c3a499a56759b21f677dcf7b67b7e/transformer_engine/pytorch/module/rmsnorm.py#L172-L174

By the way, do we have similar feature to limit # of SMs in triton?

brunomazzottiamd: Thanks for your response. Let's keep the NV names for the sake of compatibility. I can fix this PR, to use the correct environment variable in backward pass.

This feature isn't implemented on Triton side, but I think it's not that hard to do it. We can tweak this function:

```python
def get_num_sms():
    current_device_index = torch.cuda.current_device()
    current_device = torch.cuda.get_device_properties(current_device_index)
    num_sms = current_device.multi_processor_count
    return num_sms
```

Something like this: 

```python
def get_num_sms(sm_margin=None):
    current_device_index = torch.cuda.current_device()
    current_device = torch.cuda.get_device_properties(current_device_index)
    num_sms = current_device.multi_processor_count
    if sm_margin is not None and sm_margin > 0:
        num_sms = max(num_sms - sm_margin, 1)
    return num_sms
```

Does it make sense to you?

wangye805: Thanks for the updated get_num_sms function proposal. However, even if we update this get_num_sms with sm_margin, triton kernel launch will still take all available SMs (CUs in rocm). Is my understanding correct? If so, perhaps we don't need that...

brunomazzottiamd: Yes, you are correct. We also need to add `sm_margin` argument to the Triton wrappers and propagate it to `get_num_sms` function. I'm baking an implementation of this feature and testing it right now. I'll publish a new commit soon, so you can take a look and let me know what you think about it.

brunomazzottiamd: Ye, please take a look at my last commit: e3a99f03360df1feed54269b86b999b4150d0a6c. I think it implements "SM margin" feature on Triton side. I executed the RMSNorm test suite with distinct values of "SM margin" and everything is passing. On MI300X, no "SM margin" took around 4 min 45 sec to run while "SM margin = 300" (only 4 CUs available) took round 9 min to run. I think it's working as expected.

wangye805 (approved):
LGTM

PR #150, opened 2025-03-17
[ROCm] make ck float to bf16 conversion as a compile-time option

wangye805:
# Description

With default value 3, unchanged.

Fixes # (issue)

## Type of change

- [x] Infra/Build change
- [ ] Code refractor

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,6 +73,7 @@ def setup_common_extension() -> CMakeExtension:
         if os.getenv("NVTE_AOTRITON_PATH"):
             aotriton_path = Path(os.getenv("NVTE_AOTRITON_PATH"))
             cmake_flags.append(f"-DAOTRITON_PATH={aotriton_path}")
+        cmake_flags.append(f"-DCK_FUSED_ATTN_FLOAT_TO_BFLOAT16_DEFAULT={os.getenv('NVTE_CK_FUSED_ATTN_FLOAT_TO_BFLOAT16_DEFAULT', 3)}")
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -73,6 +73,7 @@ def setup_common_extension() -> CMakeExtension:
         if os.getenv("NVTE_AOTRITON_PATH"):
             aotriton_path = Path(os.getenv("NVTE_AOTRITON_PATH"))
             cmake_flags.append(f"-DAOTRITON_PATH={aotriton_path}")
+        cmake_flags.append(f"-DCK_FUSED_ATTN_FLOAT_TO_BFLOAT16_DEFAULT={os.getenv('NVTE_CK_FUSED_ATTN_FLOAT_TO_BFLOAT16_DEFAULT', 3)}")
```
ipanfilo, line 76: We have 3 defined in several places - here and in ck_fused_attn/CMakeLists.txt
I think it might be better to have one place with default and only pass an option to CMake if env is set

wangye805: Thanks. I removed the default value definition in ck_fused_attn/CMakeLists.txt

ipanfilo: Will it work if cmake called not from setup.py then?

wangye805: Yeah, if we are compiling libck_fused_attn.a, we can add -DCK_FUSED_ATTN_FLOAT_TO_BFLOAT16_DEFAULT=3 to the cmake command

ipanfilo (approved):

PR #153, opened 2025-03-20
[Fix] hipblaslt_gemm handle unsafe usage on multi stream scene

RuibinCheung:
# Description

- Fix hipblaslt_gemm handle unsafe usage on multi stream scene.
- Update .gitignore to ignore hip files.
- Update gemm workspace from 32MB to 76MB (which is recommended by hipblaslt team.).

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

## Changes

- Fix hipblaslt_gemm handle unsafe usage on multi stream scene.
- Update .gitignore to ignore hip files.
- Update gemm workspace from 32MB to 76MB (which is recommended by hipblaslt team.).

# Checklist:

### DIFF
--- a/transformer_engine/common/gemm/cublaslt_gemm.cu
+++ b/transformer_engine/common/gemm/cublaslt_gemm.cu
@@ -60,12 +60,13 @@ namespace transformer_engine {
 
 #ifdef __HIP_PLATFORM_AMD__
 //Forward declaration. The implementation is in rocm_gemm.cu
+// stream_offset -1 means torch default stream.
 void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
                  const Tensor *inputBias, Tensor *outputPreGelu, int m, int n, int k, int lda,
                  int ldb, int ldd, bool transa, bool transb, bool grad,
                  void* workspace, size_t workspaceSize, bool accumulate, bool use_split_accumulator,
                  int math_sm_count, int m_split, int n_split, bool gemm_producer,
-                 const Tensor *inputCounter, hipStream_t stream);
+                 const Tensor *inputCounter, hipStream_t stream, int stream_offset = -1);
 #else // Use cublasLt
 void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
                  const Tensor *inputBias, Tensor *outputPreGelu, int m, int n, int k, int lda,
@@ -314,6 +315,77 @@ static void init_streams_and_events() {
 
 }  // namespace transformer_engine
 
+#ifdef __HIP_PLATFORM_AMD__
+// stream_offset = -1 means torch default stream
+void nvte_cublas_gemm_with_stream_offset(const NVTETensor A, const NVTETensor B, NVTETensor D, const NVTETensor bias,
+                      NVTETensor pre_gelu_out, bool transa, bool transb, bool grad,
+                      NVTETensor workspace, bool accumulate, bool use_split_accumulator,
+                      int math_sm_count, cudaStream_t stream, int stream_offset) {
+  NVTE_API_CALL(nvte_cublas_gemm_with_stream_offset);
+
+  using namespace transformer_engine;
+  const Tensor *inputA = reinterpret_cast<const Tensor *>(A);
+  const Tensor *inputB = reinterpret_cast<const Tensor *>(B);
+  Tensor *outputD = reinterpret_cast<Tensor *>(D);
+  const Tensor *biasTensor = reinterpret_cast<const Tensor *>(bias);
+  Tensor *outputGelu = reinterpret_cast<Tensor *>(pre_gelu_out);
+  Tensor *wspace = reinterpret_cast<Tensor *>(workspace);
+
+  const int m = transa ? inputA->data.shape[0] : inputA->data.shape[1];
+  const int k = transa ? inputA->data.shape[1] : inputA->data.shape[0];
+  const int n = transb ? inputB->data.shape[1] : inputB->data.shape[0];
+  int lda, ldb, ldd;
+  if (transa && !transb) {  // TN
+    lda = k;
+    ldb = k;
+    ldd = m;
+  } else if (!transa && !transb) {  // NN
+    lda = m;
+    ldb = k;
+    ldd = m;
+  } else if (!transa && transb) {  // NT
+    lda = m;
+    ldb = n;
+    ldd = m;
+  } else {  // TT
+    NVTE_ERROR("TT layout not allowed.");
+  }
+
+  bool nvte_log_gemm_config = false;
+  if (const char* env_p = std::getenv("NVTE_LOG_GEMM_CONFIG") ) {
+    if (env_p != nullptr && std::string(env_p) == "1")
+      nvte_log_gemm_config = true;
+  }
+
+  if (nvte_log_gemm_config) {
+    float A_scale_inv, B_scale_inv;
+    (void)cudaMemcpy(&A_scale_inv, inputA->scale_inv.dptr, sizeof(float), cudaMemcpyDeviceToHost);
+    (void)cudaMemcpy(&B_scale_inv, inputB->scale_inv.dptr, sizeof(float), cudaMemcpyDeviceToHost);
+    std::cout << "m=" << m << " k=" << k << " n=" << n 
+        << " transa=" << (transa?"T":"N")
+        << " transb=" << (transb?"T":"N")
+        << " A_type=" << (int)inputA->data.dtype
+        << " B_type=" << (int)inputB->data.dtype
+        << " D_type=" << (int)outputD->data.dtype
+        << " bias_type=" << (int)biasTensor->data.dtype
+        << " grad=" << grad
+        << " bias=" << (biasTensor->data.dptr != nullptr)
+        << " gelu=" << (outputGelu->data.dptr != nullptr)
+        << " use_fp8=" << ( is_fp8_dtype(inputA->data.dtype) || is_fp8_dtype(inputB->data.dtype) )
+        << " A_scale_inverse = " <<  A_scale_inv
+        << " B_scale_inverse = " <<  B_scale_inv
+        << " accumulate=" << accumulate
+        << std::endl;
+  }
+
+  cublas_gemm(inputA, inputB, outputD,  biasTensor, outputGelu, m, n, k, lda, ldb, ldd,
+              transa, transb,
+              grad,
+              wspace->data.dptr, wspace->data.shape[0], accumulate, use_split_accumulator,
+              math_sm_count, 0, 0, false, nullptr, stream, stream_offset);
+}
+#endif // __HIP_PLATFORM_AMD__
+
 void nvte_cublas_gemm(const NVTETensor A, const NVTETensor B, NVTETensor D, const NVTETensor bias,
                       NVTETensor pre_gelu_out, bool transa, bool transb, bool grad,
                       NVTETensor workspace, bool accumulate, bool use_split_accumulator,
@@ -459,9 +531,15 @@ void nvte_multi_stream_cublas_gemm(const NVTETensor *A, const NVTETensor *B, NVT
   }
 
   for (int i = 0; i < num_gemms; i++) {
+#ifdef __HIP_PLATFORM_AMD__
+    nvte_cublas_gemm_with_stream_offset(A[i], B[i], D[i], bias[i], pre_gelu_out[i], transa, transb, grad,
+                                        workspace[i % num_streams], accumulate, use_split_accumulator, math_sm_count,
+                                        compute_streams[i % num_streams], i % num_streams);
+#else
     nvte_cublas_gemm(A[i], B[i], D[i], bias[i], pre_gelu_out[i], transa, transb, grad,
                      workspace[i % num_streams], accumulate, use_split_accumulator, math_sm_count,
                      compute_streams[i % num_streams]);
+#endif // __HIP_PLATFORM_AMD__
   }
 
   // record events on compute streams
--- a/transformer_engine/common/gemm/cublaslt_gemm.cu
+++ b/transformer_engine/common/gemm/cublaslt_gemm.cu
@@ -314,7 +314,7 @@ static void init_streams_and_events() {
 
 }  // namespace transformer_engine
 
-// -1 means the stream from outer rather than compute_streams
+// compute_stream_offset = -1 means the stream from outer rather than compute_streams
 static void cublas_gemm_ex(const NVTETensor A, const NVTETensor B, NVTETensor D, const NVTETensor bias,
                            NVTETensor pre_gelu_out, bool transa, bool transb, bool grad,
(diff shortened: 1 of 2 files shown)


transformer_engine/pytorch/module/base.py
```diff
@@ -52,6 +55,10 @@
 
 
 def get_cublas_workspace_size_bytes() -> None:
+    """Return 75MiB if using AMD GPU."""
+    if IS_HIP_EXTENSION:
+        return 79_691_776
```
wangye805: Why do we need 75MB for AMD GPU?

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -403,126 +403,150 @@ namespace {
...
   {
-    return pool.size();
+    return _pool.size();
   }
 
 private:
-  std::mutex mt;
-  using Pool = std::vector<std::forward_list<hipblasLtHandle_t>>;
+  std::mutex _mt;
+  using Pool = std::vector<std::unordered_map<hipStream_t, hipblasLtHandle_t>>;
```
wangye805: I don't quite understand why the pool need stream id as part of the key. One handle released from one stream id can definitely reused by another stream id.

wangye805 (changes requested):
I took a look at how pytorch is handling this handle pool issue. It seems that they are using unique workspace for given stream id and handle id: https://github.com/pytorch/pytorch/blob/2c4bc65366a2cfce00fbddb2efa19e7b337e9b60/aten/src/ATen/cuda/CublasHandlePool.cpp#L202-L217

Can we first understand the root cause of this group-gemm bug? Was it because we need to use different handle for different stream id? Or different workspace for different (handle, stream) pair but handle can be shared across stream ids?

ipanfilo (changes requested):
Most of PR changes are not related to the problem. 
Only rocm_gemm.cu needs changes and it does not need object and variables renamings that just create noise that do not help reviewing.
Please limit PR to necessary changes only

RuibinCheung:
> I took a look at how pytorch is handling this handle pool issue. It seems that they are using unique workspace for given stream id and handle id: https://github.com/pytorch/pytorch/blob/2c4bc65366a2cfce00fbddb2efa19e7b337e9b60/aten/src/ATen/cuda/CublasHandlePool.cpp#L202-L217
> 
> Can we first understand the root cause of this group-gemm bug? Was it because we need to use different handle for different stream id? Or different workspace for different (handle, stream) pair but handle can be shared across stream ids?

You are right. The handle is related to workspace. When it set **workspace size to 0** the bug can't be reproduced. But I think we can't align to PyTorch because calling `hipMalloc` directly is not acceptable it will hinder PyTorch memory allocator manage device memory.

The CUDA create handle on stack when calling `cublas_gemm` every time. I think that align to CUDA is an acceptable way.

RuibinCheung:
> Most of PR changes are not related to the problem. Only rocm_gemm.cu needs changes and it does not need object and variables renamings that just create noise that do not help reviewing. Please limit PR to necessary changes only

Done.

transformer_engine/common/gemm/rocm_gemm.cu
RuibinCheung: ACK. Thx.

transformer_engine/pytorch/module/base.py
RuibinCheung: The 32MB and 4MB is recomended by cuBLASLt. This 76MB is recommended by hipBLASLt team. 
I have do some benchmark and I found it will improve about 3-5% performance on gemm config of our model.

RuibinCheung: I have moved this change to another PR(https://github.com/ROCm/TransformerEngine/pull/155).

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -414,15 +414,15 @@ public:
...
+    
     if (!pool[device_id].empty())
     {
       hipblasLtHandle_t h = pool[device_id].front();
       pool[device_id].pop_front();
       return h;
     }
 
-    return nullptr;
+      return nullptr;
```
ipanfilo: Wrong align

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -435,18 +435,20 @@ public:
...
       std::cout << "[ERROR] Attempt to store handles to invalid pool" << std::endl;
     }
-    for (unsigned int i=0; i<pool.size(); i++)
+    for (unsigned int device_id=0; device_id<pool.size(); device_id++)
     {
-      if (handles[i] != nullptr)
+      for (auto it = handles[device_id].begin(); it != handles[device_id].cend(); it++)
       {
-        pool[i].push_front(handles[i]);
+        if (it->second != nullptr) {
```
ipanfilo: Why nullptr entry would be added to the map?

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -435,18 +435,20 @@ public:
     return h;
   }
 
-  void store(const std::vector<hipblasLtHandle_t>& handles)
+  void store(const std::vector<std::unordered_map<hipStream_t, hipblasLtHandle_t>>& handles)
```
ipanfilo: When handles go to/from pool it is better to call synchronize on corresponding stream because it is not guaranteed that client had waited for execution completion before destroying the thread.
It is undesired to call long HIP methods under lock_guard.
It may also require special handling if synchronize is called on store when Pool is destroyed before HandleCache (see DESTROY_HIPBLASLT_HANDLES_POOL in code).

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -487,30 +489,49 @@ private:
 
 thread_local static class HandleCache {
 public:
-  hipblasLtHandle_t get(int device_id) const
+  hipblasLtHandle_t get(int device_id, hipStream_t stream) const
   {
-    return d.empty() ? nullptr : d[device_id];
```
ipanfilo: Since streams are per device, vector is not needed - single stream:handle container is sufficient

transformer_engine/common/gemm/rocm_gemm.cu
RuibinCheung: ACK. Thx.

transformer_engine/common/gemm/rocm_gemm.cu
RuibinCheung: ACK. Thx.

transformer_engine/common/gemm/rocm_gemm.cu
RuibinCheung: Removed. Thx.

transformer_engine/common/gemm/rocm_gemm.cu
RuibinCheung: ACK. Thx.

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -314,6 +315,77 @@ static void init_streams_and_events() {
 
 }  // namespace transformer_engine
 
+#ifdef __HIP_PLATFORM_AMD__
+// stream_offset = -1 means torch default stream
+void nvte_cublas_gemm_with_stream_offset(const NVTETensor A, const NVTETensor B, NVTETensor D, const NVTETensor bias,
```
wangye805: No need to add an extra function definition which is quite similar to hipblaslt_gemm in rocm_gemm.cu. 

Since we already add static (global) variable hipblaslt_handles, we can add one two more static global variables, bool is_grouped_gemm, and stream_offset to make hipblaslt_gemm goes another route in the grouped gemm path

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -459,9 +531,15 @@ void nvte_multi_stream_cublas_gemm(const NVTETensor *A, const NVTETensor *B, NVT
   }
 
   for (int i = 0; i < num_gemms; i++) {
+#ifdef __HIP_PLATFORM_AMD__
```
wangye805: Also need to guard by USE_HIPBLASLT?

transformer_engine/common/gemm/cublaslt_gemm.cu
RuibinCheung: The static global variables only can be accessed on the same cpp file. We can use it to route the different gemm path.

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -374,15 +376,34 @@ void nvte_cublas_gemm(const NVTETensor A, const NVTETensor B, NVTETensor D, cons
         << std::endl;
   }
 
-  cublas_gemm(inputA, inputB, outputD,  biasTensor, outputGelu, m, n, k, lda, ldb, ldd,
 #ifdef __HIP_PLATFORM_AMD__
+  cublas_gemm(inputA, inputB, outputD,  biasTensor, outputGelu, m, n, k, lda, ldb, ldd,
```
ipanfilo: Just guard new parameter not whole method call

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -60,12 +60,13 @@ namespace transformer_engine {
 
 #ifdef __HIP_PLATFORM_AMD__
 //Forward declaration. The implementation is in rocm_gemm.cu
+// stream_offset -1 means torch default stream.
```
ipanfilo: Please rephrase, it is not torch API

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -314,11 +315,12 @@ static void init_streams_and_events() {
 
 }  // namespace transformer_engine
 
-void nvte_cublas_gemm(const NVTETensor A, const NVTETensor B, NVTETensor D, const NVTETensor bias,
+void nvte_cublas_gemm_ex(const NVTETensor A, const NVTETensor B, NVTETensor D, const NVTETensor bias,
```
ipanfilo: Make it static to clearly indicate it is for internal use only. And maybe make it not nvte_* but cublas_gemm_ex

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -314,11 +315,12 @@ static void init_streams_and_events() {
...
                       NVTETensor pre_gelu_out, bool transa, bool transb, bool grad,
                       NVTETensor workspace, bool accumulate, bool use_split_accumulator,
-                      int math_sm_count, cudaStream_t stream) {
-  NVTE_API_CALL(nvte_cublas_gemm);
+                      int math_sm_count, cudaStream_t stream, int stream_offset = -1) {
+  NVTE_API_CALL(nvte_cublas_gemm_ex);
```
ipanfilo: Not needed - it is only called from other NVTE API

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1004,6 +1004,16 @@ static inline int getIntEnv(const char *name, int defval, int minval)
 
 } //namespace
 
+static std::once_flag init_flag;
+static hipblasLtHandle_t hipblaslt_handles[num_streams];
+
+// Warning: only call once per device!
```
ipanfilo: Put comment that if follows multistrteam logic in cublas_gemm.

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1820,14 +1832,20 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
 #ifdef USE_HIPBLASLT
   if (use_hipblaslt || !use_rocblas)
   {
+    // Init hipblaslt handles (once, globally)
+    std::call_once(init_flag, init_hipblaslt_handles);
```
ipanfilo: Do it only if stream_offset is valid and not -1

transformer_engine/common/gemm/cublaslt_gemm.cu
RuibinCheung: ACK. Thx.

transformer_engine/common/gemm/cublaslt_gemm.cu
RuibinCheung: ACK. Thx.

transformer_engine/common/gemm/cublaslt_gemm.cu
RuibinCheung: ACK. Thx.

transformer_engine/common/gemm/cublaslt_gemm.cu
RuibinCheung: ACK. Thx.

transformer_engine/common/gemm/rocm_gemm.cu
RuibinCheung: ACK. Thx.

transformer_engine/common/gemm/rocm_gemm.cu
RuibinCheung: ACK. Thx.

transformer_engine/common/gemm/cublaslt_gemm.cu
RuibinCheung: ACK. Thx.

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -314,11 +314,11 @@ static void init_streams_and_events() {
 
 }  // namespace transformer_engine
 
-void nvte_cublas_gemm(const NVTETensor A, const NVTETensor B, NVTETensor D, const NVTETensor bias,
-                      NVTETensor pre_gelu_out, bool transa, bool transb, bool grad,
-                      NVTETensor workspace, bool accumulate, bool use_split_accumulator,
-                      int math_sm_count, cudaStream_t stream) {
-  NVTE_API_CALL(nvte_cublas_gemm);
+// -1 means the stream from outer rather than compute_streams
```
ipanfilo: Comment. What -1?

transformer_engine/common/gemm/rocm_gemm.cu
ipanfilo: Not fixed. And for code cleanness, please move corresponding variables and do handles initialization here

ipanfilo (changes requested):

transformer_engine/common/gemm/cublaslt_gemm.cu
RuibinCheung: ACK. Thx.

transformer_engine/common/gemm/rocm_gemm.cu
RuibinCheung: ACK. Thx.

ipanfilo (approved):

wangye805 (approved):
LGTM

PR #156, opened 2025-03-24
[Feat] optim TE_HIPBLASLT_ALGO_LOAD logic

RuibinCheung:
# Description

We try to use hipblaslt-bench to do tuning offline and we found the best algo id of hipblaslt-bench generally not be included in the return value of `hipblasLtMatmulAlgoGetHeuristic` when algo count is small.  So it cannot apply the best solution on runtime.

Generally we think the best algo id from hipblaslt-bench is better than TE auto tuning because of hipblaslt-bench owns more extensive search scope. 

We hope to apply the best algo id directly on TE to avoid tuning again.

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/common/gemm/rocm_gemm.cu
+++ b/transformer_engine/common/gemm/rocm_gemm.cu
@@ -1184,11 +1184,8 @@ void hipblaslt_gemm(const Tensor *inputA,
           ws_size_min
         )) {
           cached_algo.algo = algo_arr[0].algo;
-          if (ws_size_min != cached_algo.ws_size_min)
-          {
-            cached_algo.ws_size_min = ws_size_min;
-            algoCache.store(gemm_cfg, cached_algo);
-          }
+          cached_algo.ws_size_min = ws_size_min;
+          algoCache.store(gemm_cfg, cached_algo);
         }
       }
       
--- a/transformer_engine/common/gemm/rocm_gemm.cu
+++ b/transformer_engine/common/gemm/rocm_gemm.cu
@@ -1161,7 +1161,6 @@ void hipblaslt_gemm(const Tensor *inputA,
   {
     bool logTuning = getIntEnv("TE_HIPBLASLT_LOG_TUNING", 0, 0) != 0;
 
-#if HIP_VERSION >= 60000000
     // Find algo base algo_id directly if tuning file is set.
     if (cached_algo.hasId())
     {
@@ -1184,15 +1183,15 @@ void hipblaslt_gemm(const Tensor *inputA,
           ws_size_min
         )) {
           cached_algo.algo = algo_arr[0].algo;
+          cached_algo.ws_size_min = algo_arr[0].workspaceSize;
           algoCache.store(gemm_cfg, cached_algo);
         }
       }
-      
+
       if (logTuning && !cached_algo.algo.has_value()) {
         std::cout << "[WARNING] Cannot get corresponding solution from cached algoId " << cached_algo.algoId << std::endl;
       }
     }
-#endif // #if HIP_VERSION >= 60000000
 
     int firstAlgo = getIntEnv("TE_HIPBLASLT_ALGO_SELECTION", 0, 0);
     int tuneLoopCount = getIntEnv("TE_HIPBLASLT_TUNING_RUN_COUNT", 0, 0);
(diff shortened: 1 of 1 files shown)


transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1158,11 +1159,47 @@ void hipblaslt_gemm(const Tensor *inputA,
...
+            algoCache.store(gemm_cfg, cached_algo);
+          }
+        }
+      }
+      
+      if (logTuning && !cached_algo.algo.has_value()) {
+        std::cout << "[WARNING] Cannot get corresponding solution from cached algoId " << cached_algo.algoId << std::endl;
+      }
+    }
+
```
ipanfilo, line 1195: Why does this code needs fallback to old scheme with searching for ID in the list returned by hipblasLtMatmulAlgoGetHeuristic()?
Also, since which ROCm version this hipblaslt_ext API is available?

RuibinCheung: 1. The ID may be invalid because of it from user input. It's a better way which fallback to old scheme than throw a runtime error.
2. I noticed that was supported on ROCm 6.0.0. Details: https://rocm.docs.amd.com/projects/hipBLASLt/en/docs-6.0.0/usage.html#getalgosfromindex

ipanfilo: If ID is invalid, it won't be found with old scheme as well, will it?
TE does not throw error in this case but continues as if there were no entry for the problem

RuibinCheung: Yeah, you are right. For the old scheme, if ID is invalid, it will chose a default algo. 

https://github.com/ROCm/TransformerEngine/blob/25a33da640da941fc190db649325e3cfc1b751e4/transformer_engine/common/gemm/rocm_gemm.cu#L1171-L1197

Here the minimum of algo count is 1. It means we will owns a default algo. So it will not cause throw error when calling `hipblasltMatmul`

ipanfilo: Old scheme is under this if https://github.com/ROCm/TransformerEngine/blob/240c71d7b964952d424ca33aadc80904e2a1b23b/transformer_engine/common/gemm/rocm_gemm.cu#L1227
New changes make it unnecessary, as well as algoTotalCount increasing based on cached_algo.index

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1158,11 +1159,45 @@ void hipblaslt_gemm(const Tensor *inputA,
   GemmAlgoCache::Algo cached_algo;
   if (algoCache.find(gemm_cfg, workspaceSize, cached_algo) == 0 || !cached_algo.algo.has_value())
   {
+    bool logTuning = getIntEnv("TE_HIPBLASLT_LOG_TUNING", 0, 0) != 0;
+
+#if HIP_VERSION >= 60000000
```
ipanfilo: Since ROCm before 6.0 is not supported by this module anymore, this guard is not needed

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1158,11 +1159,45 @@ void hipblaslt_gemm(const Tensor *inputA,
...
+          operationDesc, 
+          static_cast<const void*>(&one),
+          Adesc, 
+          Bdesc, 
+          static_cast<const void*>(&beta),
+          Ddesc,
+          Ddesc,
+          algo_arr[0].algo,
+          ws_size_min
+        )) {
```
ipanfilo, line 1184: Need check that ws_size_min <= workspaceSize and do not need algoCache.store unless update WS size value there

transformer_engine/common/gemm/rocm_gemm.cu
RuibinCheung: ACK. Thx.

transformer_engine/common/gemm/rocm_gemm.cu
RuibinCheung: ACK. Thx.

transformer_engine/common/gemm/rocm_gemm.cu
RuibinCheung: The type of algo of `algoCache.find` is reference rather than reference of pointer. So the assignment value of `cached_algo.algo = algo_arr[0].algo` will not work because the real cached algo in algoCache and cached_algo are in two different memory address.
So we should store the algo whatever the workspace size has changed or not.

ipanfilo: We should not, cached_algo.algo is not stored.

transformer_engine/common/gemm/rocm_gemm.cu
ipanfilo: Not completed

transformer_engine/common/gemm/rocm_gemm.cu
RuibinCheung: I wonder why we should not store cached_algo.algo ? If we don't store it we have to call `hipblaslt_ext::getAlgosFromIndex` to get it on every function call because the `cached_algo.algo.has_value()` is false.

ipanfilo: Is it based on some tests that you made? Can you please share the reproducer for this?

transformer_engine/common/gemm/rocm_gemm.cu
VeeraRajasekhar: https://github.com/ROCm/TransformerEngine/blob/240c71d7b964952d424ca33aadc80904e2a1b23b/transformer_engine/common/gemm/rocm_gemm.cu#L1227

These changes were removed in this PR, isn't this conversation resolved?

transformer_engine/common/gemm/rocm_gemm.cu
VeeraRajasekhar: Should I add the suggested changes by you @ipanfilo for this conversation, as we didn't see any reproducer from rubin.

VeeraRajasekhar:
@ipanfilo, Can I rebase with latest changes from our dev in this branch? or should I create a new PR?

transformer_engine/common/gemm/rocm_gemm.cu
ipanfilo: Requested changes are not completed.
...as well as algoTotalCount increasing based on cached_algo.index

ipanfilo:
> @ipanfilo, Can I rebase with latest changes from our dev in this branch? or should I create a new PR?

Both are good for me

VeeraRajasekhar:
getting tracked in https://github.com/ROCm/TransformerEngine/pull/310

PR #180, opened 2025-05-02
update hipify_torch submodule for version 2

jeffdaily:
Update the hipify_torch submodule to conditionally use version 2 of the hipify mappings.

### DIFF
--- a/build_tools/utils.py
+++ b/build_tools/utils.py
@@ -416,9 +416,23 @@ def uninstall_te_wheel_packages():
         ]
     )
 
+def detect_hipify_v2():
+    try:
+        from torch.utils.hipify import __version__
+        from packaging.version import Version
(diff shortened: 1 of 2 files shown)


build_tools/utils.py
```diff
@@ -416,9 +416,23 @@ def uninstall_te_wheel_packages():
         ]
     )
 
+def detect_hipify_v2():
+    try:
+        from torch.utils.hipify import __version__
```
wenchenvincent, line 421: What's the pytorch version where hipify is upgraded to v2?

jeffdaily: It is pending.  torch PR https://github.com/pytorch/pytorch/pull/151845 will make the change.  This TE PR is to get ahead of the change without breaking current "v1" behavior.

wenchenvincent (approved):
LGTM.

PR #197, opened 2025-06-02
Allows compilation with either Hipblaslt or Rocblas disabled

gabeweisz:
# Description

The documentation states that the environment variables NVTE_USE_HIPBLASLT is used to enable or disable Hipblaslt and NVTE_USE_ROCBLAS is used to enable or disable rocblas, but setup.py was previously enabling each if these was set at all (not checking 0 or 1), and CMakefiles.txt was defaulting both to on, so it was impossible to disable either. 
A [Rocblas refactoring](https://github.com/ROCm/rocBLAS/pull/1572) has removed F8 support, making it necessary to build TE with rocblas disabled.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Use the values of NVTE_USE_HIPBLASLT and NVTE_USE_ROCBLAS to control whether or not hipblaslt and rocblas support is enabled, respectively
- Fix an error where the multimap STL was missing if NVTE_USE_ROCBLAS  was false

# Checklist:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -82,6 +82,7 @@ The default installation above supports both rocBlas and hipBlasLt in GEMM compu
   export NVTE_USE_HIPBLASLT=1
   export NVTE_USE_ROCBLAS=1
 
+If neither of the aforementioned environment variables are set or both of them are set, then both backends are built. If only one of these environment variables is set, then the other backend is not built.
 When both GEMM backends are supported the aforementioned env variables can be used to select which backend to use. If none is set hipBlasLt is used by default. The hipBlasLt backed has not yet supported all the GEMM configurations in the pytorch unit tests. 
 
 Test
--- a/setup.py
+++ b/setup.py
@@ -80,13 +80,10 @@ def setup_common_extension() -> CMakeExtension:
         #  If one is set and the other is not, then only the corresponding backend is built
         enable_hipblaslt = os.getenv("NVTE_USE_HIPBLASLT") is not None
         enable_rocblas = os.getenv("NVTE_USE_ROCBLAS") is not None
-        if enable_hipblaslt and enable_rocblas:
-            cmake_flags.append("-DUSE_HIPBLASLT=ON")
-            cmake_flags.append("-DUSE_ROCBLAS=ON")
-        elif enable_hipblaslt:
+        if enable_hipblaslt and not enable_rocblas:
             cmake_flags.append("-DUSE_HIPBLASLT=ON")
             cmake_flags.append("-DUSE_ROCBLAS=OFF")
-        elif enable_rocblas:
+        elif enable_rocblas and not enable_hipblaslt:
             cmake_flags.append("-DUSE_HIPBLASLT=OFF")
             cmake_flags.append("-DUSE_ROCBLAS=ON")
         else:
--- a/setup.py
+++ b/setup.py
@@ -75,10 +75,8 @@ def setup_common_extension() -> CMakeExtension:
     cmake_flags = []
     if rocm_build():
         cmake_flags.append("-DUSE_ROCM=ON")
-        if os.getenv("NVTE_USE_HIPBLASLT") is not None:
-            cmake_flags.append("-DUSE_HIPBLASLT=ON")
-        if os.getenv("NVTE_USE_ROCBLAS") is not None:
-            cmake_flags.append("-DUSE_ROCBLAS=ON")
+        cmake_flags.append("-DUSE_HIPBLASLT=" + ("ON" if bool(int(os.getenv("NVTE_USE_HIPBLASLT", "1"))) else "OFF"))
+        cmake_flags.append("-DUSE_ROCBLAS=" + ("ON" if bool(int(os.getenv("NVTE_USE_ROCBLAS", "1"))) else "OFF"))
         if os.getenv("NVTE_AOTRITON_PATH"):
             aotriton_path = Path(os.getenv("NVTE_AOTRITON_PATH"))
             cmake_flags.append(f"-DAOTRITON_PATH={aotriton_path}")
--- a/setup.py
+++ b/setup.py
@@ -78,9 +78,21 @@ def setup_common_extension() -> CMakeExtension:
         # HIPBLASLT and ROCBLAS support are controlled as follows:
         #  If neither NVTE_USE_HIPBLASLT nor NVTE_USE_ROCBLAS are set, or both are set, both frameworks are enabled
         #  If one is set and the other is not, then only the corresponding backend is built
-        #  Note that the "no environment variable set" case relies on defaults set in transformer_engine/common/CMakeLists.txt
         enable_hipblaslt = os.getenv("NVTE_USE_HIPBLASLT") is not None
         enable_rocblas = os.getenv("NVTE_USE_ROCBLAS") is not None
+        if enable_hipblaslt and enable_rocblas:
+            cmake_flags.append("-DUSE_HIPBLASLT=ON")
+            cmake_flags.append("-DUSE_ROCBLAS=ON")
+        elif enable_hipblaslt:
+            cmake_flags.append("-DUSE_HIPBLASLT=ON")
+            cmake_flags.append("-DUSE_ROCBLAS=OFF")
+        elif enable_rocblas:
+            cmake_flags.append("-DUSE_HIPBLASLT=OFF")
+            cmake_flags.append("-DUSE_ROCBLAS=ON")
+        else:
+            cmake_flags.append("-DUSE_HIPBLASLT=ON")
+            cmake_flags.append("-DUSE_ROCBLAS=ON")
+
         if enable_hipblaslt or enable_rocblas:
             cmake_flags.append("-DUSE_HIPBLASLT=" + ("ON" if enable_hipblaslt else "OFF"))
             cmake_flags.append("-DUSE_ROCBLAS=" + ("ON" if enable_rocblas else "OFF"))
--- a/setup.py
+++ b/setup.py
@@ -75,8 +75,18 @@ def setup_common_extension() -> CMakeExtension:
     cmake_flags = []
     if rocm_build():
         cmake_flags.append("-DUSE_ROCM=ON")
-        cmake_flags.append("-DUSE_HIPBLASLT=" + ("ON" if bool(int(os.getenv("NVTE_USE_HIPBLASLT", "1"))) else "OFF"))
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -75,10 +75,8 @@ def setup_common_extension() -> CMakeExtension:
     cmake_flags = []
     if rocm_build():
         cmake_flags.append("-DUSE_ROCM=ON")
-        if os.getenv("NVTE_USE_HIPBLASLT") is not None:
-            cmake_flags.append("-DUSE_HIPBLASLT=ON")
-        if os.getenv("NVTE_USE_ROCBLAS") is not None:
-            cmake_flags.append("-DUSE_ROCBLAS=ON")
+        cmake_flags.append("-DUSE_HIPBLASLT=" + ("ON" if bool(int(os.getenv("NVTE_USE_HIPBLASLT", "1"))) else "OFF"))
+        cmake_flags.append("-DUSE_ROCBLAS=" + ("ON" if bool(int(os.getenv("NVTE_USE_ROCBLAS", "1"))) else "OFF"))
```
ipanfilo: If start checking NVTE_USE_HIPBLASLT/ROCBLAS env variables values, IMO it should be done everywhere and also described in README.rst

gabeweisz: Where else should it be checked? Are you saying that cmake should be checking these env variables directly inste4ad of USE_HIPBLASLT or USE_ROCBLAS?

Readme.rst does already say that these variables are used to [enable/disable the backends](https://github.com/ROCm/TransformerEngine/blob/a01f7f11bda6efda9fc45a104d7f8bef4dbbc8e9/README.rst?plain=1#L82)

The whole point of this PR is that the current documentations says that rocblas or hipblaslt can be disabled with these environment variables, but they currently cannot be because the environment variables described in readme.rst are not handled correctly.

ipanfilo: Historically TE build infrastructure and runtime code do not check the value of those variables but rather if they are set or not. So whenever those variables are checked they are checked just for existence.
How it is currently made, it is indeed not possible to disable any backend by setting env variables.
To fix this functionality with current env usage model, the change should rather be 
if USE_HIPBLASLT is not None or USE_ROCBLAS is not None: 
    if USE_HIPBLASLT is None: cmake.flags.append(-DUSE_HIPBLASLT=OFF)
    if USE_ROCBLAS is None: cmake.flags.append(-DUSE_ROCBLAS=OFF)

gabeweisz: Understood. I changed the code this way because this is what the documentation said (at least in my reading), but I will switch the code and documentation to follow your suggestion, and clarify readme.rst

gabeweisz: Updated as requested

README.rst
```diff
@@ -82,6 +82,7 @@ The default installation above supports both rocBlas and hipBlasLt in GEMM compu
   export NVTE_USE_HIPBLASLT=1
   export NVTE_USE_ROCBLAS=1
 
+If neither of the aforementioned environment variables are set or both of them are set, then both backends are built. If only one of these environment variables is set, then the other backend is not built.
```
ipanfilo, line 85: If only one of them is set then the only corresponding backend is built

setup.py
```diff
@@ -75,10 +75,18 @@ def setup_common_extension() -> CMakeExtension:
     cmake_flags = []
     if rocm_build():
         cmake_flags.append("-DUSE_ROCM=ON")
+        # HIPBLASLT and ROCBLAS support are controlled as follows:
+        #  If neither NVTE_USE_HIPBLASLT nor NVTE_USE_ROCBLAS are set, or both are set, both frameworks are enabled
+        #  If one is set and the other is not, then the framework that whose variable is set is enabled and the other is disabled
+        enable_hipblaslt = True
```
ipanfilo: assign variables with (os.getenv() is not None)
Then if any variable is True, set flags. If both are False, no need to set any flags

gabeweisz: This is more elegant, thanks for the tip

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -6,6 +6,7 @@
 #include <type_traits>
 #include <transformer_engine/gemm.h>
 #include <transformer_engine/transformer_engine.h>
+#include <map>
```
gabeweisz, line 9: This was needed when Hipblaslt was enabled but Rocblas was not. I don't think it hurts anything to always include it

setup.py
```diff
@@ -75,10 +75,17 @@ def setup_common_extension() -> CMakeExtension:
...
-        if os.getenv("NVTE_USE_ROCBLAS") is not None:
-            cmake_flags.append("-DUSE_ROCBLAS=ON")
+        # HIPBLASLT and ROCBLAS support are controlled as follows:
+        #  If neither NVTE_USE_HIPBLASLT nor NVTE_USE_ROCBLAS are set, or both are set, both frameworks are enabled
+        #  If one is set and the other is not, then only the corresponding backend is built
+        enable_hipblaslt = os.getenv("NVTE_USE_HIPBLASLT") is not None
+        enable_rocblas = os.getenv("NVTE_USE_ROCBLAS") is None
+
+        # If neither environment variable is set, then both frameworks will be built based on the defaults
+        # in transformer_engine/common/CMakeLists.txt
```
ipanfilo: This comment duplicates one on lines 78-80

setup.py
```diff
@@ -75,10 +75,17 @@ def setup_common_extension() -> CMakeExtension:
...
         cmake_flags.append("-DUSE_ROCM=ON")
-        if os.getenv("NVTE_USE_HIPBLASLT") is not None:
-            cmake_flags.append("-DUSE_HIPBLASLT=ON")
-        if os.getenv("NVTE_USE_ROCBLAS") is not None:
-            cmake_flags.append("-DUSE_ROCBLAS=ON")
+        # HIPBLASLT and ROCBLAS support are controlled as follows:
+        #  If neither NVTE_USE_HIPBLASLT nor NVTE_USE_ROCBLAS are set, or both are set, both frameworks are enabled
+        #  If one is set and the other is not, then only the corresponding backend is built
+        enable_hipblaslt = os.getenv("NVTE_USE_HIPBLASLT") is not None
+        enable_rocblas = os.getenv("NVTE_USE_ROCBLAS") is None
```
ipanfilo: is not None

ipanfilo (changes requested):

setup.py
gabeweisz: The goal of this comment was to point out that it relies on transformer_engine/common/CMakeLists.txt for correct behavior when no flags are passed, but I will consolidate

setup.py
gabeweisz: Oops thanks for catching

setup.py
```diff
@@ -75,10 +75,15 @@ def setup_common_extension() -> CMakeExtension:
...
-            cmake_flags.append("-DUSE_HIPBLASLT=ON")
-        if os.getenv("NVTE_USE_ROCBLAS") is not None:
-            cmake_flags.append("-DUSE_ROCBLAS=ON")
+        # HIPBLASLT and ROCBLAS support are controlled as follows:
+        #  If neither NVTE_USE_HIPBLASLT nor NVTE_USE_ROCBLAS are set, or both are set, both frameworks are enabled
+        #  If one is set and the other is not, then only the corresponding backend is built
+        #  Note that the "no environment variable set" case relies on defaults set in transformer_engine/common/CMakeLists.txt
+        enable_hipblaslt = os.getenv("NVTE_USE_HIPBLASLT") is not None
+        enable_rocblas = os.getenv("NVTE_USE_ROCBLAS") is not None
+        if enable_hipblaslt or enable_rocblas:
```
wangye805: This section of codes is correct logically but can we make it more readable as:
if enable_hipblaslt and enable_rocblas:
//enable both backend
elif enable_hipblaslt or enable_rocblas:
//enable one backend and disable the other
else:
//enable both backend

Then this can corresponds to your comment section

wangye805 (changes requested):

setup.py
gabeweisz: Readability is good. I changed the code to explicitly list the four possible code paths.

setup.py
```diff
@@ -75,10 +75,27 @@ def setup_common_extension() -> CMakeExtension:
...
+        elif enable_rocblas:
+            cmake_flags.append("-DUSE_HIPBLASLT=OFF")
+            cmake_flags.append("-DUSE_ROCBLAS=ON")
+        else:
+            cmake_flags.append("-DUSE_HIPBLASLT=ON")
+            cmake_flags.append("-DUSE_ROCBLAS=ON")
+
+        if enable_hipblaslt or enable_rocblas:
+            cmake_flags.append("-DUSE_HIPBLASLT=" + ("ON" if enable_hipblaslt else "OFF"))
+            cmake_flags.append("-DUSE_ROCBLAS=" + ("ON" if enable_rocblas else "OFF"))
```
wangye805: These lines should be deleted with your newly added control logic

wangye805 (dismissed):

setup.py
gabeweisz: Yes, sorry for missing that

setup.py
ipanfilo: I think that was was meant is that we actually have only two special cases to handle: (hipblaslt=ON, rocblas=OFF) and (hipblaslt=OFF, rocblas=ON). Other two combinations - are default behaviour that don't event need adding any flags. 
Otherwise I cannot agree that those 4 ifs are more readable than one if with two lines of code

gabeweisz: You two duke it out and tell me what version of this code will be acceptable. We need something like this to be able to turn off hipblaslt or rocblas.

wangye805: @ipanfilo is more familiar with this section than me. You can follow his advice and ignore my comments

gabeweisz: @ipanfilo - do you want me to go back to commit c9f493715afb1e186232368eccf676a8e8551138? Or do you want something else?

ipanfilo: For this piece of code c9f4937 looks good for me. 
Or you may shorten the ifs to have 
if enable_hipblaslt and not enable rocblas: 
 ...
elseif enable_rocblas and not enable_hipblaslt:
...

ipanfilo (commented):
Put my comment regarding the controversial piece of code. Sorry for confusion

setup.py
gabeweisz: c9f4937 is my personal favorite, but I don't want anyone looking at this code to have to think about what is minor functionality, so I will go with the latter suggestion of putting the "if both true" and "if both false" cases into the else clause. Could one of you please approve so I can close this out (assuming you don't see any typos in the new version)?

ipanfilo (approved):

PR #199, opened 2025-06-04
Fixed attention benchmark script and added it to CI.

VeeraRajasekhar:
# Description

**Added --use_ck_bwd_v3**
- Enables CK backward-v3 kernels only when the flag is passed.
- parse_results() now looks for kernel_func when v3 is active, ensuring the correct fwd/bwd kernel times are captured in times.csv.

**Added --run_sanity_checks**
Opt-in verification new  flag post-processes profiler output, ensuring every backend that should have run produced a valid results.stats.csv, and that fused vs CK kernel timings agree within a configurable tolerance.

**ci/pytorch.sh** calls the benchmark with both --use_ck_bwd_v3 and --run_sanity_checks when TEST_LEVEL ≥ 3, automatically gating performance regressions.
Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/benchmarks/attention/benchmark_attention_rocm.py
+++ b/benchmarks/attention/benchmark_attention_rocm.py
@@ -73,7 +73,7 @@ columns = [
     ]
 
 output_csv="times.csv"
-
+cwd = os.getcwd()
 # Runs benchmark with warmup iterations and profiles using rocprof
 def benchmark_dot_product_attention(model, attention, column_name, dirname):
     config = model_configs[model]
@@ -145,6 +145,10 @@ def parse_helper(model, dirname, fwd_search_pattern, bwd_search_pattern, column_
     # Extract kernel timing values    
     fwd_values = df[df["Name"].str.contains(fwd_search_pattern)]["AverageNs"].to_numpy()
     bwd_values = df[df["Name"].str.contains(bwd_search_pattern)]["AverageNs"].to_numpy()
+
+    if len(bwd_values) == 0:
+        return False  # CK V3 not supported or kernel_func not found
+    
     t_attn_avg = np.empty(len(fwd_values) + len(bwd_values))
     t_attn_avg[:len(fwd_values)] = fwd_values
     t_attn_avg[len(fwd_values):] = bwd_values
@@ -154,19 +158,27 @@ def parse_helper(model, dirname, fwd_search_pattern, bwd_search_pattern, column_
     df_times.loc[model, f"{column_name} Kernels (bwd)"] = t_attn_avg[len(fwd_values):].sum() / 1e6
     df_times.loc[model, f"{column_name} Kernels (fwd+bwd)"] = t_attn_avg.sum() / 1e6
 
+    return True
+
 # Parses profiler logs for different attention backends
 def parse_results(model, df_times, perf_dir_flash_attn, perf_dir_fused_attn, perf_dir_fused_ck, perf_dir_fused_aotriton, use_ck_bwd_v3):
     if perf_dir_flash_attn:
         parse_helper(model, perf_dir_flash_attn, "FmhaFwd", "FmhaBwd", "FlashAttention", df_times)
 
     if perf_dir_fused_attn:
-        bwd_pattern = "kernel_func" if use_ck_bwd_v3 else "FmhaBwd"
-        parse_helper(model, perf_dir_fused_attn, "FmhaFwd", bwd_pattern, "FusedAttention", df_times)
+        ck_v3_success = False
+        if use_ck_bwd_v3:
+            ck_v3_success = parse_helper(model, perf_dir_fused_ck, "FmhaFwd", "kernel_func", "FusedAttention", df_times)
+        if not ck_v3_success:
+            parse_helper(model, perf_dir_fused_ck, "FmhaFwd", "FmhaBwd", "FusedAttention", df_times)
 
-    if perf_dir_fused_ck:
-        bwd_pattern = "kernel_func" if use_ck_bwd_v3 else "FmhaBwd"
-        parse_helper(model, perf_dir_fused_ck, "FmhaFwd", bwd_pattern, "FusedAttention CK", df_times)
-
+    if perf_dir_fused_attn:
+        ck_v3_success = False
+        if use_ck_bwd_v3:
+            ck_v3_success = parse_helper(model, perf_dir_fused_ck, "FmhaFwd", "kernel_func", "FusedAttention CK", df_times)
+        if not ck_v3_success:
+            parse_helper(model, perf_dir_fused_ck, "FmhaFwd", "FmhaBwd", "FusedAttention CK", df_times)
+    
     if perf_dir_fused_aotriton:
         parse_helper(model, perf_dir_fused_aotriton, "attn_fwd", "bwd", "FusedAttention AOTriton", df_times)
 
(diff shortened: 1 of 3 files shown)


benchmarks/attention/benchmark_attention_rocm.py
```diff
@@ -174,7 +172,8 @@ def parse_results(model, df_times, filename_flash_attn, filename_fused_attn, fil
 
 
 def main():
-    
+    cwd = os.path.dirname(__file__)
+    os.chdir(cwd)
```
wenchenvincent: Do we have to change to the directory of this script? It will save the results in that directory then. But the user would usually expect to have the result in the directory where they run this script.

wangye805 (changes requested):
Apart from the path issue Wen mentioned, can we add this to CI so that we can keep monitoring whether this script failed with our new PRs?

benchmarks/attention/benchmark_attention_rocm.py
VeeraRajasekhar: Okay, I can change this. I thought there were too many directory in the root level, so changed it to that. I can revert back this.

VeeraRajasekhar:
> Apart from the path issue Wen mentioned, can we add this to CI so that we can keep monitoring whether this script failed with our new PRs?

Do you mean, adding a new row here
image

OR

Add it to in the ci/pytorch.sh file?

wangye805:
> > Apart from the path issue Wen mentioned, can we add this to CI so that we can keep monitoring whether this script failed with our new PRs?
> 
> Do you mean, adding a new row here image
> 
> OR
> 
> Add it to in the ci/pytorch.sh file?

I guess adding to ci/pytorch.sh is easier, right?

benchmarks/attention/benchmark_attention_rocm.py
VeeraRajasekhar: @wenchenvincent can I make just the times.csv in the directory where the user ran the script and all the rocprof & intermediate outputs to the benchmark/attention directory and print these details in the terminal after the run? how does this sound?

wenchenvincent: I suggest saving the rocprof results also in the directory where the user ran the script. Sometimes, the user might want to inspect the raw data.

VeeraRajasekhar:
> > > Apart from the path issue Wen mentioned, can we add this to CI so that we can keep monitoring whether this script failed with our new PRs?
> > 
> > 
> > Do you mean, adding a new row here image
> > OR
> > Add it to in the ci/pytorch.sh file?
> 
> I guess adding to ci/pytorch.sh is easier, right?

yes

benchmarks/attention/benchmark_attention_rocm.py
```diff
@@ -153,35 +155,51 @@ def parse_helper(model, filename, fwd_search_pattern, bwd_search_pattern, column
...
 
-    if filename_fused_attn:
-        parse_helper(model, filename_fused_attn, "FmhaFwd", "FmhaBwd", "FusedAttention", df_times)
+    if perf_dir_fused_attn:
+        parse_helper(model, perf_dir_fused_attn, "FmhaFwd", "FmhaBwd", "FusedAttention", df_times)
 
-    if filename_fused_ck:
-        parse_helper(model, filename_fused_ck, "FmhaFwd", "FmhaBwd", "FusedAttention CK", df_times)
+    if perf_dir_fused_ck:
+        bwd_pattern = "kernel_func" if use_ck_bwd_v3 else "FmhaBwd"
```
wangye805: Is there other ways to search the v3 pattern? For example, search the kernel launched from TE ck_fused_attn lib? Currently "kernel_func" is fine for ck in MI300/MI308 but after our AITER migration, this may change

wangye805 (approved):
LGTM

ci/pytorch.sh
```diff
@@ -93,6 +106,8 @@ fi
 #Master script mode: prepare testing prerequisites first
 start_message
 install_prerequisites
+# Run benchmark
+run_benchmark
```
ipanfilo: Why do we run benchmarks unconditionally as the first thing to test?

ci/pytorch.sh
```diff
@@ -81,6 +81,19 @@ run_test_config_mgpu(){
...
+    echo "\n============= Running benchmarks attention script ============="
+    BENCH_SCRIPT="$DIR/../benchmarks/attention/benchmark_attention_rocm.py"
+
+    if [ ! -f "$BENCH_SCRIPT" ]; then
+        echo "Benchmark script not found: $BENCH_SCRIPT"
+        return
+    fi
+
+    python "$BENCH_SCRIPT" --use_ck_bwd_v3
+    echo "================================================================\n"
```
wenchenvincent: I think we need to put in some sanity checks to see if the benchmark result makes sense.

ci/pytorch.sh
VeeraRajasekhar: Modified it to run at the end of the script and under TEST_LEVEL=3

benchmarks/attention/benchmark_attention_rocm.py
wenchenvincent: If we have the trace, we would be able to find the kernel by the temporal correlation. However, it will be complex. I think for simplicity, we can just use the kernel name for search pattern for now.

VeeraRajasekhar: okay wen

ci/pytorch.sh
VeeraRajasekhar: Any suggestions or ideas for this @wenchenvincent ?

wangye805:
A minor issue: can you modify the title to make it more expressive? I cannot relate the current title to attention benchmarking :-)

ci/pytorch.sh
VeeraRajasekhar: Added sanity checks

benchmarks/attention/benchmark_attention_rocm.py
```diff
@@ -176,14 +188,83 @@ def parse_results(model, df_times, perf_dir_flash_attn, perf_dir_fused_attn, per
...
+                rel_fwd = abs(f_fwd - c_fwd) / max(f_fwd, c_fwd)
+                rel_bwd = abs(f_bwd - c_bwd) / max(f_bwd, c_bwd)
+                if rel_fwd < tol and rel_bwd < tol:
+                    print(f"  [OK ] Fused vs CK diff ≤ {tolerance_pct}% "
+                          f"(fwd {rel_fwd*100:.2f} %, bwd {rel_bwd*100:.2f} %)")
+                else:
+                    ok_overall = False
+                    print(f"  [FAIL] Fused vs CK kernel time diff > {tolerance_pct}% "
+                          f"(fwd {rel_fwd*100:.2f} %, bwd {rel_bwd*100:.2f} %)")
+        print("-" * 60)
```
wenchenvincent, line 266: Did you return the status code?

VeeraRajasekhar: No, I am printing an error statement, I initially planned to do a **sys.exit**, but didn't checkin this. Should I do sys.exit and break CI run, if anything is failed?

wenchenvincent: I think you would want ci/pytorch.sh to be able to catch the error if there is a failure when running benchmarking_attention_rocm.py. So try to do it in a way that is similar to how other tests are run.

I think you might want to do something similar to this line: https://github.com/ROCm/TransformerEngine/blob/dev/ci/pytorch.sh#L32
To do that, you do need to be able to return a status from running benchmarking_attention_rocm.py

ci/pytorch.sh
```diff
@@ -90,7 +90,7 @@ run_benchmark() {
         return
     fi
 
-    python "$BENCH_SCRIPT" --use_ck_bwd_v3
+    python "$BENCH_SCRIPT" --use_ck_bwd_v3 --run_sanity_checks
```
wenchenvincent: After changing the BENCH_SCRIPT to be able to return a status. Change it to something like:

python "$BENCH_SCRIPT" --use_ck_bwd_v3 --run_sanity_checks || test_run_error "benchmark "

And we also need to assign a test level to testing the benchmark script.

VeeraRajasekhar: The test_level is already added while invoking in line 150. Working on adding return status

ci/pytorch.sh
```diff
@@ -81,6 +81,19 @@ run_test_config_mgpu(){
     fi
 }
 
+run_benchmark() {
```
ipanfilo, line 84: Add `check_test_filter "benchmark" || return` here so benchmark can be skip if needed

VeeraRajasekhar: Added

ci/pytorch.sh
VeeraRajasekhar: Added return status and corresponding error messages

wenchenvincent (approved):
I edited the title of the PR to be more descriptive. Also, next time, let the reviewers mark the conversation as resolved.

PR #200, opened 2025-06-06
Add wheels build support

ipanfilo:
# Description

Fixes # [12089](https://github.com/ROCm/frameworks-internal/issues/12089)

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

## Changes

- Modify build scripts to build and install wheels
- Add docker file

# Checklist:


build_tools/jax.py
```diff
@@ -63,7 +63,9 @@ def setup_jax_extension(
         xla_home,
     ])
 
-    if rocm_build():
+    # If NVTE_RELEASE_BUILD is set, we assume not building but sources packaging
+    # and we do not hipify the sources
+    if rocm_build() and not bool(int(os.getenv("NVTE_RELEASE_BUILD", "0"))):
```
wenchenvincent, line 68: Let's change the env var to `NVTE_WHEEL_BUILD` to be more precise.

setup.py
```diff
@@ -150,12 +150,13 @@ def setup_requirements() -> Tuple[List[str], List[str], List[str]]:
         assert bool(
             int(os.getenv("NVTE_RELEASE_BUILD", "0"))
         ), "NVTE_RELEASE_BUILD env must be set for metapackage build."
+        te_cuda_vers = "rocm" if rocm_build() else "cu12"
```
wenchenvincent, line 158: Do we need to specify the rocm version as well?

build_tools/build_ext.py
```diff
@@ -207,7 +207,8 @@ def run(self) -> None:
         def build_extensions(self):
             # BuildExtensions from PyTorch and PaddlePaddle already handle CUDA files correctly
             # so we don't need to modify their compiler. Only the pybind11 build_ext needs to be fixed.
-            if "pytorch" not in get_frameworks() and "paddle" not in get_frameworks():
+            ext_names = [ext.name for ext in self.extensions]
```
wenchenvincent, line 210: Why do we need this change?

build_tools/utils.py
```diff
@@ -279,6 +279,21 @@ def cuda_version() -> Tuple[int, ...]:
     return tuple(int(v) for v in version)
 
 
+def rocm_version() -> Tuple[int, ...]:
```
wenchenvincent: It seems that this function is not used.

build_tools/wheel_utils/Dockerfile.rocm.x86
```diff
@@ -0,0 +1,38 @@
...
+
+# Setup dev packages
+RUN dnf group install -y "Development Tools" && dnf install -y git cmake llvm-toolset
+RUN dnf clean all
+RUN rm -rf /var/cache/dnf/*
+
+ENV CMAKE_PREFIX_PATH=${ROCM_PATH}/lib/cmake
+ENV NVTE_RELEASE_BUILD=1
+
+ARG GPU_TARGETS="gfx90a;gfx942;gfx950;gfx1100"
```
wenchenvincent: As discussed, let's just use gfx942 as the default for now.

build_tools/jax.py
wangye805: @wenchenvincent NV upstream is using NVTE_RELEASE_BUILD. We can keep the same to be consistent with our upstream

build_tools/utils.py
```diff
@@ -394,6 +409,35 @@ def copy_common_headers(
         shutil.copy(path, new_path)
 
 
+def copy_hipify_tools(
```
wangye805, line 397: Why do we need to copy the hipify_tool and then delete it during release building? Can we use the hipify tool directly in 3rdparty dir?

build_tools/wheel_utils/build_wheels.sh
```diff
@@ -18,52 +20,84 @@ mkdir -p /wheelhouse/logs
 # Generate wheels for common library.
 git config --global --add safe.directory /TransformerEngine
 cd /TransformerEngine
+git pull
```
wenchenvincent: Why do we need `git pull` if we just check out a branch in the next command?

build_tools/wheel_utils/Dockerfile.rocm.x86
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
```
wenchenvincent: Let's also add `manylinux` in the Dockerfile name explicitly. `Dockerfile.manylinux.rocm.x86`

build_tools/jax.py
ipanfilo: It is NV variable so let's keep it to avoid further maintenance issues

setup.py
ipanfilo: I was told that different ROCm versions packages are going to be located separately on AMD repo.
Also, building script and even whole TE code do not depend on the ROCm version so hardcoding it does not make sense from this perspective. 
For CUDA it is little bit different - build docker file installs specific version of it, so version can be hardcoded here to match what is installed.
For ROCm, we rely on pre-installed version of it so it should be detected runtime
Thus, I'd avoid use version here unless it is really needed for something.

build_tools/build_ext.py
ipanfilo: Real build environment depends not on installed frameworks but on actually engaged extensions. For monolithic build from source the list of extensions reflects the list of frameworks but for installing from wheel, every extension is built separately.  With single integration framework installed it is not a problem too but on unified Pytorch+JAX image, it gives incorrect positive results when build JAX extension

build_tools/wheel_utils/build_wheels.sh
ipanfilo: Checkout does not pull recent changes from origin. Having pull there allows to specify desired branch and every time build from its tip.

build_tools/wheel_utils/Dockerfile.rocm.x86
ipanfilo: I'll change to Dockerfile.rocm.manylinux.x86 then. Prefer to use rocm as the major specifier

build_tools/utils.py
ipanfilo: I keep it for now till finalize that no version info is needed.

setup.py
wenchenvincent: There are some places in TE code that would build based on different ROCm versions:
https://github.com/ROCm/TransformerEngine/blob/dev/transformer_engine/common/amd_detail/hip_float8.h#L16

It means that the binary would be a bit different for different ROCm versions.

ipanfilo: No doubts. The binary could be different despite event w/o code change just because ROCm interfaces can change. The wheel must be installed with the same ROCm version it was built with. Adding version to wheel tag does not ensure correct usage, it only helps to distinct wheel files if they're located together. I think some other way like adding custom metadata may be less disruptive but still give user way to check what what the package build agains

wenchenvincent: Yeah, it would be great to add a mechanism to check the rocm version when installing the wheels. Not sure how much effort it will take. I am thinking, for now, we at least distinguish that in the wheel name. We can have a follow-up PR to add the checking mechanism.

build_tools/wheel_utils/build_wheels.sh
wenchenvincent: In this case, we did a fresh clone of the repo and I think the checkout would get the tip of the branch. If we didn't do a fresh clone, then it is true that the checkout might be stale.

build_tools/build_ext.py
wenchenvincent: What's the difference between the installed frameworks and the actually engaged extensions?

ipanfilo: Installed frameworks is what it is: list of frameworks installed on the system for which integration extensions can be built. Actual list of extensions to build may have all or only some of possible ones. Because TE setup.py builds list of extensions to build based on list of frameworks, the both lists match. But when source wheel is installed, it contains only single extensions, no matter what is frameworks list.

build_tools/wheel_utils/build_wheels.sh
ipanfilo: There are 2 possible scenarios - source tree is mapped to container from host in which case it supposedly has fresh data or it can use local tree checked out at time of image creation in which case pull is desired

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -75,9 +73,6 @@ endforeach()
 add_library(ck_fused_attn STATIC ${ck_fused_attn_SOURCES})
 set(CK_FUSED_ATTN_COMPILE_OPTIONS)
 list(APPEND CK_FUSED_ATTN_COMPILE_OPTIONS -Wno-undefined-func-template -DCK_TILE_FMHA_FWD_FAST_EXP2=1 -DCK_TILE_FMHA_FWD_SPLITKV_API=0 -DCK_TILE_FMHA_FWD_APPENDKV_API=0 -DCK_TILE_FLOAT_TO_BFLOAT16_DEFAULT=${CK_FUSED_ATTN_FLOAT_TO_BFLOAT16_DEFAULT} -fgpu-flush-denormals-to-zero -Wno-float-equal -ftemplate-backtrace-limit=0  -fPIC  -Wno-gnu-line-marker -Wunused-variable -Wuninitialized -Werror)
-foreach(rocm_arch ${CK_FUSED_ATTN_TARGET_GPUS})
```
wenchenvincent, line 78: With those removed, how are we passing the build targets to CK fused attention?

transformer_engine/common/CMakeLists.txt
```diff
@@ -461,10 +470,6 @@ else()
   # Ask hcc to generate device code during compilation so we can use
   # host linker to link.
   set(HIP_HCC_FLAGS "${HIP_HCC_FLAGS} -fno-gpu-rdc -Wno-defaulted-function-deleted")
-  foreach(rocm_arch ${CMAKE_HIP_ARCHITECTURES})
-    # if CMAKE_CXX_FLAGS has --offload-arch set already, better to rm first
```
wenchenvincent, line 465: With these removed, how are we passing the build targets?

transformer_engine/common/ck_fused_attn/CMakeLists.txt
ipanfilo: Because of https://github.com/ROCm/TransformerEngine/blob/d0c927192b93f44e63f48566f8284be6daea33e1/transformer_engine/common/CMakeLists.txt#L251
That seems to be desired way to instruct HIP of target architectures.

transformer_engine/common/CMakeLists.txt
ipanfilo: Because of https://github.com/ROCm/TransformerEngine/blob/d0c927192b93f44e63f48566f8284be6daea33e1/transformer_engine/common/CMakeLists.txt#L251

build_tools/wheel_utils/build_wheels.sh
wangye805: Do we need to guard our changes by a if-else to make NV flow unchanged

transformer_engine/common/ck_fused_attn/CMakeLists.txt
wangye805: But this GPU_TARGETS are set at the parent level/dir of this transformer_engine/common/ck_fused_attn/CMakeLists.txt. There were no --offload-arch specified to the object/library ck_fused_attn here. Will this cause some problems?

build_tools/utils.py
ipanfilo: Pytorch/JAX extra release building is just packaging of necessary files to built the library on target machine.

build_tools/wheel_utils/build_wheels.sh
ipanfilo: It is good question. It little bit puzzling because we need TE first to determine if it is ROCm or NV. Let me look into it.

transformer_engine/common/ck_fused_attn/CMakeLists.txt
ipanfilo: Variables set in parent CMakeLists.txt remain visible during add_subdirectory. It is the same way how CK_TILE_FLOAT_TO_BFLOAT16_DEFAULT is passed. 
It is only different for AOTriton which has to set GPU_TARGETS itself

wenchenvincent: Is `GPU_TARGETS` the new preferred way for selecting build targets for ROCm? Is there a minimum CMake version required for it to work? It seems that I can find it in the ROCm docs (https://rocm.docs.amd.com/en/latest/conceptual/cmake-packages.html) but cannot quite find it in the CMake docs..

ipanfilo: I don't think it is CMake itself but rather CMake HIP package - it expects GPU_TARGETS to be set by caller or tries to detect it from GPUs on the system. For development build when you build and run on the same system - it works fine, for wheels build may cause erros. That is why you it is in ROCm doccumentation

build_tools/wheel_utils/build_wheels.sh
ipanfilo: Done

wangye805 (approved):
LGTM

setup.py
```diff
@@ -125,13 +125,18 @@ def setup_requirements() -> Tuple[List[str], List[str], List[str]]:
...
+                test_reqs.extend(["numpy", "onnxruntime", "torchvision", "prettytable"])
         if "jax" in frameworks:
-            install_reqs.extend(["jax", "flax>=0.7.1"])
-            test_reqs.extend(["numpy", "praxis"])
+            if rocm_build():
+                from build_tools.jax import jax_install_requires
+                install_reqs.extend(jax_install_requires(["flax>=0.7.1"]))
+            else:
+                install_reqs.extend(["jax", "flax>=0.7.1"])
+                test_reqs.extend(["numpy", "praxis"])
```
wenchenvincent, line 139: How does test_reqs work?

ipanfilo: Well, I think it should install dependencies when build and install tests through setup.py. But this feature is deprecated in setuptools and it not used for ROCm builds

wenchenvincent (approved):
LGTM

PR #207, opened 2025-06-20
Support math_sm_count for GEMM

ipanfilo:
# Description

Add math_sm_count support for GEMM

Fixes: [#11921](https://github.com/ROCm/frameworks-internal/issues/11921)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add math_sm_count support in ROCm cublas_gemm
- Create basic test
- Fix pytorch CI script

# Checklist:

### DIFF
--- a/tests/pytorch/test_gemm_sm_count.py
+++ b/tests/pytorch/test_gemm_sm_count.py
@@ -50,4 +50,4 @@ def test_gemm_sm_count():
     with torch.cuda.stream(torch.cuda.Stream()):
         constrained_timing = _run_gemm_timing("Constrained", gemm_parameters)
     
-    assert full_timing  < constrained_timing * 5, "GEMM performance is not changed with constrained SM count"
+    assert full_timing * 5  < constrained_timing, "GEMM performance is not changed with constrained SM count"
--- a/tests/pytorch/test_gemm_sm_count.py
+++ b/tests/pytorch/test_gemm_sm_count.py
@@ -25,10 +25,10 @@ def _run_gemm_timing(name: str, gemm_parameters: Dict) -> float:
     
     timing = timing[2:]  # Discard the initial timings to avoid warm-up effects
     mean = sum(timing) / len(timing)
-    disp = [abs(t - mean) for t in timing]
-    logging.info(f"{name} timing: {mean} ms with variance {sum(disp)/len(timing)}")
-    if any(d > 0.2 * mean for d in disp):
-        raise RuntimeError(f"High timing variance detected: {timing}")
+    abs_dev = [abs(t - mean) for t in timing]
+    logging.info(f"{name} timing: {mean} ms with deviation {sum(abs_dev)/len(timing)}")
+    if any(d > 0.2 * mean for d in abs_dev):
+        raise RuntimeError(f"High timing deviation detected: {timing}")
     return mean
 
 
--- a/tests/pytorch/test_gemm_sm_count.py
+++ b/tests/pytorch/test_gemm_sm_count.py
@@ -1,0 +1,53 @@
+# Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
+# License for AMD contributions = MIT. See LICENSE for more information
+
+import os
+import torch
+from typing import Dict
+
+from transformer_engine.pytorch.cpp_extensions import general_gemm
+from transformer_engine.pytorch.module.base import get_workspace
+import logging
+
+
+def _run_gemm_timing(name: str, gemm_parameters: Dict) -> float:
+    """ Run GEMM operations and return average execution time."""
+    start = torch.cuda.Event(enable_timing=True)
+    end = torch.cuda.Event(enable_timing=True)
+
+    timing = []
+    for _ in range(5):
+        start.record()
+        general_gemm(**gemm_parameters)
+        end.record()
+        torch.cuda.current_stream().synchronize()
+        timing.append(start.elapsed_time(end))
+    
+    timing = timing[2:]  # Discard the initial timings to avoid warm-up effects
+    mean = sum(timing) / len(timing)
+    disp = [abs(t - mean) for t in timing]
+    logging.info(f"{name} timing: {mean} ms with variance {sum(disp)/len(timing)}")
+    if any(d > 0.2 * mean for d in disp):
+        raise RuntimeError(f"High timing variance detected: {timing}")
+    return mean
+
+
+def test_gemm_sm_count():
+    """ Test Math SM count for GEMM operation by comparing performance with different SM counts."""
+    M = 2304
+    N = 4096
+    K = 768
+    datatype = torch.float32
+    A = torch.randn((K, M), device="cuda", dtype=datatype)
+    B = torch.randn((N, K), device="cuda", dtype=datatype)
+    gemm_parameters = {'A': A, 'B': B, 'layout': "NN", 'workspace': get_workspace()}
+  
+    with torch.cuda.stream(torch.cuda.Stream()):
+        full_timing = _run_gemm_timing("Full", gemm_parameters)
+    
+    cu_count = torch.cuda.get_device_properties(torch.cuda.current_device()).multi_processor_count
+    os.environ['NVTE_EXT_MARGIN_SM'] = str(95 * cu_count // 100) # Use 5% of SMs
+    with torch.cuda.stream(torch.cuda.Stream()):
+        constrained_timing = _run_gemm_timing("Constrained", gemm_parameters)
+    
+    assert full_timing  < constrained_timing * 5, "GEMM performance is not changed with constrained SM count"
--- a/transformer_engine/common/gemm/rocm_gemm.cu
+++ b/transformer_engine/common/gemm/rocm_gemm.cu
@@ -35,6 +35,8 @@ template<typename T>
 struct CacheEntry {
   T value;
   hipEvent_t event;
+
+  constexpr CacheEntry() : value(), event(nullptr) {}
   
   bool isValid() const { return event != nullptr; }
 
@@ -52,16 +54,9 @@ struct CacheEntry {
     {
       return false;
     }
-    else if (err == hipErrorCapturedEvent)
-    {
-      std::cerr << "[ERROR] hipEventQuery failed: " << hipGetErrorString(err) << std::endl;
-      NVTE_ERROR("Invalid event");
-      return false;
-    }
     else
     {
-      std::cerr << "[ERROR] hipEventQuery failed with error " << err << ": " << hipGetErrorString(err) << std::endl;
-      NVTE_ERROR("Invalid event");
+      NVTE_ERROR("Invalid event: err=", std::to_string(err), " ", hipGetErrorString(err));
       return false;
     }
   }
@@ -71,7 +66,7 @@ template<typename T, typename K>
 class ObjCache {
 public:
   using Data = std::unordered_map<K, std::unordered_map<hipStream_t, CacheEntry<T>>>;
-  static constexpr CacheEntry<T> invalidEntry = {};
+  static constexpr CacheEntry<T> invalidEntry{};
 
   const CacheEntry<T>& get(const K& key, const hipStream_t stream) const
   {
@@ -120,7 +115,7 @@ public:
     data[key][stream] = item; 
   }
 
-  ObjCache(void (*offload)(const Data&)): offload(offload) {}
+  ObjCache(void (*a_offload)(const Data&)): offload(a_offload) {}
 
   ~ObjCache()
   {
@@ -157,7 +152,7 @@ class ObjPool: public ObjCache<T, K> {
       {
         for (const auto &it2: it.second)
         {
-          this->data[it.first][it2.first] = it2.second;
+          ObjCache<T, K>::set(it.first, it2.first, it2.second);
         }
       }
     }
@@ -1181,7 +1176,6 @@ void hipblaslt_gemm(const Tensor *inputA,
 
         NVTE_CHECK_CUDA(hipStreamSynchronize(stream));
         hipStream_t &profilingStream = stream; // Reuse the stream for profiling
-        NVTE_CHECK_CUDA(hipStreamCreateWithFlags(&profilingStream, hipStreamNonBlocking));
         using tuning_clock = std::chrono::steady_clock;
         tuning_clock::now(); //the first call takes little longer so do it outside the loop
         tuning_clock::duration bestTime = tuning_clock::duration::max();
@@ -1361,13 +1355,16 @@ bool get_service_stream(int math_sm_count, hipStream_t stream, struct ServiceStr
   }
 
   ServiceStreamKey key = make_service_stream_key(device_id, math_sm_count);
-  auto streamEntry = service_stream_cache.get(key, stream);
+  CacheEntry<hipStream_t> streamEntry = service_stream_cache.get(key, stream);
   if (!streamEntry.isValid()) {
     /* There is no entry in the cache, try the following:
       * 1. Try to acquire any available stream form the cache.
       * 2. If not available, try to acquire any available stream form the pool.
-      * 3. If still not available, create a new stream and event.
-    */
+      * 3. If still not available, create a new stream and event. */
+    bool b_log = false;
+    if (const char* env_p = std::getenv("NVTE_LOG_MATH_SM_COUNT") ) {
+      b_log = (env_p != nullptr) && (std::string(env_p) == "1");
+    }
     streamEntry = service_stream_cache.acquire(key, stream);
     if (!streamEntry.isValid()) {
       streamEntry = service_stream_pool.acquire(key, stream);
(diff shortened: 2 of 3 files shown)


tests/pytorch/test_gemm_sm_count.py
```diff
@@ -0,0 +1,53 @@
...
+        start.record()
+        general_gemm(**gemm_parameters)
+        end.record()
+        torch.cuda.current_stream().synchronize()
+        timing.append(start.elapsed_time(end))
+    
+    timing = timing[2:]  # Discard the initial timings to avoid warm-up effects
+    mean = sum(timing) / len(timing)
+    disp = [abs(t - mean) for t in timing]
+    logging.info(f"{name} timing: {mean} ms with variance {sum(disp)/len(timing)}")
```
wangye805: nit: this disp formula is called mean absolute deviation. In statistics, when people say variance, it usually means sqrt(avg(t-mean)^2). We can name it deviation?

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -27,10 +27,149 @@
...
+    {
+      return true;
+    }
+    else if (err == hipErrorNotReady)
+    {
+      return false;
+    }
+    else if (err == hipErrorCapturedEvent)
+    {
+      std::cerr << "[ERROR] hipEventQuery failed: " << hipGetErrorString(err) << std::endl;
```
wangye805: Why do we need both std::cerr and NVTE_ERROR? Can we just use NVTE_ERROR to throw one single error message?

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -27,10 +27,149 @@
...
 #include "../util/logging.h"
 
+namespace transformer_engine {
+
 namespace {
 
+template<typename T> 
+struct CacheEntry {
+  T value;
+  hipEvent_t event;
```
wangye805, line 37: For safety and compatibility with different c++ version, shall we make a constructor to make event initialized to nullptr if not specified?

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -27,10 +27,149 @@
...
+    }
+    return invalidEntry;
+  }
+
+  void set(const K& key, hipStream_t stream, const CacheEntry<T>& item)
+  { 
+    data[key][stream] = item; 
+  }
+
+  ObjCache(void (*offload)(const Data&)): offload(offload) {}
```
wangye805: Emm, it took me quite sometime to understand the these two offload: one refer to the member function defined in protected, one refer to the argument. Can we distinguish them by make all member variable end with "_" or start with "m", or some other special notation?

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1048,7 +1180,7 @@ void hipblaslt_gemm(const Tensor *inputA,
                     << tuneLoopCount << " loops " << std::endl;
 
         NVTE_CHECK_CUDA(hipStreamSynchronize(stream));
-        hipStream_t profilingStream;
+        hipStream_t &profilingStream = stream; // Reuse the stream for profiling
         NVTE_CHECK_CUDA(hipStreamCreateWithFlags(&profilingStream, hipStreamNonBlocking));
```
wangye805: Once we reuse the stream, do we still need hipStreamCreateWithFlags?

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1175,16 +1306,119 @@ void hipblaslt_gemm(const Tensor *inputA,
...
+  NVTE_CHECK_CUDA(hipGetDevice(&device_id));
+  NVTE_CHECK_CUDA(hipDeviceGetAttribute(&device_cu_count, hipDeviceAttributeMultiprocessorCount, device_id));
+  if (math_sm_count < 0 || math_sm_count > device_cu_count)
+  {
+    std::cerr << "[WARNING] Invalid math_sm_count: " << math_sm_count << std::endl;
+    return false; // Invalid math_sm_count
+  }
+  else if (math_sm_count == device_cu_count)
+  {
+    return false; // math_sm_count == device_cu_count is equivalent to math_sm_count == 0
```
wangye805, line 1345: Why math_sm_count==device_cu_count is equivalent to math_sm_count==0?

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1175,16 +1306,119 @@ void hipblaslt_gemm(const Tensor *inputA,
...
+      if (math_sm_count % 32 != 0)
+      {
+        mask[maskSize-1] = (1UL << (math_sm_count % 32)) - 1;
+      }
+      NVTE_CHECK_CUDA(hipExtStreamCreateWithCUMask(&streamEntry.value, maskSize, mask.data()));
+      NVTE_CHECK_CUDA(hipEventCreateWithFlags(&streamEntry.event, hipEventDisableTiming));
+      std::cout << "[DEBUG] Created service stream for device " << device_id
+                << " with " << math_sm_count << " CUs" << std::endl;
+    }
+  service_stream_cache.set(key, stream, streamEntry);
```
wangye805: nit: indent not correct for this line

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1175,16 +1306,119 @@ void hipblaslt_gemm(const Tensor *inputA,
...
+    {
+      const uint32_t maskSize = (math_sm_count + 31) / 32;
+      std::vector<uint32_t> mask(maskSize, (uint32_t)-1);
+      if (math_sm_count % 32 != 0)
+      {
+        mask[maskSize-1] = (1UL << (math_sm_count % 32)) - 1;
+      }
+      NVTE_CHECK_CUDA(hipExtStreamCreateWithCUMask(&streamEntry.value, maskSize, mask.data()));
+      NVTE_CHECK_CUDA(hipEventCreateWithFlags(&streamEntry.event, hipEventDisableTiming));
+      std::cout << "[DEBUG] Created service stream for device " << device_id
```
wangye805: Let's make an env to control the debug message printing. Maybe NVTE_LOG_GEMM_SM_COUNT?

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1175,16 +1306,119 @@ void hipblaslt_gemm(const Tensor *inputA,
...
                  int ldb, int ldd, bool transa, bool transb, bool grad,
                  void *workspace, size_t workspaceSize, bool accumulate, bool use_split_accumulator,
                  int math_sm_count, int m_split, int n_split, bool gemm_producer,
                  const Tensor *inputCounter, hipStream_t stream, int compute_stream_offset)
-{  
-  // ROCBLAS support has been removed. HIPBLASLT is the only GEMM backend supported.
-  
-  // Check compute_stream_offset valid.
+{
+  struct ServiceStreamCtl ss_ctl;
```
wangye805: nit: this definition looks like c-style? Can we directly use c++ style as "ServiceStreamCtl ss_ctl"?

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1175,16 +1306,119 @@ void hipblaslt_gemm(const Tensor *inputA,
...
+  NVTE_CHECK_CUDA(hipStreamIsCapturing(stream, &captureStatus));
+  if (captureStatus != hipStreamCaptureStatusNone)
+  {
+    std::cerr << "[WARNING] Cannot use math_sm_count with captured stream" << std::endl;
+    return false; // Cannot use service stream with captured stream
+  }
+
+  ServiceStreamKey key = make_service_stream_key(device_id, math_sm_count);
+  auto streamEntry = service_stream_cache.get(key, stream);
+  if (!streamEntry.isValid()) {
```
wangye805, line 1359: Based on my understanding, this implementation checks on host that whether an existing cu-masked stream exists with all tasks finished (by querying the associated hip event). If there is an existing cu-masked stream but with unfinished tasks, a new stream will be created. 

Why do we need such a stream with all tasks finished during querying? For general pytorch tasks, this makes sense as we want to start launching tasks asap without knowledge of other tasks in previously used stream. 

For most of LLM training tasks with TE, there will be a large time gap between host launching gemm kernels and device running gemms, maybe even 10~20 transformer layers. With 4 gemms in fwd and 8 gemms in bwd, we will need to create 40-160 streams in total. Also, in TE, only this gemm module will deal with this cache. 

Therefore, I believe we can add to existing streams without querying whether previous tasks finished or not

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -27,10 +27,149 @@
...
+    }
+
+    void store(const typename ObjCache<T, K>::Data &cache)
+    {
+      std::lock_guard<std::mutex> lock(mt);
+      for (const auto &it: cache)
+      {
+        for (const auto &it2: it.second)
+        {
+          this->data[it.first][it2.first] = it2.second;
```
wangye805: Maybe use the set function defined above?

wangye805 (changes requested):

tests/pytorch/test_gemm_sm_count.py
ipanfilo: Yes, I know that variance is sqrt(avg(d^2)) and thank you to to point to correct terminology

transformer_engine/common/gemm/rocm_gemm.cu
ipanfilo: Well, with our usage it does not rely on implicit initialization because it is initialized once to invalidEntry and all other cases use copy-constructor and explicit members assignment. However, there is no harm in default constructor in this case and I'll add it for consistency with isValid() method which relies on event value

transformer_engine/common/gemm/rocm_gemm.cu
ipanfilo: Because it means using all CUs which effectively removes limitation

transformer_engine/common/gemm/rocm_gemm.cu
ipanfilo: No, it first check if existing service stream exists for current stream and reuses it w/o conditions.
And when it does not exists, it tries to find idle service stream created for other streams.
Thus, we don't need to create 40-160 service streams unless client app creates so many of them or use dozens of threads.

transformer_engine/common/gemm/rocm_gemm.cu
wangye805: Thanks

transformer_engine/common/gemm/rocm_gemm.cu
ipanfilo: We don't. It was an error

transformer_engine/common/gemm/rocm_gemm.cu
wangye805: Ah, I see it here: https://github.com/ROCm/TransformerEngine/blob/475a0eec707934da1b4f3eb2872a0e7d673a6a19/transformer_engine/common/include/transformer_engine/gemm.h#L68. Thanks

transformer_engine/common/gemm/rocm_gemm.cu
wangye805: Ah, I see. Your implementation checks isValid() instead of isAvailable() for the stream. Got it. Thanks

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1175,16 +1300,130 @@ void hipblaslt_gemm(const Tensor *inputA,
...
+  NVTE_CHECK_CUDA(hipStreamWaitEvent(ctl.stream, ctl.start_event, 0));
+  return true; 
+}
+
+void release_service_stream(hipStream_t stream, struct ServiceStreamCtl &ctl)
+{
+    NVTE_CHECK_CUDA(hipEventRecord(ctl.end_event, ctl.stream));
+    NVTE_CHECK_CUDA(hipStreamWaitEvent(stream, ctl.end_event, 0));
+    //TODO: when event are really destroyed (documentation says on devide synchronize) and how much overhead is to create them
+    //May need to store event in eventPool and reuse them after thy are recorded
```
wangye805, line 1409: What's a typical cost for event create and destroy? If in the order of 20ns, probably we can keep the current codes unchanged. Thanks

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -27,10 +27,144 @@
...
+template<typename T, typename K>
+class ObjPool: public ObjCache<T, K> {
+  public:
+    const CacheEntry<T>& get(const K& key, const hipStream_t stream) const
+    {
+      std::lock_guard<std::mutex> lock(mt);
+      return ObjCache<T, K>::get(key, stream);
+    }
+
+    CacheEntry<T> acquire(const K& key, const hipStream_t stream, bool get_available = true)
```
wangye805, line 142: I saw the get_available is always true in our usage. Maybe we can get rid of it, unless we plan for future get_available=false usage?

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -27,10 +27,144 @@
...
+
+    if (auto itr = key_item.find(stream); itr != key_item.end())
+    {
+      auto ret = itr->second;
+      key_item.erase(itr);
+      return ret;
+    }
+    
+    if (!get_available)
+      return invalidEntry;
```
wangye805, line 101: Same here for get_available, if we always assign get_available=false, we can remove this if

transformer_engine/common/gemm/rocm_gemm.cu
ipanfilo: The idea behind using template here was to eventually switch hipblaslt handle caching to it too not to have two set of classes to do similar things. This parameter was added with bearing that in mind. So I'd like to keep it for now

transformer_engine/common/gemm/rocm_gemm.cu
ipanfilo: See my comment above

transformer_engine/common/gemm/rocm_gemm.cu
wangye805: Ah, okay, got it

wangye805 (approved):
LGTM

tests/pytorch/test_gemm_sm_count.py
```diff
@@ -0,0 +1,53 @@
...
+  
+    with torch.cuda.stream(torch.cuda.Stream()):
+        full_timing = _run_gemm_timing("Full", gemm_parameters)
+    
+    cu_count = torch.cuda.get_device_properties(torch.cuda.current_device()).multi_processor_count
+    os.environ['NVTE_EXT_MARGIN_SM'] = str(95 * cu_count // 100) # Use 5% of SMs
+    with torch.cuda.stream(torch.cuda.Stream()):
+        constrained_timing = _run_gemm_timing("Constrained", gemm_parameters)
+    
+    assert full_timing  < constrained_timing * 5, "GEMM performance is not changed with constrained SM count"
```
wenchenvincent: Why do we use `full_timing  < constrained_timing * 5`? Supposedly, while constrained timing use 5% CU, it would be mostly certain that `full_timing  < constrained_timing`

transformer_engine/common/gemm/rocm_gemm.cu
ipanfilo: It is rather in the order of 200ns

tests/pytorch/test_gemm_sm_count.py
ipanfilo: Good catch. It should be opposite full_timing * 5 < constrained_timing

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1175,16 +1300,130 @@ void hipblaslt_gemm(const Tensor *inputA,
...
                  int math_sm_count, int m_split, int n_split, bool gemm_producer,
                  const Tensor *inputCounter, hipStream_t stream, int compute_stream_offset)
-{  
-  // ROCBLAS support has been removed. HIPBLASLT is the only GEMM backend supported.
-  
-  // Check compute_stream_offset valid.
+{
+  ServiceStreamCtl ss_ctl;
+  bool use_service_stream =
+      (math_sm_count != 0) ? get_service_stream(math_sm_count, stream, ss_ctl) : false;
```
wenchenvincent, line 1425: `math_sm_count != 0` This condition might redundant as `get_service_stream` is already taking care of it.

wenchenvincent (approved):
LGTM. One minor comment to attend to.

transformer_engine/common/gemm/rocm_gemm.cu
ipanfilo: It is made this way not to call get_service_stream() every time - in most cases math_sm_count is 0. Indeed, get_service_stream also validates math_sm_count but its analysis is deeper.

PR #216, opened 2025-06-25
Added drop in replacement for tex.quantize

sudhu2k:
# Description

This PR introduces new functionality to the TransformerEngine, specifically providing Triton-based drop-in replacements for cast_transpose operations for tex.quantize method. This is motivated by the need to optimize and improve performance for quantization and dequantization routines within the framework, as cast_transpose is a critical component in these operations.

Fixes [12520](https://github.com/ROCm/frameworks-internal/issues/12520)

## Type of change

- [x ] New feature (non-breaking change which adds functionality)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Adds a new drop-in replacement for tex.quantize method to call the cast_transpose triton kernel
- Adds unit tests to test out the new quantize method in cast.py (transformer_engine.pytorch.cpp_extensions.cast)

# Checklist:

- [x ] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x ] The functionality is complete
- [x ] I have commented my code, particularly in hard-to-understand areas
- [ x] I have made corresponding changes to the documentation
- [x ] My changes generate no new warnings
- [x ] I have added tests that prove my fix is effective or that my feature works
- [ x] New and existing unit tests pass locally with my changes


sudhu2k:
From the Readme.rst file:

Known Issue with ROCm 6.4 PyTorch Release
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Using the docker image ``rocm/pytorch:rocm6.4_ubuntu22.04_py3.10_pytorch_release_2.5.1`` triggers a failure in the unit-test ``tests/pytorch/test_permutation.py`` (tracked in Jira ticket SWDEV-534311).  

Rebuilding PyTorch at commit ``f929e0d602a71aa393ca2e6097674b210bdf321c`` resolves the issue.

I'm rebuilding pytorch to run the tests again.

wangye805:
> From the Readme.rst file:
> 
> Known Issue with ROCm 6.4 PyTorch Release ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 
> Using the docker image `rocm/pytorch:rocm6.4_ubuntu22.04_py3.10_pytorch_release_2.5.1` triggers a failure in the unit-test `tests/pytorch/test_permutation.py` (tracked in Jira ticket SWDEV-534311).
> 
> Rebuilding PyTorch at commit `f929e0d602a71aa393ca2e6097674b210bdf321c` resolves the issue.
> 
> I'm rebuilding pytorch to run the tests again.

Which docker image are you using? I believe if we use the current CI image, we should be fine

sudhu2k:
> > From the Readme.rst file:
> > Known Issue with ROCm 6.4 PyTorch Release ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> > Using the docker image `rocm/pytorch:rocm6.4_ubuntu22.04_py3.10_pytorch_release_2.5.1` triggers a failure in the unit-test `tests/pytorch/test_permutation.py` (tracked in Jira ticket SWDEV-534311).
> > Rebuilding PyTorch at commit `f929e0d602a71aa393ca2e6097674b210bdf321c` resolves the issue.
> > I'm rebuilding pytorch to run the tests again.
> 
> Which docker image are you using? I believe if we use the current CI image, we should be fine

I'm using compute-artifactory.amd.com:5000/rocm-plus-docker/framework/private/te-ci:rocm-6.4_0_ubuntu22_py310_torch25_jax0435qa_fa273

tests/pytorch/test_cast.py
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
```
wangye805: Let's put all triton kernel pytests under tests/pytorch/triton_kernels/. 

Also, let's remove the old test_cast_transpose.py under this dir or rename it to this test_cast.py

tests/pytorch/test_cast.py
```diff
@@ -0,0 +1,51 @@
...
+# License for AMD contributions = MIT. See LICENSE for more information
+
+import os
+import pytest
+import torch
+
+from transformer_engine.pytorch.triton_kernels.cast import quantize_triton
```
wangye805: Lets follow the other triton kernels naming convention to name it te_quantize_triton, instead of quantize_triton

tests/pytorch/test_cast.py
```diff
@@ -0,0 +1,51 @@
...
+from transformer_engine.pytorch.tensor.float8_tensor import Float8Quantizer
+import transformer_engine_torch as tex
+from triton_kernels.test_common import fill_uniform, get_tolerances
+
+@pytest.mark.parametrize("shape",
```
wangye805: I thought we will make shapes same as that in cpp unit tests: https://github.com/ROCm/TransformerEngine/blob/236178e5d758727fc93a40224c11b829bece358e/tests/cpp/operator/test_cast.cu#L76-L94

tests/pytorch/test_cast.py
```diff
@@ -0,0 +1,51 @@
...
+                        (217, 256),
+                        ])
+@pytest.mark.parametrize("in_dtype", [torch.float32, torch.float16, torch.bfloat16])
+@pytest.mark.parametrize("out_dtype", [tex.DType.kFloat8E4M3, tex.DType.kFloat8E5M2])
+def test_quantize(shape, in_dtype, out_dtype):
+    input_tensor = fill_uniform(shape, dtype=in_dtype)
+
+    scale_tensor = torch.rand(1, dtype=torch.float32, device='cuda') * 3.0 - 2.0
+    amax_tensor = torch.zeros(1, dtype=torch.float32, device='cuda')
+    quantizer = Float8Quantizer(scale=scale_tensor, amax=amax_tensor, fp8_dtype=out_dtype)
```
wangye805: We can name one quantizer to tex_quantizer, the other quantizer to be triton_quantizer, instead of quantizer2 for readability

tests/pytorch/test_cast.py
```diff
@@ -0,0 +1,51 @@
...
+
+    output_tensor  = quantize_triton(input_tensor, quantizer=quantizer)
+    
+    quantizer2 = Float8Quantizer(scale=scale_tensor, amax=amax_tensor, fp8_dtype=out_dtype)
+    te_quantized_out = tex.quantize(input_tensor, quantizer2)
```
wangye805: Both runs are using te. One is pytorch c++ extension, one is using triton. We can distinguish them by tex (extension) and triton. Lets name them quantized_out_tex and quantized_out_triton

tests/pytorch/test_cast.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # Check for cyclic imports
+    os.environ['NVTE_USE_CAST_TRANSPOSE_TRITON'] = '1'
+    quantizer2.quantize(input_tensor)
```
wangye805: Why do we need that?

tests/pytorch/test_cast.py
```diff
@@ -0,0 +1,51 @@
...
+    
+    atol, rtol = get_tolerances(in_dtype)
+    assert torch.allclose(output_tensor, te_quantized_out, atol=atol, rtol=rtol), 'Quantized results do not match!'
```
wangye805: output_tensor and te_quantized_out are both python QuantizedTensor object. How does torch.allclose work on them? Does it check both the cast output and transpose output?

transformer_engine/pytorch/module/base.py
```diff
@@ -1049,7 +1050,11 @@ def get_weight_workspace(
             if hasattr(out, "quantize_"):
                 out.quantize_(tensor, noop_flag=skip_update_flag)
             else:
-                tex.quantize(tensor, quantizer, out, skip_update_flag)
+                use_cast_transpose_triton =  bool( int(os.environ.get('NVTE_USE_CAST_TRANSPOSE_TRITON', '0')) )
+                if use_cast_transpose_triton:
```
wangye805: Or more concisely:
quantize_func = te_quantize_triton if use_cast_transpose_triton else tex.quantize
then quanize_func(tensor, quantizer, out, skip_update_flag)

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -80,7 +82,11 @@ def update_quantized(
             src = src.contiguous()
 
         # Launch cast kernel
-        tex.quantize(src, self, dst, noop_flag)
+        use_cast_transpose_triton =  bool( int(os.environ.get('NVTE_USE_CAST_TRANSPOSE_TRITON', '0')) )
```
wangye805, line 87: Same here. Try to make it more concise

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -106,7 +112,7 @@ def make_empty(
         # Allocate FP8 data transpose if needed
         data_transpose = None
         if self.columnwise_usage:
-            inner_dim = data.size(-1)
+            inner_dim = data.size(-1) if len(data.size()) > 0 else 1
```
wangye805: Why do we want to change it? 

Any changes applied just to our rocm flow need to use IS_HIP_EXTENSION to guard it. We need to make sure the upstream NV flow do not change

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -59,7 +61,11 @@ def update_quantized(
             src = src.contiguous()
 
         # Launch cast kernel
-        tex.quantize(src, self, dst, noop_flag)
+        use_cast_transpose_triton =  bool( int(os.environ.get('NVTE_USE_CAST_TRANSPOSE_TRITON', '0')) )
```
wangye805: Same here, make it more concise

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -6,6 +6,7 @@
 from __future__ import annotations
 from collections.abc import Iterable
 import math
+import os
```
wangye805: Do not forget to update our copyright info for some new files we touched

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -6,6 +6,7 @@
 from __future__ import annotations
```
wangye805: In fact, we don't have any mxfp8 cast_transpose or quantize kernels for rocm TE. We shouldn't change this file.

transformer_engine/pytorch/tensor/quantized_tensor.py
```diff
@@ -190,7 +191,13 @@ def forward(
         quantizer: Quantizer,
     ) -> QuantizedTensor:
         # pylint: disable=missing-function-docstring
-        return tex.quantize(tensor, quantizer)
+        use_cast_transpose_triton =  bool( int(os.environ.get('NVTE_USE_CAST_TRANSPOSE_TRITON', '0')) )
```
wangye805, line 197: Same here, to make it more concise

transformer_engine/pytorch/triton_kernels/cast.py
```diff
@@ -0,0 +1,72 @@
...
+    # Construct no-op flag if needed
+    if noop_flag is None:
+        noop_flag = _empty_tensor()
+
+    if out.size().numel() == 0:
+        # Return empty output if the quantized tensor has no elements
+        return out
+        
+    if input_tensor.nelement() > 0:
+        if out.get_metadata()["data_transpose"] is not None:
```
wangye805: Do we need the scaling_mode filtering like the switch-case statement: https://github.com/ROCm/TransformerEngine/blob/236178e5d758727fc93a40224c11b829bece358e/transformer_engine/common/util/cast_kernels.cuh#L1251-L1280

wangye805: And also, do we need to first make sure the output is fp8?

transformer_engine/pytorch/triton_kernels/cast_transpose.py
```diff
@@ -69,19 +69,27 @@ def _cast_transpose_triton(A, noop_ptr, C, T, stride_am, stride_an, stride_bn, s
 
 def te_cast_transpose_noop_triton(input, noop_flag, input_scale, cast_out, trans_out, amax_out, scale_inv_out, otype):
 
-    M, N = input.shape
+    row_length = input.shape[-1] if len(input.shape) > 0 else 1
```
wangye805, line 74: Why do we need to change this file?

wangye805:
We also need to guard this non_tn_fp8_gemm_supported() function (https://github.com/ROCm/TransformerEngine/blob/236178e5d758727fc93a40224c11b829bece358e/transformer_engine/pytorch/csrc/extensions/util.cpp#L12) because it's used to determine whether to generate transpose data in quantizer: https://github.com/ROCm/TransformerEngine/blob/236178e5d758727fc93a40224c11b829bece358e/transformer_engine/pytorch/csrc/extensions/quantizer.cpp#L106-L140

wangye805 (changes requested):

transformer_engine/pytorch/triton_kernels/cast_transpose.py
sudhu2k: Because some tests were passing input of shape with dimensions more than 2, our kernel wrappers were assuming the input to be of shape M, N. Pasting the message I sent in the group chat:

Hi!, I was running sanity checks on the integration with NVTE_USE_CAST_TRANSPOSE_TRITON = 1. I saw that one of the sanity checks were using input of size -> (config.seq_len, config.batch_size, config.hidden_size)
 
But the method:
te_cast_transpose_noop_triton(input, noop_flag, input_scale, cast_out, trans_out, amax_out, scale_inv_out, otype):
 
    M, N = input.shape
 
is written to operate on 2D shape. Seems like tex.quantize can handle any shape, should I try reshaping it?
But then, the kernel expects proper strides, so what should I do here?
Reshaped and modified the stride, the tests are passing now.

tests/pytorch/test_cast.py
sudhu2k: If you check the quantized_tensor.py line 197, 
https://github.com/ROCm/TransformerEngine/blob/fe6d21b324bdb288d1fc47cedc553a5e223b79b8/transformer_engine/pytorch/tensor/quantized_tensor.py#L194-L200

While integrating triton quantize, I had to import quantize function from cast.py inside _QuantizeFunc and cast.py also imports quantized_tensor.py for the Quantizer abstract class. This created a cyclic import and the line `quantizer2.quantize(input_tensor)` was throwing an error because it uses _QuantizeFunc forward method. Ideally _QuantizeFunc should be in a different file. I added quantizer2.quantize(input_tensor) just in case anyone tried modifying the file ignoring the cyclic import. Let me know what you think

tests/pytorch/test_cast.py
sudhu2k: **output_tensor and te_quantized_out are both python QuantizedTensor object** -- Right! 

When all_close method is called with the args, the flow is like -> https://github.com/ROCm/TransformerEngine/blob/fe6d21b324bdb288d1fc47cedc553a5e223b79b8/transformer_engine/pytorch/tensor/float8_tensor.py#L350

calls
https://github.com/ROCm/TransformerEngine/blob/fe6d21b324bdb288d1fc47cedc553a5e223b79b8/transformer_engine/pytorch/tensor/quantized_tensor.py#L352
calls 

https://github.com/ROCm/TransformerEngine/blob/fe6d21b324bdb288d1fc47cedc553a5e223b79b8/transformer_engine/pytorch/tensor/quantized_tensor.py#L374-L377

 Which in turn calls arg.dequantize(dtype=arg.dtype) to dequantize the cast output. 
 
 
 It doesn't check the transpose, but it only checks the cast part of it. I guess we should have a check for transpose too. I'll look into that!

transformer_engine/pytorch/triton_kernels/cast.py
sudhu2k: Yes, I think I should add the case statements to even check if it's a regular fp8 or not. For the scale mode filtering, we would JUST be checking the instance of the quantizer object.
I'll also add an error that mxfp8 isn't supported yet for the triton function.

transformer_engine/pytorch/tensor/float8_tensor.py
sudhu2k: There is a testcase, which passes ndim as 0 (tests a scalar input). In that case accessing index -1 would throw index error.

data_transpose = None
        if self.columnwise_usage:
>           inner_dim = data.size(-1)
E           IndexError: Dimension specified as -1 but tensor has no dimensions

So I added this check. But this was incorrect, as the shape of the transpose were varying. I've fixed this in the new commit.

tests/pytorch/test_cast.py
wangye805: I still does not get your point. 

If you remove these two lines, will we get cyclic import issue? If so, we will need to come up with a better way to resolve it because these two lines seems to be irrelevant to this test. In other words, it is confusing to me why we want to quantize the input tensor in this test. 

If you just want to test whether we still have cyclic import issue, we should start another pytest to test it dedicatedly.

tests/pytorch/test_cast.py
wangye805: Thanks for addressing this. In addition to your current torch.allclose check on _transpose, can you also check whether they are both none? I didn't follow the details but I was quite curious about the transpose output for just one dimensional shape like (16) and (16000), and the 4-dimensional shape (5, 4, 3, 160)?

transformer_engine/pytorch/tensor/float8_tensor.py
wangye805: In the new commit, did you skip the scalar testcase? Or how do you resolve it?

transformer_engine/pytorch/triton_kernels/cast_transpose.py
wangye805: Are you saying that your change can make te_cast_transpose_noop_triton function now support not just two-dimensional inputs, but also scalar, 1-dim, and higher dimensional inputs? If so, let's comment on this function and summarize in the PR message

tests/pytorch/triton_kernels/test_cast.py
```diff
@@ -0,0 +1,56 @@
...
+    # Check for cyclic imports
+    os.environ['NVTE_USE_CAST_TRANSPOSE_TRITON'] = '1'
+    tex_quantizer.quantize(input_tensor)
+    
+    atol, rtol = get_tolerances(in_dtype)
+    assert torch.allclose(quantized_out_triton, quantized_out_tex, atol=atol, rtol=rtol), 'Quantized results do not match!'
+    assert torch.allclose(
+        quantized_out_triton._transpose, quantized_out_tex._transpose, atol=0.0, rtol=0.0
+    ), 'Transposed quantized results do not match!'
+    assert torch.allclose(quantized_out_triton._get_quantizer().scale, quantized_out_tex._get_quantizer().scale, atol=atol, rtol=rtol), 'Scale results do not match!'
```
wangye805: Do we have different scale and scale inv for the _transpose part?

tests/pytorch/triton_kernels/test_cast.py
```diff
@@ -0,0 +1,56 @@
...
+    quantized_out_tex = tex.quantize(input_tensor, tex_quantizer)
+
+    # Check for cyclic imports
+    os.environ['NVTE_USE_CAST_TRANSPOSE_TRITON'] = '1'
+    tex_quantizer.quantize(input_tensor)
+    
+    atol, rtol = get_tolerances(in_dtype)
+    assert torch.allclose(quantized_out_triton, quantized_out_tex, atol=atol, rtol=rtol), 'Quantized results do not match!'
+    assert torch.allclose(
+        quantized_out_triton._transpose, quantized_out_tex._transpose, atol=0.0, rtol=0.0
```
wangye805: Why do we use atol=0 and rtol=0 for the _transpose part comparison? Did our cpp unit tests also enforce the exact match?

tests/pytorch/triton_kernels/test_cast.py
```diff
@@ -0,0 +1,56 @@
...
+    
+    tex_quantizer = Float8Quantizer(scale=scale_tensor, amax=amax_tensor, fp8_dtype=out_dtype)
+    quantized_out_tex = tex.quantize(input_tensor, tex_quantizer)
+
+    # Check for cyclic imports
+    os.environ['NVTE_USE_CAST_TRANSPOSE_TRITON'] = '1'
+    tex_quantizer.quantize(input_tensor)
+    
+    atol, rtol = get_tolerances(in_dtype)
+    assert torch.allclose(quantized_out_triton, quantized_out_tex, atol=atol, rtol=rtol), 'Quantized results do not match!'
```
wangye805: As you already mentioned, torch.allclose will issue a dequantize for the quantized_tensor object, can we directly compare their contents, instead of the dequantized output?

wangye805 (changes requested):
How about the "te_cast_transpose_dbias_triton" integration? Do we want to leave it for the next PR?

transformer_engine/pytorch/triton_kernels/cast_transpose.py
sudhu2k: Yes! It supports scalar, 1-dim and higher dim inputs. I'll add a comment.

sudhu2k:
> How about the "te_cast_transpose_dbias_triton" integration? Do we want to leave it for the next PR?

Does it mean the implementation of dbias_dact and other activations from cast.cpp? If so, I think we can leave it for another PR. Let me know you think.

tests/pytorch/triton_kernels/test_cast.py
sudhu2k: We can compare the uint8 bit level representation of the fp8 values, but I think if the dequantized outputs match then the bit level uint8 representations would also match. But note that we are doing bit level uint8 comparison for the transpose though.

tests/pytorch/triton_kernels/test_cast.py
sudhu2k: Since the _transpose part is all uint8 bit level representations of the fp8 values, I'm comparing the unit8 values of the _transpose with atol and rtol of 0. So, I'm expecting exact same bit level representations of fp8 type. The cpp unit tests didn't check the transpose values, it just compares output, amax, scale_inv.

tests/pytorch/triton_kernels/test_cast.py
sudhu2k: No we don't, the scale and scale_inv is shared for the original and transpose data.

tests/pytorch/test_cast.py
sudhu2k: > If you remove these two lines, will we get cyclic import issue?

No, it is just to verify and to make sure to throw error for any future developers who tries to move the import statement in quantized_tensor.py.
But yeah, I guess it would be another pytest and it might be a trivial issue as it might would be caught while developing or by any other test using cast.py. I'll remove it for now.

tests/pytorch/test_cast.py
sudhu2k: Yes I can add the None checks for the transpose.
About the shapes of the transpose, it goes as follows:

shape: (16)
transpose_shape: (16)

shape: (16000)
transpose_shape: (16000)

shape: (768, 1024)
transpose_shape: (1024, 768)

shape: (1, 128)
transpose_shape: (128, 1)

shape: (5, 4, 3, 160)
transpose_shape: (160, 5, 4, 3)

When I was working on this, I thought transpose of shape (16) would be (1, 16) and that's what I was getting initially but had to change to match the cpp output.
Also, for multi-dimensional output, I believe we can take transpose from any dimension, I'm not sure why it's the last dim first and rest of the dim later.

transformer_engine/pytorch/tensor/float8_tensor.py
sudhu2k: I resolved it using a function that I wrote inside the cast.py file _setup_conditional_transpose_storage.

https://github.com/ROCm/TransformerEngine/blob/523accb3d2b8a0cfe1012724440cd46a83eed795/transformer_engine/pytorch/triton_kernels/cast.py#L24-L47

https://github.com/ROCm/TransformerEngine/blob/523accb3d2b8a0cfe1012724440cd46a83eed795/transformer_engine/pytorch/triton_kernels/cast.py#L68-L75

tests/pytorch/test_cast.py
wangye805: Sounds good. Thanks. Let's just make it the same as the hipified cuda kernel

wangye805:
> > How about the "te_cast_transpose_dbias_triton" integration? Do we want to leave it for the next PR?
> 
> Does it mean the implementation of dbias_dact and other activations from cast.cpp? If so, I think we can leave it for another PR. Let me know you think.

Yeah, we can leave it for the next PR

wangye805 (approved):
LGTM

wangye805 (approved):

PR #230, opened 2025-07-14
Initial implementation of FP8 RMSNorm support

Micky774:
# Description

Implements FP8 support for RMSNorm Triton kernel.

Fixes https://github.com/ROCm/frameworks-internal/issues/12855

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Updates `test_fusible_ops::test_rmsnorm` to more explicitly check quantized results
- Updates `_rmsnorm_fwd_triton` with support for FP8 output quantization
- Updates `te_rmsnorm_fwd_triton` with support for FP8 output quantization

# Checklist:

### DIFF
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -1,5 +1,5 @@
 # Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-#
+# Copyright (c) 2022-2025, Advanced Micro Devices, Inc. All rights reserved.
 # See LICENSE for license information.
 
 from __future__ import annotations
@@ -16,9 +16,6 @@ import transformer_engine.common.recipe
 import transformer_engine.pytorch as te
 from transformer_engine.pytorch.fp8 import FP8GlobalStateManager
 import transformer_engine.pytorch.ops as te_ops
-from transformer_engine.pytorch.triton_kernels.common import (
-    te_dtype_to_torch_dtype,
-)
 from transformer_engine.pytorch.ops.fused import (
     BackwardLinearAdd,
     ForwardLinearBiasActivation,
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -23,6 +23,7 @@ from transformer_engine.pytorch.ops.fused import (
 )
 from transformer_engine.pytorch.tensor import QuantizedTensor
 from transformer_engine.pytorch.tensor.float8_tensor import Float8Tensor, Float8Quantizer
+from transformer_engine.pytorch.tensor.mxfp8_tensor import MXFP8Tensor, MXFP8Quantizer
 from transformer_engine.pytorch.utils import is_bf16_compatible
 import transformer_engine_torch as tex
 
@@ -1252,11 +1253,16 @@ class TestBasicOps:
         assert y_test.dtype == dtype
         # Expected numerical error
         tols = dtype_tols(dtype)
+
+        # Explicit checks for quantization
         if quantized_compute:
             tols = dtype_tols(y_test._quantizer.dtype)
-            assert isinstance(y_test, Float8Tensor)
+            expected_tensor_cls = {
+                Float8Quantizer:Float8Tensor,
+                MXFP8Quantizer:MXFP8Tensor
+            }[type(y_test._quantizer)]
+            assert isinstance(y_test, expected_tensor_cls)
             y_test = y_test.dequantize(dtype=torch.float32)
-            y_test = y_test.to(device="cpu")
 
         # Check results
         y_test = y_test.to(dtype=torch.float64, device="cpu")
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -1,5 +1,7 @@
-# Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# This file was modified for portability to AMDGPU
 # Copyright (c) 2022-2025, Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
 # See LICENSE for license information.
 
 from __future__ import annotations
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -7,18 +7,18 @@ from __future__ import annotations
 from collections.abc import Iterable
 import math
 from typing import Optional
-import os
 
 import pytest
 import torch
-from torch.utils.cpp_extension import IS_HIP_EXTENSION
 
 import transformer_engine
 import transformer_engine.common.recipe
 import transformer_engine.pytorch as te
 from transformer_engine.pytorch.fp8 import FP8GlobalStateManager
 import transformer_engine.pytorch.ops as te_ops
-from transformer_engine.pytorch.ops._common import is_float8_tensor
+from transformer_engine.pytorch.triton_kernels.common import (
+    te_dtype_to_torch_dtype,
+)
 from transformer_engine.pytorch.ops.fused import (
     BackwardLinearAdd,
     ForwardLinearBiasActivation,
@@ -1256,7 +1256,7 @@ class TestBasicOps:
         # Expected numerical error
         tols = dtype_tols(dtype)
         if quantized_compute:
-            tols = dtype_tols(tex.DType.kFloat8E4M3)
+            tols = dtype_tols(y_test._quantizer.dtype)
             assert isinstance(y_test, Float8Tensor)
             y_test = y_test.dequantize(dtype=torch.float32)
             y_test = y_test.to(device="cpu")
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -1252,14 +1252,16 @@ class TestBasicOps:
             y_test = forward(x_test)
         y_test.backward(dy_test)
 
+        assert y_test.dtype == dtype
         # Expected numerical error
         tols = dtype_tols(dtype)
         if quantized_compute:
             tols = dtype_tols(tex.DType.kFloat8E4M3)
             assert isinstance(y_test, Float8Tensor)
+            y_test = y_test.dequantize(dtype=torch.float32)
+            y_test = y_test.to(device="cpu")
 
         # Check results
-        assert y_test.dtype == dtype
         y_test = y_test.to(dtype=torch.float64, device="cpu")
         dx_test = x_test.grad.to(dtype=torch.float64, device="cpu")
         dw_test = op.weight.grad.to(dtype=torch.float64, device="cpu")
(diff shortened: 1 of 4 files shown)


transformer_engine/pytorch/triton_kernels/rmsnorm.py
```diff
@@ -113,9 +137,50 @@ def _rmsnorm_fwd_triton(output_ptr, input_ptr, g_ptr, rsigma_ptr, input_row_stri
             rms_norm = row * norm_factor * g
 
             output_ptrs = output_ptr + row_idx * output_row_stride + col_offsets
-            output_ptrs = tl.multiple_of(output_ptrs, (16, ))
+            output_ptrs = tl.multiple_of(output_ptrs, (8, ))
+            if IS_FP8:
+                amax_temp = tl.max(tl.abs(rms_norm), axis=-1)
+                amax = amax_temp if amax_temp > amax else amax
+                rms_norm = rms_norm * scale
             tl.store(output_ptrs, rms_norm.to(output_ptr.type.element_ty), mask=mask)
```
Micky774: Currently getting a segmentation fault at this line (note the other `tl.store(output_ptrs, ...)` line would probably segfault too, but isn't directly tested right now. cc: @wenchenvincent @ipanfilo 

Edit: I suspect it's because Triton is finding the "fake dtype" from `output_ptr` since it is the `Float8Tensor` wrapper, and not a tensor directly.

Micky774:
@wenchenvincent @ipanfilo should be ready for review now. Note that currently the correctness of the `quantizer.amax` value is **not** tested (nor was it before this PR) and I'm not sure what the best way to test it would be. Once we finish the triton kernel refactors, maybe we could create a generic test comparing RMS/LayerNorm triton kernel quantization with `tex` quantization?

transformer_engine/pytorch/triton_kernels/rmsnorm.py
```diff
@@ -47,15 +66,15 @@ def _rmsnorm_fwd_triton(output_ptr, input_ptr, g_ptr, rsigma_ptr, input_row_stri
             for blk_idx in tl.range(0, n_cols_blks, num_stages=2):
                 cols = blk_idx * BLOCK_SIZE + col_offsets
                 input_ptrs = row_input_ptr + cols
-                input_ptrs = tl.multiple_of(input_ptrs, (16, ))
+                input_ptrs = tl.multiple_of(input_ptrs, (8, ))
```
ipanfilo: I wonder why do are we sure of this alignment and why is it changed?

Micky774: Honestly not sure -- I assumed it was originally in consideration of the smallest bit-width being 16 hence demoting to 8 to account for FP8 (at which point, is it really worth including at all?). Open to suggestions here.

ipanfilo: From my understanding, it sets pointer alignment hint so matching float16 bit-width is just coincident. I expect that with BLOCK_SIZE with FP8 is 2x of FP16 one, thus align should not decrease. Does it work with original value?

Micky774: Yes it does, as well as with 32 and 64...

tests/pytorch/test_fusible_ops.py
```diff
@@ -7,18 +7,18 @@
 from collections.abc import Iterable
 import math
 from typing import Optional
-import os
 
 import pytest
 import torch
-from torch.utils.cpp_extension import IS_HIP_EXTENSION
```
ipanfilo, line 14: Does this test match the upstream variant now?

Micky774: Yes, with minor additions to explicitly check quantization outcomes. I can remove the additions if we want to keep it as exact to the upstream test as possible. The use of `IS_HIP_EXTENSION` was added in the previous `rmsnorm` refactor PR I authored.

ipanfilo: If it is modified from upstream, there should be AMD copyright there

Micky774: Copyright statement added.

tests/pytorch/test_fusible_ops.py
```diff
@@ -1,24 +1,21 @@
 # Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-#
+# Copyright (c) 2022-2025, Advanced Micro Devices, Inc. All rights reserved.
```
ipanfilo, line 2: Please follow other upstream files for copyright format

Micky774: Does it look right now? I've used the same format as in `test_numerics.py`

transformer_engine/pytorch/triton_kernels/rmsnorm.py
```diff
@@ -91,7 +117,11 @@ def _rmsnorm_fwd_triton(output_ptr, input_ptr, g_ptr, rsigma_ptr, input_row_stri
                 g += 1
             rms_norm = x * norm_factor * g
             output_ptrs = row_output_ptr + cols
-            tl.store(output_ptrs, rms_norm.to(output_ptr.type.element_ty), mask=mask)
+            if IS_FP8:
+                amax_temp = tl.max(tl.abs(rms_norm), axis=-1)
+                amax = amax_temp if amax_temp > amax else amax
```
wenchenvincent: Not sure if this would result in a branch or not.
Try with `amax = tl.maximum(amax, amax_temp)`

wenchenvincent:
I think we also need to add tests in tests/pytorch/triton_kernels

transformer_engine/pytorch/triton_kernels/rmsnorm.py
```diff
@@ -114,8 +144,40 @@ def _rmsnorm_fwd_triton(output_ptr, input_ptr, g_ptr, rsigma_ptr, input_row_stri
...
+                rms_norm = rms_norm * scale
+            tl.store(output_ptrs, rms_norm.to(output_type), mask=mask)
+    if IS_FP8:
+        tl.store(amax_ptr + row_start, amax)
 
+@triton.jit
+def _rmsnorm_fwd_fp8_meta_triton(
```
wenchenvincent: I think we can fuse this kernel into the _rmsnorm_fwd_triton. Is it because you want to be able to handle both fp8 and mxfp8 so that it is not fused?

transformer_engine/pytorch/triton_kernels/rmsnorm.py
Micky774: Updated!

transformer_engine/pytorch/triton_kernels/rmsnorm.py
Micky774: It was mainly to avoid race conditions wrt `amax` since afaik Triton doesn't directly support synchronization barriers. Do you know of a better way to handle this?

wenchenvincent: Do you mean the race condition of writing to `amax` from different workgroups? The `atomic_max` should be able to avoid the race condition. Let me know if you need a bit more discussion on that.

Micky774: Updated

wenchenvincent:
@Micky774 What error were you seeing with using the quantizer?

Micky774:
I'm specifically getting 
```traceback
RuntimeError: /home/meekzain/TransformerEngine/transformer_engine/common/transformer_engine_hip.cpp:165 in function CheckOutputTensor: Assertion failed: t.scale_inv.dptr != nullptr. FP8 scaling factor output z_scale_inverse must be allocated
```

which _looks_ like it may be saying that the `tex` implementation is failing to set the appropriate metadata on the output? I'm not always getting this error, and I'm trying to sort out exactly what triggers it.


Edit: This seems to be because the `tex` implementation requires any `ln_out` provided to it to be carrying FP8 metadata. Removing this from tests for now.

Micky774:
Otherwise I'm also getting 
```traceback
RuntimeError: /home/meekzain/TransformerEngine/transformer_engine/common/normalization/common_hip.h:210 in function getKernel: Unavailable kernel for this normalization config.
```

Edit: this is caused by the choice of `fp8_dtype` -- it seems the `tex` implementation only supports `tex.DType.kFloat8E4M3`

Micky774:
@wenchenvincent Currently we fail to match the `tex` implementation based on the existing tolerances (by single bits), but when using the tolerances from `test_fusible_ops` we pass. Which tolerances do we prefer?

wenchenvincent:
> Otherwise I'm also getting
> 
> ```
> RuntimeError: /home/meekzain/TransformerEngine/transformer_engine/common/normalization/common_hip.h:210 in function getKernel: Unavailable kernel for this normalization config.
> ```
> 
> Edit: this is caused by the choice of `fp8_dtype` -- it seems the `tex` implementation only supports `tex.DType.kFloat8E4M3`

Right. I think for normalization kernels, only the forward kernel might output fp8. The backward kernel does not output fp8. E5M2 is only used for backward pass.

tests/pytorch/triton_kernels/test_rmsnorm.py
```diff
@@ -151,29 +163,29 @@ def test_rmsnorm_fwd_triton(M, N, in_dtype, out_dtype, zero_centered_gamma, quan
...
         input_tensor,
         gamma_tensor,
         epsilon,
-        ln_out_hipified,
+        None,
         quantizer_hip, torch_dtype_to_te_dtype(out_dtype),
         fwd_ln_sm_margin,
         zero_centered_gamma
     )
-    atol, rtol = get_tolerances(in_dtype)
```
wenchenvincent, line 149: I noticed that the tolerances was based on `in_dtype`. I think it should be based on `out_dtype`. If you change that, would the original tolerances work?

wenchenvincent:
> @wenchenvincent Currently we fail to match the `tex` implementation based on the existing tolerances (by single bits), but when using the tolerances from `test_fusible_ops` we pass. Which tolerances do we prefer?

`test_fusible_ops` might use higher tolerances as there might be more ops in the tests. As I mentioned in a comment, I noticed that previously, you set the tolerances was based on in_dtype. I think it should be based on out_dtype. If you change that, would the original tolerances work?

tests/pytorch/triton_kernels/test_rmsnorm.py
Micky774: No, even setting to the output dtype the tolerances are too tight. Specifically the `uint8` representations of the underlying data have a difference of 1 at most, but still fail under usual tolerances.

Micky774:
> > @wenchenvincent Currently we fail to match the `tex` implementation based on the existing tolerances (by single bits), but when using the tolerances from `test_fusible_ops` we pass. Which tolerances do we prefer?
> 
> `test_fusible_ops` might use higher tolerances as there might be more ops in the tests. As I mentioned in a comment, I noticed that previously, you set the tolerances was based on in_dtype. I think it should be based on out_dtype. If you change that, would the original tolerances work?

fwiw in `test_fusible_ops` they use the same tolerances even when comparing a reference implementations against the singular layers directly (e.g. the `TestBasicOps` class), so I'm not sure if it is a relaxation _necessarily_ due to fusing multiple ops.

Micky774:
@wenchenvincent actually we have that `torch.finfo(torch.float8_e4m3fn).eps = 0.125` so I think the FP8 tolerances from `test_fusible_ops` should be safe to use.

wenchenvincent (approved):
LGTM.

PR #246, opened 2025-08-04
Fix dropout when using a new-style rng

### LINKED ISSUE #244: [Issue]: TE dropout requires old-style PRNGs
### Problem Description

Jax changed the preferred type for RNG seeds from an array of ints to an opaque type in [August 2023](https://docs.jax.dev/en/latest/jep/9263-typed-keys.html#overview).

TE attention still requires the old-style keys in [transformer_engine/jax/cpp_extensions/attention.py](https://github.com/ROCm/TransformerEngine/blob/c97ecae53e8cd1d8b925530eef25036ef43f9181/transformer_engine/jax/cpp_extensions/attention.py#L208).
Any applications using new-style keys will see an error when this line of code attempts to convert the seed to an int.

It may be possible to work around this issue easily by changing the line to:
`seed = jax.random.key_data(seed).astype(self.rng_state_dtype)`

### Operating System

Ubuntu 22.04, Ubuntu 24.04

### CPU

Any

### GPU

MI300x series

### ROCm Version

Any 6.4

### ROCm Component

_No response_

### Steps to Reproduce

Run TE DotProductAttention with dropout != 0.0 (for example set it to 0.2) and a PRNG created with  jax.random.key() instead of jax.random.PRNGKey()

### (Optional for Linux users) Output of /opt/rocm/bin/rocminfo --support

_No response_

### Additional Information

_No response_

gabeweisz:
# Description

Previously TE would crash with a type error [at this line of code](https://github.com/ROCm/TransformerEngine/blob/c97ecae53e8cd1d8b925530eef25036ef43f9181/transformer_engine/jax/cpp_extensions/attention.py#L208) if new-style RNGs were used. This PR fixes the error by converting a new-style RNG to an int properly.

Fixes #244 

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Uses Jax's [recommended mechanism](https://docs.jax.dev/en/latest/jep/9263-typed-keys.html) to convert a new-style RNG to int data when a new-style RNG is used
- Includes tests for new-style RNGs

# Checklist:

### DIFF
--- a/tests/jax/test_fused_attn.py
+++ b/tests/jax/test_fused_attn.py
@@ -388,10 +388,15 @@ class FusedAttnRunner:
         self.cp_size = self.mesh.shape.get(self.mesh_resource.cp_resource, 1)
         self.tp_size = self.mesh.shape.get(self.mesh_resource.tp_resource, 1)
 
-        if self.use_old_rng:
-            key = jax.random.PRNGKey(0)
+        # only support new-style RNGs on AMD hardware since they will crash otherwise
+        if is_hip_extension():
+            if self.use_old_rng:
+                key = jax.random.PRNGKey(0)
+            else:
+                key = jax.random.key(0)
         else:
             key = jax.random.key(0)
+
         q_key, k_key, v_key, bias_key, dropout_key = jax.random.split(key, 5)
 
         q_shape = (self.batch_size, self.max_seqlen_q, self.num_heads_q, self.head_dim)
@@ -587,7 +592,7 @@ class FusedAttnRunner:
                 case _:
                     raise ValueError(f"Unknown {self.seq_desc_format=}")
 
-        self.dropout_rng = dropout_key
+        self.dropout_rng = dropout_key if self.dropout_prob > 0 else None
         self.scaling_factor = 1.0 / sqrt(self.head_dim)
 
         # Setup distributed sharding specs
@@ -630,10 +635,15 @@ class FusedAttnRunner:
             self.bias_pspec = PartitionSpec()
         self.bias_sharding = NamedSharding(self.mesh, self.bias_pspec)
 
-        if jnp.issubdtype(self.dropout_rng.dtype, jax.dtypes.prng_key):
-            self.dropout_rng_pspec = PartitionSpec()
+        # New-style RNG fix is only applied for AMD GPUs
+        if is_hip_extension():
+            if self.dropout_rng is not None and jnp.issubdtype(self.dropout_rng.dtype, jax.dtypes.prng_key):
+                self.dropout_rng_pspec = PartitionSpec()
+            else:
+                self.dropout_rng_pspec = PartitionSpec(None,)
         else:
             self.dropout_rng_pspec = PartitionSpec(None,)
+
         self.dropout_rng_sharding = NamedSharding(self.mesh, self.dropout_rng_pspec)
 
         self.logit_scale_pspec = PartitionSpec(None, None, self.mesh_resource.cp_resource, None)
@@ -863,7 +873,7 @@ class FusedAttnRunner:
 
             # Assume all batch has the same actual_seqlen, probably needs to extend the tests
             bias_mask = self.mask[0, 0]
-
+            
             # Assert all masked dbias are 0s
             assert_allclose(
                 jnp.where(bias_mask, primitive_dbias, 0),
@@ -936,11 +946,11 @@ class FusedAttnRunner:
         pytest.param(0.1, id="DROP_0.1"),
     ],
 )
+# Only testing old-style RNGs by default to reduce the # of tests but leaving the hooks in place
 @pytest.mark.parametrize(
     "use_old_rng",
     [
         pytest.param(True, id="Old-style rng"),
-        pytest.param(False, id="New-style rng"),
     ],
 )
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


gabeweisz:
@wangye805 - this RNG-fixing PR is ready. Do you have any comments on it?

tests/jax/test_fused_attn.py
```diff
@@ -931,6 +936,13 @@ def check_dqkv(primitive, reference, pad, idx):
         pytest.param(0.1, id="DROP_0.1"),
     ],
 )
+@pytest.mark.parametrize(
+    "use_old_rng",
+    [
+        pytest.param(True, id="Old-style rng"),
+        pytest.param(False, id="New-style rng"),
+    ],
```
Micky774, line 954: Adding this as an additional parameter doubles the test parameter space. Is this strictly necessary, or might it be sufficient to add a singular non-regression test to check the exact error that this PR aims to fix?

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -205,6 +205,10 @@ def check_seed(self, seed, dropout_probability, is_training):
                 f"casted to dtype {self.rng_state_dtype}. "
                 "Please use threefry/rbg/unsafe_rbg PRNG implementations to remove this warning."
             )
+            if jnp.issubdtype(seed.dtype, jax.dtypes.prng_key):
+                # New-style RNGs cannot directly be cast to integers
+                # See https://docs.jax.dev/en/latest/jep/9263-typed-keys.html for details
+                seed = jax.random.key_data(seed)
```
Micky774: While this serves as an immediate hotfix, we probably should address the underlying problem and update to avoid relying on directly accessing the raw buffer. Is this necessary due to a technical restraint, or is it just going to require some effort to update our consumption of the random seed?

tests/jax/test_fused_attn.py
gabeweisz: I thought it would be useful to test everything, but I can reduce the number of tests that use this. Since the error is a type-cast error, we probably only need one

transformer_engine/jax/cpp_extensions/attention.py
gabeweisz: My guess is that it would require a significant effort to change CK / AITER C++ libraries to change the consumption of the random seed.

gabeweisz: CK's C++ code requires a [64-bit int as a random seed](https://github.com/ROCm/composable_kernel/blob/b4f3487d8423a0e14bbb90e1cf8305d3560f3d17/example/ck_tile/01_fmha/fmha_fwd.cpp#L424). I am sure AITER is similar. I do not think Jax opaque PRNGs can be passed to C++ as a 64-bit int, but if you think they can we can report this to the CK/AITER team

gabeweisz: We could potentially use randint() instead of jax.random.key_data() to get seed for CK. I am not sure which version is better, but this version is suggested in the Jax docs

tests/jax/test_fused_attn.py
```diff
@@ -387,7 +388,10 @@ def _setup_inputs(self):
         self.cp_size = self.mesh.shape.get(self.mesh_resource.cp_resource, 1)
         self.tp_size = self.mesh.shape.get(self.mesh_resource.tp_resource, 1)
 
-        key = jax.random.PRNGKey(0)
+        if self.use_old_rng:
```
wangye805: Can we use the is_hip_extension() to guard our rocm side change?

tests/jax/test_fused_attn.py
```diff
@@ -583,7 +587,7 @@ def generate_random_segment_ids(
                 case _:
                     raise ValueError(f"Unknown {self.seq_desc_format=}")
 
-        self.dropout_rng = dropout_key if self.dropout_prob > 0 else None
+        self.dropout_rng = dropout_key
```
wangye805: Why do we need dropout_key when dropout_prob ==0 ?

tests/jax/test_fused_attn.py
```diff
@@ -626,9 +630,10 @@ def to_dp_shardings(x):
             self.bias_pspec = PartitionSpec()
         self.bias_sharding = NamedSharding(self.mesh, self.bias_pspec)
 
-        self.dropout_rng_pspec = PartitionSpec(
-            None,
-        )
+        if jnp.issubdtype(self.dropout_rng.dtype, jax.dtypes.prng_key):
```
wangye805: Same here, please use the is_hip_extension to guard the rocm side changes

tests/jax/test_fused_attn.py
```diff
@@ -858,7 +863,7 @@ def check_dqkv(primitive, reference, pad, idx):
 
             # Assume all batch has the same actual_seqlen, probably needs to extend the tests
             bias_mask = self.mask[0, 0]
-            
+
```
wangye805: Let's not introduce unnecessary reformatting

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -205,6 +205,10 @@ def check_seed(self, seed, dropout_probability, is_training):
                 f"casted to dtype {self.rng_state_dtype}. "
                 "Please use threefry/rbg/unsafe_rbg PRNG implementations to remove this warning."
             )
+            if jnp.issubdtype(seed.dtype, jax.dtypes.prng_key):
```
wangye805: Same here, use is_hip_extension to guard rocm side changes

wangye805 (changes requested):

tests/jax/test_fused_attn.py
gabeweisz: I will make that change - but I will point out that this code is broken on Nvidia TE too

tests/jax/test_fused_attn.py
gabeweisz: There was some code that failed when dropout_key was none

tests/jax/test_fused_attn.py
gabeweisz: My text editor did this - this was not on purpose

tests/jax/test_fused_attn.py
gabeweisz: The check on line 633 failed when it was non - I'll change it there

tests/jax/test_fused_attn.py
Micky774: That makes sense. @wangye805 do you think we can do without the hip guard in that case? Especially seeing as it would break on CPU backend too?

wangye805 (approved):

PR #253, opened 2025-08-06
FIX LayerNorm clamping

Micky774:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Clamp LayerNorm to avoid NaNs
- Add non-regression test for layernorm
- Fix bug where LayerNorm would create a transpose buffer even when quantizer doesn't request it
- Improved LayerNorm general testing
- Improved RMSNorm test explicitness

# Checklist:

### DIFF
--- a/tests/pytorch/triton_kernels/test_layernorm.py
+++ b/tests/pytorch/triton_kernels/test_layernorm.py
@@ -33,6 +33,7 @@ from test_common import (
     fill_uniform,
     get_tolerances,
     compare_results,
+    dtype_tols,
 )
 
 def compute_ref_stats(data: torch.Tensor, epsilon: float):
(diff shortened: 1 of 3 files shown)


Micky774:
Note that currently layernorm tests do not include cases where fake output dtypes are used in the creation of quantized tensors, so I think it would be worth taking some time to generally refactor the layernorm and rmsnorm fwd tests into a singular norm fwd test to avoid needing to keep the two sets of tests in sync with each other. We could probably do something similar for the bwd tests but I'm more familiar with the fwd hence why I'm only suggesting them. What do you think @wenchenvincent ?

wenchenvincent:
> Note that currently layernorm tests do not include cases where fake output dtypes are used in the creation of quantized tensors, so I think it would be worth taking some time to generally refactor the layernorm and rmsnorm fwd tests into a singular norm fwd test to avoid needing to keep the two sets of tests in sync with each other. We could probably do something similar for the bwd tests but I'm more familiar with the fwd hence why I'm only suggesting them. What do you think @wenchenvincent ?

Sounds good.

Micky774:
Great! This PR is just a quick patch, so I'll open a follow-up PR for the full refactor.

wenchenvincent (approved):
LGTM.

PR #255, opened 2025-08-07
Refactor Triton kernel tests

Micky774:
# Description

Refactors triton kernel tests (in particular `test_rmsnorm.py` and `test_layernorm.py`) into a common suite to minimize redundancy and increase robustness since they share similar API, use cases, and most importantly failure cases. This PR includes several bug-fixes to `te_layernorm_fwd_triton` and `te_rmsnorm_fwd_triton`

Fixes https://github.com/ROCm/frameworks-internal/issues/13256

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Removed `test_rmsnorm.py` and `test_layernorm.py`
- Introduced `TestNorms` test class into `tests/pytorch/triton_kernels/test_common.py`
- Introduced helper functions for comparison
- Unified old `fwd` and `bwd` tests
- Updated tests to include correct `out_dtype` semantics
- Updated `te_layernorm_fwd_triton` API to match HIP kernel implementation
- Updated `te_layernorm_fwd_triton` to correctly clear existing transpose caches when making new transpose
- Updated `te_layernorm_fwd_triton` to correctly handle being passed a quantized-tensor for `ln_out`
- Fixed transpose indexing bug in `_rmsnorm_fwd_triton` when using the blocked version of the algorithm
- Updated `te_compare_results` API to match that of `compare_results`
- Introduced `use_pytorch_semantics` kwarg in `te_compare_results` to toggle between upstream Cpp-style mismatch detection, and pytorch-style mismatch detection (both still include rounding error mitigation)
- Removed `compare_results` in favor of directly using `te_compare_results`
- Creates auto-tune wrapped, and auto-tune free implementations of underlying normalization kernels.
- Adds backwards-compatible `autotune=True` variable to `te_{rms, layer}norm_fwd_triton` functions
- Dispatches to auto-tune free implementations for testing resulting in significant CI speedup

# Checklist:


Micky774:
@wenchenvincent @ipanfilo Ready for review.

tests/pytorch/triton_kernels/test_common.py
```diff
@@ -13,16 +15,95 @@
...
+test_shapes = [
+    (2048, 4096),
+    (768, 2048),
+    (256, 1024),
+    (128, 768),
+    (64, 512),
+    (173, 409),
+    (71, 3571),
+    (29, 17389),
+]
```
Micky774: My only concern going into this is that the test shapes make the cross product of the parameter space very large, however we should keep them if we want parity with the CPP tests I guess.

Note that these are the shapes used for testing `RMSNorm` -- they use _different_ shapes for `LayerNorm` though I'm convinced these ought to be sufficient. I'd go so far as to suggest that we _cut down_ on a few of these for sake of run time.

Thoughts?

Micky774:
There are also one or two non-deterministic failures with the `LayerNorm` kernel, even with the relaxed tolerance...

tests/pytorch/triton_kernels/test_common.py
wenchenvincent: Very likely that some shapes were added by upstream to cover bugs exposed in production workload.

Micky774:
Note that the tests are passing -- there was an issue with the CI hence why it displays as still in progress.

Edit: current CI failures are unrelated

Micky774:
@wenchenvincent ready for review

Micky774:
Note that last passing sGPU test was off of [66c3057](https://github.com/ROCm/TransformerEngine/pull/255/commits/66c3057f7c13103bf8c8ae5a842cde7e08b157bc) and the subsequent changes have been structural rather than functional.

Micky774:
@wenchenvincent @ipanfilo @wangye805

I wanted to ask about our testing function again. I previously opened [an issue](https://github.com/ROCm/frameworks-internal/issues/13356) describing the differences in implementation details between our `te_compare_results` and `torch.allclose`. With my Triton test refactor PR, the difference really matters since in general we have consistent failures using our `te_compare_results` in certain areas, and consistent failures using `torch.allclose` in others. The point is, it kind of feels like we are "cheating" if we simply choose the provider that works best for our pass rate. Being consistent across the board (using either) results in problems though.

I'm honestly not sure why upstream uses the corresponding `te_compare_results` semantics, but we should make an active decision on how we want to handle our triton kernel correctness testing. I generally think that `torch` semantics are more "standard" in the ecosystem and should be favored. With that being said, our backward functions do not pass using `torch` semantics alone. I did, however, try modifying the `te_compare_results` function with a `use_torch_semantics` arg to instead calculate `mismatch = torch.abs(diff) > atol + rtol * torch.abs(r)` _as well as_
subsequently filter out rounding errors, and that works.

So to summarize, our testing is inconsistent and brittle at best, but I think we can somewhat standardize it by switching to `torch` + rounding error mitigation as our default. This corresponds to `compare_results(provider="te", use_torch_semantics=True, ...)`.

tests/pytorch/triton_kernels/test_common.py
```diff
@@ -188,12 +189,12 @@ def te_compare_results(t, r, atol, rtol, msg):
 
 
 # Call PyTorch tensor comparison function or TE tensor comparison function.
-def compare_results(provider, actual, expected, atol, rtol, msg):
+def compare_results(provider, actual, expected, atol, rtol, msg, use_torch_semantics=False):
```
wenchenvincent: As we now add the option of `use_torch_semantics`, shall we remove `provider` option?

Micky774: I kept it only for backwards compatibility, but I'm happy to remove it if we want to move away from torch calls. We currently dispatch to torch for scalar checks in our casting tests.

Micky774:
@wenchenvincent I've updated tests to use `te_compare_results` directly, optionally with `use_torch_semantics=True` depending on whether they called torch previously. PR should be ready for review. Let me know if you have any other concerns.

wenchenvincent (approved):
LGTM.

PR #258, opened 2025-08-12
[Fix] If 'ninja' is not found, it installs it via pip

AllenFarcas:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- If 'ninja' is not found, then 'ninja' will be installed via pup from setup.py
- Added checks with corresponding errors in build_ext.py and CMakeLists.txt

# Checklist:

### DIFF
--- a/build_tools/build_ext.py
+++ b/build_tools/build_ext.py
@@ -79,9 +79,10 @@ class CMakeExtension(setuptools.Extension):
 
         # Check whether parallel build is restricted
         max_jobs = get_max_jobs_for_parallel_build()
-        if not found_ninja():
+        if found_ninja():
+            configure_command.append("-GNinja")
+        elif rocm_build():
             raise RuntimeError(f"This project requires the Ninja build system. Install it using 'pip install ninja'.")
-        configure_command.append("-GNinja")
         build_command.append("--parallel")
         if max_jobs > 0:
             build_command.append(str(max_jobs))
--- a/build_tools/build_ext.py
+++ b/build_tools/build_ext.py
@@ -79,8 +79,9 @@ class CMakeExtension(setuptools.Extension):
 
         # Check whether parallel build is restricted
         max_jobs = get_max_jobs_for_parallel_build()
-        if found_ninja():
-            configure_command.append("-GNinja")
+        if not found_ninja():
+            raise RuntimeError(f"This project requires the Ninja build system. Install it using 'pip install ninja'.")
+        configure_command.append("-GNinja")
         build_command.append("--parallel")
         if max_jobs > 0:
             build_command.append(str(max_jobs))
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,7 @@ import sys
 import time
 from pathlib import Path
 from typing import List, Tuple
+import subprocess
 
 import setuptools
 from wheel.bdist_wheel import bdist_wheel
@@ -117,6 +118,7 @@ def setup_requirements() -> Tuple[List[str], List[str], List[str]]:
     if not found_cmake():
         setup_reqs.append("cmake>=3.21")
     if not found_ninja():
+        subprocess.check_call([sys.executable, "-m", "pip", "install", "ninja"])
         setup_reqs.append("ninja")
     if not found_pybind11():
         setup_reqs.append("pybind11")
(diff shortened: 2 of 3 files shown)


ipanfilo (commented):
I think it should also either explicitly use ninja generator when configure cmake extension in setup.py or in CMakeLists itself there should be check for generator

AllenFarcas:
@ipanfilo I just added in build_ext.py a runtime error if ninja is not found and also added that check in the CMakeLists a check for the generator that has a fatal error if ninja is not the CMAKE_GENERATOR.

build_tools/build_ext.py
```diff
@@ -81,6 +81,8 @@ def _build_cmake(self, build_dir: Path, install_dir: Path) -> None:
         max_jobs = get_max_jobs_for_parallel_build()
         if found_ninja():
             configure_command.append("-GNinja")
+        else:
+            raise RuntimeError(f"This project requires the Ninja build system. Install it using 'pip install ninja'.")
```
Micky774:
```suggestion
        if not found_ninja():
            raise RuntimeError(f"This project requires the Ninja build system. Install it using 'pip install ninja'.")
        configure_command.append("-GNinja")
```

build_tools/build_ext.py
```diff
@@ -79,8 +79,9 @@ def _build_cmake(self, build_dir: Path, install_dir: Path) -> None:
 
         # Check whether parallel build is restricted
         max_jobs = get_max_jobs_for_parallel_build()
-        if found_ninja():
-            configure_command.append("-GNinja")
+        if not found_ninja():
```
wangye805: Use rocm_build() to guard when we have different behavior with upstream NVTE

transformer_engine/common/CMakeLists.txt
```diff
@@ -32,6 +32,10 @@ else()
   
 endif()
 
+if(NOT CMAKE_GENERATOR STREQUAL "Ninja")
```
ipanfilo: Let's move it to ROCm specific section to let NV build still using Make

build_tools/build_ext.py
ipanfilo: Throw error only on rocm build. I.e. add 
elif rocm_build():
   raise RauntimeError
to the original condition

setup.py
```diff
@@ -117,6 +118,7 @@ def setup_requirements() -> Tuple[List[str], List[str], List[str]]:
     if not found_cmake():
         setup_reqs.append("cmake>=3.21")
     if not found_ninja():
+        subprocess.check_call([sys.executable, "-m", "pip", "install", "ninja"])
```
wangye805, line 121: I don't think we need to call pip install ninja directly. 

TE installation should be able to install required packages/dependencies once we put it in setup_reqs

transformer_engine/common/CMakeLists.txt
wangye805: Need to guard by USE_ROCM

wangye805 (dismissed):

setup.py
AllenFarcas: The setup_reqs actually contains ninja if it is not installed and yet, it does not install it. I searched more info on this and found that setup_requires is considered a legacy mechanism. PEP 518 introduced the pyproject.toml file and a mechanism to specify your build-time dependencies. 

Do you think that leaving this install of ninja here is ok @wangye805 @ipanfilo @wenchenvincent ?

https://stackoverflow.com/questions/73257839/setup-py-install-is-deprecated-warning-shows-up-every-time-i-open-a-terminal-i

https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html#summary

wangye805: Was this feature already deprecated? In other words, was the root cause of setup_reqs contains ninja but not installed due to that this feature deprecated? 

If so, we can temporarily keep this "pip install ninja". 

Just notice NV upstream already switched to pyproject.toml: https://github.com/NVIDIA/TransformerEngine/pull/1852/files

AllenFarcas: This feature was already deprecated. It appears for quite some time, setup.py wasn't able to reliably install and ensure the "setup_reqs" are met. Due to this unreliability, we currently can patch with the "pip install ninja".

ipanfilo: Indeed setup_reqs does not work. So I think install it with pip here is OK. One possible thing to improve is make ninja>=1.10 as it is minimal version for aotriton building but I think pip install will actually have something newer anyways

ipanfilo (approved):

PR #259, opened 2025-08-13
FIX Accumulate `int` overflow in workspace memory calculation

Micky774:
# Description

Adds static cast to `std::accumulate` arg to ensure correct templating to prevent integer overflow. Includes non-regression test.

Fixes https://github.com/ROCm/frameworks-internal/issues/13179?reload=1?reload=1

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/tests/pytorch/fused_attn/test_fused_attn.py
+++ b/tests/pytorch/fused_attn/test_fused_attn.py
@@ -266,6 +266,8 @@ def test_dot_product_mem_calc(workspace_opt):
         pytest.skip("This test requires the CK fused attention backend.")
 
     os.environ["NVTE_CK_USES_FWD_V3"] = "1"
+    os.environ["NVTE_FUSED_ATTN_CK"] = "0"
+    os.environ["NVTE_FUSED_ATTN_AOTRITON"] = "1"
     _, _ = _run_dot_product_attention(
         dtype,
         config,
@@ -277,6 +279,8 @@ def test_dot_product_mem_calc(workspace_opt):
         is_training,
     )
     del os.environ["NVTE_CK_USES_FWD_V3"]
+    del os.environ["NVTE_FUSED_ATTN_CK"]
+    del os.environ["NVTE_FUSED_ATTN_AOTRITON"]
 
 
 @pytest.mark.skipif(get_cudnn_version() < (8, 9, 1), reason="cuDNN 8.9.1+ is required.")
--- a/tests/pytorch/fused_attn/test_fused_attn.py
+++ b/tests/pytorch/fused_attn/test_fused_attn.py
@@ -241,6 +241,48 @@ if is_bf16_compatible():  # bf16 requires sm_80 or higher
     param_types.append(torch.bfloat16)
 param_types_lean = [torch.bfloat16]
 
+@pytest.mark.parametrize("workspace_opt", [True, False])
+def test_dot_product_mem_calc(workspace_opt):
+    """
+    Non-regression test for memory workspace calculation integer overflow issue.
+    """
+    ckpt_attn = False
+    pad_between_seqs = False
+    if not is_bf16_compatible():
+        pytest.skip("This test requires bf16 support.")
+    dtype = torch.bfloat16
+    config = ModelConfig(16, 128, 8, 128, 8192, 8192, 0.0, "causal", "no_bias")
+    qkv_layout = "sbhd_sbhd_sbhd"
+    available_backends, _ = _get_attention_backends(
+        config,
+        qkv_dtype=dtype,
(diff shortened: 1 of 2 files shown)


wangye805 (changes requested):
LGTM for current changes

For the padding, unpadding kernels, we can also run into indexing near the total number of elements of a tensor and cu_seqlen is passed in as int32_t:  https://github.com/ROCm/TransformerEngine/blob/35ea398935c4660720bb31437be5179d64812393/transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp#L340

Can you help check whether overflow won't happen in places like this in fused_attn_ck.cpp and inside transformer_engine/common/ck_fused_attn/

Thanks

tests/pytorch/fused_attn/test_fused_attn.py
```diff
@@ -241,6 +241,48 @@ def test():
     param_types.append(torch.bfloat16)
 param_types_lean = [torch.bfloat16]
 
+@pytest.mark.parametrize("workspace_opt", [True, False])
+def test_dot_product_mem_calc(workspace_opt):
```
wenchenvincent, line 246: What's the failure mode of this test with this fix?

Micky774: The amount of memory required to properly run this config can satisfied by an MI300, so it should it should be able to run without issue after calculating the memory allocation correctly. Hence the failure mode is an inability to allocate enough memory (due to incorrect calclulations) resulting in a runtime error.

tests/pytorch/fused_attn/test_fused_attn.py
```diff
@@ -241,6 +241,48 @@ def test():
...
+        window_size=config.window_size,
+        pad_between_seqs=pad_between_seqs,
+    )
+    _, fused_attn_supported, _ = available_backends
+    if not fused_attn_supported:
+        pytest.skip("This test requires the CK fused attention backend.")
+    is_training = config.head_dim_qk <= 128 and config.head_dim_v <= 128
+    # FusedAttention backend
+    os.environ["NVTE_CK_USES_FWD_V3"] = "1"
+    os.environ["NVTE_FUSED_ATTN_BACKEND"] = "1"
```
ipanfilo: This is CUDA only env variable while the whole test is ROCm specific.

tests/pytorch/fused_attn/test_fused_attn.py
```diff
@@ -241,6 +241,48 @@ def test():
...
+    qkv_layout = "sbhd_sbhd_sbhd"
+    available_backends, _ = _get_attention_backends(
+        config,
+        qkv_dtype=dtype,
+        qkv_layout=qkv_layout,
+        window_size=config.window_size,
+        pad_between_seqs=pad_between_seqs,
+    )
+    _, fused_attn_supported, _ = available_backends
+    if not fused_attn_supported:
```
ipanfilo: Fused attention is not necessary CK. You have to check the third attribute returned by _get_attention_backends

tests/pytorch/fused_attn/test_fused_attn.py
ipanfilo, line 246: Need to skip on CUDA

ipanfilo (changes requested):

Micky774:
@ipanfilo I think I have addressed your concerns, let me know if you have any others!

wangye805 (approved):

tests/pytorch/fused_attn/test_fused_attn.py
```diff
@@ -252,23 +253,19 @@ def test_dot_product_mem_calc(workspace_opt):
...
-    if not fused_attn_supported:
+    if FusedAttnBackend["CK"] not in fused_attn_backends:
         pytest.skip("This test requires the CK fused attention backend.")
-    is_training = config.head_dim_qk <= 128 and config.head_dim_v <= 128
-    # FusedAttention backend
+
     os.environ["NVTE_CK_USES_FWD_V3"] = "1"
-    os.environ["NVTE_FUSED_ATTN_BACKEND"] = "1"
-    os.environ["NVTE_FUSED_ATTN_CK"] = "1"
-    os.environ["NVTE_FUSED_ATTN_AOTRITON"] = "0"
```
ipanfilo: While CK is a default backend so the test will work as expected, I'd keep  os.environ["NVTE_FUSED_ATTN_CK"] = "1" and  os.environ["NVTE_FUSED_ATTN_AOTRITON"] = "0" for sanity.

ipanfilo (approved):

PR #260, opened 2025-08-13
AITER V3 asm kernels installation

ipanfilo:
# Description

Install pre-compiled AITER V3 kernels hsaco.

Fixes [#13238](https://github.com/ROCm/frameworks-internal/issues/13238)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Install AITER V3 kernels
- Use env variable to tell AITER location of  V3 kernels HSACO

# Checklist:


transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
```diff
@@ -5,13 +5,47 @@
...
+      case 95: // Gfx950
+        arh_str = "gfx950/"; // trailing slash is mandatory
+        break;
+      default:
+        // Unsupported V3 architecture
+        return;
+    }
+    Dl_info info;
+    dladdr((void*)set_aiter_asm_dir, &info);
+    setenv("AITER_ASM_DIR",
```
wenchenvincent, line 42: I am thinking of the case where there is preset AITER_ASM_DIR in a docker container, where they could use something from AITER directly and they could use a different AITER commit. In that case, shall we override or not?

ipanfilo: I set override flag to 0 to give user opportunity to override dafault behaviour. Indeed, it bears risk of misbehaviour, especially if someone set AITER_ASM_DIR globally. May be flexibility of using 0 is not worth of risk. I can change it to 1

wenchenvincent: Thanks Ilya. I was also debating about this. Is there a way we can detect if the env var is preset? And if so, we might want to print a warning?

ipanfilo: Are you thinking about scenario when an app uses TE and also AITER directly so changing env here may cause side effects?
How realistic is it?

wenchenvincent: Not the same app. But in the same docker image, it could have TE installed and AITER installed (There are ppl who use the AITER python API). And it is possible it would interfere.

ipanfilo: I think interfere is only possible if the same app uses two copies of AITER - one by TE and the other directly or through another library. Otherwise I do not see any problem - the env is set only for current process and it is used only to pass V3 kernels location from TE to AITER

wenchenvincent (approved):
LGTM. Thanks!

PR #267, opened 2025-08-15
FIX Update intra-seq padding detection in CK Fused Attention backend

Micky774:
# Description

Fixes https://github.com/ROCm/frameworks-internal/issues/13341

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Adds check to ensure that only non-trivial seq-padding arrays count towards determining whether the context uses inter-sequence padding.

# Checklist:


transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -724,9 +732,10 @@ void fused_attn_ck_fwd_impl(
         window_size_left, window_size_right,
         devPtrO,
         o_stride[1], o_stride[2],
-        devPtrSoftmaxAux,
+        devPtrSoftmaxLSEWithoutPadding,
         nvte_ck_uses_fwd_v3,
         stream));
+    add_padding_softmax_lse(b, h, s_q, max_tokens_q, is_ragged, devPtrSoftmaxLSEWithoutPadding, devPtrCuSeqlensQ, devPtrSeqOffsetsQ, devPtrSoftmaxAux, stream);
```
wangye805: Here devPtrSeqOffsetsQ will be nullptr... We can simply put devPtrCuSeqlensQ into the place of devPtrSeqOffsetsQ

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -1299,7 +1308,7 @@ void fused_attn_ck_fwd_qkvpacked(
   bool is_padding = (attn_mask_type == NVTE_Mask_Type::NVTE_PADDING_MASK || 
                      attn_mask_type == NVTE_Mask_Type::NVTE_PADDING_CAUSAL_MASK ||
                      attn_mask_type == NVTE_Mask_Type::NVTE_PADDING_CAUSAL_BOTTOM_RIGHT_MASK);
-  bool pad_between_seqs = (is_ragged && !(input_cu_seqlens_padded->data.shape.empty())) || (!is_ragged && is_padding && !(input_cu_seqlens->data.shape.empty()));
+  bool pad_between_seqs = (devPtrCuSeqlens!=devPtrSeqOffsets || !input_cu_seqlens_padded->data.shape.empty()) && (is_ragged || is_padding);
```
wangye805: Sorry I didn't quite understand the logic here. 

I thought pad_between_seqs happens in the following cases:
1). is_ragged (THD format) and (cu_seqlen_ptr not the same as cu_seqlen_padded_ptr) and (cu_seqlen_padded is not null or size 0 or none)
2). is_ragged false (BSHD/SBHD format), and getting padding mask, and (cu_seqlen is not null, size 0 or none)

Was your underlying logic the same? Or can you comment on how you get this boolean expressions?

wangye805 (changes requested):
We will also need workspace allocation and buffer for softmax-lse in the bwd pass

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -638,6 +642,10 @@ void fused_attn_ck_fwd_impl(
     workspace_next = static_cast<void *>(static_cast<int8_t *>(workspace_next) + o_storage_bytes);
     // reset the final results since padded places need to be 0
     NVTE_CHECK_CUDA(cudaMemsetAsync(devPtrO, 0, o_storage_bytes, stream));
+  }else if(pad_between_seqs){
```
ipanfilo: I think it should is_ragged, right?

ipanfilo (changes requested):

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
wangye805: Right, should be if(is_ragged)

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: Good catch -- I simply misread both conjunctive clauses as using `input_cu_seqlens_padded`, not noticing the latter used  `input_cu_seqlens` instead. Similarly for other calculations within this file. I've updated it, please let me know if you spot any other errors.

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: Updated

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: Good catch, fixed

wangye805:
> We will also need workspace allocation and buffer for softmax-lse in the bwd pass

You will also need https://github.com/ROCm/TransformerEngine/blob/8e5afb2a1099d638a88622d3ca614f135ab957a2/transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp#L948-L949 and https://github.com/ROCm/TransformerEngine/blob/8e5afb2a1099d638a88622d3ca614f135ab957a2/transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp#L1067 applied to fused_attn_ck_bwd_impl when is_ragged true but pad_between_seqs false.

wangye805:
Let's also add a pytest (probably can be based on the customer benchmark scripts)?

Micky774:
> Let's also add a pytest (probably can be based on the customer benchmark scripts)?

For this, are you suggesting a pytest where we measure the peak memory usage and threshold it against a heuristic?

Micky774:
@wangye805 PR is ready for review w/ implementation of bwd

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -1112,6 +1127,8 @@ void fused_attn_ck_bwd_impl(
     add_padding(dtype, b, hg, s_kv, d_v, max_tokens_kv, is_ragged, v_stride[0], v_stride[1], v_stride[2], devPtrdVWithoutPadding, devPtrCuSeqlensKV, devPtrSeqOffsetsKV, devPtrdV, stream);
 
   }else if(is_ragged){
+    remove_padding_softmax_lse(b, h, s_q, max_tokens_q, is_ragged, devPtrSoftmaxAux, devPtrCuSeqlensQ, devPtrSeqOffsetsQ, devPtrSoftmaxLSEWithoutPadding, stream);
```
wangye805, line 1157: Use devPtrCuSeqlensQ instead of devPtrSeqOffsetsQ in case devPtrSeqOffsetsQ is null/None/empty?

wangye805 (changes requested):

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: Done!

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -1112,6 +1127,8 @@ void fused_attn_ck_bwd_impl(
     add_padding(dtype, b, hg, s_kv, d_v, max_tokens_kv, is_ragged, v_stride[0], v_stride[1], v_stride[2], devPtrdVWithoutPadding, devPtrCuSeqlensKV, devPtrSeqOffsetsKV, devPtrdV, stream);
 
   }else if(is_ragged){
+    remove_padding_softmax_lse(b, h, s_q, max_tokens_q, is_ragged, devPtrSoftmaxAux, devPtrCuSeqlensQ, devPtrCuSeqlensQ, devPtrSoftmaxLSEWithoutPadding, stream);
```
wangye805: Then we should pass devPtrSoftmaxLSEWithoutPadding to ck_attn_varlen_bwd in line 1147

wangye805 (changes requested):

Micky774:
@wangye805 I've adjusted with your feedback, but I've also changed how we determine `pad_between_seqs`. I add an additional check to determine whether the `cu_seqlen` array contains identical values to `cu_seqlen_padded`. I'm not super familiar with the C level Tensor API, so please let me know if there is a better way to go about data access.

This change allows tests to _actually_ reach the proper padding-free THD branch (whereas before it remained untouched by all the tests). All tests in `pytorch/fused_attn/test_fused_attn.py` pass locally.

Micky774:
After discussion, we decided to relax the check for `pad_between_seqs` again and enforce the choice of passing `cu_seqlen_padded=cu_seqlen` when trying to disable padding (reflected in the test as setting `cu_seqlen_padded=None` so that it later points to `cu_seqlen`).

tests/pytorch/fused_attn/test_fused_attn.py
```diff
@@ -1140,8 +1146,8 @@ def get_dummy_cuda_rng_tracker() -> CudaRNGStatesTracker:
         max_seqlen_kv=config.max_seqlen_kv,
         cu_seqlens_q=cu_seqlens_q,
         cu_seqlens_kv=cu_seqlens_kv,
-        cu_seqlens_q_padded=cu_seqlens_q_after_pad if backend == "FusedAttention" else None,
-        cu_seqlens_kv_padded=cu_seqlens_kv_after_pad if backend == "FusedAttention" else None,
+        cu_seqlens_q_padded=cu_seqlens_q_padded,
```
wangye805, line 1185: I would suggest we keep the original _run_dot_product_attention behavior and add a new rocm specific one in which cu_seqlen_q_padded points to cu_seqlen_q if pad_between_seqs is false. In that sense, we can test both scenarios that when cu_seqlens_q_padded is a clone and that cu_seqlens_q_padded is just a reference

wangye805 (changes requested):

tests/pytorch/fused_attn/test_fused_attn.py
Micky774: Done.

tests/pytorch/fused_attn/test_fused_attn.py
```diff
@@ -315,8 +315,9 @@ def test_dot_product_mem_calc(workspace_opt):
 @pytest.mark.parametrize("qkv_layout", [None])
 @pytest.mark.parametrize("swa", [False])
 @pytest.mark.parametrize("pad_between_seqs", [False])
+@pytest.mark.parametrize("share_cu_seqlens_ref", [False])
```
wangye805, line 318: Can we add rocm guard to only enable those new pytests on our rocm machine, not NV

wangye805 (changes requested):

tests/pytorch/fused_attn/test_fused_attn.py
Micky774: Added a guard to `test_dpa_qkv_layout_thd` -- `test_dpa_qkv_layout_thd_mqa_gqa` is already a ROCm-specific test so we should be good there.

wangye805 (approved):

wangye805 (changes requested):
Need to address the pytest failures: 
image

Micky774:
> Need to address the pytest failures:

Fixed now.

wangye805:
> > Need to address the pytest failures:
> 
> Fixed now.

Emm, it looks like we have another fused-attn related failure in the jax side: 
image

Can you also help run cp pytests locally as well? Thanks

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -567,6 +567,10 @@ void fused_attn_ck_fwd_impl(
       (*workspace_size)+= max_tokens_q*h*sizeof(float);
       // request q, k, v, o buffer without padding
       (*workspace_size)+= q_storage_bytes + k_storage_bytes + v_storage_bytes + o_storage_bytes;
+    }else if(is_ragged){
+      // softmax_lse buffer needed for pad_between_seq only
```
wangye805: The comment message is not right. Should be: softmax_lse buffer needed for is_ragged as well

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
wangye805, line 1157: And you also need to change the devPtrSoftmaxAux in line 1147 to devPtrSoftmaxLSEWithoutPadding

wangye805 (changes requested):

Micky774:
> Can you also help run cp pytests locally as well? Thanks

Looks like `tests/jax/test_distributed_fused_attn.py` passes now. I suspect it was a consequence of using the `devPtrSoftmaxAux` pointer instead of `devPtrSoftmaxLSEWithoutPadding` as you pointed out in your comment. I tried testing `tests/pytorch/fused_attn/test_fused_attn_with_cp.py` locally as well to make sure we're all set there, but I keep running into rather opaque errors citing that `torch.distributed.launch returned non-zero exit status 1`...might need to wait on CI for that one.

wangye805:
> > Can you also help run cp pytests locally as well? Thanks
> 
> Looks like `tests/jax/test_distributed_fused_attn.py` passes now. I suspect it was a consequence of using the `devPtrSoftmaxAux` pointer instead of `devPtrSoftmaxLSEWithoutPadding` as you pointed out in your comment. I tried testing `tests/pytorch/fused_attn/test_fused_attn_with_cp.py` locally as well to make sure we're all set there, but I keep running into rather opaque errors citing that `torch.distributed.launch returned non-zero exit status 1`...might need to wait on CI for that one.

I recall that tests/jax/test_distributed_fused_attn.py failure/hang is flaky? Maybe you can run several times to verify whether the hang/failure really goes away? 

For the pytorch side failure, we can run it again when no one else is using that local machine.

Micky774:
> I recall that tests/jax/test_distributed_fused_attn.py failure/hang is flaky? Maybe you can run several times to verify whether the hang/failure really goes away?

AFAIK that flaky failure/hang is only when JIT is enabled -- I ran w/ JIT disabled as we do on our CI tests. That shouldn't be flaky. If it is, then that's a _new_ issue to be tracked too...

README.rst
```diff
@@ -261,6 +261,10 @@ NVTE_FUSED_ATTN=0 will use the TE unfused attention even if NVTE_FUSED_ATTN_CK o
 Fused attention backends are chosen according to the match results between the actual problem config and the support matrix of the specific backend.
 For the scenario that both backends are enabled and match the problem configuration, the CK backend will be chosen with higher priority.
 
+Note that when using `THD` format tensors with CK Fused Attention, one should pass `None` for `cu_seqlens_q_padded, cu_seqlens_kv_padded`
+to indicate that there is no padding between sequences. Otherwise, passing proper tensors will indicate padding between sequences. This is the case
+for both the `FusedAttention` and `DotProductAttention` modules.
+
```
ipanfilo, line 267: It is a requirement for CK only, isn't it?

tests/pytorch/fused_attn/test_fused_attn.py
```diff
@@ -1129,6 +1146,12 @@ def get_dummy_cuda_rng_tracker() -> CudaRNGStatesTracker:
         k = inp[1]
         v = inp[2]
         d_out = out_grad
+        if not pad_between_seqs and share_cu_seqlens_ref:
+            cu_seqlens_q_padded = cu_seqlens_q
```
ipanfilo: Where is this path tested? It requires pad_between_seqs = False and share_cu_seqlens_ref = True which is not in tests configs.

ipanfilo: And documentation says None should be passed in this case

ipanfilo (changes requested):

README.rst
Micky774: Yes I believe so, and I indicate that in the note as well

tests/pytorch/fused_attn/test_fused_attn.py
Micky774: It's included in `test_dpa_qkv_layout_thd` and `test_dpa_qkv_layout_thd_mqa_gqa` parameterizations. Using `None` implicitly performs this sharing later on anyways, so I figured it would be worth setting it up explicitly here to make clear what is being tested, but I could switch to `None` if preferred.

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -18,6 +18,29 @@
 namespace transformer_engine {
 namespace fused_attn_rocm {
 
+bool _get_pad_between_seqs(
```
wangye805: Currently we don't have a coding style like the naming convention. But let's make it consistent with other functions in this file and other source code files, to just use get_pad_between_seqs. Otherwise, we probably need to change other internal used functions

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -18,6 +18,29 @@
...
+  // First we check whether we have a ragged array with a non-trivial
+  // input_cu_seqlens_padded tensor
+  bool pad_between_seqs = (
+    is_ragged
+    && input_cu_seqlens->data.dptr!=input_cu_seqlens_padded->data.dptr
+    && !input_cu_seqlens_padded->data.shape.empty()
+  );
+  // Next we guard against both pointers being null, which would indicate that
+  // this call is an initial workspace allocation call provided with dummy data
+  pad_between_seqs |= (
```
wangye805: Let's use logical operator (||) instead of bitwise operator for readability.

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -18,6 +18,29 @@
...
+    is_ragged
+    && input_cu_seqlens->data.dptr==nullptr
```
wangye805: Regarding the second checking condition, since jax is using TensorWrapper(nullptr, std::vector<size_t>{num_segments + 1}, DType::kInt32) for cu_seqlen/cu_seqlen_padded, let's also check cu_seqlen_padded->data.shape is not empty

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -18,6 +18,29 @@
...
+){
+  // First we check whether we have a ragged array with a non-trivial
+  // input_cu_seqlens_padded tensor
+  bool pad_between_seqs = (
+    is_ragged
+    && input_cu_seqlens->data.dptr!=input_cu_seqlens_padded->data.dptr
+    && !input_cu_seqlens_padded->data.shape.empty()
+  );
+  // Next we guard against both pointers being null, which would indicate that
+  // this call is an initial workspace allocation call provided with dummy data
```
wangye805: Mention this scenario is from jax side workspace buffer allocation

wangye805 (changes requested):

tests/pytorch/fused_attn/test_fused_attn.py
ipanfilo: I think that test should pass to TE whatever is suggested in documentation.
And test_dpa_qkv_layout_thd and test_dpa_qkv_layout_thd_mqa_gqa parameterizations do not include [False, True]

README.rst
ipanfilo: But tests behave the same way for CK and AOT

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: Done

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: Done

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: Good idea, updated

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: Done

tests/pytorch/fused_attn/test_fused_attn.py
Micky774: Gotcha, updated! And good catch, I corrected the test parameterization.

README.rst
Micky774: I've added some skips for these test configs if CK is not an available backend. Let me know if you had something else in mind for this comment.

wangye805 (approved):

tests/pytorch/fused_attn/test_fused_attn.py
```diff
@@ -806,39 +811,68 @@ def test_dpa_qkv_layout(dtype, model_configs, model, qkv_layout):
...
+        config,
+        qkv_dtype=dtype,
+        qkv_layout=qkv_layout,
+        window_size=config.window_size,
+        pad_between_seqs=pad_between_seqs,
+    )
+    if not IS_HIP_EXTENSION and share_cu_seqlens_ref:
+        pytest.skip("share_cu_seqlens_ref is a ROCm TE specific implementation detail.")
+    elif share_cu_seqlens_ref and FusedAttnBackend["CK"] not in fused_attn_backends:
+        pytest.skip("This test is only required for the CK fused attention backend.")
```
ipanfilo: Minor. I think it will be more clear if all this block is under 'if share_cu_seqlens_ref'

tests/pytorch/fused_attn/test_fused_attn.py
```diff
@@ -806,39 +811,68 @@ def test_dpa_qkv_layout(dtype, model_configs, model, qkv_layout):
...
+    _, _, fused_attn_backends = _get_attention_backends(
+        config,
+        qkv_dtype=dtype,
+        qkv_layout=qkv_layout,
+        window_size=config.window_size,
+        pad_between_seqs=pad_between_seqs,
+    )
+    if share_cu_seqlens_ref and FusedAttnBackend["CK"] not in fused_attn_backends:
+        pytest.skip("This test is only required for the CK fused attention backend.")
+
```
ipanfilo, line 863: Please make skipping conditions and ("pad_between_seqs", "share_cu_seqlens_ref") parameters list the same as test_dpa_qkv_layout_thd

Micky774: Updated the parameter list (I thought I already did but I missed it, good catch), but I don't think we need to change the skip condition here since `not IS_HIP_EXTENSION` is already skipped for on this test via decorator.

tests/pytorch/fused_attn/test_fused_attn.py
Micky774: Updated

ipanfilo (approved):

PR #268, opened 2025-08-15
Fix NCCL error in test_torch_fsdp2

VeeraRajasekhar:
# Description

Synchronize all ranks (and all DeviceMesh subgroups) before teardown to fix intermittent NCCL/RCCL async exceptions at exit.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add per-axis barriers: iterate mesh.ndim - 1 … 0 and call dist.barrier(group=mesh.get_group(dim)) to drain collectives on each DeviceMesh dimension.
- Ensure dist.destroy_process_group() is called by all ranks.

# Checklist:

### DIFF
--- a/ci/pytorch.sh
+++ b/ci/pytorch.sh
@@ -83,6 +83,7 @@ run_test_config_mgpu(){
         run 3 distributed/test_fusible_ops.py
         run 3 distributed/test_numerics.py
         run 3 fused_attn/test_fused_attn_with_cp.py
+        run 3 distributed/test_torch_fsdp2.py
     fi
 }
 
--- a/ci/pytorch.sh
+++ b/ci/pytorch.sh
@@ -82,8 +82,8 @@ run_test_config_mgpu(){
         run 3 test_sanity_import.py
         run 3 distributed/test_fusible_ops.py
         run 3 distributed/test_numerics.py
-        run 3 fused_attn/test_fused_attn_with_cp.py
         run 3 distributed/test_torch_fsdp2.py
+        run 3 fused_attn/test_fused_attn_with_cp.py
     fi
 }
 
--- a/tests/pytorch/distributed/run_fsdp2_model.py
+++ b/tests/pytorch/distributed/run_fsdp2_model.py
@@ -171,11 +171,7 @@ def _train(args):
         if LOCAL_RANK == 0:
             print(f"Rank {LOCAL_RANK}: Iteration {iteration} completed.")
 
-    # Sync all ranks before clean up
-    if isinstance(mesh, torch.distributed.device_mesh.DeviceMesh):
-        for dim in range(mesh.ndim - 1, -1, -1):
-            dist.barrier(group=mesh.get_group(dim))
-
+    dist.barrier(device_ids=[torch.cuda.current_device()])
     dist.destroy_process_group()
     if LOCAL_RANK == 0:
         print(f"Rank {LOCAL_RANK}: Done...")
--- a/tests/pytorch/distributed/run_fsdp2_model.py
+++ b/tests/pytorch/distributed/run_fsdp2_model.py
@@ -1,3 +1,4 @@
+#!/usr/bin/python3
 # This file was modified for portability to AMDGPU
 # Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
 # Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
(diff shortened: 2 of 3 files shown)


tests/pytorch/distributed/run_fsdp2_model.py
```diff
@@ -171,6 +171,11 @@ def _train(args):
         if LOCAL_RANK == 0:
             print(f"Rank {LOCAL_RANK}: Iteration {iteration} completed.")
 
+    # Sync all ranks before clean up
```
wangye805: Emm, I thought mesh will always be a DeviceMesh according to line 135 to line 148.

tests/pytorch/distributed/run_fsdp2_model.py
```diff
@@ -171,6 +171,11 @@ def _train(args):
...
+    if isinstance(mesh, torch.distributed.device_mesh.DeviceMesh):
+        for dim in range(mesh.ndim - 1, -1, -1):
```
wangye805: Why do we need dist.barrier for each mesh dim? I thought one single dist.barrier() for one pytorch process is enough?

wangye805 (changes requested):

wenchenvincent:
I think FSDP might have already implemented the synchronization between the processes. Needing the extra barrier might indicate that there might be some problems with the implementation.

tests/pytorch/distributed/run_fsdp2_model.py
VeeraRajasekhar: Yes, you are right, I can remove it.

tests/pytorch/distributed/run_fsdp2_model.py
VeeraRajasekhar: No, we don't need to, I have changed to use one single barrier which solved the issue.

Thanks.

wangye805:
> I think FSDP might have already implemented the synchronization between the processes. Needing the extra barrier might indicate that there might be some problems with the implementation.

Emm, @VeeraRajasekhar, let's try it on NV machines to see whether we can reproduce

ci/pytorch.sh
```diff
@@ -83,6 +83,7 @@ run_test_config_mgpu(){
         run 3 distributed/test_fusible_ops.py
         run 3 distributed/test_numerics.py
         run 3 fused_attn/test_fused_attn_with_cp.py
+        run 3 distributed/test_torch_fsdp2.py
```
ipanfilo, line 85: Please move it up to be with other distributed tests

tests/pytorch/distributed/run_fsdp2_model.py
ipanfilo, line 1: Add AMD copyright

tests/pytorch/distributed/test_torch_fsdp2.py
```diff
@@ -1,3 +1,5 @@
+# This file was modified for portability to AMDGPU
+# Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
```
wangye805: If there is no changes below (no difference between ours and NV upstream), we do not need to have copyright info

tests/pytorch/distributed/run_fsdp2_model.py
```diff
@@ -171,6 +172,8 @@ def _train(args):
         if LOCAL_RANK == 0:
             print(f"Rank {LOCAL_RANK}: Iteration {iteration} completed.")
 
+    if torch_version() < (2, 6, 0):
```
wangye805, line 180: Add a comment to explain there is a bug in torch which may result in test failure

wangye805 (changes requested):

tests/pytorch/distributed/run_fsdp2_model.py
VeeraRajasekhar: Added a comment

tests/pytorch/distributed/test_torch_fsdp2.py
VeeraRajasekhar: Updated this

wangye805 (approved):

wangye805:
Thanks @VeeraRajasekhar. 

By the way, next time, let's not use forced push once we already start the reviewing process. You can push in new fix commit and we can squash them in merging. Force pushing can lose our history and make it difficult for reviewers to track

PR #274, opened 2025-08-19
update hipify_torch submodule to fix v2 mappings

jeffdaily:
Incorrect duplicate cublas mappings were found for version 2 of the mappings file.  Update hipify_torch submodule to pick up the changes before torch's hipify v2 is merged upstream.


ipanfilo:
I wonder what about v1 mappings? They have the same duplicated entries.

jeffdaily:
RIght, I saw v1 mappings are also duplicated but it's not breaking the build.  It's been that way for a long time.  The v2 builds do break, however, which is why I'm trying to fix it.

ipanfilo (approved):

PR #278, opened 2025-08-21
Normalization kernels for mxfp8

alextmagro:
# Description

Forward normalization kernels for mxfp8 for RMSNorm and LayerNorm

Fixes # (issue)
#12995

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Tuned and general kernels for RMS and Layer normalization

# Checklist:

### DIFF
--- a/tests/cpp/operator/CMakeLists.txt
+++ b/tests/cpp/operator/CMakeLists.txt
@@ -10,7 +10,10 @@ list(APPEND test_cuda_sources
             test_cast_dbias.cu
             test_cast_dbias_dgelu.cu
             test_cast_gated_swiglu.cu
+            test_cast_mxfp8_gated_swiglu.cu
             test_qdq.cu
+            test_cast_mxfp8.cu
+            test_dequantize_mxfp8.cu
             test_cast_transpose.cu
             test_transpose.cu
             test_cast_transpose_current_scaling.cu
@@ -19,16 +22,16 @@ list(APPEND test_cuda_sources
             test_cast_transpose_dgeglu.cu
             test_act.cu
             test_normalization.cu
+            test_normalization_mxfp8.cu
             test_multi_cast_transpose.cu
             test_multi_padding.cu
             test_causal_softmax.cu
             test_swizzle.cu
-            test_cast_mxfp8_gated_swiglu.cu
-            test_cast_mxfp8.cu
-            test_normalization_mxfp8.cu
-            test_dequantize_mxfp8.cu
-            test_cublaslt_gemm.cu
 	          ../test_common.cu)
+if(NOT USE_CUDA)
+  list(APPEND test_cuda_sources
+              test_cublaslt_gemm.cu)
+endif()
 
 if(USE_CUDA)
   add_executable(test_operator ${test_cuda_sources})
--- a/tests/cpp/operator/CMakeLists.txt
+++ b/tests/cpp/operator/CMakeLists.txt
@@ -25,15 +25,10 @@ list(APPEND test_cuda_sources
             test_swizzle.cu
             test_cast_mxfp8_gated_swiglu.cu
             test_cast_mxfp8.cu
+            test_normalization_mxfp8.cu
             test_dequantize_mxfp8.cu
+            test_cublaslt_gemm.cu
 	          ../test_common.cu)
-if(USE_CUDA)
-  list(APPEND test_cuda_sources
-              test_normalization_mxfp8.cu)
-else()
-  list(APPEND test_cuda_sources
-              test_cublaslt_gemm.cu)
-endif()
 
 if(USE_CUDA)
   add_executable(test_operator ${test_cuda_sources})
--- a/tests/cpp/operator/test_normalization_mxfp8.cu
+++ b/tests/cpp/operator/test_normalization_mxfp8.cu
@@ -53,7 +53,8 @@ void dequantize_1x_kernel(InputType* input_ptr, ScaleType* scale_ptr, OutputType
   const size_t blocks_per_tile_Y = tile_size_Y / block_size_Y;
   const size_t blocks_per_tile_X = tile_size_X / block_size_X;
 #ifdef __HIP_PLATFORM_AMD__
-    const auto [_, __, ___, blocks_per_row] = get_scale_tensor_dims(rows, cols, scaling_mode_x, scaling_mode_y);
+    const auto scale_dims = get_scale_tensor_dims(rows, cols, scaling_mode_x, scaling_mode_y);
+    const size_t blocks_per_row = std::get<3>(scale_dims);
 #else
   const size_t blocks_per_row = (cols + block_size_X - 1) / block_size_X;
 #endif
--- a/tests/cpp/operator/test_normalization_mxfp8.cu
+++ b/tests/cpp/operator/test_normalization_mxfp8.cu
@@ -186,6 +186,7 @@ void performTest(const size_t N, const size_t H, const bool zero_centered_gamma,
   cudaGetDeviceProperties(&prop, 0);
 
 #ifndef __HIP_PLATFORM_AMD__
+// TODO: Guard all MXFP8 tests for hip_version >= gfx9.5
   if (getDeviceComputeCapability() < blackwellComputeCapability) {
     GTEST_SKIP();
   }
--- a/tests/cpp/operator/test_normalization_mxfp8.cu
+++ b/tests/cpp/operator/test_normalization_mxfp8.cu
@@ -1,4 +1,6 @@
 /*************************************************************************
+ * This file was modified for portability to AMDGPU
+ * Copyright (c) 2022-2025, Advanced Micro Devices, Inc. All rights reserved.
  * Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  *
  * See LICENSE for license information.
@@ -17,6 +19,7 @@
 #include <gtest/gtest.h>
 
 #include <transformer_engine/normalization.h>
+#include <transformer_engine/cast.h>
 #include <transformer_engine/transformer_engine.h>
 #include "../test_common.h"
 
@@ -39,7 +42,7 @@ std::map<NormType, std::string> normToString = {
 
 template <typename InputType, typename ScaleType, typename OutputType>
 void dequantize_1x_kernel(InputType* input_ptr, ScaleType* scale_ptr, OutputType* output_ptr,
-  size_t rows, size_t cols, size_t scaling_mode_x, size_t scaling_mode_y){
+  size_t rows, size_t cols, size_t scaling_mode_x, size_t scaling_mode_y) {
 
   const size_t block_size_Y = scaling_mode_x;   // mind the mapping Y <-- x
   const size_t block_size_X = scaling_mode_y;   //              and X <-- y
@@ -49,7 +52,11 @@ void dequantize_1x_kernel(InputType* input_ptr, ScaleType* scale_ptr, OutputType
   const size_t tiles_num_X = (cols + tile_size_X - 1) / tile_size_X;
   const size_t blocks_per_tile_Y = tile_size_Y / block_size_Y;
   const size_t blocks_per_tile_X = tile_size_X / block_size_X;
+#ifdef __HIP_PLATFORM_AMD__
+    const auto [_, __, ___, blocks_per_row] = get_scale_tensor_dims(rows, cols, scaling_mode_x, scaling_mode_y);
+#else
   const size_t blocks_per_row = (cols + block_size_X - 1) / block_size_X;
+#endif
 
   #pragma omp parallel for proc_bind(spread) schedule(static)
   for (size_t t = 0; t < tiles_num_Y * tiles_num_X; ++t) {
@@ -133,7 +140,11 @@ void compute_ref_stats(NormType norm_type,
       compute_t current = static_cast<compute_t>(data[i * H + j]);
       sum_sq += (current - m) * (current - m);
     }
+#ifdef __HIP_PLATFORM_AMD__
+    rsigma[i] = 1.0f / sqrtf((sum_sq / H) + epsilon);
+#else
     rsigma[i] = rsqrtf((sum_sq / H) + epsilon);
+#endif
   }
 }
 
@@ -173,9 +184,11 @@ void performTest(const size_t N, const size_t H, const bool zero_centered_gamma,
   cudaDeviceProp prop;
   cudaGetDeviceProperties(&prop, 0);
 
+#ifndef __HIP_PLATFORM_AMD__
   if (getDeviceComputeCapability() < blackwellComputeCapability) {
     GTEST_SKIP();
   }
+#endif
 
   using WeightType = InputType;
   DType itype = TypeInfo<InputType>::dtype;
@@ -279,7 +292,7 @@ void performTest(const size_t N, const size_t H, const bool zero_centered_gamma,
   compareResults("output_rowwise", dequantized_output, ref_output.get(), true, atol, rtol, false);
   if (is_training)
     compareResults("output_colwise", dequantized_output, ref_output.get(), false, atol, rtol, false);
-}
+  }
 
 std::vector<std::pair<size_t, size_t>> test_cases = {
   {32, 32},
--- a/tests/cpp/operator/test_normalization_mxfp8.cu
+++ b/tests/cpp/operator/test_normalization_mxfp8.cu
@@ -53,8 +53,8 @@ void dequantize_1x_kernel(InputType* input_ptr, ScaleType* scale_ptr, OutputType
   const size_t blocks_per_tile_Y = tile_size_Y / block_size_Y;
   const size_t blocks_per_tile_X = tile_size_X / block_size_X;
 #ifdef __HIP_PLATFORM_AMD__
-    const auto scale_dims = get_scale_tensor_dims(rows, cols, scaling_mode_x, scaling_mode_y);
-    const size_t blocks_per_row = std::get<3>(scale_dims);
+  const auto scale_dims = get_scale_tensor_dims(rows, cols, scaling_mode_x, scaling_mode_y);
+  const size_t blocks_per_row = std::get<3>(scale_dims);
 #else
   const size_t blocks_per_row = (cols + block_size_X - 1) / block_size_X;
 #endif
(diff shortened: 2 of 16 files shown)


tests/cpp/operator/test_cast_mxfp8.cu
```diff
@@ -69,10 +71,17 @@ void scale_block(const ProcessingMethod processing_method,
...
+#ifndef __HIP_PLATFORM_AMD__
             if (isinf(elt) || isnan(elt)) {
                 continue;
             }
             amax = std::max(amax, std::abs(elt));
+#else
+            if (std::isinf(elt) || std::isnan(elt)) {
+                continue;
+            }
+            amax = fmaxf(amax, fabsf(elt));
```
ipanfilo: ROCm should support std::max, shouldn't it?

tests/cpp/operator/test_normalization_mxfp8.cu
```diff
@@ -49,7 +52,11 @@ void dequantize_1x_kernel(InputType* input_ptr, ScaleType* scale_ptr, OutputType
   const size_t tiles_num_X = (cols + tile_size_X - 1) / tile_size_X;
   const size_t blocks_per_tile_Y = tile_size_Y / block_size_Y;
   const size_t blocks_per_tile_X = tile_size_X / block_size_X;
+#ifdef __HIP_PLATFORM_AMD__
+    const auto [_, __, ___, blocks_per_row] = get_scale_tensor_dims(rows, cols, scaling_mode_x, scaling_mode_y);
```
ipanfilo: ident

tests/cpp/operator/test_normalization_mxfp8.cu
```diff
@@ -276,7 +289,7 @@ void performTest(const size_t N, const size_t H, const bool zero_centered_gamma,
   compareResults("output_rowwise", dequantized_output, ref_output.get(), true, atol, rtol, false);
   if (is_training)
     compareResults("output_colwise", dequantized_output, ref_output.get(), false, atol, rtol, false);
-}
+  }
```
ipanfilo: format

transformer_engine/common/common.h
```diff
@@ -189,6 +189,7 @@ using e8m0_t = uint8_t;
 using bf16 = hip_bfloat16;
 using fp8e4m3 = te_hip_fp8_e4m3;
 using fp8e5m2 = te_hip_fp8_e5m2;
+using e8m0_t = uint8_t;
```
ipanfilo: Just move it out of ifdef

transformer_engine/common/normalization/common.cpp
```diff
@@ -65,7 +65,11 @@ template <typename KernelParamsType>
 TeNormalizationPlan<KernelParamsType>::TeNormalizationPlan(
     NVTE_Norm_Type NormType, NVTE_Norm_Stage NormStage, DType wtype, DType itype, DType otype,
     DType ctype, const size_t batch_size, const size_t hidden_size, const size_t sm_count,
-    const bool zero_centered_gamma, const bool is_tuned)
+#ifdef __HIP_PLATFORM_AMD__
+    const bool zero_centered_gamma, const NVTEScalingMode mode, const bool training, const bool is_tuned)
```
ipanfilo: Can just parameters different between CUDA and ROCm be under ifdef?

transformer_engine/common/normalization/common.cpp
```diff
@@ -101,6 +116,11 @@ void TeNormalizationPlan<ForwardKernelParams>::execute(Tensor* z, void* x_dptr,
   kernel_params.amax = z->amax.dptr;
   kernel_params.scale = z->scale.dptr;
   kernel_params.scale_inv = z->scale_inv.dptr;
+#ifdef __HIP_PLATFORM_AMD__
+  if (kernel_params.mxfp8_out) {
+    kernel_params.z_tensor = z;
```
ipanfilo: I wonder if kernel_params.z can be used instead and TensorWrapper is created where needed. Similar to mxfp8_buffer

ipanfilo (commented):
Please run lint on the code, I think it produces many new errors

tests/cpp/operator/test_cast_mxfp8.cu
alextmagro: Hi Ilya, this is from a previous commit / PR, actually. I think I need to rebase again and will force push over this PR. 

To address your question though, surprisingly, our hipify-torch tool does not support std::max. It casts std::max and std::abs  to ::max and ::abs, which then calls the only hostside max and abs functions found by default by amd-clang, which are for integers. This causes an implicit cast to ints, and an off by one error. Very odd, but a consistent issue throughout the cpp tests.

transformer_engine/common/normalization/common.cpp
alextmagro: Do you mean adding the members of z that we need to kernel_params, then calling TensorWrapper within the functions later to recreate z? We could do that. I don't think it would be a good idea to pass z directly to the functions that call the kernels, as that would require a number of larger changes.

transformer_engine/common/normalization/common.cpp
alextmagro: Yes -- the compiler resolves macro definitions before touching functions, so as long as the function calls are properly guarded as well we can do this. I considered adding these additional parameters as optional, but thought this would be easier to resolve any merge conflicts later on.

tests/cpp/operator/CMakeLists.txt
```diff
@@ -25,15 +25,10 @@ list(APPEND test_cuda_sources
             test_swizzle.cu
             test_cast_mxfp8_gated_swiglu.cu
             test_cast_mxfp8.cu
+            test_normalization_mxfp8.cu
```
ipanfilo: For further IFU merging simplification, please check normalization_mxfp8 goes where it is in upstream

tests/cpp/operator/CMakeLists.txt
```diff
@@ -25,15 +25,10 @@ list(APPEND test_cuda_sources
...
             test_dequantize_mxfp8.cu
+            test_cublaslt_gemm.cu
```
ipanfilo: Has it verified to work on NV?

tests/cpp/operator/test_normalization_mxfp8.cu
```diff
@@ -173,9 +185,11 @@ void performTest(const size_t N, const size_t H, const bool zero_centered_gamma,
   cudaDeviceProp prop;
   cudaGetDeviceProperties(&prop, 0);
 
+#ifndef __HIP_PLATFORM_AMD__
   if (getDeviceComputeCapability() < blackwellComputeCapability) {
     GTEST_SKIP();
   }
+#endif
```
ipanfilo, line 193: on AMD platform should skip if below gfx950

tests/cpp/operator/CMakeLists.txt
alextmagro: I will double check this and verify test_cublaslt_gem.cu works on NV.

tests/cpp/operator/test_normalization_mxfp8.cu
alextmagro: Do we have a macro or function already that can determine if we are above a certain architecture? If so, What do you think of us filtering all mxfp8 tests in c++ and jax in a separate ticket? We currently don't guard against gfx950 support for these tests, as they work on MI300s.

ipanfilo: Hmm, we have inconsistency there then because MXFP8 GEMM tests are guarded for gfx950 only. I'll check if they work on gfx942 then

alextmagro: Did you mention before that those require a specific hipblaslt mxfp8 configuration? For example, if you have to pass the scales to hipblaslt I wouldn't expect them to work, but if you are just passing the fp8 values they might. These tests only touch software enabled mxfp8, so should work anywhere with fp8 support.

transformer_engine/common/normalization/common.cpp
ipanfilo: My inquiry was to keep under ifdef only parameters that are different between CUDA and ROCm, i.e. training and is_tuned. Not zero_centered_gamma and mode

transformer_engine/common/normalization/common.cpp
ipanfilo: kernel_params.z_tensor is used for instance in launch_general_ in ln_fwd_cuda_kernel which creates TensorWrapper around mxfp8_buffer. Can the same be done for kernel_params.z?

transformer_engine/common/normalization/common.h
```diff
@@ -43,18 +43,28 @@ struct LaunchParams {
...
+#ifdef __HIP_PLATFORM_AMD__
+  size_t mxfp8_buffer_bytes = 0;
+#endif
   int multiprocessorCount;
   cudaStream_t stream;
 
   KernelParamsType params;
 
   size_t getTotalWorkspaceBytes(const bool _is_layernorm = true) const {
+#ifdef __HIP_PLATFORM_AMD__
```
ipanfilo, line 62: ident

transformer_engine/common/normalization/common.cpp
alextmagro: Understood, I will make that change. mode and training are the different variables, I have placed them before is_tuned because that is how our hash map is set up.

transformer_engine/common/normalization/common.cpp
alextmagro: Unfortunately, I don't think it would be easy. MXFP8_buffer is a raw device pointer, which makes it easy to wrap. z_tensor requires a lot of other info that isn't already present in kernel_params like colwise data and scales, strides, etc. that are used in quantization. We could pass the entire z tensor for other normalization plans instead of its members, but we would deviate from upstream.

tests/cpp/operator/test_normalization_mxfp8.cu
ipanfilo: Based on offline discussion, filtering should now allow gfx94x and gfx95x with TODO to remove gfx94x when gfx95x is enabled for CI

transformer_engine/common/normalization/common.cpp
ipanfilo: I see. NVTETensor, however, does not fit the idea of kernel_params - it is not something passing to kernel. Can it be part of launch params then? 
Also if mxfp8_out is True, fwd kernels use mxfp8_buffer as output, so it can be moved to launch_params too and passed to kernels as kernel_params.z

transformer_engine/common/normalization/common.h
```diff
@@ -118,6 +132,14 @@ struct ForwardKernelParams : public KernelParamsBase {
 
   // Whether to compute scale and amax
   bool fp8_out;
+
+#ifdef __HIP_PLATFORM_AMD__
+  // MXFP8 requires an additional buffer, and the full tensor
+  bool mxfp8_out;
+  void *mxfp8_buffer;
+  Tensor *z_tensor;
+  bool training;
```
ipanfilo: Where training is used?

transformer_engine/common/util/cast.cu
```diff
@@ -143,6 +143,24 @@ void nvte_quantize_dbias_dsrelu(const NVTETensor input, const NVTETensor activat
       activation_input, input, nullptr, output, dbias, workspace, stream);
 }
 
+#ifdef __HIP_PLATFORM_AMD__
+void nvte_quantize_norm(const NVTETensor input, NVTETensor output, cudaStream_t stream) {
```
ipanfilo: It is  only needed for MXFP8 out case, why not use mxfp8_quantize directly w/o modification of quantize_helper?
And I doubt it should be part of exported TE API

transformer_engine/common/normalization/layernorm/ln_fwd_cuda_kernel.cu
```diff
@@ -56,6 +64,14 @@ static void launch_tuned_(LaunchParams<ForwardKernelParams> &launch_params,
     (void)cudaLaunchCooperativeKernel((void *)kernel, grid, block, (void **)&params_,  // NOLINT(*)
                                       Kernel_traits::SMEM_BYTES_FWD, stream);
   }
+#ifdef __HIP_PLATFORM_AMD__
+  if (launch_params.params.mxfp8_out) {
+    const size_t rows = launch_params.params.rows;
+    const size_t cols = launch_params.params.cols;
+    auto mxfp8_buffer = TensorWrapper(reinterpret_cast<float*>(launch_params.params.mxfp8_buffer), {rows, cols}, DType::kFloat32);
```
ipanfilo: mxfp8_buffer is allocated with using sizeof(compute_t), let's use the same type here for consistency.

ipanfilo (changes requested):

alextmagro:
Hi @ipanfilo, I have made the changes as discussed, and rerun tests. Please let me know if anything else needs fixing, thanks!

transformer_engine/common/normalization/common.h
```diff
@@ -43,18 +45,32 @@ struct LaunchParams {
   size_t workspace_bytes = 0;
   size_t barrier_bytes = 0;
   size_t dgamma_part_bytes = 0;
+#ifdef __HIP_PLATFORM_AMD__
+  size_t mxfp8_buffer_bytes = 0;
+#endif
   int multiprocessorCount;
   cudaStream_t stream;
 
+  // TE MXFP8 quantization parameters
```
ipanfilo, line 54: Aren't those parameters HIP specific? Why are they separated from mxfp8_buffer_bytes

transformer_engine/common/normalization/common.cpp
```diff
@@ -141,6 +159,13 @@ void TeNormalizationPlan<KernelParamsType>::_set_workspace() {
       (void)cudaMemsetAsync(_launch_params.params.barrier, 0, _launch_params.barrier_bytes,
                             _launch_params.stream);
     }
+#ifdef __HIP_PLATFORM_AMD__
+    if constexpr (std::is_same_v<KernelParamsType, ForwardKernelParams>) {
+      size_t offset = _launch_params.workspace_bytes + _launch_params.barrier_bytes;
+      // TODO: Make more general than float for compute_t
```
ipanfilo, line 168: Why typecast is needed if params.z is void*?

transformer_engine/common/normalization/common.h
```diff
@@ -245,7 +269,11 @@ class TeNormalizationPlan : public NormalizationPlanBase {
  public:
   TeNormalizationPlan(NVTE_Norm_Type NormType, NVTE_Norm_Stage NormStage, DType wtype, DType itype,
                       DType otype, DType ctype, const size_t batch_size, const size_t hidden_size,
-                      const size_t sm_count, const bool zero_centered_gamma, const bool is_tuned);
+                      const size_t sm_count, const bool zero_centered_gamma, const NVTEScalingMode mode
```
ipanfilo: Should mode and training be under ifdef instead of is_tuned?
And is there specific reason to insert new parameters before is_tuned not to the end of the parameters list?

transformer_engine/common/include/transformer_engine/cast.h
```diff
@@ -1,4 +1,6 @@
 /*************************************************************************
+ * This file was modified for portability to AMDGPU
+ * Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
```
ipanfilo: If there are no other modifications anymore, lets revert it

transformer_engine/common/normalization/common.cpp
```diff
@@ -101,6 +114,11 @@ void TeNormalizationPlan<ForwardKernelParams>::execute(Tensor* z, void* x_dptr,
   kernel_params.amax = z->amax.dptr;
   kernel_params.scale = z->scale.dptr;
   kernel_params.scale_inv = z->scale_inv.dptr;
+#ifdef __HIP_PLATFORM_AMD__
+  if (kernel_params.mxfp8_out) {
+    _launch_params.z_tensor = z;
```
ipanfilo, line 120: Also set kernel_params.z to nullptr and comment where real value is assigned to it

transformer_engine/common/include/transformer_engine/cast.h
alextmagro: Done

transformer_engine/common/normalization/common.cpp
alextmagro: Done.

transformer_engine/common/normalization/common.h
alextmagro: mode and training are the new parameters, but they can be after is_tuned. I got the order a bit mixed up, they should be at the end to align with the rest of the functions.

transformer_engine/common/normalization/common.cpp
alextmagro: You're right, it isn't needed here. I was following the style of the other casts, but since workspace_dptr has already been cast to byte* we don't need it.

transformer_engine/common/normalization/common.h
alextmagro: The other parameters are from upstream, and used in other normalization plans. We are just increasing the size of the workspace buffer to store the intermediate values.

transformer_engine/common/normalization/common.cpp
```diff
@@ -488,12 +516,18 @@ NormalizationPlanBase* NormalizationPlanRegistry::getNormalizationPlan(
 #endif
   if (NormStage == NVTE_Norm_Stage::Forward) {
     plan = std::make_unique<TeNormalizationPlan<ForwardKernelParams>>(
-        NormType, NormStage, wtype, itype, otype, ctype, batch_size, hidden_size, sm_count,
-        zero_centered_gamma, is_tuned);
+        NormType, NormStage, wtype, itype, otype, ctype, batch_size, hidden_size, sm_count,zero_centered_gamma, is_tuned
```
ipanfilo: Minor: it probably should be split to two lines like it was original to fit 100 width

transformer_engine/common/normalization/common.cpp
```diff
@@ -488,12 +516,18 @@ NormalizationPlanBase* NormalizationPlanRegistry::getNormalizationPlan(
...
+#ifdef __HIP_PLATFORM_AMD__
+        , mode, training
+#endif
+      );
   } else {
     plan = std::make_unique<TeNormalizationPlan<BackwardKernelParams>>(
-        NormType, NormStage, wtype, itype, otype, ctype, batch_size, hidden_size, sm_count,
-        zero_centered_gamma, is_tuned);
+        NormType, NormStage, wtype, itype, otype, ctype, batch_size, hidden_size, sm_count, zero_centered_gamma, is_tuned
```
ipanfilo: The same a line 519

transformer_engine/common/normalization/common.h
ipanfilo: mxfp8_buffer_bytes above is HIP specific and guarded
Why z_tensor and training are separated from it and not guarded?

alextmagro: I had kept mxfp8_buffer_bytes where it was so it was defined alongside the other workspace variables, but will move it down and guard everything.

ipanfilo (approved):

PR #279, opened 2025-08-21
[Feat] Add transpose cache to LayerNorm kernel

AllenFarcas:
# Description

Fixes  https://github.com/ROCm/frameworks-internal/issues/13313

## Type of change

- [ ] Infra/Build change
- [X] Code refactoring

## Changes

- Added transpose cache for LayerNorm kernel FP8 implementation (following RMSNorm kernel implementation of transpose cache)
- Refactored variable names in both LayerNorm and RMSNorm to improve clarity and readability as well as unify variable names

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/triton_kernels/layernorm.py
+++ b/transformer_engine/pytorch/triton_kernels/layernorm.py
@@ -456,13 +456,13 @@ def _layernorm_bwd_dwdb_triton_v2(
     tl.store(FINAL_DW + cols, sum_dw.to(FINAL_DW.type.element_ty), mask=cols < N)
     tl.store(FINAL_DB + cols, sum_db.to(FINAL_DB.type.element_ty), mask=cols < N)
 
-def te_layernorm_fwd_triton(input: torch.Tensor, 
-                            weight: torch.Tensor, 
-                            bias: torch.Tensor,  
+def te_layernorm_fwd_triton(input: torch.Tensor,
+                            weight: torch.Tensor,
+                            bias: torch.Tensor,
                             eps: float,
-                            ln_out: torch.Tensor, 
-                            quantizer: Quantizer, 
-                            out_dtype: TE_DType,
+                            ln_out: torch.Tensor,
+                            quantizer: Quantizer,
+                            otype: tex.DType,
                             sm_margin: int,
                             zero_centered_gamma: bool):
     if sm_margin is not None and sm_margin > 0:
@@ -473,7 +473,7 @@ def te_layernorm_fwd_triton(input: torch.Tensor,
     device = input.device
     M, N = input.shape
 
-    IS_MFP8 = isinstance(quantizer, MXFP8Quantizer)
+    IS_MXFP8 = isinstance(quantizer, MXFP8Quantizer)
     MAKE_TRANSPOSE = False
 
     # Create empty tensors for mu and rsigma
@@ -481,11 +481,11 @@ def te_layernorm_fwd_triton(input: torch.Tensor,
     rsigma = torch.empty((M,), dtype=torch.float32, device=device)
 
     torch_out_dtype = (
-        out_dtype if isinstance(out_dtype, torch.dtype)
-        else te_dtype_to_torch_dtype(out_dtype)
+        otype if isinstance(otype, torch.dtype)
+        else te_dtype_to_torch_dtype(otype)
     )
     # Create ln_out
-    if quantizer is None or IS_MFP8:
+    if quantizer is None or IS_MXFP8:
         ln_out = torch.empty(M, N, dtype=torch_out_dtype, device=device)
     else:
         if ln_out is None:
@@ -567,7 +567,7 @@ def te_layernorm_fwd_triton(input: torch.Tensor,
         )
 
     # For MXFP8, we do regular layernorm and then quantize it separately
-    if IS_MFP8:
+    if IS_MXFP8:
         ln_out = te_quantize_triton(ln_out, quantizer)
     
     # Reduce and find amax if "not APPLY_ATOMIC" is True.
--- a/transformer_engine/pytorch/triton_kernels/layernorm.py
+++ b/transformer_engine/pytorch/triton_kernels/layernorm.py
@@ -57,17 +57,21 @@ def _layernorm_fwd_triton(
     n_rows,
     n_cols,
     eps,
+    out_transpose_ptr,
+    out_transpose_stride,
     ZERO_CENTERED_GAMMA: tl.constexpr,
     BLOCK_SIZE: tl.constexpr,
     IS_FP8: tl.constexpr,
     APPLY_ATOMIC: tl.constexpr,
     PERSISTENT: tl.constexpr,
     FP8_MAX: tl.constexpr,
+    MAKE_TRANSPOSE: tl.constexpr
 ):
 
     # program id
     pid = tl.program_id(0)
     num_tiles = tl.num_programs(0)
+    col_offsets = tl.arange(0, BLOCK_SIZE)
 
     if PERSISTENT:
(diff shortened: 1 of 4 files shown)


transformer_engine/pytorch/triton_kernels/rmsnorm.py
```diff
@@ -357,21 +359,22 @@ def te_rmsnorm_bwd_triton(dz, x, rsigma, gamma, sm_margin, zero_centered_gamma):
...
-    quantizer,
-    otype,
-    sm_margin,
-    zero_centered_gamma
+    input: torch.Tensor,
+    weight: torch.Tensor,
+    eps: float,
+    ln_out: torch.Tensor,
+    quantizer: Quantizer,
+    out_dtype: TE_DType,
```
Micky774: The correct variable here is actually `otype` to match the HIP kernels, and the correct typing is `tex.DType` (note that `TE_DType` is actually a conversion dict).

transformer_engine/pytorch/triton_kernels/layernorm.py
```diff
@@ -474,13 +472,20 @@ def te_layernorm_fwd_triton(input: torch.Tensor,
         )
     device = input.device
     M, N = input.shape
-    
+
+    IS_MFP8 = isinstance(quantizer, MXFP8Quantizer)
```
Micky774: This should be `IS_MXFP8` -- it was a typo on my end in the RMSNorm side

Micky774 (changes requested):

transformer_engine/pytorch/triton_kernels/rmsnorm.py
AllenFarcas: This is how it was implemented in LayerNorm and I used the same notation in the RMSNorm for uniformity.

`out_dtype: TE_DType`

Micky774: It's wrong in LayerNorm, please change it there too

transformer_engine/pytorch/triton_kernels/layernorm.py
```diff
@@ -147,16 +143,17 @@ def _layernorm_fwd_triton(
                 amax = amax_temp if amax_temp > amax else amax
                 y_block = y_block * scale
                 y_block = tl.clamp(y_block, -FP8_MAX, FP8_MAX)
-            tl.store(y_ptr_start + col_offsets, y_block.to(y_ptr.type.element_ty))
+            tl.store(y_ptr_start + cols, y_block.to(y_ptr.type.element_ty))
```
brunomazzottiamd: **Suggestion:** What about doing `y_block = y_block.to(y_ptr.type.element_ty)` before storing it? By this way, I think the code will be less repetitive since now you're casting twice (for y and for the transpose cache). This suggestion is also applicable to the last masked iteration.

transformer_engine/pytorch/triton_kernels/layernorm.py
```diff
@@ -506,11 +511,24 @@ def te_layernorm_fwd_triton(input: torch.Tensor,
         amax_out = quantizer.amax
         scale_inv = ln_out._scale_inv
         cast_out = ln_out._data
+        MAKE_TRANSPOSE = quantizer.columnwise_usage
+        tl_dtype = te_dtype_to_triton_dtype(quantizer.dtype)
```
brunomazzottiamd: **Suggestion:** `tl_dtype` variable is only used inside `if MAKE_TRANSPOSE:` statement. I'd move `tl_dtype` declaration and assignment to that `if` statement, thus reducing variable scope.

transformer_engine/pytorch/triton_kernels/layernorm.py
```diff
@@ -506,11 +511,24 @@ def te_layernorm_fwd_triton(input: torch.Tensor,
...
+            out_transpose_stride = ln_out._transpose.stride(0)
+        else:
+            out_transpose_ptr = None
+            out_transpose_stride = None
     else:
         scale = None
         amax_out = None
         scale_inv = None
         cast_out = ln_out
+        out_transpose_ptr = None
```
brunomazzottiamd: **Suggestion:** You're assigning `None` to `out_transpose_ptr` and `out_transpose_stride` in two distinct places. I'd be more defensive and let them begin as `None` and just change their values in case the feature is enabled.

Something like this:
```python
out_transpose_ptr, out_transpose_stride = None, None

if IS_FP8 and MAKE_TRANSPOSE:
   out_transpose_ptr = ...
   out_transpose_stride = ...
```

transformer_engine/pytorch/triton_kernels/layernorm.py
```diff
@@ -546,7 +567,7 @@ def te_layernorm_fwd_triton(input: torch.Tensor,
         )
 
     # For MXFP8, we do regular layernorm and then quantize it separately
-    if isinstance(quantizer, MXFP8Quantizer):
+    if IS_MXFP8:
         ln_out = te_quantize_triton(ln_out, quantizer)
```
brunomazzottiamd, line 575: After reviewing `layernorm.py` source file, I think the transpose cache feature only makes sense for `fp8` data types. Am I right?

If this is the case then I suggest adding a [`static_assert`](https://triton-lang.org/main/python-api/generated/triton.language.static_assert.html#triton.language.static_assert) to the Triton kernel, in order to guarantee this condition at kernel compile time. Something like this:

```python
tl.static_assert(
    not MAKE_TRANSPOSE or IS_FP8,
    msg="Transpose cache requires fp8 data type."
)
```

Please double check my Boolean logic!

transformer_engine/pytorch/triton_kernels/rmsnorm.py
```diff
@@ -457,7 +460,7 @@ def te_rmsnorm_fwd_triton(
         FP8_MAX,
         MAKE_TRANSPOSE,
     )
-    if IS_MFP8:
+    if IS_MXFP8:
         out = quantizer.quantize(out)
 
     return out, None, rsigma
```
brunomazzottiamd: **Suggestion:** Since you're touching this file, please add an empty new line at the end of it.

Micky774 (approved):

Micky774 (changes requested):
Code looks good, but I'm going to block pending on a level 3 CI run ([link](https://ml-ci-internal.amd.com/job/TransformerEngine/job/TransformerEngine/job/PR-279/3/)) that I just launched. Once that's green I'll update my review.

brunomazzottiamd (approved):
@AllenFarcas, I have some minor suggestions for you. It's nothing that important. I think the most relevant suggestion is the one related to the kernel compile time assertion. Feel free to ignore my nitpicks if you wish, no worries at all.

Please keep us posted about your layernorm optimization ideas regarding a 2-pass implementation!

brunomazzottiamd:
@AllenFarcas, how easy is it to create a new unit test that verifies the transpose cache was correctly populated?

I've been thinking about something like this:

1. create a random input $x$, with the required conditions to trigger `MAKE_TRANSPOSE=True` ($x$ can have any shape, we don't care and don't want to test multiple shapes)
2. compute $y = layernorm(x)$
3. assert that the transpose cache of $y$ exists and it's bit-identical to $y^T$

The other way around, you can create $x$ with the required conditions to set `MAKE_TRANSPOSE=False` and assert that the transpose cache of $y$ is absent.

Micky774:
> @AllenFarcas, how easy is it to create a new unit test that verifies the transpose cache was correctly populated?

We already have tests in place which compare the transpose cache to the transpose cache produced by the HIP kernels, so I don't think we need to add any new tests there.

Micky774 (approved):
CI Failures seem unrelated -- lvl 3 sGPU tests pass, so I'm satisfied.

PR #295, opened 2025-09-02
Fix AITER ASM kernels for editable install

Micky774:
# Description

Add symlink for AITER ASM in editable installs

Fixes https://github.com/ROCm/frameworks-internal/issues/13520

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
```diff
@@ -40,7 +40,7 @@ void set_aiter_asm_dir() {
     Dl_info info;
     dladdr((void*)set_aiter_asm_dir, &info);
     setenv("AITER_ASM_DIR",
-           (std::filesystem::path(info.dli_fname).parent_path() / "aiter" / arh_str).c_str(), 1);
+           (std::filesystem::path(info.dli_fname).parent_path() / "3rdparty" / "aiter" / "hsa" / arh_str).c_str(), 1);
```
wangye805: Is this change needed for pip editable installation? Will it break the non-editable installation? 

And I also recall that editable installation will be depreciated with newer setuptool? @ipanfilo can confirm. Thanks

Micky774: Yes, this is needed for the editable install. Ah, I think you're right that it would break regular install since we have 
```
  install(DIRECTORY
        ${__AITER_SOURCE_DIR}/hsa/${ARCH}/fmha_v3_fwd
        DESTINATION ${CMAKE_INSTALL_PREFIX}/transformer_engine/aiter/${ARCH}/
        PATTERN "codegen.py" EXCLUDE)
  install(DIRECTORY
        ${__AITER_SOURCE_DIR}/hsa/${ARCH}/fmha_v3_bwd
        DESTINATION ${CMAKE_INSTALL_PREFIX}/transformer_engine/aiter/${ARCH}/
        PATTERN "codegen.py" EXCLUDE)
```
in our `CMakeLists.txt`

I think we're going to try to take a look at getting our editable install working for `setuptools>80.0.0` anyways right? So this is still something that we'd benefit from resolving. Perhaps we update the output directory that the ASM files are copied to and set it to `${CMAKE_INSTALL_PREFIX}/transformer_engine/3rdparty/aiter/hsa/${ARCH}/` to match the editable case?

ipanfilo: The suggested change does not correctly solve editable install too. AITER asm kernels are copied to transformer_engine/aiter.
The problem with editable install is that libtransformer_engine.so is not under transformer engine but in root, while the rest components (AITER, AOTriton) are under subdir

ipanfilo (changes requested):
The change incorrectly solves editable install problem.
As a w/a, you can copy/link transformer_engine/aiter direcroty to TE root after editable install

Micky774:
@ipanfilo @wangye805 ready for review

setup.py
```diff
@@ -216,7 +241,7 @@ def setup_requirements() -> Tuple[List[str], List[str], List[str]]:
         long_description=long_description,
         long_description_content_type="text/x-rst",
         ext_modules=ext_modules,
-        cmdclass={"build_ext": CMakeBuildExtension, "bdist_wheel": TimedBdist},
+        cmdclass=cmdclass,
```
ipanfilo, line 240: But cmdclass is not defined in this method, is it?

setup.py
```diff
@@ -171,6 +194,8 @@ def setup_requirements() -> Tuple[List[str], List[str], List[str]]:
         setup_requires, install_requires, test_requires = setup_requirements()
         ext_modules = [setup_common_extension()]
         cmdclass = {"build_ext": CMakeBuildExtension, "bdist_wheel": TimedBdist}
+        if rocm_build():
+            cmdclass["develop"] = develop
```
ipanfilo, line 194: Does "develop" affect editable install only?

setup.py
```diff
@@ -49,6 +50,28 @@
...
+        if (
+            int(os.getenv("NVTE_FUSED_ATTN_CK", "1")) and
+            int(os.getenv("NVTE_FUSED_ATTN", "1"))
+        ):
+            # Ensure that the AITER ASM kernels are properly available at runtime
+            # by creating a symlink to them. This is only necessary for editable
+            # mode since our C++ code assumes the AITER ASM kernel paths relative
+            # to trasnformer_engine.so, which is different in editable installs.
+            import triton
+            arh_str = triton.runtime.driver.active.get_current_target().arch
```
ipanfilo: Why not just link or even move aiter dir? It is not actually needed under transformer_engine for editable install

setup.py
Micky774: Yes, editable install is equivalent to calling `python setup.py develop` and it does not affect regular install

setup.py
Micky774: It is, see lines 184 (not in diff) and 198

setup.py
Micky774: Very good point on just linking `aiter` directly. I prefer a symlink for now just so we avoid needless IO as opposed to moving the kernels.

setup.py
```diff
@@ -49,6 +50,25 @@
...
+            # Ensure that the AITER ASM kernels are properly available at runtime
+            # by creating a symlink to them. This is only necessary for editable
+            # mode since our C++ code assumes the AITER ASM kernel paths relative
+            # to trasnformer_engine.so, which is different in editable installs.
+            project_dir = Path(__file__).parent
+            asm_src_dir = project_dir / 'transformer_engine' / 'aiter'
+            # Must be synced with
+            # TransformerEngine/transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
+            asm_target_dir = project_dir / 'aiter'
+            if asm_src_dir.is_dir() and not asm_target_dir.is_dir():
```
ipanfilo, line 70: If symlink already exists, what will be is_dir() result, and, if it is false, will it correctly overwrite the symlink on the next line?

ipanfilo: A reason not to move it?

setup.py
ipanfilo: My bad, I thought it is separate function. 
You change changes behaviour of release wheel building (with NVTE_BUILD_METAPACKAGE). You need to validate that it works OK after the change, or, original code seems have unused cmdclass assignments on lines 181 and 193, so you may correct line 181 one

setup.py
Micky774: > If symlink already exists, what will be is_dir() result, and, if it is false, will it correctly overwrite the symlink on the next line?

If the sym link already exists, then `asm_target_dir.is_dir() == True` hence nothing further is done.

> A reason not to move it?

Just to avoid unnecessary IO out of principle, but I'm not married to a symlink solution.

setup.py
Micky774: Okay I've adjusted it to be equivalent to the old functionality in case we don't go down the ROCm build path. I'm not sure what exactly goes on with the build metapackage path, so I figure this is safer and we can revisit later if we ever need to (i.e. in case originally the `cmdclass` being overwritten was a bug).

wangye805 (approved):

ipanfilo (approved):

Issue #293, opened 2025-09-02
[Bug] what():  hipGetDeviceProperties failed with error: no ROCm-capable device is detected


caaatch22:
**Describe the bug**

Similar to https://github.com/ROCm/TransformerEngine/pull/222 and https://github.com/ROCm/TransformerEngine/pull/183

**Steps/Code to reproduce bug**

Build lastest transformer_engine with rocm7.0.0
 
run:
`HIP_VISIBLE_DEVICES="" python -c "import transformer_engine"`

result:
`terminate called after throwing an instance of 'std::runtime_error'
  what():  hipGetDeviceProperties failed with error: no ROCm-capable device is detected
Aborted (core dumped)`

if I run:
`HIP_VISIBLE_DEVICES="0" python -c "import transformer_engine"`
result:
`(python3.10.13) [root@e01-cn-ioy49f8z90q ~]# HIP_VISIBLE_DEVICES="0" python -c "import transformer_engine"
[WARNING  | transformer_engine.pytorch.dot_product_attention.utils]: Supported flash-attn versions are >= 2.1.1, <= 2.8.0.post2. Found flash-attn 2.8.3.`
(which exist normally)

**Expected behavior**

Should not  throw this  error in import phase, many RL framework with ray would import megatron.core(which import TE) without gpu set, e.g. in remote worker: @ray.remote(cpu=1) ...

caaatch22:
With TE build from 5571d1c106dc21ec8e250cd5d838274a0add4568, we will not encounter this issue

ipanfilo:
TE commit 54dd2bdc4508c895a6ff7098fb86e1bfc88f5364

PR #299, opened 2025-09-04
Don't call HIP on TE load

ipanfilo:
# Description

Do not call HIP on TE load to determine FP8 FNUZ/OCP limits

Fixes  [Import error when there is no GPU #13532](https://github.com/ROCm/frameworks-internal/issues/13532) / ROCm/TransformerEngine#293
CI failure

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Updated FP8 numeric traits and QuantizationLimits to postpone FNUZ/OCP limit selection on host path till actual usage

# Checklist:

### DIFF
--- a/ci/jax.sh
+++ b/ci/jax.sh
@@ -50,6 +50,7 @@ run_test_config() {
     NVTE_CK_USES_FWD_V3=1 NVTE_CK_USES_BWD_V3=1 run 1 test_fused_attn.py # Using FAv3 for forward and backward pass
     run_default_fa 1 test_helper.py
     run_default_fa 1 test_layer.py #it effectevly always uses unfused attention
+    run_default_fa 1 test_sanity_import.py
     run_default_fa 1 test_sharding.py
     run_default_fa 1 test_softmax.py
 }
--- a/tests/jax/test_sanity_import.py
+++ b/tests/jax/test_sanity_import.py
@@ -11,13 +11,20 @@ if __name__ == "__main__":
 
 def test_lazy_init():
     import jax, os, pytest, subprocess, sys
-    dev_count = jax.local_device_count()
-    if dev_count < 2:
-        pytest.skip("Test requires several visible devices")
+    if not transformer_engine.jax.is_hip_extension():
+        pytest.skip("This is ROCm test")
     os.environ["NVTE_FRAMEWORK"] = "jax"
-    dev_count_test = subprocess.run([sys.executable, '-c', 
-                                     "import transformer_engine, jax, os; " +
-                                     "os.environ['HIP_VISIBLE_DEVICES']='0'; "+
-                                     "exit(jax.local_device_count())"]
-                                     ).returncode
-    assert dev_count_test==1, "Changing visible devices after import did not affect device count"
+    ret = subprocess.run(
+        [sys.executable, "-c",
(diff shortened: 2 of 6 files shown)


ci/jax.sh
```diff
@@ -50,6 +50,7 @@ run_test_config() {
     NVTE_CK_USES_FWD_V3=1 NVTE_CK_USES_BWD_V3=1 run 1 test_fused_attn.py # Using FAv3 for forward and backward pass
     run_default_fa 1 test_helper.py
     run_default_fa 1 test_layer.py #it effectevly always uses unfused attention
+    run_default_fa 1 test_sanity_import.py
```
wenchenvincent, line 53: Was this test not run on CI before?

tests/jax/test_sanity_import.py
```diff
@@ -11,13 +11,20 @@
...
-                                     "import transformer_engine, jax, os; " +
-                                     "os.environ['HIP_VISIBLE_DEVICES']='0'; "+
-                                     "exit(jax.local_device_count())"]
-                                     ).returncode
-    assert dev_count_test==1, "Changing visible devices after import did not affect device count"
+    ret = subprocess.run(
+        [sys.executable, "-c",
+         "import os; os.environ['HIP_VISIBLE_DEVICES']=''; import transformer_engine"]
+         ).returncode
+    assert ret == 0, "Failed to import TE witout visible devices"
```
wenchenvincent: Nit: without

transformer_engine/common/utils.cuh
```diff
@@ -991,18 +991,24 @@ struct Numeric_Traits;
...
-#else
+#ifndef __HIP_PLATFORM_AMD__
   static constexpr double maxNorm = 448;
-#endif //__HIP_PLATFORM_AMD__
+#elif defined(__HIP_DEVICE_COMPILE__)
+  static constexpr double maxNorm = te_fp8_fnuz() ? 240 : 448;
+#elif defined(TE_DYNAMIC_HIP_FP8_TYPE)
+ // dummy declaration for correct translation;
+ // it is not defined anywhere and it's usage should be eliminated before linking
+  static double maxNorm;
```
wenchenvincent, line 1001: Is this `maxNorm` not used?

ci/jax.sh
ipanfilo: It was on level 3 in mGPU section because originally tests required mGPU

transformer_engine/common/utils.cuh
ipanfilo: No. Numeric_Traits_fp8e4m3 is used instead for fp8e4m3 datatype to implement FNUZ/OCP difference. It, however, should be declared because translator translates all branches of constexpr condition

wenchenvincent: Thanks. So it seems that the `maxNorm` is only used at device side and thus it is safe to set it to an undefined value on the host side, right?

ipanfilo: Right, it is not used on device. Just to clarify - it is not set to undefined value but not set at all. I.e. if host code really uses it, linker will fail.

wenchenvincent (approved):

wenchenvincent:
Let's wait for CI to finish. A bit strange that it was not started yet..

PR #304, opened 2025-09-04
Fix GPU fault due to invalid argument with current quantization

ipanfilo:
# Description

Fixes [13564](https://github.com/ROCm/frameworks-internal/issues/13564)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Properly pass tensors to triton kernel when Float8CurrentScalingQuantizer is used

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/triton_kernels/layernorm.py
+++ b/transformer_engine/pytorch/triton_kernels/layernorm.py
@@ -476,8 +476,7 @@ def te_layernorm_fwd_triton(input: torch.Tensor,
     M, N = input.shape
 
     #TODO: Check if need distinct Current and Delayed quantization
-    IS_FP8 = (isinstance(quantizer, Float8Quantizer) or 
-              isinstance(quantizer, Float8CurrentScalingQuantizer))
+    IS_FP8 = isinstance(quantizer, (Float8Quantizer, Float8CurrentScalingQuantizer))
     IS_MXFP8 = isinstance(quantizer, MXFP8Quantizer)
     assert (quantizer is None or IS_FP8 or IS_MXFP8), "Unsupported quantizer type"
 
--- a/transformer_engine/pytorch/triton_kernels/layernorm.py
+++ b/transformer_engine/pytorch/triton_kernels/layernorm.py
@@ -7,7 +7,7 @@ import os
 
 import torch
 
-from ..tensor.float8_tensor import Float8Quantizer
+from ..tensor.float8_tensor import Float8Quantizer, Float8CurrentScalingQuantizer
 from ..constants import TE_DType
 from ..tensor.mxfp8_tensor import MXFP8Quantizer
 from ..tensor.quantized_tensor import Quantizer
@@ -474,13 +474,21 @@ def te_layernorm_fwd_triton(input: torch.Tensor,
         )
     device = input.device
     M, N = input.shape
-    
+
+    # MXFP8 is handled regularly, hence other quantizers are considered FP8
+    IS_FP8_DELAYED = isinstance(quantizer, Float8Quantizer)
+    IS_FP8 = IS_FP8_DELAYED or isinstance(quantizer, Float8CurrentScalingQuantizer)
+    IS_MXFP8 = isinstance(quantizer, MXFP8Quantizer)
+    assert (quantizer is None or IS_FP8 or IS_MXFP8), "Unsupported quantizer type"
+
+    MAKE_TRANSPOSE = False
+
     # Create empty tensors for mu and rsigma
     mu = torch.empty((M,), dtype=torch.float32, device=device)
     rsigma = torch.empty((M,), dtype=torch.float32, device=device)
     torch_out_dtype = te_dtype_to_torch_dtype(out_dtype)
     # Create ln_out
-    if quantizer is None or isinstance(quantizer, MXFP8Quantizer):
+    if quantizer is None or IS_MXFP8:
         ln_out = torch.empty(M, N, dtype=torch_out_dtype, device=device)
     else:
         if ln_out is None:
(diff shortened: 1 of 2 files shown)


Micky774 (commented):
Is there a reason to distinguish between `IS_FP8_DELAYED` and `IS_FP8` or should we just have 
```python
IS_FP8 = isinstance(quantizer, (Float8Quantizer, Float8CurrentScalingQuantizer))
```

transformer_engine/pytorch/triton_kernels/layernorm.py
```diff
@@ -474,13 +474,21 @@ def te_layernorm_fwd_triton(input: torch.Tensor,
...
     M, N = input.shape
-    
+
+    # MXFP8 is handled regularly, hence other quantizers are considered FP8
+    IS_FP8_DELAYED = isinstance(quantizer, Float8Quantizer)
+    IS_FP8 = IS_FP8_DELAYED or isinstance(quantizer, Float8CurrentScalingQuantizer)
+    IS_MXFP8 = isinstance(quantizer, MXFP8Quantizer)
+    assert (quantizer is None or IS_FP8 or IS_MXFP8), "Unsupported quantizer type"
+
+    MAKE_TRANSPOSE = False
```
sudhu2k: I don't see MAKE_TRANSPOSE being used anywhere in layernorm.py

sudhu2k (changes requested):

transformer_engine/pytorch/triton_kernels/layernorm.py
Micky774: This is for merging into release 2.2 which doesn't have some of the triton norm updates (e.g. layernorm fused transpose which requires `MAKE_TRANSPOSE`) so we can go without the `MAKE_TRANSPOSE` flag.

ipanfilo: Oh, yes. MAKE_TRANSPOSE is from another commit cherry-pick which I didn't fully cleaned up

ipanfilo:
> Is there a reason to distinguish between `IS_FP8_DELAYED` and `IS_FP8` or should we just have
> 
> ```python
> IS_FP8 = isinstance(quantizer, (Float8Quantizer, Float8CurrentScalingQuantizer))
> ```

I made them separate bearing in mind there may be further changes specific for CURRENT vs DELAYED scaling. Here I'd like to get your input

Micky774:
> I made them separate bearing in mind there may be further changes specific for CURRENT vs DELAYED scaling. Here I'd like to get your input

In that case, let's go without it for now. I think eventually we'll want to support current scaling directly (i.e. we don't compute `amax`) since even though it's not a huge computation burden, the role of the atomic operation may be outsized, but I don't like having a dangling/unused constant. We can introduce it back when we introduce separate behavior for the two scaling modes.

transformer_engine/pytorch/triton_kernels/layernorm.py
```diff
@@ -474,13 +474,19 @@ def te_layernorm_fwd_triton(input: torch.Tensor,
         )
     device = input.device
     M, N = input.shape
-    
+
+    #TODO: Check if need distinct Current and Delayed quantization
+    IS_FP8 = (isinstance(quantizer, Float8Quantizer) or 
+              isinstance(quantizer, Float8CurrentScalingQuantizer))
```
Micky774: Nit:
```python
    IS_FP8 = isinstance(quantizer, (Float8Quantizer, Float8CurrentScalingQuantizer))
```

transformer_engine/pytorch/triton_kernels/rmsnorm.py
```diff
@@ -378,8 +382,11 @@ def te_rmsnorm_fwd_triton(
             f"The shape of `weight` must be feature-aligned, "
             f"but {weight.shape[0]=} while {input.shape[1]=}"
         )
-    IS_FP8 = isinstance(quantizer, Float8Quantizer)
-    IS_MFP8 = isinstance(quantizer, MXFP8Quantizer)
+
+    #TODO: Check if need distinct Current and Delayed quantization
+    IS_FP8 = (isinstance(quantizer, Float8Quantizer) or 
+              isinstance(quantizer, Float8CurrentScalingQuantizer))
```
Micky774: Nit:
```python
    IS_FP8 = isinstance(quantizer, (Float8Quantizer, Float8CurrentScalingQuantizer))
```

Micky774 (approved):
Minor nit, otherwise looks good. Thanks!

sudhu2k (approved):
LGTM

Micky774 (changes requested):
We should add explicit tests for our `Float8CurrentScalingQuantizer` support.

ipanfilo:
> We should add explicit tests for our `Float8CurrentScalingQuantizer` support.

Please elaborate.  The numerical tests that were failing use CurrentScalingQuantizer. Moreover, it was used even before 2.2 changes that broke Triton path. Adding of extra tests to validate that with CurrentScaling we don't keep Amax history is out of scope of this crash fixing

Micky774:
> > We should add explicit tests for our `Float8CurrentScalingQuantizer` support.
> 
> Please elaborate. The numerical tests that were failing use CurrentScalingQuantizer. Moreover, it was used even before 2.2 changes that broke Triton path. Adding of extra tests to validate that with CurrentScaling we don't keep Amax history is out of scope of this crash fixing

I meant extending `test/pytorch/triton_kernels/test_norms.py` to include it as a quantizer so we explicitly test API, but I forgot that this was already caught by our CI meaning we have some coverage. You're right that a more comprehensive testing is out of scope for this.

Micky774 (approved):

PR #308, opened 2025-09-10
Ifu dev 250517 v2.4

alextmagro:
IFU 2.3 and 2.4, will run test level 3 on Jenkins and finish confluence page shortly.


build_tools/build_ext.py
```diff
@@ -135,25 +136,29 @@ def run(self) -> None:
...
+                    os.remove(ext)
 
         def build_extensions(self):
-            # BuildExtensions from PyTorch already handle CUDA files correctly
-            # so we don't need to modify their compiler. Only the pybind11 build_ext needs to be fixed.
-            ext_names = [ext.name for ext in self.extensions]
-            if "transformer_engine_pytorch" not in ext_names:
+            # For core lib + JAX install, fix build_ext from pybind11.setup_helpers
+            # to handle CUDA files correctly.
+            if "pytorch" not in get_frameworks():
```
ipanfilo: ext_names should be used instead of get_frameworks() - it is a bug fix for multiframework environment

build_tools/build_ext.py
```diff
@@ -176,10 +182,14 @@ def _compile_fn(obj, src, ext, cc_args, extra_postargs, pp_opts) -> None:
...
-                            _, nvcc_bin = cuda_path()
+                            nvcc_bin = nvcc_path()
                         original_compiler = self.compiler.compiler_so
 
-                        if os.path.splitext(src)[1] in [".cu", ".cuh", ".hip"]:
+                        if (
+                            os.path.splitext(src)[1] in [".cu", ".cuh"]
+                            and not framework_extension_only
+                        ):
+                            nvcc_bin = nvcc_path()
```
ipanfilo: Nvcc_bin correctly set above. This line should be removed

build_tools/utils.py
```diff
@@ -256,7 +279,38 @@ def cuda_path() -> Tuple[str, str]:
...
+@functools.lru_cache(maxsize=None)
+def get_cuda_include_dirs() -> Tuple[str, str]:
+    """Returns the CUDA header directory."""
+
+    # If cuda is installed via toolkit, all necessary headers
+    # are bundled inside the top level cuda directory.
+    if cuda_toolkit_include_path() is not None:
+        return [cuda_toolkit_include_path()]
+
+    if rocm_build():
```
ipanfilo: It should be the first check in the function

build_tools/jax.py
```diff
@@ -43,52 +43,43 @@ def setup_jax_extension(
...
     # Compile flags
     cxx_flags = ["-O3"]
-    nvcc_flags = ["-O3"]
-
-    if rocm_build():
-        # Pybind11 extension does not know about HIP so specify necessary parameters here
-        rocm_home, _ = rocm_path()
-        macros=[("USE_ROCM",None)]
-        cxx_flags.extend(["-D__HIP_PLATFORM_AMD__", "-I{}/include".format(str(rocm_home))])
-        nvcc_flags.extend([f"--offload-arch={arch}" for arch in
```
ipanfilo, line 82: How is archs list passed to hipcc then?

build_tools/pytorch.py
```diff
@@ -27,87 +29,43 @@ def setup_pytorch_extension(
...
 
     # Compiler flags
-    cxx_flags = [
-        "-O3",
-        "-fvisibility=hidden",
-    ]
-    if rocm_build():
-        nvcc_flags = [
-            "-O3",
-            "-U__HIP_NO_HALF_OPERATORS__",
```
ipanfilo, line 59: Are these defines still defined by default?

build_tools/pytorch.py
```diff
@@ -27,87 +29,43 @@ def setup_pytorch_extension(
...
-            "--use_fast_math",
-        ]
+        cxx_flags.append("-g0")
 
-    if rocm_build():
-        ##TODO: Figure out which hipcc version starts to support this parallel compilation
-        nvcc_flags.extend(["-parallel-jobs=4"])
-        # Pytorch extension can get the ROCm architecture from the environment variable
-        if os.getenv("NVTE_ROCM_ARCH") is not None:
-            os.environ["PYTORCH_ROCM_ARCH"] = os.getenv("NVTE_ROCM_ARCH")
```
ipanfilo, line 85: how are target architectures passed to the extension?

build_tools/utils.py
```diff
@@ -197,7 +198,7 @@ def rocm_build() -> bool:
             else:
                 raise FileNotFoundError(f"Could not find hipcc at {hipcc_bin}")
         else:
-            cuda_path()
+            nvcc_bin = nvcc_path()
```
ipanfilo: nvcc_bin is unused

build_tools/utils.py
```diff
@@ -207,7 +208,7 @@ def rocm_build() -> bool:
 
     # Try to detect CUDA
     try:
-        cuda_path()
+        nvcc_bin = nvcc_path()
```
ipanfilo: nvcc_bin is unused

examples/jax/encoder/common.py
```diff
@@ -29,3 +32,23 @@ def is_fp8_supported():
         # only GFX9.4 machines support fp8
         return gpu_arch == 94
     return gpu_arch >= 90
+
+
+@lru_cache
+def is_mxfp8_supported():
+    """Return if FP8 has hardware supported"""
```
ipanfilo, line 39: ROCm path is needed

setup.py
```diff
@@ -84,11 +82,15 @@ def setup_common_extension() -> CMakeExtension:
...
+        if bool(int(os.getenv("NVTE_ENABLE_NVSHMEM", "0"))):
+            assert (
+                os.getenv("NVSHMEM_HOME") is not None
+            ), "NVSHMEM_HOME must be set when compiling with NVTE_ENABLE_NVSHMEM=1"
+            cmake_flags.append("-DNVTE_ENABLE_NVSHMEM=ON")
+
         if bool(int(os.getenv("NVTE_BUILD_ACTIVATION_WITH_FAST_MATH", "0"))):
             cmake_flags.append("-DNVTE_BUILD_ACTIVATION_WITH_FAST_MATH=ON")
 
-        cmake_flags.append("-DUSE_ROCM=OFF")
```
ipanfilo, line 110: I think USE_ROCM is ON by default so this line is necessary

ipanfilo (changes requested):

tests/cpp/operator/test_normalization.h
```diff
@@ -0,0 +1,182 @@
+/*************************************************************************
```
ipanfilo, line 1: AMD copyright

setup.py
alextmagro: OK, I have readded it into the else block for if rocm_build:

examples/jax/encoder/common.py
alextmagro: Added -- I assumed gpu_arch >= 95 was the correct logic for MI350+.

build_tools/utils.py
alextmagro: I have removed the variable assignment.

build_tools/utils.py
alextmagro: I have removed the variable assignment.

build_tools/pytorch.py
alextmagro: One of the biggest changes in V2.4 was a refactor to improve build time by moving all CUDA kernels from pytorch/csrc and jax/csrc to common/. This means that we are building jax and pytorch as CPU only cpp extensions, rather than cuda extensions. We no longer have a target, and thus don't need an architecture, nvcc_flags, or hipcc anymore.

build_tools/utils.py
alextmagro: I have moved the check up.

build_tools/build_ext.py
alextmagro: Done.

tests/cpp/operator/test_normalization.cu
```diff
@@ -391,6 +261,7 @@ INSTANTIATE_TEST_SUITE_P(
     ::testing::Values(DType::kFloat32, DType::kBFloat16, DType::kFloat16),
     ::testing::Values(DType::kFloat32, DType::kBFloat16, DType::kFloat16, DType::kFloat8E4M3),
     ::testing::ValuesIn(test_cases),
+    ::testing::Values(false, true),
```
ipanfilo, line 266: On ROCm platforms cudnn_zero_centered_gamm_in_weight_dtype should always be false

tests/cpp/operator/test_normalization_mxfp8.cu
```diff
@@ -342,6 +287,7 @@ INSTANTIATE_TEST_SUITE_P(
     ::testing::Values(DType::kFloat8E5M2, DType::kFloat8E4M3),
     ::testing::ValuesIn(test_cases),
     ::testing::Values(true, false),
+    ::testing::Values(true, false),
```
ipanfilo, line 299: On ROCm platform cudnn_zero_centered_gamm_in_weight_dtype should always be false

tests/cpp/operator/test_normalization_mxfp8.cu
```diff
@@ -261,11 +201,15 @@ void performTest(const size_t N, const size_t H, const bool zero_centered_gamma,
...
+                     ref_output.get(),
                      ref_mu_ptr,
                      ref_rsigma_ptr,
                      N, H,
-                     ref_output.get(),
-                     zero_centered_gamma);
+                     nullptr, // amax
+                     1.f, // scale
+                     zero_centered_gamma,
+                     true, // CuDNN is the only MXFP8 backend currently
```
ipanfilo, line 220: No CuDNN support on ROCm

tests/cpp/operator/test_normalization_mxfp8.cu
```diff
@@ -197,19 +127,25 @@ void performTest(const size_t N, const size_t H, const bool zero_centered_gamma,
...
+  Tensor mu("mu", std::vector<size_t>{ N }, DType::kFloat32);
+  Tensor rsigma("rsigma", std::vector<size_t>{ N }, DType::kFloat32);
   Tensor workspace;
 
 
   fillUniform(&input);
   fillUniform(&gamma);
   fillUniform(&beta);
 
+  if (zero_centered_gamma_in_weight_dtype) {
```
ipanfilo, line 148: Should it be guarded?

tests/jax/test_custom_call_compute.py
```diff
@@ -40,389 +53,852 @@
     (2048, 2048, 1024),
     (2048, 1024, 1024),
 ]
-
-FP8_COMPUTE_TYPE = [jnp_float8_e4m3_type, jnp_float8_e5m2_type]
-    
-LN_CASES = [(512, 1024)]
+FP8_COMPUTE_TYPE = [jnp.float8_e4m3fn, jnp.float8_e5m2]
```
ipanfilo: Breaks gfx942 functionality - FNUZ FP8 data types are used there.  jnp_float8_... values should be used instead of jnp.float_.... datatypes

tests/jax/test_distributed_layernorm.py
```diff
@@ -70,12 +100,21 @@ def test_layernorm(
         dtype,
         zero_centered_gamma,
         shard_weights,
+        fp8_recipe,
+        use_shardy,
     ):
+        jax.config.update("jax_use_shardy_partitioner", use_shardy)
         epsilon = 1e-6
         ln_type = "layernorm"
+        q_dtype = jnp.float8_e4m3fn
```
ipanfilo: Float8 types on ROCm are detected runtime with get_jnp_float8_e4m3_type()

tests/jax/utils.py
```diff
@@ -15,18 +15,18 @@
...
 from jax import nn as jax_nn
 from jax import random as jax_random
+import pytest
 
 from transformer_engine.jax.attention import (
     AttnMaskType,
     canonicalize_attn_mask_type,
     make_swa_mask,
 )
-from transformer_engine.jax.fp8 import get_jnp_float8_e4m3_type, get_jnp_float8_e5m2_type, DType as TEDType
```
ipanfilo, line 29: It is needed for gfx942

tests/pytorch/test_float8blockwisetensor.py
```diff
@@ -0,0 +1,570 @@
...
+
+# Types that can be interpreted as tensor dims
+DimsType = Union[Iterable[int], int]
+
+# TODO replace with call to fp8.py when recipe added.
+recipe_available = not IS_HIP_EXTENSION and (get_device_compute_capability() >= (9, 0) and float(torch.version.cuda) >= 12.8)
+reason_for_no_recipe = "Quantize kernels require TMA and are only relevant with GEMMS."
+
+
+@pytest.mark.skipif(not recipe_available, reason=reason_for_no_recipe)
```
ipanfilo, line 53: It is probably more clear to add skipif(IS_HIP_EXTENSION) rather that use recepie_available for it

tests/pytorch/test_multi_tensor.py
```diff
@@ -216,3 +219,43 @@ def test_multi_tensor_unscale_l2norm(input_size_pair, applier, repeat, in_type,
     if per_tensor:
         torch.testing.assert_close(norm_per_tensor, normab.broadcast_to(norm_per_tensor.shape))
     assert overflow_buf.item() == 0
+
+
+@pytest.mark.parametrize("input_size_pair", input_size_pairs + [(1, 1)])
+@pytest.mark.parametrize("applier", appliers)
+@pytest.mark.parametrize("repeat", [1, 55])
+@pytest.mark.parametrize("fp8_dtype", [torch.float8_e4m3fn, torch.float8_e5m2])
```
ipanfilo: Dynamic float8 type detection is required on ROCm

tests/pytorch/test_numerics.py
```diff
@@ -2327,6 +2543,12 @@ def test_kv_cache_accuracy(dtype, bs, model_key, use_RoPE, input_format, module,
         pytest.skip("FusedAttention and FlashAttention do not support FP32")
     if use_RoPE:
         pytest.skip("KV cache does not support starting positions for RoPE")
+    if (
```
ipanfilo, line 2564: Guard for ROCm

transformer_engine/common/CMakeLists.txt
```diff
@@ -116,53 +116,69 @@ include_directories(${PROJECT_SOURCE_DIR}/..)
 set(transformer_engine_SOURCES)
 if(USE_CUDA)
   list(APPEND transformer_engine_SOURCES
-       cudnn_utils.cpp
```
ipanfilo, line 119: It looks like original identation was correct

ipanfilo (changes requested):

transformer_engine/common/common.h
```diff
@@ -263,11 +298,41 @@ TRANSFORMER_ENGINE_TYPE_NAME(__nv_fp8_e8m0)
 #endif // #ifdef __HIP_PLATFORM_AMD__
 #undef TRANSFORMER_ENGINE_TYPE_NAME
 
+template <typename T>
+struct TypeExtrema;
+
+template <>
+struct TypeExtrema<fp8e4m3> {
+  static constexpr float max = 448.0f;
```
ipanfilo, line 307: This one requires dynamic delayed initialization for gfx942/gfx950

transformer_engine/common/fused_attn/utils.h
ipanfilo, line 1: AMD copyright

transformer_engine/common/fused_attn_rocm/utils.cpp
```diff
@@ -221,5 +221,71 @@ void generateMatrixStrides(
     }
 }
 
+__global__ void populate_rng_state_kernel(int64_t *rng_state_dst, const int64_t *const seed,
```
ipanfilo, line 224: Do fused_attn/utils.cu and fused_attn_rocm/utils.cpp have a lot similarities. Does it make sense to merge them?

transformer_engine/common/fused_attn/utils.h
```diff
@@ -150,6 +150,40 @@ DType get_ragged_offset_dtype(NVTE_QKV_Layout_Group layout_group, int64_t num_at
...
+  FusedAttnOffsetManager() {}
+  size_t offset_ = 0;
+};
+
+__global__ void populate_rng_state_kernel(int64_t *rng_state_dst, const int64_t *const seed,
+                                          int64_t offset);
+
+__global__ void get_runtime_num_segments_kernel(int32_t *cu_seqlen, size_t len, uint32_t *out);
+
+#ifndef USE_ROCM
```
ipanfilo: Is this file used on ROCm platform? Shouldn't fused_attn_rocm/utils.h be used?

transformer_engine/common/fused_attn_rocm/utils.h
ipanfilo, line 1: Copyright

transformer_engine/common/normalization/common.cpp
```diff
@@ -551,6 +556,13 @@ bool use_cudnn_norm_fwd() { return _cudnn_norm_fwd_flag(); }
 bool use_cudnn_norm_bwd() { return _cudnn_norm_bwd_flag(); }
 #endif
 
+bool& _zero_centered_gamma_in_weight_dtype() {
+  static bool flag = transformer_engine::getenv<bool>("NVTE_ZERO_CENTERED_GAMMA_IN_WTYPE");
+  return flag;
+}
+
+bool& use_zero_centered_gamma_in_weight_dtype() { return _zero_centered_gamma_in_weight_dtype(); }
```
ipanfilo, line 564: Its declaration at line 44 is guarded with !__HIP_PLATFORM_AMD__

transformer_engine/common/recipe/fp8_block_scaling.cu
```diff
@@ -0,0 +1,257 @@
+/*************************************************************************
```
ipanfilo, line 1: AMD copyright

transformer_engine/common/transformer_engine.cpp
```diff
@@ -503,3 +483,13 @@ void nvte_destroy_quantization_config(NVTEQuantizationConfig config) {
     delete reinterpret_cast<transformer_engine::QuantizationConfig *>(config);
   }
 }
+
+int nvte_is_non_tn_fp8_gemm_supported() {
```
ipanfilo, line 487: Is it called from outside the library? If not it can be guarded for ROCm

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
```diff
@@ -0,0 +1,563 @@
+/*************************************************************************
```
ipanfilo, line 1: AMD copyright

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
```diff
@@ -0,0 +1,495 @@
+/*************************************************************************
```
ipanfilo, line 1: AMD copyright

transformer_engine/jax/attention.py
```diff
@@ -287,8 +287,7 @@ def is_fused_attn_kernel_available(
     kv_num_heads,
     q_max_seqlen,
     kv_max_seqlen,
-    head_dim_qk,
```
ipanfilo, line 290: It is ROCm feature - MLA support

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -114,8 +113,7 @@ class FusedAttnHelper:
     kv_num_heads: int
     q_max_seqlen: int
     kv_max_seqlen: int
-    head_dim_qk: int
```
ipanfilo, line 117: ROCm feature

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -0,0 +1,467 @@
...
+)
+
+
+__all__ = ["gemm"]
+
+
+num_cublas_streams = 4
+
+
+def get_cublas_workspace_size_bytes() -> None:
```
ipanfilo, line 32: Refer similar TE pytorch function for ROCm values

transformer_engine/jax/cpp_extensions/misc.py
```diff
@@ -38,8 +40,8 @@ def te_dtype_to_jax_dtype(te_dtype):
         TEDType.kBFloat16: jnp.bfloat16,
         TEDType.kInt32: jnp.int32,
         TEDType.kInt64: jnp.int64,
-        TEDType.kFloat8E4M3: get_jnp_float8_e4m3_type(),
```
ipanfilo, line 41: Broken gfx942 functionality

transformer_engine/jax/csrc/extensions.h
```diff
@@ -271,40 +93,41 @@ XLA_FFI_DECLARE_HANDLER_SYMBOL(ScaledUpperTriangMaskedSoftmaxForwardHandler);
...
     NVTE_Mask_Type mask_type, NVTE_QKV_Layout qkv_layout, DType dtype, bool is_training,
     bool deterministic, size_t max_segments_per_seq, int64_t window_size_left,
     int64_t window_size_right);
 
-void FusedAttnBackward(cudaStream_t stream, void **buffers, const char *opaque, size_t opaque_len);
+// Grouped GEMM
+XLA_FFI_DECLARE_HANDLER_SYMBOL(GroupedGemmHandler);
 
-XLA_FFI_DECLARE_HANDLER_SYMBOL(FusedAttnBackwardHandler);
+// Cudnn helpers
```
ipanfilo, line 127: Should be guarded

transformer_engine/common/CMakeLists.txt
alextmagro: Reverted -- looks the extra space got removed somehow.

tests/pytorch/test_numerics.py
alextmagro: Added

tests/pytorch/test_float8blockwisetensor.py
alextmagro: I added it to the recipe specifically because the TODO mentions that it should be replaced with the call to fp8.py when recipe is added, which it already has been. Since we guard for HIP in fp8.py, I thought when this TODO was resolved the test would still be skipped gracefully. Then, when we implement fp8 block scaling we only need to remove the one guard.

tests/cpp/operator/test_normalization_mxfp8.cu
alextmagro: OK, I will skip tests that have cudnn_zero_centered_gamm_in_weight_dtype = true. That should take care of guarding within the test as well.

tests/cpp/operator/test_normalization_mxfp8.cu
alextmagro: The use_cudnn variable is only used to determine whether zero_centered_gamma_in_weight_dtype is used or not in the reference computation, in this line:
`const bool zero_centered_gamma_in_weight_dtype = use_cudnn && cudnn_zero_centered_gamma_in_weight_dtype;`
since we are always setting zero_centered_gamma_in_weight_dtype to false, I am wondering if it is ok to leave this as is.

tests/cpp/operator/test_normalization.cu
alextmagro: I have changed our current guard for use_cudnn to skip cudnn_zero_centered_gamm_in_weight_dtype = true tests.

tests/cpp/operator/test_normalization.h
alextmagro: I have added missing copyrights throughout.

transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -17,12 +17,12 @@ NVTE_Fused_Attn_Backend GetFusedAttnBackend(DType q_dtype, DType kv_dtype,
                                             NVTE_Mask_Type mask_type, float dropout_probability,
                                             size_t q_attn_heads, size_t kv_attn_heads,
                                             size_t q_max_seqlen, size_t kv_max_seqlen,
-                                            size_t qk_head_dim, size_t v_head_dim,
```
ipanfilo, line 20: ROCm feature

transformer_engine/jax/csrc/extensions/cublas.cpp
```diff
@@ -0,0 +1,24 @@
+/*************************************************************************
```
ipanfilo, line 1: AMD copyright

transformer_engine/jax/csrc/extensions/utils.h
```diff
@@ -26,23 +23,10 @@ namespace jax {
...
-                           cudaStream_t stream);
-#else
-void PopulateRngStateAsync(void *rng_state_dst, 
-                           const void *const seed,
-                           size_t batch_size, 
-                           size_t num_heads, 
-                           size_t q_max_seqlen, 
-                           size_t kv_max_seqlen,
-                           cudaStream_t stream);
+#ifdef USE_ROCM
```
ipanfilo: Why is it for ROCm only?

build_tools/build_ext.py
alextmagro: I have made this fix, although I am not sure this bug fix is necessary anymore for multiframework since the extensions no longer have CUDA files.

tests/jax/test_custom_call_compute.py
alextmagro: I have reverted the changes for this throughout the codebase.

transformer_engine/common/common.h
alextmagro: I'm not sure what you mean by this, could you elaborate? fp8e4m3 is defined as 
`using fp8e4m3 = te_hip_fp8_e4m3`
where te_hip_fp8_e4m3 is defined in hip_float8.h and dynamically set based on the HIP version used.

transformer_engine/common/fused_attn_rocm/utils.cpp
alextmagro: I think there is a lot of overlap between fused_attn_rocm/ and fused_attn/ now. I am wondering if merging these folders together would be a good idea for a different ticket.

transformer_engine/common/fused_attn/utils.h
alextmagro: It isn't, this is an artifact from me trying (unsuccessfully) to directly use some of the code in fused_attn/. I have removed it.

transformer_engine/common/normalization/common.cpp
alextmagro: I have removed nvte_enable_zero_centered_gamma_in_weight_dtype from the HIP build

transformer_engine/common/transformer_engine.cpp
alextmagro: This is called by pytorch/crsc/quantizer.

transformer_engine/jax/attention.py
alextmagro: This was an odd change from upstream -- head_dim_qk and head_dim_v are used in release 2.2, and 2.5, but 2.3 and 2.4 limit it to just head_dim due to a bug in cudnn. I was unsure if I should follow upstream or stick with what should work for AMD. Can you confirm i should revert these changes?

transformer_engine/jax/cpp_extensions/gemm.py
alextmagro: Done.

transformer_engine/jax/csrc/extensions.h
alextmagro: Done

transformer_engine/jax/csrc/extensions/utils.h
alextmagro: GetCudnnRuntimeVersion() usage has been replaced with cudnnGetVersion() which don't have access to.

transformer_engine/jax/quantize/helper.py
```diff
@@ -0,0 +1,491 @@
+# Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ipanfilo, line 3: AMD copyright

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -0,0 +1,1631 @@
+# Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ipanfilo, line 3: AMD copyright

transformer_engine/pytorch/csrc/extensions.h
```diff
@@ -382,14 +381,20 @@ void fused_multi_row_padding(at::Tensor input, at::Tensor output,
                              std::vector<size_t> input_row_list,
                              std::vector<size_t> padded_input_row_list);
 /***************************************************************************************************
- * swizzle
+ * NVSHMEM APIs
```
ipanfilo, line 386: Is it available on ROCm?

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -136,139 +157,191 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
...
+  m.def("thd_grad_correction", &transformer_engine::pytorch::thd_grad_correction,
         "Correct the THD format gradients of context parallelism in backward pass",
         py::call_guard<py::gil_scoped_release>());
-  m.def("thd_get_partitioned_indices", &thd_get_partitioned_indices,
+  m.def("thd_get_partitioned_indices", &transformer_engine::pytorch::thd_get_partitioned_indices,
         "Generate partitioned indices for inputs in THD format",
         py::call_guard<py::gil_scoped_release>());
 
+  // nvshmem functions
+  m.def("init_nvshmem_backend", &transformer_engine::pytorch::init_nvshmem_backend,
```
ipanfilo, line 289: Is it available on ROCm platform?

transformer_engine/jax/quantize/helper.py
alextmagro: I think I added copyrights to this and backends.py with a previous commit.

transformer_engine/pytorch/csrc/extensions/pybind.cpp
alextmagro: We don't have shmem yet, I will guard for this.

build_tools/jax.py
```diff
@@ -11,11 +11,7 @@
 
 import setuptools
 
-<<<<<<< HEAD
-from .utils import rocm_build, rocm_path, hipify, cuda_path, all_files_in_dir
-=======
-from .utils import get_cuda_include_dirs, all_files_in_dir, debug_build_enabled
->>>>>>> 42b51c40c4e39adce9640cf98f8a3f5869f5f270
+from .utils import rocm_build, rocm_path, hipify, all_files_in_dir, get_cuda_include_dirs, debug_build_enabled
```
wangye805: Why not keep both lines to minimize our code diff with NV upstream

build_tools/pytorch.py
```diff
@@ -10,17 +10,15 @@
...
 
-<<<<<<< HEAD
 from .utils import (
     rocm_build,
     hipify,
     all_files_in_dir,
     cuda_archs,
     cuda_version,
+    get_cuda_include_dirs,
+    debug_build_enabled,
```
wangye805, line 23: Same suggestion here: keep one import for rocm and one for NV to minimize the code-diff

build_tools/pytorch.py
```diff
@@ -58,70 +47,9 @@ def setup_pytorch_extension(
     if rocm_build() and not bool(int(os.getenv("NVTE_RELEASE_BUILD", "0"))):
         current_file_path = Path(__file__).parent.resolve()
         base_dir = current_file_path.parent
-        sources = hipify(base_dir, csrc_source_files, sources, include_dirs)
+        sources = hipify(base_dir, csrc_source_files, sources, include_dirs[1:])
```
wangye805, line 57: Why do we skip the index-0 for include_dirs?

tests/jax/test_custom_call_compute.py
```diff
@@ -203,21 +177,6 @@ def test_act_grad(self, shape, activation_type):
         assert_allclose(prim_grad, ref_grad, dtype=x.dtype)
 
     @pytest.mark.skipif(not is_fp8_supported, reason=reason)
-<<<<<<< HEAD
-    def test_qdq(self):
```
wangye805, line 69: Was this test_qdq removed entirely or moved to other places

tests/jax/test_distributed_fused_attn.py
```diff
@@ -267,7 +228,6 @@ def test_cross_attn(
             seqlen,
             seqlen,
             hidden,
-            hidden,
```
wangye805, line 183: This is the MLA feature we cherry-picked from v2.6. Basically we need both head_dim_qk and head_dim_v

tests/jax/test_distributed_fused_attn.py
```diff
@@ -280,7 +240,6 @@ def test_cross_attn(
             num_head,
             num_head,
             hidden,
-            hidden,
```
wangye805, line 196: Same issue for MLA

tests/jax/test_distributed_fused_attn.py
```diff
@@ -335,6 +294,8 @@ def impl_test_context_parallel_attn(
         use_scan_ring=False,
     ):
         if qkv_layout.is_thd():
+            if is_hip_extension():
+                pytest.skip("THD + ring on Rocm doesn't support context parallelism.")
```
wangye805, line 306: Why do we not support CP with thd + ring?

tests/jax/test_distributed_fused_attn.py
```diff
@@ -370,7 +331,6 @@ def impl_test_context_parallel_attn(
             num_head,
             num_kv_heads,
             hidden,
-            hidden,
```
wangye805, line 286: Same MLA issue

tests/jax/test_distributed_fused_attn.py
```diff
@@ -401,7 +361,6 @@ def check_has_backend_for_mask(mask_type):
                 seqlen,
                 seqlen,
                 hidden,
-                hidden,
```
wangye805, line 318: Need this for MLA

tests/jax/test_fused_attn.py
```diff
@@ -297,8 +298,7 @@ class FusedAttnRunner:
     max_seqlen_kv: int
     num_heads_q: int
     num_heads_kv: int
-    head_dim_qk: int
-    head_dim_v: int
```
wangye805, line 301: Need to bring back head_dim_qk and head_dim_v

tests/pytorch/test_numerics.py
```diff
@@ -1309,7 +1301,6 @@ def test_fp8_linear_without_transpose_cache_accuracy(dtype, bs, model, fp8_model
             bias=True,
             params_dtype=dtype,
             device="cuda",
-            keep_fp8_weight_transpose_cache=False
```
wangye805, line 1307: keep_fp8_weight_transpose_cache is a unique feature from us

tests/pytorch/test_numerics.py
```diff
@@ -1718,7 +1712,6 @@ def test_fp8_layernorm_mlp_without_transpose_cache_accuracy(dtype, bs, model, ac
         normalization=normalization,
         params_dtype=dtype,
         device="cuda",
-        keep_fp8_weight_transpose_cache = False,
```
wangye805, line 1608: Same here

transformer_engine/common/__init__.py
```diff
@@ -158,32 +160,36 @@ def load_framework_extension(framework: str):
     if framework == "torch":
         extra_dep_name = "pytorch"
 
+    te_cuda_vers = "rocm" if te_rocm_build else "cu12"
```
wangye805, line 163: Does this cu12 used in some version checking later in kernels/pytorch extension/jax extension?

transformer_engine/common/include/transformer_engine/fused_attn.h
```diff
@@ -646,6 +647,23 @@ void nvte_fused_attn_bwd(const NVTETensor Q, const NVTETensor K, const NVTETenso
...
+ *  \param[in]     batch_size                Batch size.
+ *  \param[in]     num_heads                 # of attention heads.
+ *  \param[in]     q_max_seqlen              Max sequence length used for computing for Q.
+ *                                           it may be >= max(seqlen_q_i) for i=0,...batch_size-1.
+ *  \param[in]     kv_max_seqlen             Max sequence length used for computing for K and V.
+ *                                           it may be >= max(seqlen_kv_i) for i=0,...batch_size-1.
+ *  \param[in]     stream                    CUDA stream used for this operation.
+ */
+void nvte_populate_rng_state_async(void *rng_state_dst, const void *const seed,
+                                   size_t batch_size, size_t num_heads, size_t q_max_seqlen,
```
wangye805, line 664: You can minimize the code change by just put ifndef on batzh_size and num_heads

transformer_engine/jax/attention.py
wangye805: I think you need to keep the MLA feature

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -136,12 +136,8 @@ def forward(
         fsdp_group: Union[dist_group_type, None],
         module: torch.nn.Module,
         skip_fp8_weight_update: bool,
-<<<<<<< HEAD
-        keep_fp8_weight_transpose_cache: bool,
```
wangye805, line 127: ROCm feature

wangye805 (changes requested):

build_tools/pytorch.py
ipanfilo: I think index[0] is <hip_root>/include

examples/jax/encoder/common.py
```diff
@@ -6,18 +6,21 @@
...
     from transformer_engine.jax.util import is_mi200
 
+
 @lru_cache
 def is_bf16_supported():
     """Return if BF16 has hardware supported"""
     gpu_arch = get_device_compute_capability(0)
     if is_hip_extension():
         # only GFX9.4 and MI200 machines support bf16
-        return gpu_arch == 94 or is_mi200()
```
ipanfilo, line 20: Update comment to GFX9.4/9.5 and check explicitly for >=94 and <100 because some Radeon arch are >=100

examples/jax/encoder/common.py
ipanfilo: Not >=95 but ==95 or >=95 and < 100 - some Radeon have arch>=100

tests/cpp/operator/test_normalization_mxfp8.cu
ipanfilo: Better to ifdef parameters here and assert in test that cudnn_zero_centered_gamma_in_weight_dtype is false. Current approach generates unnecessary test cases that are always skipped

build_tools/utils.py
ipanfilo: Where?

tests/pytorch/test_multi_tensor.py
alextmagro: Done.

build_tools/pytorch.py
alextmagro: include_dirs[0] is /opt/rocm/include -- we don't want to hipify these files.

tests/pytorch/test_numerics.py
alextmagro: Yes, I readded it to the code but missed these tests. Will readd it and double check that I didn't miss anywhere else in the test directory. Thanks.

transformer_engine/common/__init__.py
alextmagro: cu12 is used as the extension to the package names such as transformer_engine_{te_cuda_vers}, and is checked in the pytorch/common __init__.py files. It doesn't look like JAX checks for it.

build_tools/jax.py
alextmagro: Good idea, will make that change.

tests/jax/test_custom_call_compute.py
alextmagro: This was moved to line 579 in the same file

transformer_engine/common/include/transformer_engine/fused_attn.h
alextmagro: NVTE_Fused_Attn_Backend backend is not used for ROCm. If I remember correctly, NVTETensor points to a void*, but the code change would get messy with multiple ifndefs

tests/pytorch/test_numerics.py
alextmagro: Can you double check you are on the most recent commit? I have readded keep_fp8_weight_transpose_cache

tests/jax/test_distributed_fused_attn.py
alextmagro: Sorry, this skip condition is incorrect and I missed fixing it. What we don't support is shardy sharding, as that is a JAX 0.5 > feature. I have moved this skip to the shardy sharding tests.

tests/cpp/operator/test_normalization.cu
ipanfilo: Better to ifdef parameters here and assert in test that cudnn_zero_centered_gamma_in_weight_dtype is false. Current approach generates unnecessary test cases that are always skipped

tests/cpp/operator/test_normalization_mxfp8.cu
ipanfilo: It might be left as is but let's put comments before the call

tests/pytorch/test_multi_tensor.py
```diff
@@ -216,3 +219,43 @@ def test_multi_tensor_unscale_l2norm(input_size_pair, applier, repeat, in_type,
...
+@pytest.mark.parametrize("epsilon", [0.0, 100.0])
+def test_multi_tensor_compute_scale_and_scale_inv(
+    input_size_pair, applier, repeat, fp8_dtype, pow_2_scales, epsilon
+):
+    sizea, sizeb = input_size_pair
+    device = torch.device("cuda")
+    overflow_buf = torch.zeros(1, dtype=torch.int32, device=device)
+    a = torch.randn([sizea], dtype=torch.float32, device=device).abs()
+    b = torch.randn([sizeb], dtype=torch.float32, device=device).abs()
+    max_fp8 = torch.finfo(fp8_dtype).max
```
ipanfilo, line 239: Does torch.finfo support tex.DType?

tests/pytorch/test_multi_tensor.py
```diff
@@ -216,3 +219,43 @@ def test_multi_tensor_unscale_l2norm(input_size_pair, applier, repeat, in_type,
...
+        overflow_buf,
+        [amax_list, scale_list, scale_inv_list],
+        max_fp8,
+        pow_2_scales,
+        epsilon,
+    )
+
+    for amax, scale, scale_inv in zip(amax_list, scale_list, scale_inv_list):
+        scale_ref, scale_inv_ref, _ = scale_from_amax_tensor(
+            torch.float32, amax, fp8_dtype, eps=epsilon, pow_2_scales=pow_2_scales
```
ipanfilo, line 259: does it support test.DType?

transformer_engine/common/common.h
ipanfilo: FNUZ and OCP formats of FP8E4M3 have different max values - 240 and 448 correspondingly. You can refer Quantized_Limits and Numeric_Traits in transformer_engine/common/utils.cuh

transformer_engine/jax/csrc/extensions/utils.h
ipanfilo: Make it inline constexpr then and return whatever was GetCudnnRuntimeVersion() value. This way you don't need modify utils.cpp and allow some compiler optimization when cudnnGetVersion() is used directly

build_tools/utils.py
alextmagro: Sorry, I meant up a function level. The logic is now in setup_jax_extension and setup_pytorch_extension

examples/jax/encoder/common.py
alextmagro: Ah, didn't know that. I have added the upper bound of 100.

tests/pytorch/test_multi_tensor.py
alextmagro: No, this is what was causing the errors in test_multi_tensor. I have checked in a fix.

transformer_engine/jax/csrc/extensions/utils.h
alextmagro: Done. Added a todo to make consteval if we move to c++20.

transformer_engine/common/common.h
```diff
@@ -303,7 +303,17 @@ struct TypeExtrema;
...
 template <>
 struct TypeExtrema<fp8e4m3> {
+#ifndef __HIP_PLATFORM_AMD__
   static constexpr float max = 448.0f;
+#elif defined(__HIP_DEVICE_COMPILE__)
+  static constexpr float maxNorm = te_fp8_fnuz() ? 240.0f : 448.0f;
+#elif defined(TE_DYNAMIC_HIP_FP8_TYPE)
+ // dummy declaration for correct translation;
+ // it is not defined anywhere and it's usage should be eliminated before linking
+  static float maxNorm;
```
ipanfilo: is maxNorm used by GPU kernels only?

transformer_engine/jax/csrc/extensions/utils.h
```diff
@@ -21,11 +21,16 @@ namespace transformer_engine {
...
 int GetCudaRuntimeVersion();
+#ifndef USE_ROCM
 size_t GetCudnnRuntimeVersion();
-int GetDeviceComputeCapability(int gpu_id);
-#ifdef USE_ROCM
-size_t cudnnGetVersion();
+#else
+// use a very high number for cudnn in rocm to enable all filtering
+//(99, 0, 0)
+inline size_t GetCudnnRuntimeVersion() { return 990000; }
```
ipanfilo, line 29: Can it be constexpr too since moved ot header?

alextmagro: I was unsure on this. Since we have a pybind around GetCudnnRuntimeVersion() I wasn't confident it was safe to do so.

transformer_engine/common/common.h
alextmagro: Yes, does that affect how we should dynamically define the variable?

ipanfilo: If it is not used on host, it's OK - no dynamic calculation is needed. In this case, however, no need to check for TE_DYAMIC_HIP_FP8_TYPE but make dummy declarartion just under '#else'

transformer_engine/common/common.h
```diff
@@ -307,10 +307,6 @@ struct TypeExtrema<fp8e4m3> {
   static constexpr float max = 448.0f;
 #elif defined(__HIP_DEVICE_COMPILE__)
   static constexpr float maxNorm = te_fp8_fnuz() ? 240.0f : 448.0f;
-#elif defined(TE_DYNAMIC_HIP_FP8_TYPE)
- // dummy declaration for correct translation;
- // it is not defined anywhere and it's usage should be eliminated before linking
-  static float maxNorm;
 #else
   static constexpr float maxNorm = 240.0f;
```
ipanfilo: It should be dummy value here, so if at some moment maxNorm is accessed in host path, linking will fail

tests/cpp/operator/test_normalization_mxfp8.cu
```diff
@@ -216,7 +216,7 @@ void performTest(const size_t N, const size_t H, const bool zero_centered_gamma,
                      nullptr, // amax
                      1.f, // scale
                      zero_centered_gamma,
-                     true, // CuDNN is the only MXFP8 backend currently
+                     true, // CuDNN is the only CUDA MXFP8 backend currently. ROCm supports MXFP8.
```
ipanfilo: It does not resolve confusion of using CuDNN on ROCm.
Please keep this comment in upstream variant and add one before the call with what you explained in this PR that CuDNN flag does not have effect on ROCm so we don't worry that it is true.

ipanfilo (commented):
Please make sure you also restored MLA and keep_fp8_transpose_cache code, not only function interfaces

tests/cpp/operator/test_normalization_mxfp8.cu
alextmagro: Understood, I have updated the comment.

transformer_engine/common/common.h
alextmagro: I have changed to be a dummy variable here, but I am curious why linking would fail if used hostside. Wouldn't maxNorm be zero-initialized if accessed since it is static?

tests/jax/test_distributed_fused_attn.py
alextmagro: THD + ring is failing after all. I am not sure why we don't support it, but all tests fail with the parameters:

pytest.param(
        QKVLayout.THD_THD_THD, AttnMaskType.PADDING_CAUSAL_MASK, id="THD_SEPARATE-PADDING_CAUSAL")

I have readded this skip.

transformer_engine/pytorch/module/layernorm_mlp.py
```diff
@@ -188,180 +204,176 @@ def forward(
         fsdp_group: Union[dist_group_type, None],
         module: torch.nn.Module,
         skip_fp8_weight_update: bool,
-        keep_fp8_weight_transpose_cache: bool,
+        symmetric_ar_type: str,
+        debug: Optional[bool] = False,
+        keep_fp8_weight_transpose_cache: Optional[bool] = True,
```
ipanfilo: Why is it changed?

transformer_engine/pytorch/module/linear.py
```diff
@@ -106,7 +118,9 @@ def forward(
         fsdp_group: Union[dist_group_type, None],
         module: torch.nn.Module,
         skip_fp8_weight_update: bool,
-        keep_fp8_weight_transpose_cache: bool,
+        symmetric_ar_type: str,
+        debug: Optional[bool] = False,
+        keep_fp8_weight_transpose_cache: Optional[bool] = True,
```
ipanfilo: Why is it changed

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -124,7 +136,9 @@ def forward(
         fsdp_group: Union[dist_group_type, None],
         module: torch.nn.Module,
         skip_fp8_weight_update: bool,
-        keep_fp8_weight_transpose_cache: bool,
+        symmetric_ar_type: str,
+        debug: Optional[bool] = False,
+        keep_fp8_weight_transpose_cache: bool = True,
```
ipanfilo, line 141: Why default is needed?

alextmagro: Default value is needed as it comes after another parameter with a default value now. As all calling functions have a default of True for keep_fp8_weight_transpose_cache behavior is unchanged. 

I misunderstood that in python we needed Optional variables to follow other Optional variables as an extension, so I have removed the typing.

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -59,92 +59,170 @@ inline void CreateCublasHandle(cublasLtHandle_t *handle) {
   NVTE_CHECK_CUBLAS(cublasLtCreate(handle));
 }
 
+/* Parameters for cuBLAS GEMM
+ *
+ * cuBLAS follows the BLAS convention of column-major ordering. This
+ * is different than the row-major that is typically used in
+ * Transformer Engine.
+ *
+ */
```
wangye805, line 68: These changes should be incorporated into transformer_engine/common/rocm_gemm.cu since there are corresponding GemmParam and CanonicalizeGemmInput as well

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -167,18 +244,33 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
...
+  const int ldd = m;
+
   // Return immediately if GEMM is trivial
   if (m <= 0 || n <= 0) {
     return;
   }
   NVTE_CHECK(k > 0);
 
-  const GemmParam &param = CanonicalizeGemmInput(*inputA, transa, *inputB, transb, k, lda, ldb);
+  const GemmParam param = CanonicalizeGemmInput(*inputA, transa, *inputB, transb, m, n, k);
```
wangye805, line 272: Changes inside this function should also be included into hipblaslt_gemm function inside transformer_engine/common/gemm/rocm_gemm.cu.

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -626,3 +721,13 @@ void nvte_multi_stream_cublas_gemm(const NVTETensor *A, const NVTETensor *B, NVT
     NVTE_CHECK_CUDA(cudaStreamWaitEvent(stream, cublas_event[s]));
   }
 }
+
+#ifndef __HIP_PLATFORM_AMD__
```
wangye805, line 725: Why do we disable cublaslt handle manager directly? I thought it help NV side to resolve some memory leaking issue

transformer_engine/common/normalization/common.cpp
wangye805: Start next line with "const bool zero_centered_gamma, const bool is_tuned" to reduce code diff with NVTE upstream

wangye805, line 72: Why NV upstream do not need to pass into scaling mode and training bool variable?

alextmagro:
42b51c40c is the target upstream commit

transformer_engine/common/recipe/__init__.py
```diff
@@ -7,6 +7,7 @@
 """This module provides predefined FP8 recipes."""
 from __future__ import annotations
 import warnings
+import os
 from enum import Enum
 #from typing import Literal, Optional, Union, Callable, NamedTuple
```
wangye805: Remove this unnecessary comment

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
```diff
@@ -0,0 +1,565 @@
...
+                        reinterpret_cast<float*>(scale_inv_t.dptr), row_length, num_rows,
+                        scale_stride_x, scale_stride_y, scale_t_stride_x, scale_t_stride_y, epsilon,
+                        pow_2_scale);
+              }  // full-tile
+              )  // return_transpose
+          )      // OutputType
+      )          // InputType
+  NVTE_CHECK_CUDA(cudaGetLastError());
+}
+#endif // #ifndef __HIP_PLATFORM_AMD__
```
wangye805: If we are blocking the only function in this file, why not just not compiling this function in cmakelists.txt and leave this cpp file untouched?

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
```diff
@@ -0,0 +1,497 @@
...
+                  reinterpret_cast<const InputType*>(input.dptr),
+                  reinterpret_cast<OutputType*>(output.dptr),
+                  reinterpret_cast<OutputType*>(output_t.dptr),
+                  reinterpret_cast<float*>(scale_inv.dptr),
+                  reinterpret_cast<float*>(scale_inv_t.dptr), row_length, num_rows, scale_stride_x,
+                  scale_stride_y, scale_t_stride_x, scale_t_stride_y, epsilon, rowwise_option,
+                  columnwise_option, pow2_scale);)  // kAligned
+          )                                         // OutputType
+      )                                             // InputType
+  NVTE_CHECK_CUDA(cudaGetLastError());
```
wangye805, line 490: Same issue here: why not just leave this file untouched and do not compile it in cmakelists.txt?

wangye805 (changes requested):

wangye805:
@alextmagro , this file is already removed in NV upstream: https://github.com/ROCm/TransformerEngine/blob/d962f0f538bfca0aa5ab9abb4b0229e1742ea768/transformer_engine/pytorch/attention.py#L7. Do we still need to keep it?

wangye805:
Are we able to release this guard in v2.4? https://github.com/ROCm/TransformerEngine/blob/ad1d28f5ad4542498c967445130ceb135e6c7b19/transformer_engine/pytorch/cpp_extensions/__init__.py#L12

transformer_engine/pytorch/cross_entropy.py
```diff
@@ -24,7 +24,13 @@ class CrossEntropyFunction(torch.autograd.Function):
 
     @staticmethod
     def forward(
-        ctx, _input, target, label_smoothing=0.0, reduce_loss=False, dist_process_group=None
+        ctx,
```
wangye805, line 25: Now this file has the same content as NV upstream. We can remove the copyright to minimize our code diff with NV upstream

wangye805:
This file no longer exits in NV upstream in v2.4: https://github.com/ROCm/TransformerEngine/blob/ad1d28f5ad4542498c967445130ceb135e6c7b19/transformer_engine/pytorch/csrc/type_shim.h#L26. Do we still need to keep it?

wangye805:
This function is not used anymore in NV upstream: https://github.com/ROCm/TransformerEngine/blob/d962f0f538bfca0aa5ab9abb4b0229e1742ea768/transformer_engine/pytorch/__init__.py#L30. Do we still need it?

wangye805:
In transformer_engine/pytorch/setup.py, NV upstream has those *-cu12 setup_requires but we don't: https://github.com/NVIDIA/TransformerEngine/blob/3cd6870ceb809f173a31e3cac5034e56cf879156/transformer_engine/pytorch/setup.py#L58

Is this expected?

transformer_engine/pytorch/triton/permutation.py
wangye805: Those rocm-specific definitions are not used anymore. Shall we remove them?

transformer_engine/pytorch/cross_entropy.py
alextmagro: Done!

transformer_engine/common/recipe/__init__.py
alextmagro: Done.

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
alextmagro: We already skip compiling it, I had just left this as a reminder but have removed the #ifndef

alextmagro:
> Are we able to release this guard in v2.4?
> 
> https://github.com/ROCm/TransformerEngine/blob/ad1d28f5ad4542498c967445130ceb135e6c7b19/transformer_engine/pytorch/cpp_extensions/__init__.py#L12

Looks like we should be able to.

alextmagro:
> In transformer_engine/pytorch/setup.py, NV upstream has those *-cu12 setup_requires but we don't: https://github.com/NVIDIA/TransformerEngine/blob/3cd6870ceb809f173a31e3cac5034e56cf879156/transformer_engine/pytorch/setup.py#L58
> 
> Is this expected?

I would say so, as *-cu12 requires are specifically CUDA libraries that we don't use.

transformer_engine/common/gemm/cublaslt_gemm.cu
alextmagro: I have updated this and GemmParam.

transformer_engine/common/normalization/common.cpp
alextmagro: Done.

alextmagro: Those variables are used only for MXFP8 -- as NV doesn't have MXFP8 normalization with TE (they use cudnn) they don't need these variables.

transformer_engine/common/gemm/cublaslt_gemm.cu
alextmagro: This code calls createCublasHandle, which is guarded against. nvte_cublas_handle_init is called in the #ifndef USE_ROCM path of JAX as well.

setup.py
```diff
@@ -126,7 +129,15 @@ def setup_requirements() -> Tuple[List[str], List[str], List[str]]:
     """
 
     # Common requirements
-    setup_reqs: List[str] = []
+    setup_reqs: List[str] = [
```
ipanfilo: While it is rather ignored by modern setuptools, let's guard it anyway

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -190,89 +190,155 @@ static hipDataType get_hipblaslt_dtype(const transformer_engine::DType t) {
...
 
 GemmParam CanonicalizeGemmInput(const transformer_engine::Tensor &A, const cublasOperation_t transA,
                                 const transformer_engine::Tensor &B, const cublasOperation_t transB,
-                                const int k, const int lda, const int ldb) {
+                                const int m, const int n, const int k) {
   using namespace transformer_engine;
-  NVTE_CHECK(A.scaling_mode == B.scaling_mode,
-             "Inputs A and B to GEMM need to have the same scaling mode!");
+  NVTE_CHECK(
+      A.scaling_mode == B.scaling_mode ||
```
ipanfilo: I don't think we support anything except A.scaling_mode == B.scaling_mode

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -190,89 +190,155 @@ static hipDataType get_hipblaslt_dtype(const transformer_engine::DType t) {
...
+    } else {
+      NVTE_CHECK(A.has_columnwise_data(), "Input A is missing column-wise usage");
+    }
+    ret.A = is_A_transposed ? A.data.dptr : A.columnwise_data.dptr;
+    ret.transA = transA;
+    ret.Atype = is_A_transposed ? A.data.dtype : A.columnwise_data.dtype;
+    ret.A_scale_inv = is_A_transposed ? A.scale_inv.dptr : A.columnwise_scale_inv.dptr;
+    ret.lda = is_A_transposed ? k : m;
+  } else if (A.scaling_mode == NVTE_BLOCK_SCALING_1D || A.scaling_mode == NVTE_BLOCK_SCALING_2D) {
+    // FP8 block scaling
```
ipanfilo: Is it supported on ROCm?

transformer_engine/common/include/transformer_engine/gemm.h
```diff
@@ -119,6 +119,15 @@ namespace transformer_engine {
 
 constexpr int num_streams = 4;
 
+/*! \brief TE/JAX cudaGraph requires the cuBLAS initialization to happen outside of the capturing
+ * region. This function is a helper to call cublasCreate() which allocate memory for the handle.
+ * The function will be called in the initialize phase of the related XLA custom calls.
+ */
+
+#ifndef __HIP_PLATFORM_AMD__
```
ipanfilo, line 129: It should have AMD copyright then

transformer_engine/pytorch/triton/permutation.py
ipanfilo: It looks like the file can be reset to upstream version

hipify_custom_map.json
```diff
@@ -5,7 +5,8 @@
         "util/cuda_runtime.h" : "util/hip_runtime.h",
         "ATen/cudnn/Handle.h" : "ATen/miopen/Handle.h",
         "CUfunc_cache" : "hipFuncCache_t", 
-        "<nvtx3/nvToolsExt.h>" : "<roctracer/roctx.h>"
+        "<nvtx3/nvToolsExt.h>" : "<roctracer/roctx.h>",
+        "cublasLtHandle_t" : "hipblasLtHandle_t"
```
ipanfilo: Where is it needed?

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -0,0 +1,1633 @@
...
+        layer_number: Optional[int] = None,
+        deterministic: bool = False,
+    ) -> None:
+        super().__init__()
+
+        self.softmax_scale = softmax_scale
+        self.attention_dropout = attention_dropout
+        self.attention_dropout_ctx = attention_dropout_ctx
+        self.attention_type = attention_type
+        self.use_FAv2_bwd = os.getenv(
```
ipanfilo: add if not IS_HIP_EXTENSION (ref. attention.py)

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -0,0 +1,1633 @@
+# This file was modified for portability to AMDGPU
+# Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
```
ipanfilo: Please update to 2022-2025 - this file copied code from earlier years than 2025

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -0,0 +1,3577 @@
+# This file was modified for portability to AMDGPU
+# Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
```
ipanfilo: Update to 2022-2025

alextmagro: I have updated the copyrights.

transformer_engine/pytorch/triton/permutation.py
alextmagro: I have reset the file to upstream.

transformer_engine/common/gemm/rocm_gemm.cu
alextmagro: You are right, we don't support BLOCK_SCALING yet, so I will remove reference to it in rocm_gemm.

hipify_custom_map.json
alextmagro: Artifact from testing, I have removed now.

wangye805 (approved):

tests/cpp/test_common.cu
```diff
@@ -731,6 +731,13 @@ std::pair<double, double> getTolerances(const DType type) {
 
 template <typename T>
 void generate_data_uniformly(T* data, const size_t size, std::mt19937* gen) {
+#ifdef __HIP_PLATFORM_AMD__
+  // TODO: Introduce a parallel RNG library (Random123, PCG, rocRAND)
+  std::uniform_real_distribution<> dis(-2.0, 1.0);
+  for (int i = 0; i < size; i++) {
+    data[i] = static_cast<T>(dis(*gen));
+  }
+#else
```
ipanfilo, line 740: What is a problem with original generator and should it be a part of IFU?

alextmagro: Original generator produces non-deterministic test data depending on the # of OMP threads. It also is only marginally faster as the threads have to generate numbers anyway when discard() is called. 

It could be a separate PR, but I added it to the IFU because the new test data generation method is causing the 2 GEMM failures I noticed, and we will need this fix in release 2.4.

ipanfilo: Please confirm - new method is causing 2 failures or original one?

alextmagro: original one, specifically when OMP_NUM_THREADS=16. They don't show up by default locally because we use OMP_NUM_THREADS=28 by default on our dev cluster. It only occurs on nodes with 128 vCPUs.

ipanfilo (approved):

PR #309, opened 2025-09-11
Hotfix removal of env variable in test

Micky774:
# Description

Originally even if test was run with `NVTE_CK_USES_FWD_V3=1` it would be deleted in the first test run and consequently prevents the use of FWD_V3 kernels. Setting the env variable in the test is not necessary since CI runs it with `NVTE_CK_USES_FWD_V3=1` anyways.

This PR will also update the AITER submodule commit to include fixes necessary for the tests to pass. This will fix https://github.com/ROCm/frameworks-internal/issues/13708

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/tests/pytorch/fused_attn/test_fused_attn.py
+++ b/tests/pytorch/fused_attn/test_fused_attn.py
@@ -288,6 +288,7 @@ def test_dot_product_mem_calc(workspace_opt):
     if FusedAttnBackend["CK"] not in fused_attn_backends:
         pytest.skip("This test requires the CK fused attention backend.")
 
+    nvte_ck_uses_fwd_v3 = os.environ["NVTE_CK_USES_FWD_V3"]
     os.environ["NVTE_CK_USES_FWD_V3"] = "1"
     os.environ["NVTE_FUSED_ATTN_CK"] = "1"
     os.environ["NVTE_FUSED_ATTN_AOTRITON"] = "0"
@@ -301,7 +302,7 @@ def test_dot_product_mem_calc(workspace_opt):
         pad_between_seqs,
         is_training,
     )
-    del os.environ["NVTE_CK_USES_FWD_V3"]
+    os.environ["NVTE_CK_USES_FWD_V3"] = nvte_ck_uses_fwd_v3
     del os.environ["NVTE_FUSED_ATTN_CK"]
     del os.environ["NVTE_FUSED_ATTN_AOTRITON"]
 
--- a/tests/pytorch/fused_attn/test_fused_attn.py
+++ b/tests/pytorch/fused_attn/test_fused_attn.py
@@ -288,8 +288,6 @@ def test_dot_product_mem_calc(workspace_opt):
     if FusedAttnBackend["CK"] not in fused_attn_backends:
         pytest.skip("This test requires the CK fused attention backend.")
 
-    nvte_ck_uses_fwd_v3 = os.environ["NVTE_CK_USES_FWD_V3"]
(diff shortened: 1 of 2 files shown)


wangye805 (approved):

ipanfilo (changes requested):
I think it breaks the modified test. Those variables were set on purpose

Micky774:
> I think it breaks the modified test. Those variables were set on purpose

I don't think it does actually. The V3 env variable doesn't affect the actual memory size calculations which were producing the original error. And even then, this file is run with the V3 flags as well, so it should still be covered.

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -557,6 +557,12 @@ void fused_attn_ck_fwd_impl(
       nvte_log_ck_config = true;
   }
   bool nvte_ck_uses_fwd_v3 = getenv<int>("NVTE_CK_USES_FWD_V3", 0);
+  if(layout==NVTE_QKV_Layout::NVTE_BSHD_BSHD_BSHD){
```
wangye805: Quick question, was our aiter fwd_v3 only supporting BSHD_BSHD_BSHD? I thought it may also support BS3HD, or BSHD_BS2HD. Or maybe it just does not support the SBHD family (SB3HD, SBH3D, SBHD_SB2HD, and so on). 

Let's confirm with aiter team and test those config on our side

wangye805 (changes requested):

tests/pytorch/fused_attn/test_fused_attn.py
ipanfilo, line 1: I have another question about this test: what does workspace_opt do there?

Micky774: Hmm I'm not sure. I don't think it affects workspace buffer size calculation. We can probably fix it to a single variable instead of running for both.

wangye805: @ipanfilo , its a NV upstream env for cudnn: https://github.com/ROCm/TransformerEngine/blob/ec21e1f35f5a36c1def9c89c4865eba3947f0665/transformer_engine/pytorch/attention.py#L5497

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: I've updated the guard to account for the specific failures.

wangye805 (approved):
Thanks for testing those layout format completely. Let's make sure all pytests can pass in CI before mering. 

Please also give those failed attn configs in our pytest to aiter team for their fixing.

tests/pytorch/fused_attn/test_fused_attn.py
ipanfilo: Ok then - seems we don't need it for ROCm tests. Since it exists not in this tests only, it may be different PR to remove running with different workspace_opt values

ipanfilo (approved):

Micky774:
Note that the tests are currently failing since this PR is waiting for https://github.com/ROCm/aiter/pull/994 to be merged, and with the updates there CI should be green.

Micky774:
The AITER PR has been merged, and the submodule commit has been bumped in this PR. Once CI is green it should be good to merge.

PR #318, opened 2025-09-18
Add scaling mode to GEMM algo cache key

ipanfilo:
# Description

Add scaling mode to GEMM algo cache key and to algo store not to reuse the same entry for Tensor and Block scaling

Fixes [#13781](https://github.com/ROCm/frameworks-internal/issues/13781)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/transformer_engine/common/gemm/rocm_gemm.cu
+++ b/transformer_engine/common/gemm/rocm_gemm.cu
@@ -506,6 +506,8 @@ public:
     int m, n, k;
     int lda, ldb, ldd;
     hipblasOperation_t transa, transb;
+    //fp8_scale is int instead of hipblasLtMatmulMatrixScale_t for compatibility with old hipblasLt
+    int fp8_scale;
     hipblasLtEpilogue_t epilogue;
 
     Key(int deviceCap_,
@@ -513,13 +515,13 @@ public:
         hipDataType d_type_, hipDataType bias_type_,
         int m_, int n_, int k_, int lda_, int ldb_, int ldd_,
         hipblasOperation_t transa_, hipblasOperation_t transb_,
-        hipblasLtEpilogue_t epilogue_):
+        int fp8_scale_, hipblasLtEpilogue_t epilogue_):
         deviceCap(deviceCap_),
         a_type(a_type_), b_type(b_type_),
         d_type(d_type_), bias_type(bias_type_),
         m(m_), n(n_), k(k_), lda(lda_), ldb(ldb_), ldd(ldd_),
         transa(transa_), transb(transb_),
-        epilogue(epilogue_) {}
+        fp8_scale(fp8_scale_), epilogue(epilogue_) {}
 
     Key() {}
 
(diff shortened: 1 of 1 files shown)


transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -506,20 +506,22 @@ public:
     int m, n, k;
     int lda, ldb, ldd;
     hipblasOperation_t transa, transb;
+    //fp8_scale is int instead of hipblasLtMatmulMatrixScale_t for compatibility with old hipblasLt
+    int fp8_scale;
```
wangye805: fp8_scale_mode? fp8_scale sounds like a float value for the scale

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -745,6 +747,22 @@ protected:
         continue;
       }
 
+      //Check and filter out compute and scale types
+      if (computeNameMapper.getValue(comp, "comp") != HIPBLAS_COMPUTE_32F ||
+        typeNameMapper.getValue(scale, "scale") != HIP_R_32F)
```
wangye805, line 752: Add a comment or debug printing for this if condition? For example, why comp cannot be HIPBLAS_COMPUTE_32F and why scale cannot be HIP_R_32F?

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -745,6 +747,22 @@ protected:
...
+        continue;
+      }
+
+#if HIPBLASLT_VERSION_MAJOR > 0 || HIPBLASLT_VERSION_MINOR >= 15
+      if (cfg.fp8_scale < 0 || cfg.fp8_scale >= (int)HIPBLASLT_MATMUL_MATRIX_SCALE_END)
+#else
+      if (cfg.fp8_scale != 0)
+#endif
+      {
+        continue;
```
wangye805, line 764: Similarly, add a comment or debug printing?

wangye805 (changes requested):

transformer_engine/common/gemm/rocm_gemm.cu
ipanfilo: In our code comp type and scale data type are passed to hipblasLt as hardcoded constants. Thus, if read values are different from HIPBLAS_COMPUTE_32F/HIP_R_32F this entry will never be used even if some few other types are supported by hipblaslt itself. It is minor optimization not to store hardcoded values. 
The whole 'if' statement is not new but moved here from line 770

wangye805: I saw computeNameMapper was only defined with HIPBLAS_COMPUTE_32F. Why do we choose to skip instead of error out for this if branch?

wangye805 (changes requested):

transformer_engine/common/gemm/rocm_gemm.cu
ipanfilo: Because it is not a failure itself to have different types but they are just not used. Like I said it is not a validity checking but optimization not to store unused parameters combinations. In fact, for comp it will throw error because of unrecognized name similar to any other values that it cannot parse. 
But if the value is recognized but we know it is not used - no error

wangye805: I see. Thanks

wangye805 (approved):

PR #326, opened 2025-10-01
Fix MXFP8 GEMM test failures

ipanfilo:
# Description

Fix MXFP8 GEMM related functionality and UT

Fixes [#12532](https://github.com/ROCm/frameworks-internal/issues/12532)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Fix MXFP8 GEMM reference result calculation
- Use OMP to speedup reference GEMM calculation
- Suppress massive compilation warnings
- Fix uninitialized scaling value that caused sporadic algo cache loading error
 
# Checklist:

### DIFF
--- a/tests/cpp/operator/test_cublaslt_gemm.cu
+++ b/tests/cpp/operator/test_cublaslt_gemm.cu
@@ -274,9 +274,9 @@ void performTest(const TestParams& params) {
   const bool a_colwise = !params.transa && isFp8Type(atype);
   const bool b_colwise = params.transb && isFp8Type(btype);
   Tensor A("A", params.transa ? TShape{ params.m, params.k } : TShape{ params.k, params.m },
-    atype, true, a_colwise, params.scaling_mode);
+    atype, (!a_colwise || !use_mxfp8), a_colwise, params.scaling_mode);
   Tensor B("B", params.transb ? TShape{ params.k, params.n } : TShape{ params.n, params.k },
-    btype, true, b_colwise, params.scaling_mode);
+    btype, (!b_colwise || !use_mxfp8), b_colwise, params.scaling_mode);
 
   Tensor D("D", TShape{ params.n, params.m }, dtype);
   Tensor bias;
@@ -291,7 +291,7 @@ void performTest(const TestParams& params) {
   //initialize the data and scale inv of A, B
   //fillUniform does not initialize columnwise data if rowwise data exist
   fillUniform(&A);
-  if (a_colwise) {
+  if (a_colwise && !use_mxfp8) {
     // A must be of shape k, m
     cpu_rowwise_to_columnwise(params.k, params.m,
       A.rowwise_cpu_dptr<A_Type>(), A.columnwise_cpu_dptr<A_Type>());
@@ -299,7 +299,7 @@ void performTest(const TestParams& params) {
     A.from_cpu();
   }
   fillUniform(&B);
-  if (b_colwise) {
+  if (b_colwise && !use_mxfp8) {
     // B must be of shape k, n
     cpu_rowwise_to_columnwise(params.k, params.n,
       B.rowwise_cpu_dptr<B_Type>(), B.columnwise_cpu_dptr<B_Type>());
--- a/tests/cpp/operator/test_cublaslt_gemm.cu
+++ b/tests/cpp/operator/test_cublaslt_gemm.cu
@@ -3,17 +3,15 @@
  *
  * License for AMD contributions = MIT. See LICENSE for more information
  ************************************************************************/
+#include <cmath>
+#include <iostream>
+#include <string>
+#include <cuda_bf16.h>
+#include <cuda_runtime.h>
+#include <gtest/gtest.h>
+#include <transformer_engine/cast.h>
 #include <transformer_engine/gemm.h>
 #include <transformer_engine/transformer_engine.h>
-#include <gtest/gtest.h>
-#include <cuda_runtime.h>
-#include <cuda_bf16.h>
-#include <memory>
-#include <iostream>
-#include <iomanip>
-#include <random>
-#include <cstring>
-#include <cmath>
 #include "../test_common.h"
 
 using namespace transformer_engine;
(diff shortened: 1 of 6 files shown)


tests/cpp/operator/test_cublaslt_gemm.cu
```diff
@@ -81,12 +81,14 @@ void compute_ref(
...
   size_t m, size_t k, size_t n,
   D_Type* ref_d_data,
-  float* ref_d_amax,
+  float* ref_d_amax_ptr,
   Gelu_Type* ref_gelu_data,
   bool transa,
   bool transb){
 
-  *ref_d_amax = 0;
+  float ref_d_amax = 0;
```
wenchenvincent, line 74: What's the motivation for the changes here?

ipanfilo: I'm not sure if OMP would work correctly with direct update of *ref_d_amax while local variable definitely works

tests/cpp/operator/test_cublaslt_gemm.cu
```diff
@@ -81,12 +66,14 @@ void compute_ref(
...
+
+  #pragma omp parallel for schedule(static) collapse(2) reduction(max: ref_d_amax) proc_bind(spread)
```
alextmagro, line 76: schedule(static) is unnecessary as it is the default behavior -- also probably don't want proc_bind(spread) as reductions are involved. May want to consider something like schedule(static, n) if cache performance is a bottleneck.

Probably better to have separate loops to avoid if statements in the inner loop, and use a firstprivate variable to store a local max before doing a final reduction. I'd consider a #pragma omp simd reduction(+:val) collective for the kk loop as well. 

Could move the a_scale_inv*b_scale_inv outside of the 3rd loop too.

tests/cpp/operator/test_cublaslt_gemm.cu
```diff
@@ -106,41 +93,45 @@ void compute_ref(
...
-        float a_val = a_data[ii*k + kk];
-        float b_val = b_data[kk + jj*k];
-        float a_scale_inv_val =
-            (float)std::pow(2, a_scale_inv_data[ii * a_scale_inv_shape.data[1] + kk / 32] - 127);
-        float b_scale_inv_val =
-            (float)std::pow(2, b_scale_inv_data[kk / 32 + jj * b_scale_inv_shape.data[1]] - 127);
-        val += a_scale_inv_val * a_val * b_scale_inv_val * b_val;
+        size_t a_idx = transa ? (ii*k + kk) : (kk*m + ii);
+        size_t b_idx = transb ? (kk*n + jj) : (jj*k + kk);
+        float a_scale_inv_val = (float)std::pow(2,
```
alextmagro, line 130: Are we able to use exp2f here instead of pow?

tests/cpp/operator/test_cublaslt_gemm.cu
```diff
@@ -106,41 +93,45 @@ void compute_ref(
...
-  float* ref_d_amax,
+  float* ref_d_amax_ptr,
   Gelu_Type* ref_gelu_data,
   bool transa,
   bool transb){
 
-  *ref_d_amax = 0;
+  float ref_d_amax = 0;
+
+  #pragma omp parallel for schedule(static) collapse(2) reduction(max: ref_d_amax) proc_bind(spread)
```
alextmagro, line 123: Same comments as above here, but we also should do two optimizations here that are important for OMP -- tiling the matrix, and switching the j and k for loops depending on the transpose.

alextmagro:
Hi Ilya, Matthias and I are working on parallelizing our cpp reference code, including this test. Since we are planning on offloading this reference code given the computational time, can we revert the OMP additions for now?

ipanfilo:
> Hi Ilya, Matthias and I are working on parallelizing our cpp reference code, including this test. Since we are planning on offloading this reference code given the computational time, can we revert the OMP additions for now?

OMP change should not impede your work but it gives immediate test time improvement meanwhile

alextmagro:
> OMP change should not impede your work but it gives immediate test time improvement meanwhile

Noted -- Will resolve my other comments as well.

tests/cpp/operator/test_cublaslt_gemm.cu
```diff
@@ -30,29 +28,17 @@ std::vector<std::tuple<size_t, size_t, size_t>> test_case_sizes = {
...
 }; 
 
+std::vector<std::tuple<size_t, size_t, size_t>> test_case_sizes_mxfp8 = {
+  {2304, 768, 4096},
+}; 
+
 //  A, B, Bias, Gelu, D
 //  Bias type choose as bf16 in use_fp8, D_type otherwise
 //  Gelu type the same as Bias_Type
-//  {DType::kFloat32, DType::kFloat32, DType::kFloat32, DType::kFloat32, DType::kFloat32},
```
wenchenvincent, line 36: These comments list the configs to be tested, right? Why remove them?

tests/cpp/test_common.cu
```diff
@@ -454,6 +454,9 @@ void Tensor::set_scale_inv(float scale_inv) {
         columnwise_cpu_scale_inv_ptr<float>()[0] = scale_inv;
       } else {
         std::uniform_int_distribution<uint8_t> dis(0, 127);
+        if (rowwise_) {
```
wenchenvincent, line 457: Is this miss by upstream?

tests/cpp/operator/test_cublaslt_gemm.cu
```diff
@@ -401,49 +416,93 @@ void performTest(const TestParams& params) {
...
+  D.to_cpu();
+
+  using Ref_Type = fp16;
+  DType ref_type = TypeInfo<Ref_Type>::dtype;
+
+  Tensor A_ref("A", params.transa ? TShape{params.m, params.k} : TShape{params.k, params.m}, ref_type);
+  Tensor B_ref("B", params.transb ? TShape{params.k, params.n} : TShape{params.n, params.k}, ref_type);
+  Tensor D_ref("D", TShape{params.n, params.m}, dtype);
+
+  nvte_dequantize(A.data(), A_ref.data(), 0);
```
wenchenvincent: So this tests generates fp8 and scale data and performs dequantization and use the higher precision gemm results as the reference. While it tests the dequantization kernel was making the right assumption, shall we test the quantization kernel directly?

tests/cpp/operator/test_cublaslt_gemm.cu
ipanfilo: This comment is not maintained to match real list of configs

tests/cpp/test_common.cu
ipanfilo: Yes, but actually it can be reverted - there are more issues in upstream code with maintaining rowwise and columnwise together so this patch helps only for subset of tests

wenchenvincent (approved):
LGTM.

PR #328, opened 2025-10-03
13679 TE2.4 keep_fp8_transpose_cache refactor

sudhu2k:
# Description

Refactoring of keep_fp8_transpose_cache flag in a more systematic way.
Fixes # ([13679](https://github.com/ROCm/frameworks-internal/issues/13679))

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

There are 2 ways to initialize parameters in modules, 
`with fp8_model_init(enabled=(True/False):`

When set to True, modules created inside this context will only store FP8 copies of their parameters, not the higher precision versions.
When set to False, modules created inside this context stores both higher precision and FP8 copies of parameters.

When enabled is set to True, https://github.com/ROCm/TransformerEngine/blob/18f925ffb00ababdb3268e99af5d651f55c3fd52/transformer_engine/pytorch/module/base.py#L921-L929
the code goes inside the above block and quantizer's columnwise usage is updated based on the keep_fp8_weight_transpose_cache. So, while initializing, we don't create any transpose cache.

Note: A change in TEv2.4 made it so that even when enabled is set to True for fp8_model_init, it still goes inside get_weight_workspace to update the rowwise and columnwise usage based on the weight quantizer's usage.
https://github.com/ROCm/TransformerEngine/blob/a9c402600a559f3a3c0086a0c227ca2d73020da2/transformer_engine/pytorch/module/linear.py#L228-L251

While enabled is set to False,
FP8 parameters are created lazily during the first forward pass through the get_weight_workspace() method. Since out is None here, whatever quantizer is passed to this function will be used to create the transpose. So, we just have to make sure that the quantizer passed to this function has columnwise usage set based on keep_fp8_transpose_cache. 
Quantizers are retrieved using _get_quantizers method as shown in this block -> https://github.com/ROCm/TransformerEngine/blob/18f925ffb00ababdb3268e99af5d651f55c3fd52/transformer_engine/pytorch/module/linear.py#L1140-L1146

https://github.com/ROCm/TransformerEngine/blob/18f925ffb00ababdb3268e99af5d651f55c3fd52/transformer_engine/pytorch/module/linear.py#L1212-L1213
Hence, adding the lines to update the columnwise usage in the _get_quantizers method does the job since this would be the quantizer that'll be used when get_weight_workspace is called.

Why we reverted get_weight_workspace method in base.py? (https://github.com/ROCm/TransformerEngine/blob/18f925ffb00ababdb3268e99af5d651f55c3fd52/transformer_engine/pytorch/module/base.py#L941-L1019)

Since when create workspace is called with the updated quantizer, the workspace is cached using: `self._fp8_workspaces[cache_name] = out`

This means out tensor would now have quantizer object as the one that's initially passed which depended on keep_fp8_transpose cache. So, no need to change anything here.

## Change in transpose.cpp to check numel
Another issue is
When _transpose is None
https://github.com/ROCm/TransformerEngine/blob/a9c402600a559f3a3c0086a0c227ca2d73020da2/transformer_engine/pytorch/csrc/extensions/transpose.cpp#L96-L100
if (output.has_value()) {
out = *output;
} else {
out = allocateTorchTensor(input.size(1), input.size(0), DType::kByte);
}
output.has_value() is False, so it tries to allocate tensors

when the keep_fp8_transpose_cache flag is set to False, the _transpose is cleared and becomes a tensor([], device='cuda:0') and not None.
In this case
output.has_value() is True, so an empty tensor is passed to create a transpose which leads to
E RuntimeError: /workspace/installs/TransformerEngine/transformer_engine/common/transpose/transpose.hip:206 in function transpose: Assertion failed: output.data.dptr != nullptr. Output is not allocated.

Current tests pass, because it's only run for a single iteration. Changing the unit test to multiple iterations makes the tests fail.

# Checklist:


tests/pytorch/test_numerics.py
```diff
@@ -1330,8 +1332,8 @@ def test_fp8_linear_without_transpose_cache_accuracy(dtype, bs, model, fp8_model
             keep_fp8_weight_transpose_cache=True # defaults to True
         ).eval()
 
-    outputs = _test_granular_accuracy_with_fp8(layer, bs, dtype, config)
-    ref_outputs = _test_granular_accuracy_with_fp8(ref_layer, bs, dtype, config)
+    outputs = _test_granular_accuracy_with_fp8(layer, bs, dtype, config, num_iterations=2)
```
ipanfilo: It seems you can achieve the same with much smaller changes - just call test_granular_accuracy_with_fp8() 2 times in a loop

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -1176,6 +1174,8 @@ def __init__(
         keep_fp8_weight_transpose_cache: bool = True,
     ) -> None:
         super().__init__()
+        if not IS_HIP_EXTENSION:
```
ipanfilo: Move it before line 1198 so it is grouped with assigning of corresponding property. It can even be a part of property assignment

tests/pytorch/test_numerics.py
sudhu2k: Sure, yes that'll be the simplest solution but I was actually testing them out with higher num_iterations, for example 10, to make sure the changes are solid. I also think this is a much cleaner way because it recreates a regular training loop and even though we don't zero grad, we can see the progress on the grad. Plus, it gives us an easy way to adjust the iterations down the road. But I guess IFUs had to be done carefully for this test. Let me know what you think.

transformer_engine/pytorch/module/linear.py
```diff
@@ -228,8 +228,8 @@ def forward(
         if fp8 or debug:
             # Configure quantizer
             if weight_quantizer is not None:
-                columnwise_usage = is_grad_enabled and inp.requires_grad
-                if not columnwise_usage:
+                columnwise_usage = is_grad_enabled and inp.requires_grad and keep_fp8_weight_transpose_cache
+                if not columnwise_usage and keep_fp8_weight_transpose_cache:
                     columnwise_usage = (
                         is_fp8_activation_recompute_enabled()
```
wangye805, line 234: Do we understand why fp8 columnwise data is needed in the activation recompute flow?

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -516,7 +516,7 @@ def __init__(
         self.ub_overlap_rs = ub_overlap_rs
         self.ub_overlap_ag = ub_overlap_ag
         self.ub_name = ub_name
-        self.keep_fp8_weight_transpose_cache = keep_fp8_weight_transpose_cache
+        self.keep_fp8_weight_transpose_cache = keep_fp8_weight_transpose_cache if IS_HIP_EXTENSION else True
```
wangye805: Will this keep_fp8_weight_transpose_cache be used in the group linear module? If not, can we make this data member to be optional and true by default in https://github.com/ROCm/TransformerEngine/blob/7c9ebe185c826a4c9bc6165fd9b93e933e8961bc/transformer_engine/pytorch/module/base.py#L599, to skip this assignment?

transformer_engine/pytorch/module/linear.py
```diff
@@ -652,7 +651,8 @@ def backward(ctx, grad_output: torch.Tensor) -> Tuple[Union[torch.Tensor, None],
                 nvtx_range_pop(f"{nvtx_label}.dgrad_gemm")
 
                 if ctx.fp8 and not ctx.keep_fp8_weight_transpose_cache:
-                    clear_fp8_weight_transpose_cache(weight_fp8)
+                    clear_tensor_data(weight_fp8._transpose)
+                    weight_fp8.update_usage(columnwise_usage=False)
```
wangye805, line 654: Does update_usage truly clear the tranpose data allocation? 

Previously they use clear_tensor_data with the trick to replace with empty tensor instead of setting it to None. Probably NV upstream knows that setting it to None do not clear that buffer allocation in pytorch?

We can verify it using the memory pickle file

transformer_engine/pytorch/csrc/extensions/transpose.cpp
```diff
@@ -93,7 +93,7 @@ at::Tensor fp8_transpose(at::Tensor input, DType otype, std::optional<at::Tensor
   size_t N = static_cast<size_t>(input.size(1));
 
   at::Tensor out;
-  if (output.has_value()) {
+  if (output.has_value() && output->numel() > 0) {
```
wangye805: Use a rocm specific guard

Or, assign the deallocated transpose buffer to None or make it has_value() return false?

wangye805 (changes requested):
I only checked the c++ extension part and the linear module

sudhu2k:
When keep_fp8_weight_transpose_cache is set to True:
image

When keep_fp8_weight_transpose_cache is set to False with clear memory added:
image

When keep_fp8_weight_transpose_cache is set to False with setting transpose to None using update_usage:
image

Same amount of memory reduction with and without clearing memory manually. 288MB to 280MB, is understandable since there are 2 linear layers and _transpose data is of 4MB each.

When it's being manually cleared
image
The transpose is disposed immediately after the gemm
image

Same happens even without clearing it manually.

transformer_engine/pytorch/module/linear.py
sudhu2k: Will research and get back on this.

transformer_engine/pytorch/module/grouped_linear.py
sudhu2k: Removed the use of keep_fp8_weight_transpose_cache from grouped_linear

transformer_engine/pytorch/module/linear.py
sudhu2k: Added memory traces with this PR below. Seems like clear_tensor_data is not required.

transformer_engine/pytorch/csrc/extensions/transpose.cpp
sudhu2k: Removed this, since it seems like NV code isn't accepting empty tensors for weights' transpose and code change has been made to use update_usage to set output tensor to None instead of setting it as empty tensor.

wangye805:
> When keep_fp8_weight_transpose_cache is set to True: image
> 
> When keep_fp8_weight_transpose_cache is set to False with clear memory added: image
> 
> When keep_fp8_weight_transpose_cache is set to False with setting transpose to None using update_usage: image
> 
> Same amount of memory reduction with and without clearing memory manually. 288MB to 280MB, is understandable since there are 2 linear layers and _transpose data is of 4MB each.
> 
> When it's being manually cleared image The transpose is disposed immediately after the gemm image
> 
> Same happens even without clearing it manually.

Great work for this investigation!

I'll continue my code review for the other two modules

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -501,7 +501,6 @@ def __init__(
         ub_overlap_ag: bool = False,
         ub_name: Optional[str] = None,
         delay_wgrad_compute: bool = False,
-        keep_fp8_weight_transpose_cache: bool = True,
```
wangye805, line 504: After removing this, can you help check whether this file is completely aligned with NV upstream? If so, you can remove the amd rocm copyright info

wangye805 (approved):

tests/pytorch/test_numerics.py
ipanfilo: I do not see how loop here makes any of the aforementioned harder or not possible. Unless you want make multiple iterations w/o RGN reset, in which case yes - your approach is preferable

transformer_engine/pytorch/module/grouped_linear.py
sudhu2k: There is no amd rocm copyright info in the grouped_linear module, I guess it wasn't added. There is an empty space that needs to be removed to align with the upstream, I've added a new commit. Thanks for letting me know!

tests/pytorch/test_numerics.py
sudhu2k: Yeah, it all comes down to resetting RNG several times or once and resetting the RNG once at the beginning makes the entire loop deterministic anyways, so it doesn't matter if it's done multiple times or once. Only downside with the new approach is about the IFU and merge conflicts that may arise. If we want to make it easier to merge, I guess I can revert back. Let me know!

ipanfilo: I think IFU is important so smaller changed are preferred

sudhu2k: Done, I've reverted back the _test_granular_accuracy_with_fp8 function and modified the unit test to call the function multiple times.

tests/pytorch/test_numerics.py
```diff
@@ -1332,11 +1330,17 @@ def test_fp8_linear_without_transpose_cache_accuracy(dtype, bs, model, fp8_model
             keep_fp8_weight_transpose_cache=True # defaults to True
         ).eval()
 
-    outputs = _test_granular_accuracy_with_fp8(layer, bs, dtype, config, num_iterations=2)
-    ref_outputs = _test_granular_accuracy_with_fp8(ref_layer, bs, dtype, config, num_iterations=2)
+    num_iterations = 2
```
ipanfilo, line 1336: Add comment that was originally added on _test_granular_accuracy_with_fp8

sudhu2k: Done!

ipanfilo (approved):

PR #329, opened 2025-10-03
CI hotfix: IFU test update

Micky774:
# Description

A change to `tests/pytorch/test_cpu_offloading.py` causes a crash on our CI -- this is a quick hotfix for it.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/ci/pytorch.sh
+++ b/ci/pytorch.sh
@@ -58,7 +58,7 @@ run_test_config(){
     run_default_fa 1 test_deferred_init.py
     run_default_fa 1 test_float8tensor.py
     run_default_fa 1 test_float8_current_scaling_exact.py
-    run_default_fa 1 test_cpu_offloading.py
+    NVTE_FLASH_ATTN=0 run_default_fa 1 test_cpu_offloading.py
     run_default_fa 1 test_fused_rope.py
     run_default_fa 1 test_fusible_ops.py
     run_default_fa 3 test_gemm_autotune.py
--- a/ci/pytorch.sh
+++ b/ci/pytorch.sh
@@ -58,7 +58,7 @@ run_test_config(){
     run_default_fa 1 test_deferred_init.py
     run_default_fa 1 test_float8tensor.py
     run_default_fa 1 test_float8_current_scaling_exact.py
-    NVTE_FLASH_ATTN=0 run 1 test_cpu_offloading.py
+    test $_fus_attn = auto -o $_fus_attn = ck -o $_fus_attn = aotriton && NVTE_FLASH_ATTN=0 run 1 test_cpu_offloading.py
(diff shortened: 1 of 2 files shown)


ipanfilo (commented):
Can you please describe the problem a little bit more

tests/pytorch/test_cpu_offloading.py
```diff
@@ -29,7 +29,7 @@
 
 # Flash attention saves some internal tensor for the backward pass
 # that cannot be offloaded to CPU.
-assert os.getenv("NVTE_FLASH_ATTN") == "0"
+assert os.getenv("NVTE_FLASH_ATTN", "1") == "0"
```
ipanfilo, line 32: It may be version specifc but there are no problems with ROCm FA, so this condition can just be guarded.
The file also needs AMD copyright then

ci/pytorch.sh
```diff
@@ -58,7 +58,7 @@ run_test_config(){
     run_default_fa 1 test_deferred_init.py
     run_default_fa 1 test_float8tensor.py
     run_default_fa 1 test_float8_current_scaling_exact.py
-    run_default_fa 1 test_cpu_offloading.py
+    NVTE_FLASH_ATTN=0 run_default_fa 1 test_cpu_offloading.py
```
ipanfilo: Let's run it with default config until it causes problem on ROCm (see the test comment)

wangye805:
> Can you please describe the problem a little bit more

I have a similar question: which IFU test failed? In our dev CI, I only saw core test failures (cpp test failures)
image

ipanfilo:
> I have a similar question: which IFU test failed? In our dev CI, I only saw core test failures (cpp test failures) 

The problem is only seen on Level=1 that why it passed dev but fails in PRs. It is related to naive reading of env variable w/o handling the case when the variable does not exist. Levels 1 and 3 run cpu_offloading test with different FA configs so in one case the var is there and in the other is not.

ci/pytorch.sh
ipanfilo: NVM, it really fails with FA. Since it behaved differently with different FA backends, let't run it with $_fus_attn in "auto", "ck", "aotriton"

Micky774: Updated, should be good now if I understand our CI script correctly

ipanfilo: I think it will fail on Level 1 unfused and Level 3 flash attn configs because the test requires some fused attn backend. That is why _fus_attn has to be checked for valid values

wangye805 (approved):

ipanfilo (approved):

PR #332, opened 2025-10-06
[Fix] Missed import and dim reference

GeneDer:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,7 @@
 """Installation script."""
 
 import os
+import sys
 import time
 from pathlib import Path
 from typing import List, Tuple
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,6 @@
 """Installation script."""
 
 import os
-import sys
 import time
 from pathlib import Path
 from typing import List, Tuple
@@ -153,6 +152,8 @@ def setup_requirements() -> Tuple[List[str], List[str], List[str]]:
     if not found_cmake():
         setup_reqs.append("cmake>=3.21")
     if not found_ninja():
+        import sys
+
         subprocess.check_call([sys.executable, "-m", "pip", "install", "ninja"])
         setup_reqs.append("ninja")
     if not found_pybind11():
--- a/transformer_engine/jax/flax/transformer.py
+++ b/transformer_engine/jax/flax/transformer.py
@@ -637,7 +637,7 @@ class DotProductAttention(nn.Module):  # pylint: disable=too-few-public-methods
             dropout_rng = self.make_rng(self.dropout_rng_name)
 
         if self.scale_factor is None:
-            scale_factor = 1.0 / sqrt(self.head_dim_qk)
+            scale_factor = 1.0 / sqrt(head_dim_qk)
         else:
             scale_factor = self.scale_factor
         del self.scale_factor


setup.py
```diff
@@ -7,6 +7,7 @@
 """Installation script."""
 
 import os
+import sys
```
GeneDer: This line will fail without the import https://github.com/ROCm/TransformerEngine/blob/409b43a6ae9076f254e9eea30d0dba4d63d14032/setup.py#L155

transformer_engine/jax/flax/transformer.py
```diff
@@ -637,7 +637,7 @@ def __call__(
             dropout_rng = self.make_rng(self.dropout_rng_name)
 
         if self.scale_factor is None:
-            scale_factor = 1.0 / sqrt(self.head_dim_qk)
+            scale_factor = 1.0 / sqrt(head_dim_qk)
```
GeneDer, line 640: `self.head_dim_qk` don't exist, should be `head_dim_qk` set in this line https://github.com/ROCm/TransformerEngine/blob/409b43a6ae9076f254e9eea30d0dba4d63d14032/transformer_engine/jax/flax/transformer.py#L604

wenchenvincent:
@GeneDer The error that you were seeing occurred only after you switched to use latest dev, right?

setup.py
wenchenvincent: Good catch. It seems that it got accidentally removed in the IFU PR. Could you add `import sys` right before line 155? This way it will be less likely to be accidentally removed during the future IFUs.

GeneDer:
> @GeneDer The error that you were seeing occurred only after you switched to use latest dev, right?

@wenchenvincent for the dim reference issue yes! I think missing sys import error might be there for a while. It was revealed recently when we switched to a leaner image without ninja

transformer_engine/jax/flax/transformer.py
wenchenvincent: Did you run into an error in a model run?

GeneDer: Yes!
image

setup.py
GeneDer: Moved :)

transformer_engine/jax/flax/transformer.py
wenchenvincent: @alextmagro @wangye805 @ipanfilo I am wondering why this issue was not discovered by CI.

alextmagro: Looks like I missed this change when readding changes we cherry picked  from release v2.6 regarding head_dim, apologies for that.

As for why this isn't caught, DotProductAttention is only called within the codebase by MultiHeadAttention, which explicitly sets the scale_factor. While we have a public interface for DotProductAttention, no tests call it directly, so there is no path where scale_factor is not already set.

We do have one reference to DotProductAttention in our docs/examples, in attention/arbitrary_mask_to_post_scale_bias.py but I don't think we run these examples on CI

wangye805 (approved):

PR #333, opened 2025-10-06
[Fix] Added dbias and dgelu kernels for ROCm

AllenFarcas:
# Description

Fixes https://github.com/ROCm/frameworks-internal/issues/13667 

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Added guards for `getDeviceComputeCapability()` which is for NVIDIA platforms
- Implemented`fp8_quantize_rocm` reusing `CastVectorizedUnaryKernelLauncher` and `CastVectorizedUnaryGradKernelLauncher`
- Implemented `partial_reduce_kernel` and `reduce_dbias_rocm` to efficiently reduce large sized inputs
- Enabled `test_cast_dbias` and `test_cast_dbias_dgelu` cpp tests for ROCm

# Checklist:


transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -1209,17 +1209,74 @@ template <bool IS_DBIAS, bool IS_DACT, bool IS_ACT, typename ParamOP,
...
+    }
+  #endif //#ifndef __HIP_PLATFORM_AMD__
+
   switch (output->scaling_mode) {
     case NVTE_DELAYED_TENSOR_SCALING: {
-      if (!IS_DACT) {
-        CastVectorizedUnaryKernelLauncher<ParamOP, OP>(input, noop, output, stream);
-      } else {
-        CastVectorizedUnaryGradKernelLauncher<ParamOP, OP>(input, act_input, output, stream);
+      if constexpr (IS_DBIAS) {
```
ipanfilo: This branch is the majority of the method and it is action only on ROCm. Won't it be easier to have separate fp8_quantize_amd() and keep l_100 method untouched?

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -1250,19 +1307,21 @@ void fp8_quantize(const Tensor &input, const Tensor *act_input, const Tensor *no
...
-#endif //#ifndef __HIP_PLATFORM_AMD__
-    // Supported by the Arch < 10.0
+ #ifndef __HIP_PLATFORM_AMD__
+     // NVIDIA
+     // Supported by the Arch >= 10.0
+     if (is_supported_by_CC_100()) {
+         fp8_quantize_arch_ge_100<IS_DBIAS, IS_DACT, IS_ACT, ParamOP, OP>(input, act_input, noop, output,
+                                                                         dbias, workspace, stream);
+     } else { // Supported by the Arch < 10.0
+         fp8_quantize_arch_l_100<IS_DBIAS, IS_DACT, IS_ACT, ParamOP, OP>(input, act_input, noop, output,
```
ipanfilo: Is it different from original code?

AllenFarcas: Functionality is the same but it is a clearer separation of NVIDIA and AMD code with a single ifndef.

transformer_engine/common/util/cast_kernels.cuh
AllenFarcas: Implemented

tests/cpp/operator/test_cast_dbias.cu
```diff
@@ -149,15 +149,17 @@ class CastDBiasTestSuite : public ::testing::TestWithParam<std::tuple<transforme
 TEST_P(CastDBiasTestSuite, TestCastDBias) {
     using namespace transformer_engine;
     using namespace test;
-    // Skip tests for pre-Blackwell architectures
-    if (getDeviceComputeCapability() < blackwellComputeCapability) {
-        GTEST_SKIP();
-    }
+
```
ipanfilo: Please revert unnecessary indentation changes to minimize possible further IFU conflicts

transformer_engine/common/util/vectorized_pointwise.h
```diff
@@ -1,4 +1,6 @@
 /*************************************************************************
+ * This file was modified for portability to AMDGPU
```
ipanfilo: Based on offline discussion, should the kernels be moved to separate header?

tests/cpp/operator/test_cast_dbias.cu
```diff
@@ -149,15 +149,17 @@ class CastDBiasTestSuite : public ::testing::TestWithParam<std::tuple<transforme
...
+    #ifndef __HIP_PLATFORM_AMD__
```
wangye805: There should not be space before "#ifdef" and "#endif"

tests/cpp/operator/test_cast_dbias_dgelu.cu
```diff
@@ -164,10 +164,13 @@ class CastDBiasDGeluTestSuite : public ::testing::TestWithParam<std::tuple<trans
 TEST_P(CastDBiasDGeluTestSuite, TestCastDBiasDgelu) {
     using namespace transformer_engine;
     using namespace test;
-    // Skip tests for pre-Blackwell architectures
-    if (getDeviceComputeCapability() < blackwellComputeCapability) {
-        GTEST_SKIP();
-    }
+
+    #ifndef __HIP_PLATFORM_AMD__
```
wangye805: Same issue here

tests/cpp/operator/test_cast_float8blockwise.cu
```diff
@@ -478,9 +478,11 @@ class FusedCastFloat8VectorwiseTestSuite
   }
 
 TEST_P(FusedCastFloat8BlockwiseTestSuite, TestFusedCastFloat8Blockwise) {
-  if (getDeviceComputeCapability() < hopperComputeCapability) {
-    GTEST_SKIP();
-  }
+  #ifndef __HIP_PLATFORM_AMD__
```
wangye805: Indent

tests/cpp/operator/test_cast_float8blockwise.cu
```diff
@@ -529,9 +531,11 @@ TEST_P(FusedCastFloat8BlockwiseTestSuite, TestFusedCastFloat8Blockwise) {
 }
 
 TEST_P(FusedCastFloat8VectorwiseTestSuite, TestFusedCastFloat8Vectorwise) {
-  if (getDeviceComputeCapability() < hopperComputeCapability) {
-    GTEST_SKIP();
-  }
+  #ifndef __HIP_PLATFORM_AMD__
```
wangye805: Indent

tests/cpp/operator/test_normalization.cu
```diff
@@ -35,9 +35,11 @@ void performTest(const size_t N, const size_t H, const bool zero_centered_gamma,
     return;
   }
 
-  if (getDeviceComputeCapability() < blackwellComputeCapability && use_cudnn) {
-    GTEST_SKIP() << "cuDNN normalizations not supported on pre-Blackwell GPUs yet!";
-  }
+  #ifndef __HIP_PLATFORM_AMD__
```
wangye805: Indent

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -1228,6 +1228,79 @@ void fp8_quantize_arch_l_100(const Tensor &input, const Tensor *act_input, const
...
+        TRANSFORMER_ENGINE_TYPE_SWITCH_INPUT(
+            input.dtype(), IType,
+            TRANSFORMER_ENGINE_TYPE_SWITCH_FP8ONLY(
+                output->dtype(), OType,
+                constexpr auto UnaryOP = ActivationType<float, ParamOP, OP>::op;
+                constexpr int nvec = 16 / sizeof(IType);
+
+                if constexpr (IS_DACT) {
+                  // Case: dgelu + dbias fusion
+                  VectorizedCastDBiasDGeluKernelLauncher<nvec, ParamOP, UnaryOP>(
```
wangye805: Let's not name it DGelu as this should be able to handle other activation function

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -1228,6 +1228,79 @@ void fp8_quantize_arch_l_100(const Tensor &input, const Tensor *act_input, const
...
+                      reinterpret_cast<const IType *>(input.data.dptr),
+                      reinterpret_cast<const IType *>(act_input->data.dptr),
+                      reinterpret_cast<OType *>(output->data.dptr),
+                      reinterpret_cast<fp32 *>(workspace->data.dptr),
+                      reinterpret_cast<const fp32 *>(output->scale.dptr),
+                      reinterpret_cast<fp32 *>(output->amax.dptr),
+                      reinterpret_cast<fp32 *>(output->scale_inv.dptr), N, {}, stream);
+                } else {
+                  // Case: Simple dbias fusion
+                  VectorizedCastAndDBiasKernelLauncher<nvec, ParamOP, detail::identity>(
```
wangye805: Well, compared to the CastVectorizedUnaryKernelLauncher below, this is quite of different style. Are we able to align the style? In otherwords, just make a wrapper to take in input, act_input, ...

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -1250,19 +1323,21 @@ void fp8_quantize(const Tensor &input, const Tensor *act_input, const Tensor *no
...
-    // Supported by the Arch < 10.0
-    fp8_quantize_arch_l_100<IS_DBIAS, IS_DACT, IS_ACT, ParamOP, OP>(input, act_input, noop, output,
+ #ifndef __HIP_PLATFORM_AMD__
+     // NVIDIA
+     // Supported by the Arch >= 10.0
+     if (is_supported_by_CC_100()) {
+         fp8_quantize_arch_ge_100<IS_DBIAS, IS_DACT, IS_ACT, ParamOP, OP>(input, act_input, noop, output,
+                                                                         dbias, workspace, stream);
+     } else { // Supported by the Arch < 10.0
+         fp8_quantize_arch_l_100<IS_DBIAS, IS_DACT, IS_ACT, ParamOP, OP>(input, act_input, noop, output,
```
wangye805: Why not we enhance the fp8_quantize_arch_l_100 function?

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -1250,19 +1323,21 @@ void fp8_quantize(const Tensor &input, const Tensor *act_input, const Tensor *no
...
+                                                                        dbias, workspace, stream);
```
wangye805: Also, just curious, how does NV H100 handle these testcases? H100 is with arch <10.0. Do they also run into the same error?

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -1228,6 +1228,79 @@ void fp8_quantize_arch_l_100(const Tensor &input, const Tensor *act_input, const
...
+        // This path handles all dbias fusions for non-TMA architectures.
+        NVTE_CHECK(dbias->data.dtype == input.data.dtype, "DBias must have the same type as input.");
+        NVTE_CHECK(dbias->data.shape == std::vector<size_t>{input.flat_last_dim()},
+                   "Wrong shape of DBias.");
+        NVTE_CHECK(workspace != nullptr, "Workspace must be a tensor.");
+
+        // Workspace sizing logic (same for both dbias variants).
+        if (workspace->data.dptr == nullptr) {
+          workspace->data.shape = input.data.shape;
+          workspace->data.dtype = DType::kFloat32;
```
wangye805: Emm, you are now requesting a buffer in fp32 with the same shape as input, which is even larger than the original input. Probably fine with dbias+dact cases.

In the dbias without dact case, this workspace is not even needed. We can get the dbias by reducing on the input directly.

wangye805 (changes requested):

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -1228,6 +1228,79 @@ void fp8_quantize_arch_l_100(const Tensor &input, const Tensor *act_input, const
...
+                      reinterpret_cast<const IType *>(input.data.dptr),
+                      reinterpret_cast<OType *>(output->data.dptr),
+                      reinterpret_cast<fp32 *>(workspace->data.dptr),
+                      reinterpret_cast<const fp32 *>(output->scale.dptr),
+                      reinterpret_cast<fp32 *>(output->amax.dptr),
+                      reinterpret_cast<fp32 *>(output->scale_inv.dptr), N, {}, stream);
+                }
+
+                // The reduction step is the same for both cases.
+                reduce_dbias<IType>(reinterpret_cast<float *>(workspace->data.dptr), dbias, rows,
```
wenchenvincent: Why do we need `reduce_dbias` if we already fuse the dbias reduction?

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -1228,6 +1228,79 @@ void fp8_quantize_arch_l_100(const Tensor &input, const Tensor *act_input, const
   }
 }
 
+// ROCm FP8 quantize
+template <bool IS_DBIAS, bool IS_DACT, bool IS_ACT, typename ParamOP,
+          float (*OP)(float, const ParamOP &)>
+void fp8_quantize_amd(const Tensor &input, const Tensor *act_input, const Tensor *noop,
```
alextmagro: This function should be moved to rocm_cast_kernels.cuh
Nit: Maybe we should use the terminology rocm instead of amd here.

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -1250,19 +1323,21 @@ void fp8_quantize(const Tensor &input, const Tensor *act_input, const Tensor *no
...
+     if (is_supported_by_CC_100()) {
+         fp8_quantize_arch_ge_100<IS_DBIAS, IS_DACT, IS_ACT, ParamOP, OP>(input, act_input, noop, output,
+                                                                         dbias, workspace, stream);
+     } else { // Supported by the Arch < 10.0
+         fp8_quantize_arch_l_100<IS_DBIAS, IS_DACT, IS_ACT, ParamOP, OP>(input, act_input, noop, output,
+                                                                        dbias, workspace, stream);
+     }
+ #else
+     // AMD
+    fp8_quantize_amd<IS_DBIAS, IS_DACT, IS_ACT, ParamOP, OP>(input, act_input, noop, output,
```
alextmagro: I think we should rely on the fp8_quantize_arch_l_100 function instead, and use CastVectorizedUnaryGradKernelLauncher instead of creating new Launchers.

transformer_engine/common/util/vectorized_pointwise.h
```diff
@@ -451,6 +573,80 @@ __launch_bounds__(unary_kernel_threads) __global__
...
+    num_blocks = std::min(num_blocks, max_blocks);
+
+    switch (align) {
+      case Alignment::SAME_ALIGNED:
+        cast_dbias_kernel<nvec, true, fp32, Param, OP><<<num_blocks, threads, 0, stream>>>(
+            input, output, dbias_workspace, scale, amax, scale_inv, params, N,
+            num_aligned_elements);
+        break;
+      case Alignment::SAME_UNALIGNED:
+        cast_dbias_kernel<nvec, false, fp32, Param, OP><<<num_blocks, threads, 0, stream>>>(
```
alextmagro: Instead of creating new cast_dbias_kernel and cast_dbias_dgelu_kernel, their new functionality can be fused into unary_kernel. You can use the IS_DBIAS / IS_DACT template parameters that come into fp8_quantize_arch_l_100, and pass them through the launcher to the kernel. Setting defaults values to false will keep you from needing to make too many changes to the rest of the code.

As for the additional inputs needed, grad and dbias, I think it is better to add them as parameters to unary kernel and just set their default value to nullptr unless they are needed and handle them accordingly. You can look at the way upstream cast_fp8_2d does it or cast_mxfp8_2D_kernel

transformer_engine/common/util/cast_kernels.cuh
AllenFarcas: For GPU with compute capability <10.0 they do not have it implemented.

AllenFarcas: In the test they skip them since they have
```
if (getDeviceComputeCapability() < blackwellComputeCapability) {
        GTEST_SKIP();
    }
```

transformer_engine/common/util/cast_kernels.cuh
AllenFarcas: The existing reduce_dbias kernel is hardcoded to take a const float* workspace as its input. It assumes the data is already in FP32. Is this always the case? I was thinking maybe we could end up having FP16 for example in which case we should also modify reduce_dbias and reduce_dbias_kernel to accept any input type and up-cast correspondingly inside the kernel before the summation.

transformer_engine/common/util/vectorized_pointwise.h
```diff
@@ -279,6 +281,126 @@ __launch_bounds__(unary_kernel_threads) __global__
...
+  for (size_t tid = blockIdx.x * blockDim.x + threadIdx.x; tid < M; tid += gridDim.x * blockDim.x) {
+    loader.load(tid, N);
+#pragma unroll
+    for (int i = 0; i < nvec; ++i) {
+      const ComputeType val = static_cast<ComputeType>(loader.separate()[i]);
+      // For cast+dbias, OP is just identity, but we keep it for potential future fusions.
+      ComputeType temp = OP(val, p);
+
+      // Write pre-scaled value to workspace for reduction
+      dbias_storer.separate()[i] = temp;
```
alextmagro: This is the line you would add to unary_kernel when IS_DBIAS = true

transformer_engine/common/util/vectorized_pointwise.h
```diff
@@ -279,6 +281,126 @@ __launch_bounds__(unary_kernel_threads) __global__
...
+  for (size_t tid = blockIdx.x * blockDim.x + threadIdx.x; tid < M; tid += gridDim.x * blockDim.x) {
+    grad_loader.load(tid, N);
+    input_loader.load(tid, N);
+#pragma unroll
+    for (int i = 0; i < nvec; ++i) {
+      const ComputeType grad_val = static_cast<ComputeType>(grad_loader.separate()[i]);
+      const ComputeType input_val = static_cast<ComputeType>(input_loader.separate()[i]);
+
+      // The core dgelu fusion logic
+      const ComputeType elt = grad_val * OP(input_val, p);
```
alextmagro: This is the logic you would add to unary_kernel when IS_DACT = true

transformer_engine/common/util/vectorized_pointwise.h
AllenFarcas: My reasoning is that the clarity of the unary kernel would be degraded and the modified signature would be bloated with pointers that are often unused.

@ipanfilo @wangye805 what do you think about this?

transformer_engine/common/util/cast_kernels.cuh
AllenFarcas: The cast_dbias kernel doesn't perform the final reduction since it performs the first half of the fusion by preparing an intermediate workspace. The reduce_dbias kernel is then needed to complete the second half.

transformer_engine/common/util/cast_kernels.cuh
AllenFarcas: They use two kernel launchers: one for cast (i.e., CastVectorizedUnaryKernelLauncher) and the other for unary kernel (i.e., VectorizedUnaryKernelLauncher):

```
VectorizedUnaryKernelLauncher<nvec, ParamOP, UnaryOP>(
                reinterpret_cast<const IType *>(input.data.dptr),
                reinterpret_cast<const fp32 *>(noop->data.dptr),
                reinterpret_cast<OType *>(output->data.dptr),
                reinterpret_cast<const fp32 *>(output->scale.dptr),
                reinterpret_cast<fp32 *>(output->amax.dptr),
                reinterpret_cast<fp32 *>(output->scale_inv.dptr), N, {}, stream);
```

I used directly a call to the dbias kernel launcher so it is more straightforward.

```
VectorizedCastAndDBiasKernelLauncher<nvec, ParamOP, detail::identity>(
                      reinterpret_cast<const IType *>(input.data.dptr),
                      reinterpret_cast<OType *>(output->data.dptr),
                      reinterpret_cast<fp32 *>(workspace->data.dptr),
                      reinterpret_cast<const fp32 *>(output->scale.dptr),
                      reinterpret_cast<fp32 *>(output->amax.dptr),
                      reinterpret_cast<fp32 *>(output->scale_inv.dptr), N, {}, stream);
```

transformer_engine/common/util/cast_kernels.cuh
AllenFarcas: We leave it as NVIDIA specific as l_100 means for lower than 10.0 compute capability GPUs and to avoid IFU issues down the line, i.e., minimize impact on existing NVIDIA code and isolate big changes in code in our own files added to TE.

tests/cpp/operator/test_cast_dbias_dgelu.cu
```diff
@@ -164,10 +164,13 @@ class CastDBiasDGeluTestSuite : public ::testing::TestWithParam<std::tuple<trans
 TEST_P(CastDBiasDGeluTestSuite, TestCastDBiasDgelu) {
     using namespace transformer_engine;
     using namespace test;
-    // Skip tests for pre-Blackwell architectures
-    if (getDeviceComputeCapability() < blackwellComputeCapability) {
-        GTEST_SKIP();
-    }
+
+#ifndef __HIP_PLATFORM_AMD__
```
ipanfilo, line 168: Indent

transformer_engine/common/util/vectorized_pointwise.h
AllenFarcas: Moved to transformer_engine/common/util/rocm_cast_kernels.cuh per Alex's review.

tests/cpp/operator/test_normalization.cu
```diff
@@ -35,9 +35,11 @@ void performTest(const size_t N, const size_t H, const bool zero_centered_gamma,
     return;
   }
 
-  if (getDeviceComputeCapability() < blackwellComputeCapability && use_cudnn) {
-    GTEST_SKIP() << "cuDNN normalizations not supported on pre-Blackwell GPUs yet!";
-  }
+#ifndef __HIP_PLATFORM_AMD__
```
ipanfilo, line 38: Indent

transformer_engine/common/util/rocm_cast_kernels.cuh
```diff
@@ -385,4 +385,197 @@ __global__ void __launch_bounds__(MXFP8_THREADS_PER_CHUNK)
...
+    }
+    if (blockIdx.x == 0 && threadIdx.x == 0 && scale_inv != nullptr) {
+      reciprocal<ComputeType>(scale_inv, s);
+    }
+  }
+}
+
+template <int nvec, typename Param, fp32 (*OP)(const fp32, const Param &), typename InputType,
+          typename OutputType>
+void VectorizedCastAndDBiasKernelLauncher(const InputType *input, OutputType *output,
```
ipanfilo: Should it be just ...CastDBias..., w/o And?

tests/cpp/operator/test_cast_float8blockwise.cu
```diff
@@ -478,9 +478,11 @@ class FusedCastFloat8VectorwiseTestSuite
   }
 
 TEST_P(FusedCastFloat8BlockwiseTestSuite, TestFusedCastFloat8Blockwise) {
-  if (getDeviceComputeCapability() < hopperComputeCapability) {
-    GTEST_SKIP();
-  }
+#ifndef __HIP_PLATFORM_AMD__
```
ipanfilo, line 481: Inent

tests/cpp/operator/test_cast_float8blockwise.cu
```diff
@@ -529,9 +531,11 @@ TEST_P(FusedCastFloat8BlockwiseTestSuite, TestFusedCastFloat8Blockwise) {
 }
 
 TEST_P(FusedCastFloat8VectorwiseTestSuite, TestFusedCastFloat8Vectorwise) {
-  if (getDeviceComputeCapability() < hopperComputeCapability) {
-    GTEST_SKIP();
-  }
+#ifndef __HIP_PLATFORM_AMD__
```
ipanfilo, line 534: Indent

transformer_engine/common/util/cast_kernels.cuh
ipanfilo: Yes, this method should be moved rocm_cast_kernels.cuh and fp8_quantize_arch_l_100 moved to #ifndef __HIP_PLATFORM_AMD__ (above line 1204).

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -1228,6 +1228,79 @@ void fp8_quantize_arch_l_100(const Tensor &input, const Tensor *act_input, const
...
 }
 
+// ROCm FP8 quantize
+template <bool IS_DBIAS, bool IS_DACT, bool IS_ACT, typename ParamOP,
+          float (*OP)(float, const ParamOP &)>
+void fp8_quantize_amd(const Tensor &input, const Tensor *act_input, const Tensor *noop,
+                      Tensor *output, Tensor *dbias, Tensor *workspace,
+                      cudaStream_t stream) {
+  switch (output->scaling_mode) {
+    case NVTE_DELAYED_TENSOR_SCALING: {
```
ipanfilo: Scaling mode list is inherited from NV Arch<10 that support only delayed tensor scaling but AMD GPUs also support MXFP8. There should be branch for it too - mxfp8_quantize() might work.
And more generic is if amd function can be based on ge_100 variant and use the same kernels

tests/cpp/operator/CMakeLists.txt
```diff
@@ -29,9 +31,7 @@ list(APPEND test_cuda_sources
 	          ../test_common.cu)
 if(USE_CUDA)
   list(APPEND test_cuda_sources
-              test_cast_float8blockwise.cu
```
wenchenvincent, line 32: why was `test_cast_float8blockwise.cu` also removed?

tests/cpp/operator/test_cast_dbias.cu
```diff
@@ -149,10 +149,12 @@ class CastDBiasTestSuite : public ::testing::TestWithParam<std::tuple<transforme
 TEST_P(CastDBiasTestSuite, TestCastDBias) {
     using namespace transformer_engine;
     using namespace test;
+#ifndef __HIP_PLATFORM_AMD__
```
wenchenvincent, line 152: Do we run the test on both gfx942 and gfx950?

transformer_engine/common/util/cast_kernels.cuh
wenchenvincent: @alextmagro `fp8_quantize_arch_l_100` does not support dbias. Do you actually mean following the style in `fp8_quantize_arch_l_100` but adding the support for dbias?

transformer_engine/common/util/cast_kernels.cuh
wenchenvincent: Okay. I thought you fused the reduction into the kernel. Looking closely, the first kernel seems to only cast the bias data to fp32 and then `reduce_dbias` function reduce it. I think we can just reuse `VectorizedCastKernelLauncher` for the first part and then have a `reduce_dbias` function that could deal with input that is not necessarily fp32.

tests/cpp/operator/CMakeLists.txt
alextmagro: This is from the IFU, we don't support fp8 blockwise scaling as of now

transformer_engine/common/util/cast_kernels.cuh
alextmagro: I meant adding IS_DBIAS as an optional template parameter, then adding IS_DBIAS as a template parameter to downstream functions like CastVectorizedUnaryKernelLauncher, CastVectorizedUnaryGradKernelLauncher, and the unary_kernels. That way we don't need any new kernels or launchers, and can just add the dbias calculation directly into unary_kernel and unary_grad_kernel.

tests/cpp/operator/CMakeLists.txt
AllenFarcas: Don't know why github showed this as removed since it is still there. I only removed the test_cast_dbias.cu and test_cast_dbias_dgelu.cu ones.

```
if(USE_CUDA)
  list(APPEND test_cuda_sources
              test_cast_float8blockwise.cu)
```

ipanfilo: Because test_cast_float8blockwise.cu is removed and test_cast_float8blockwise.cu) is added

transformer_engine/common/util/cast_kernels.cuh
AllenFarcas: Implemented

transformer_engine/common/util/rocm_cast_kernels.cuh
AllenFarcas: Implemented

transformer_engine/common/util/vectorized_pointwise.h
wangye805: Currently your newly added cast_dbias_kernel looks quite similar to the unary_kernel (https://github.com/ROCm/TransformerEngine/blob/0cf1b0adb76724e6217a66ac3d40dbafdd35d81a/transformer_engine/common/util/vectorized_pointwise.h#L177) and the VectorizedCastDBiasKernelLauncher looks quite like VectorizedUnaryKernelLauncher (https://github.com/ROCm/TransformerEngine/blob/0cf1b0adb76724e6217a66ac3d40dbafdd35d81a/transformer_engine/common/util/vectorized_pointwise.h#L334). Similar relationships apply from cast_dbias_dact_kernel to unary_grad_kernel, and from VectorizedCastDBiasDActKernelLauncher to  VectorizedUnaryGradKernelLauncher. 

I think we should merge those similar kernels and launchers either by using template parameters or default arguments. Having almost identical code copies increase our future maintenance burden.

transformer_engine/common/util/cast_kernels.cuh
wangye805: Since we are not enabling this cast_dbias and cast_dbias_dgelu in e2e training in TE pytorch/jax extension, we probably would be fine to stay in the fp8_quantize_arch_l_100 as MI300 is targeting at the NV H100. 

However, one follow-up question: on MI350, are we still running through this fp8_quantize_arch_l_100? If so, that's not going to be correct. We should run similar flows as NV B200 through fp8_quantize_arch_ge_100?

tests/cpp/operator/test_cast_dbias.cu
AllenFarcas: I did run and it works. Will double check the latest commit runs on gfx942 as well.

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -826,6 +826,69 @@ void reduce_dbias(const float *workspace_ptr, Tensor *dbias, const size_t rows,
           reinterpret_cast<IType *>(dbias->data.dptr), workspace_ptr, rows, cols);
 }
 
+#ifdef __HIP_PLATFORM_AMD__
+constexpr size_t TILE_DIM = 32;
```
alextmagro: This function should be moved to rocm_cast_kernels.cuh.

Also, did you have a try with TILE_DIM = 64? I have found that to be slightly faster than 32, at least for casts.

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -826,6 +826,69 @@ void reduce_dbias(const float *workspace_ptr, Tensor *dbias, const size_t rows,
...
+    __syncthreads();
+  }
+
+  if (thread_row_in_tile == 0 && global_col < cols) {
+    partial_output[blockIdx.y * cols + global_col] = tile[0][thread_col_in_tile];
+  }
+}
+
+template <typename IType>
+void reduce_dbias_rocm(const float *workspace_ptr, Tensor *dbias, const size_t rows, const size_t cols,
```
alextmagro: This function should go to rocm_cast_kernels.cuh as well.

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -826,6 +826,69 @@ void reduce_dbias(const float *workspace_ptr, Tensor *dbias, const size_t rows,
...
+                  cudaStream_t stream) {
+    dim3 block_dim_partial(TILE_DIM, TILE_DIM);
+    dim3 grid_dim_partial(DIVUP(cols, TILE_DIM), DIVUP(rows, TILE_DIM));
+
+    const size_t partial_rows = grid_dim_partial.y;
+    float* partial_workspace;
+    cudaMalloc(&partial_workspace, partial_rows * cols * sizeof(float));
```
alextmagro: I think we have TE wrappers for memory allocation that do error handling, can you double check that?

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -1227,6 +1289,75 @@ void fp8_quantize_arch_l_100(const Tensor &input, const Tensor *act_input, const
       NVTE_ERROR("Not implemented scaling mode: " + to_string(output->scaling_mode) + ".");
   }
 }
+#else
+template <bool IS_DBIAS, bool IS_DACT, bool IS_ACT, typename ParamOP,
+          float (*OP)(float, const ParamOP &)>
+void fp8_quantize_rocm(const Tensor &input, const Tensor *act_input, const Tensor *noop,
```
alextmagro: Move this to rocm_cast_kernels.cuh as well

alextmagro (commented):
mostly LGTM, just need to move some functions to rocm_cast_kernels.cuh for licensing. Great work!

transformer_engine/common/util/cast_kernels.cuh
AllenFarcas: This one cannot be moved as it calls `CastVectorizedUnaryKernelLauncher` and `CastVectorizedUnaryGradKernelLauncher` and would mean we need a circular import as `cast_kernels.cuh` already imports `rocm_cast_kernels.cuh`

transformer_engine/common/util/cast_kernels.cuh
AllenFarcas: Added HIP_CHECK(hipMalloc(...)); and also for the free part did find HIP_CHECK(hipFree(...));

transformer_engine/common/util/cast_kernels.cuh
AllenFarcas: I can try using TILE_DIM=64. 
If I move it to `rocm_cast_kernels.cuh` then I would have to redeclare TILE_DIM there and it should coincide with the value assigned in `reduce_dbias_rocm` as it allocates the space for the partial reduction kernel.

transformer_engine/common/util/cast_kernels.cuh
AllenFarcas: This launches `reduce_dbias_kernel` from `cast_kernels.cuh`. It would be a circular import.

transformer_engine/common/util/cast_kernels.cuh
AllenFarcas: Had to also add the definition of HIP_CHECK as it was implemented in some 3rdparty/aiter header files.

tests/cpp/operator/test_cast_dbias.cu
AllenFarcas: Update: I just tested the latest commit from this PR on gfx942 and it works and the tests pass successfully.

transformer_engine/common/util/cast_kernels.cuh
AllenFarcas: Update: did try TILE_DIM=64 but had errors for the HIP_CHECK with error code 9.

tests/cpp/operator/test_cast_float8blockwise.cu
```diff
@@ -478,9 +478,11 @@ class FusedCastFloat8VectorwiseTestSuite
   }
```
wenchenvincent, line 478: Is this test file enabled or not?

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -1227,6 +1298,75 @@ void fp8_quantize_arch_l_100(const Tensor &input, const Tensor *act_input, const
...
+        return;
+      }
+
+      NVTE_CHECK(dbias, "DBias tensor must be provided when IS_DBIAS is true.");
+      NVTE_CHECK(workspace, "Workspace must be provided when IS_DBIAS is true.");
+
+      if (workspace->data.dptr == nullptr ||
+          workspace->data.dtype != DType::kFloat32 ||
+          workspace->data.shape != std::vector<size_t>{rows, cols}) {
+        workspace->data.shape = {rows, cols};
```
wenchenvincent: Why the workspace shape is still `{rows, cols}`? It is used to store the partial sum, so it should be smaller.

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -1227,6 +1298,75 @@ void fp8_quantize_arch_l_100(const Tensor &input, const Tensor *act_input, const
...
+        workspace->data.dtype = DType::kFloat32;
+        return;
+      }
+
+      workspace->amax = {};
+      workspace->scale = {};
+      workspace->scale_inv = {};
+
+      if constexpr (IS_DACT) {
```
wenchenvincent: Why do we need this again?
The reduce_bias_rocm function should take the input, and the workspace would be used to store the partial sum.

wenchenvincent: So for dgelu+dbias, you do need to call `CastVectorizedUnaryGradKernelLauncher` unfortunately. However, for dbias, you don't need to call `CastVectorizedUnaryKernelLauncher` again.

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -31,6 +31,15 @@
 #include "transformer_engine/transformer_engine.h"
 #ifdef __HIP_PLATFORM_AMD__
 #include "rocm_cast_kernels.cuh"
+#define HIP_CHECK(err)                                                     \
```
ipanfilo: Looks like duplicate of NVTE_CHECK_CUDA()

transformer_engine/common/util/cast_kernels.cuh
ipanfilo: Add reduce_dbias_kernel forward declaration to rocm_cast_kernels.cuh?

tests/cpp/operator/test_cast_float8blockwise.cu
AllenFarcas: This test is not enabled for ROCm. The CMakeFile.txt has it enabled only for NVIDIA

```
if(USE_CUDA)
  list(APPEND test_cuda_sources
              test_cast_float8blockwise.cu)
```

transformer_engine/common/util/cast_kernels.cuh
AllenFarcas: This workspace contains the full matrix of values to be reduced. It gets passed to `reduce_dbias_rocm` where a smaller, partial workspace is allocated and used in the final reduction.

transformer_engine/common/util/rocm_cast_kernels.cuh
```diff
@@ -385,4 +385,140 @@ __global__ void __launch_bounds__(MXFP8_THREADS_PER_CHUNK)
...
+void fp8_quantize_rocm(const Tensor &input, const Tensor *act_input, const Tensor *noop,
+                             Tensor *output, Tensor *dbias, Tensor *workspace,
+                             cudaStream_t stream) {
+  switch (output->scaling_mode) {
+    case NVTE_DELAYED_TENSOR_SCALING: {
+      const size_t rows = input.flat_first_dim();
+      const size_t cols = input.flat_last_dim();
+
+      if (output && output->data.dptr) {
+        if constexpr (IS_DACT) {
```
wangye805: Could IS_DACT and IS_DBIAS both true? If so, will CastVectorizedUnaryGradKernelLauncher be run twice? One in Line 463, one in Line 491?

transformer_engine/common/util/rocm_cast_kernels.cuh
```diff
@@ -385,4 +385,140 @@ __global__ void __launch_bounds__(MXFP8_THREADS_PER_CHUNK)
...
+
+        NVTE_CHECK(dbias, "DBias tensor must be provided when IS_DBIAS is true.");
+        NVTE_CHECK(workspace, "Workspace must be provided when IS_DBIAS is true.");
+
+        if (workspace->data.dptr == nullptr ||
+            workspace->data.dtype != DType::kFloat32 ||
+            workspace->data.shape != std::vector<size_t>{rows, cols}) {
+          workspace->data.shape = {rows, cols};
+          workspace->data.dtype = DType::kFloat32;
+          return;
```
wangye805, line 473: Is our cast_dbias_dgelu following NV's two pass pattern? In other words, the first pass calculates the configs and required workspace, then upstream pytorch/jax allocates the required workspace, and the second pass run the actual kernel with given workspace? 

If so, Line 462-465 will be run twice in both passes. Is this expected?

transformer_engine/common/util/rocm_cast_kernels.cuh
```diff
@@ -385,4 +385,140 @@ __global__ void __launch_bounds__(MXFP8_THREADS_PER_CHUNK)
...
+template <typename DTypeReduce, typename DBiasTypeOut>
+void reduce_dbias_rocm(const DTypeReduce *workspace_ptr, Tensor *dbias, const size_t rows,
+                       const size_t cols, cudaStream_t stream, Tensor* partial_sum_workspace) {
+  dim3 block_dim_partial(TILE_DIM, TILE_DIM);
+  dim3 grid_dim_partial(DIVUP(cols, TILE_DIM), DIVUP(rows, TILE_DIM));
+
+  const size_t partial_rows = grid_dim_partial.y;
+  float* partial_workspace = reinterpret_cast<float*>(partial_sum_workspace->data.dptr);
+
+  partial_reduce_kernel<DTypeReduce><<<grid_dim_partial, block_dim_partial, 0, stream>>>(
```
wangye805, line 442: Do we see performance improvement using this two-dimensional reduction, compared to a direct vectorized reduction

wangye805 (changes requested):

transformer_engine/common/util/rocm_cast_kernels.cuh
AllenFarcas: This was the discussed and prefered approach especially for large input sizes, where using one thread per column reduction could take very long for certain sizes. The current approach over 5 runs runs all the tests for DBias and DBiasDGelu 150-200ms faster so there was a reduction on avg. Did not test the large size shapes in isolation.

transformer_engine/common/util/rocm_cast_kernels.cuh
AllenFarcas: You are right, the second execution, after the query would run again the same code. Fixed in the latest commit.

transformer_engine/common/util/rocm_cast_kernels.cuh
AllenFarcas: Yes, both IS_DACT and IS_DBIAS can be true and the code is made to run twice: once for computing dAct and writing the results in the output tensor, and the second time to compute the dAct and write the result to the workspace for reduction. This is to reuse existing code, so make reuse of existing CastVectorizedUnaryGradKernelLauncher  without implementing new fused kernels. 

I have looked again and thought of a better way to use this. Updated in the latest commit. 

We can make one call for the CastVectorizedUnaryGradKernelLauncher to get the workspace and then use for output the CastVectorizedUnaryKernelLauncher so it takes the workspace and writes the output quantized if needed. The only other alternative is to implement a new fully fused kernel for high performance, but for now we discussed this implementation should be ok.

wangye805 (approved):

wangye805:
Please also get LGTMs from Wen and Ilya before merging. Thanks

wenchenvincent (approved):
LGTM. Let's wait for CI to complete before merging.

wenchenvincent (changes requested):
@AllenFarcas There are some core sgpu test failures. Could you make sure that they're fixed?

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -1251,17 +1251,19 @@ void fp8_quantize(const Tensor &input, const Tensor *act_input, const Tensor *no
...
-    fp8_quantize_arch_ge_100<IS_DBIAS, IS_DACT, IS_ACT, ParamOP, OP>(input, act_input, noop, output,
-                                                                     dbias, workspace, stream);
-  } else {
-#endif //#ifndef __HIP_PLATFORM_AMD__
-    // Supported by the Arch < 10.0
-    fp8_quantize_arch_l_100<IS_DBIAS, IS_DACT, IS_ACT, ParamOP, OP>(input, act_input, noop, output,
-                                                                    dbias, workspace, stream);
-#ifndef __HIP_PLATFORM_AMD__
-  }
+   // NVIDIA
```
ipanfilo: Those lines till //AMD seem just space change, Please keep indentation like in original

ipanfilo (approved):

wenchenvincent (approved):
LGTM.

PR #335, opened 2025-10-15
[TE][AITER] Add prebuilt AITER download and fallback-to-source build …

VeeraRajasekhar:
# Description

This PR introduces a new build flow to support AITER prebuilts for the fused attention component in Transformer Engine.
The new logic enables CMake to automatically download prebuilt AITER shared objects (libmha_fwd.so, libmha_bwd.so) from Artifactory based on the ROCm version, GPU architecture, and AITER commit hash.

If the requested prebuilt is unavailable (for new ROCm or AITER versions), the build gracefully falls back to compiling AITER from source.

Fixes # https://github.com/ROCm/frameworks-internal/issues/13882

## Type of change

- [X] Infra/Build change
- [ ] Code refactoring

## Changes

- Added download_aiter_prebuilt.cmake for automated prebuilt fetching, checksum verification, and cache management
- Updated ck_fused_attn/CMakeLists.txt to integrate the new prebuilt flow and handle fallback to source build
- Introduced build_tools/upload_aiter_prebuilt.sh for Artifactory upload of prebuilts.
- Updated setup.py to integrate new ENV variable.

# Checklist:


build_tools/upload_aiter_prebuilt.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/sh
+# Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
+#
+# See LICENSE for license information
+
+if [ -z "${NVTE_PREBUILT_BASE_URL:-}" ]; then
```
ipanfilo: Why is var substitution needed?

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
ipanfilo, line 1: I think CMake ExternalProject and/or FetchContent incorporate functionality needed to download binaries

build_tools/upload_aiter_prebuilt.sh
ipanfilo, line 1: It looks like it is not needed for building

setup.py
```diff
@@ -69,6 +69,8 @@ def setup_common_extension() -> CMakeExtension:
         if os.getenv("NVTE_CK_FUSED_ATTN_PATH"):
             ck_path = Path(os.getenv("NVTE_CK_FUSED_ATTN_PATH"))
             cmake_flags.append(f"-DAITER_MHA_PATH={ck_path}")
+        elif bool(int(os.getenv("NVTE_USE_PREBUILT_AITER", "0"))):
+                cmake_flags.append("-DUSE_PREBUILT_AITER=ON")
```
ipanfilo: This parameter is redundant - env variables are checked in CMakeLists.txt

transformer_engine/common/ck_fused_attn/download_aiter_prebuilt.cmake
```diff
@@ -0,0 +1,163 @@
...
+  if("${ROCM_PATH}" STREQUAL "")
+    set(ROCM_PATH "/opt/rocm")
+  endif()
+  file(READ "${ROCM_PATH}/.info/version" ROCM_VER_CONTENT)
+  string(STRIP "${ROCM_VER_CONTENT}" ROCM_VER)
+  execute_process(
+    COMMAND git -C "${TE_ROOT}/3rdparty/aiter" rev-parse --short=9 HEAD
+    OUTPUT_VARIABLE AITER_SHA OUTPUT_STRIP_TRAILING_WHITESPACE
+  )
+  if(DEFINED ENV{NVTE_ROCM_ARCH})
```
ipanfilo: Using archs as key makes more problems - single vs multiple, order, etc.
We need only upload AITER binaries built for all supported archs and use them on any system

build_tools/upload_aiter_prebuilt.sh
VeeraRajasekhar: @ipanfilo Where do you suggest is better place to place this script? in TE root dir?

setup.py
VeeraRajasekhar: Removed and checked in CMakeLists

transformer_engine/common/ck_fused_attn/download_aiter_prebuilt.cmake
VeeraRajasekhar: Removed the archs key, only using rocm version & aiter sha

build_tools/upload_aiter_prebuilt.sh
VeeraRajasekhar: removed the substitution

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
VeeraRajasekhar: Updated

build_tools/upload_aiter_prebuilt.sh
VeeraRajasekhar: Currently moved to ci directory

ipanfilo: Build_tools location is OK, I just doubt if we need upload script as a part of tree. If we keep it, it should be more generic, definitely w/o reference to artifactory

build_tools/upload_aiter_prebuilt.sh
```diff
@@ -0,0 +1,52 @@
...
+if [ -z "${NVTE_ARTIFACTORY_USER}" ] || [ -z "${NVTE_ARTIFACTORY_PASSWORD}" ]; then
+  echo "[AITER-PREBUILT] [Error] To fetch .so, set NVTE_ARTIFACTORY_USER & NVTE_ARTIFACTORY_PASSWORD" >&2
+  exit 1
+fi
+CURL_AUTH="-u ${NVTE_ARTIFACTORY_USER}:${NVTE_ARTIFACTORY_PASSWORD}"
+
+# Derive keys for cache/artifact naming
+TE_ROOT="$(git rev-parse --show-toplevel)"
+ROCM_PATH="${ROCM_PATH:-/opt/rocm}"
+ROCM_VER="$(cut -d . -f1,2,3 < "${ROCM_PATH}/.info/version")"
```
ipanfilo: TE uses major.minor only as ROCm version.

ci/upload_aiter_prebuilt.sh
```diff
@@ -0,0 +1,52 @@
...
+ROCM_VER="$(cut -d . -f1,2,3 < "${ROCM_PATH}/.info/version")"
+AITER_SHA="$(git -C "${TE_ROOT}/3rdparty/aiter" rev-parse HEAD)"
+
+# Artifact key + remote path
+KEY="rocm-${ROCM_VER}_aiter-${AITER_SHA}"
+REMOTE_PATH="aiter-prebuilts/${KEY}.tar.gz"
+ARTIFACT_URL="${NVTE_PREBUILT_BASE_URL%/}/${REMOTE_PATH}"
+
+# Local cache
+CACHE_ROOT="${TE_ROOT}/.cache/aiter-prebuilts"
```
ipanfilo: Why not using temp directory?

transformer_engine/common/ck_fused_attn/download_aiter_prebuilt.cmake
```diff
@@ -0,0 +1,95 @@
...
+function(require_env_var VAR)
+  if(NOT DEFINED ENV{${VAR}} OR "$ENV{${VAR}}" STREQUAL "")
+    message(SEND_ERROR " [AITER-PREBUILT] Environment variable ${VAR} must be set.")
+    return()
+  endif()
+endfunction()
+
+function(fetch_aiter_prebuilt __AITER_MHA_PATH_VAR)
+  # Required env vars
+  require_env_var("NVTE_PREBUILT_BASE_URL")
```
ipanfilo: Let's rename It is not NVTE_PREBUILT but NVTE_AITER_PREBUILT. Also, can user and password be part of URL or they have to be separate variables? Actuallydownload does not necessary require authentication

transformer_engine/common/ck_fused_attn/download_aiter_prebuilt.cmake
```diff
@@ -0,0 +1,95 @@
...
+  execute_process(
+    COMMAND git -C "${TE_ROOT}/3rdparty/aiter" rev-parse HEAD
+    OUTPUT_VARIABLE AITER_SHA OUTPUT_STRIP_TRAILING_WHITESPACE
+  )
+
+  set(KEY "rocm-${ROCM_VER}_aiter-${AITER_SHA}")
+  set(BASE_URL "$ENV{NVTE_PREBUILT_BASE_URL}")
+  string(REGEX REPLACE "/$" "" BASE_URL "${BASE_URL}")
+  set(ARTIFACT_URL "${BASE_URL}/aiter-prebuilts/${KEY}.tar.gz")
+  set(SHA_URL      "${BASE_URL}.sha256")
```
ipanfilo: Aren't sha files one per tgz?

transformer_engine/common/ck_fused_attn/download_aiter_prebuilt.cmake
```diff
@@ -0,0 +1,95 @@
...
+
+  # Prepare local cache
+  set(CACHE_ROOT "${TE_ROOT}/.cache/aiter-prebuilts")
```
ipanfilo: Download to build directory?

build_tools/upload_aiter_prebuilt.sh
VeeraRajasekhar: But does 7.0.0 & 7.0.2 make difference on the build files generated? if not, I will just use major.minor

ci/upload_aiter_prebuilt.sh
VeeraRajasekhar: I can use that, will change to using /tmp

transformer_engine/common/ck_fused_attn/download_aiter_prebuilt.cmake
VeeraRajasekhar: Will check if download requires authentication or not?  and remove the necessity for username and password for downloading

transformer_engine/common/ck_fused_attn/download_aiter_prebuilt.cmake
VeeraRajasekhar: yes, they are

transformer_engine/common/ck_fused_attn/download_aiter_prebuilt.cmake
VeeraRajasekhar: should I keep under build/aiter-prebuilts ?

transformer_engine/common/ck_fused_attn/download_aiter_prebuilt.cmake
VeeraRajasekhar: verified that download does not require authentication from our artifactory. changes the download cmake file.

VeeraRajasekhar: @ipanfilo And for the credentials, I got access token to upload to this repository, so I can update the script to take authorization token or I can keep the current username password?

build_tools/upload_aiter_prebuilt.sh
VeeraRajasekhar: Updated the scripts and removed reference to artifactory

ci/upload_aiter_prebuilt.sh
VeeraRajasekhar: Resolved, updated both the scripts to use tmp directories wherever required

transformer_engine/common/ck_fused_attn/download_aiter_prebuilt.cmake
VeeraRajasekhar: Resolved: Currently updated to build/aiter-prebuilts.

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -0,0 +1,141 @@
...
+cmake_minimum_required(VERSION 3.21)
+include(FetchContent)
+
+# Base URL is mandatory
+if(NOT DEFINED ENV{NVTE_AITER_PREBUILT_BASE_URL} OR "$ENV{NVTE_AITER_PREBUILT_BASE_URL}" STREQUAL "")
+  message(SEND_ERROR " [AITER-PREBUILT] ENV variable NVTE_AITER_PREBUILT_BASE_URL must be set.")
+  return()
+endif()
+
+execute_process(
```
ipanfilo: For CMake we know where TE reee root is, no need to use git for this

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -0,0 +1,141 @@
...
+    COMMAND git -C "${TE_ROOT}/3rdparty/aiter" rev-parse HEAD
+    OUTPUT_VARIABLE AITER_SHA OUTPUT_STRIP_TRAILING_WHITESPACE
+  )
+
+set(KEY "rocm-${ROCM_VER}_aiter-${AITER_SHA}")
+set(BASE_URL "$ENV{NVTE_AITER_PREBUILT_BASE_URL}")
+string(REGEX REPLACE "/$" "" BASE_URL "${BASE_URL}")
+set(FILE_URL "${BASE_URL}/aiter-prebuilts/${KEY}.tar.gz")
```
ipanfilo: Why not make aiter-prebuilts part of URL?

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -0,0 +1,141 @@
...
+
+set(CACHE_ROOT "${TE_ROOT}/build/aiter-prebuilts")
```
ipanfilo: Use CMAKE_CURRENT_BINARY_DIRE

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -0,0 +1,141 @@
...
+  # Try local cache first
+  if(EXISTS "${EXTRACT_DIR}/libmha_fwd.so" AND EXISTS "${EXTRACT_DIR}/libmha_fwd.so")
+    message(STATUS "[AITER-PREBUILT] ${EXTRACT_DIR} already exists. Skipping download.")
+    set(${__AITER_MHA_PATH_VAR} "${EXTRACT_DIR}" PARENT_SCOPE)
+    message(STATUS "[AITER-PREBUILT] Using ${__AITER_MHA_PATH_VAR}='${EXTRACT_DIR}'")
+    return()
+  endif()
+
+  # Check if prebuilt files exist in the URL provided.
+  file(DOWNLOAD "${FILE_URL}.sha256" "/tmp/aiter_prebuilt_${KEY}.sha256" STATUS sha_status LOG sha_log)
```
ipanfilo: Where is SHA256 used? Looks like it is just downloaded and that is all

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -32,18 +32,37 @@ message(STATUS "AITER V3_ASM_ARCHS: ${V3_ASM_ARCHS}")
...
-  set(ENV{CK_TILE_FLOAT_TO_BFLOAT16_DEFAULT} ${CK_FUSED_ATTN_FLOAT_TO_BFLOAT16_DEFAULT})
-  execute_process(
-    COMMAND python3 ${__AITER_TEST_DIR}/compile.py
-  )
-  # libmha_fwd.so and libmha_bwd.so will be under 3rdparty/aiter/op_tests/cpp/mha
-  set(__AITER_MHA_PATH ${__AITER_TEST_DIR})
+
+  # Try to fetch; fallback to build-from-source if fails
+  if(DEFINED ENV{NVTE_USE_PREBUILT_AITER} AND NOT "$ENV{NVTE_USE_PREBUILT_AITER}" STREQUAL "0")
+    message(STATUS "[AITER-PREBUILT] NVTE_USE_PREBUILT_AITER is set, Attempting to download prebuilt files...")
```
ipanfilo: This message is duplicated in aiter_prebuilt.cmake

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -32,18 +32,37 @@ message(STATUS "AITER V3_ASM_ARCHS: ${V3_ASM_ARCHS}")
...
+    include("${CMAKE_CURRENT_LIST_DIR}/download_aiter_prebuilt.cmake")
```
ipanfilo: Invalid file name

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -32,18 +32,37 @@ message(STATUS "AITER V3_ASM_ARCHS: ${V3_ASM_ARCHS}")
...
+    set(fetch_success TRUE)
+    download_aiter_prebuilt(__AITER_MHA_PATH)
+
+    if(NOT DEFINED __AITER_MHA_PATH OR NOT EXISTS "${__AITER_MHA_PATH}/libmha_fwd.so" OR NOT EXISTS "${__AITER_MHA_PATH}/libmha_bwd.so")
```
ipanfilo: Shouldn't __AITER_MHA_PATH be only set if be only set if libmha*.so exist? Also, is the variable cached so it is initially set on subsequent cmake calls?

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -32,18 +32,37 @@ message(STATUS "AITER V3_ASM_ARCHS: ${V3_ASM_ARCHS}")
...
-    COMMAND python3 ${__AITER_TEST_DIR}/compile.py
-  )
-  # libmha_fwd.so and libmha_bwd.so will be under 3rdparty/aiter/op_tests/cpp/mha
-  set(__AITER_MHA_PATH ${__AITER_TEST_DIR})
+
+  # Try to fetch; fallback to build-from-source if fails
+  if(DEFINED ENV{NVTE_USE_PREBUILT_AITER} AND NOT "$ENV{NVTE_USE_PREBUILT_AITER}" STREQUAL "0")
+    message(STATUS "[AITER-PREBUILT] NVTE_USE_PREBUILT_AITER is set, Attempting to download prebuilt files...")
+    include("${CMAKE_CURRENT_LIST_DIR}/download_aiter_prebuilt.cmake")
+    set(fetch_success TRUE)
```
ipanfilo: Why need setting it here and it looks duplicating functionality of __AITER_MHA_PATH

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -0,0 +1,141 @@
...
+  if(upload_code EQUAL 0)
+    message(STATUS "[AITER-PREBUILT] Upload complete.")
+  else()
+    message(FATAL_ERROR "[AITER-PREBUILT] Upload failed.")
+  endif()
+
+  file(REMOVE_RECURSE ${TMP_DIR})
+endfunction()
+
+if (CMAKE_SCRIPT_MODE_FILE)
```
ipanfilo, line 119: Could you explain how does it work?

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -0,0 +1,141 @@
...
+  message(STATUS "[AITER-PREBUILT] Downloading ${KEY}.tar.gz ...")
+  FetchContent_MakeAvailable(aiter_prebuilt)
+  message(STATUS "[AITER-PREBUILT] Successfully downloaded to ${EXTRACT_DIR}")
+  set(${__AITER_MHA_PATH_VAR} "${EXTRACT_DIR}" PARENT_SCOPE)
+  message(STATUS "[AITER-PREBUILT] Using ${__AITER_MHA_PATH_VAR}='${EXTRACT_DIR}'")
+endfunction()
+
+# Upload prebuilt zip file
+function(upload_aiter_prebuilt)  
+  file(GLOB FWD_LIST
```
ipanfilo: Should it use __AITER_MHA_PATH?

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -0,0 +1,141 @@
...
+
+  # Locate .so files
+  if (NOT EXISTS  ${FWD})
+    message(FATAL_ERROR "[AITER-PREBUILT] Missing libmha_fwd.so")
+  endif()
+  if (NOT EXISTS  ${BWD})
+    message(FATAL_ERROR "[AITER-PREBUILT] Missing libmha_bwd.so")
+  endif()
+
+  set(TMP_DIR "/tmp/aiter-prebuilts")
```
ipanfilo: Can download cache directory be used?

ipanfilo (changes requested):

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
VeeraRajasekhar: This is to upload new aiter prebuilt if its commit is updated, as we decided to upload it manually, this allows us to call upload from terminal.

VeeraRajasekhar: @ipanfilo, I can actually make changes in the CMake script to automatically upload if there is a new aiter commit or rocm version detected, by removing the need to manually upload. Should I make this change?

ipanfilo: I don't think so. It is not always necessary, moreover only properly built all target binaries should be uploaded. Also, uploading requires dealing with authentication so the question is how to get token.
I like your idea with local cache more - it built binaries are moved there automatically, the next build loop will use them

wangye805: I agree with Ilya that we should not allow customers to easily upload their built libmha_*.so's to our prebuilt dir

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -32,18 +32,37 @@ message(STATUS "AITER V3_ASM_ARCHS: ${V3_ASM_ARCHS}")
...
+    file(REMOVE_RECURSE "${__AITER_SOURCE_DIR}/aiter/jit/build")
+    # compile the libmha_fwd.so and libmha_bwd.so
+    set(ENV{AITER_LOG_MORE} 1)
+    # fp32 to bf16 cvt env still required for MI300X
+    set(ENV{CK_TILE_FLOAT_TO_BFLOAT16_DEFAULT} ${CK_FUSED_ATTN_FLOAT_TO_BFLOAT16_DEFAULT})
+    execute_process(
+      COMMAND python3 ${__AITER_TEST_DIR}/compile.py
+    )
+    # libmha_fwd.so and libmha_bwd.so will be under 3rdparty/aiter/op_tests/cpp/mha
+    set(__AITER_MHA_PATH ${__AITER_TEST_DIR})
```
wangye805: Emm, can we enable the following case:
- This is a new aiter commit so no libmha_*.so matches in our online prebuilt repo 
- Our customers built it from the scratch so there is libmha_*.so stored in 3rdparty/aiter/op_tests/cpp/mha
- Then our customers make some changes to other parts of TE and do a simple pip install without setting NVTE_CK_FUSED_ATTN_PATH ?

In other words, can our cmake script remember where we built libmha_*.so previously and pick it up in subsequent incremental building?

transformer_engine/common/ck_fused_attn/CMakeLists.txt
wangye805: __AITER_MHA_PATH is not set in this branch due to line 36

wangye805 (changes requested):

VeeraRajasekhar:
https://github.com/ROCm/TransformerEngine/pull/335#discussion_r2455653510

Hey @wangye805, its not like they can upload it to our dir. I made the script more generic, so that they can upload and download from any place they stored, its like customers can give their own URL, which will be used by the script to download and upload.

Currently the expectations by the script is 
1. Download: URL should allow public downloads.
2. Upload: is restricted by access tokens, so customer should provide their access token for their URL.

In the same way, we will use our own URL and access token for our pre-built directory.

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
VeeraRajasekhar: This is a quick check if the required aiter-prebuilt file exists in the BASE_URL provided.

FetchContent_MakeAvailable will just fail to execute if the  aiter-prebuilt file does not exist in the URL, so I am checking by downlaoding the sha256 file and checking.

ipanfilo: I see. I think the downloaded checksum should be used in FetchContent then.

transformer_engine/common/ck_fused_attn/CMakeLists.txt
ipanfilo: I do not think direct usage of 3rdparty/aiter/op_tests/cpp/mha fits proposed scheme well. But if automatically copy  *.so to __AITER_MHA_PATH (EXTRACT_DIR) after building, it should work.
Saying that, do we need some flag to clear/ignore it then in case we have local AITER modifications and need rebuild it, or it is OK to remove EXTRACT_DIR manually?

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
VeeraRajasekhar: resolved

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
VeeraRajasekhar: added to URL

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
VeeraRajasekhar: resolved

transformer_engine/common/ck_fused_attn/CMakeLists.txt
VeeraRajasekhar: removed

transformer_engine/common/ck_fused_attn/CMakeLists.txt
VeeraRajasekhar: changed

transformer_engine/common/ck_fused_attn/CMakeLists.txt
VeeraRajasekhar: removed

transformer_engine/common/ck_fused_attn/CMakeLists.txt
VeeraRajasekhar: I now added a function to check if the generated KEY matched the downloaded key. i.e., if the local aiter commit changes, it will automatically clear the cache and try to download is NVTE_AITER_PREBUILT_BASE_URL is provided, if not it will fallback to building from source and cache this latest build.

transformer_engine/common/ck_fused_attn/CMakeLists.txt
VeeraRajasekhar: modified the flow

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
VeeraRajasekhar: Updated in the FetchContent

transformer_engine/common/ck_fused_attn/CMakeLists.txt
VeeraRajasekhar: addressed this

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -0,0 +1,167 @@
...
+  OUTPUT_VARIABLE AITER_SHA OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+
+# Cache key & local paths
+set(KEY "rocm-${ROCM_VER}_aiter-${AITER_SHA}")
+set(CACHE_ROOT "${CMAKE_SOURCE_DIR}/../../build/aiter-prebuilts")
+set(EXTRACT_DIR "${CACHE_ROOT}/${KEY}")
+
+# Initialize FILE_URL
+function(init_aiter_base_url )
```
ipanfilo: Remove empty method

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -0,0 +1,167 @@
...
+endfunction()
+
+# Download prebuilt zip file
+function(download_aiter_prebuilt DOWNLOAD_SUCCESS)
+  if(NOT DEFINED ENV{NVTE_AITER_PREBUILT_BASE_URL} OR "$ENV{NVTE_AITER_PREBUILT_BASE_URL}" STREQUAL "")
+    return()
+  endif()
+
+  set(BASE_URL "$ENV{NVTE_AITER_PREBUILT_BASE_URL}")
+  string(REGEX REPLACE "/$" "" BASE_URL "${BASE_URL}")
```
ipanfilo: Can ENV{NVTE_AITER_PREBUILT_BASE_URL} be used instead of {BASE_URL}?

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -0,0 +1,167 @@
...
+    URL "${FILE_URL}"
+    URL_HASH SHA256=${AITER_SHA_CONTENT}
+    SOURCE_DIR "${EXTRACT_DIR}"
+    DOWNLOAD_EXTRACT_TIMESTAMP FALSE
+  )
+
+  # Download & extract prebuilt files
+  FetchContent_MakeAvailable(aiter_prebuilt)
+  
+  if(EXISTS "${EXTRACT_DIR}/libmha_fwd.so" AND EXISTS "${EXTRACT_DIR}/libmha_bwd.so")
```
ipanfilo: If pre-built .tgz file does not exist or corrupted, or cannot be downloaded, it has to be caught by FileContent, and his check is for when everything is correct but pre-built package does not contain necessary files, right?
Let's not check it and let build fail then

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -0,0 +1,167 @@
...
+
+# AITER commit
+execute_process(
+  COMMAND git -C "${CMAKE_SOURCE_DIR}/../../3rdparty/aiter" rev-parse HEAD
+  OUTPUT_VARIABLE AITER_SHA OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+
+# Cache key & local paths
+set(KEY "rocm-${ROCM_VER}_aiter-${AITER_SHA}")
+set(CACHE_ROOT "${CMAKE_SOURCE_DIR}/../../build/aiter-prebuilts")
```
ipanfilo: It should be under CMAKE_BINARY_DIR.

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -32,21 +32,44 @@ message(STATUS "AITER V3_ASM_ARCHS: ${V3_ASM_ARCHS}")
...
-  set(__AITER_MHA_PATH ${__AITER_TEST_DIR})
-else()
+if(DEFINED AITER_MHA_PATH)
+  message(STATUS "[AITER-PREBUILT] Using AITER_MHA_PATH=${AITER_MHA_PATH}")
   # use pre-built libmha_fwd.so libmha_bwd.so
-  set(__AITER_MHA_PATH ${AITER_MHA_PATH})
+  set(__AITER_MHA_PATH ${AITER_MHA_PATH} CACHE PATH "Cached AITER MHA path" FORCE)
+else()
+  set(AITER_CACHE_VALID FALSE)
+  set(AITER_PREBUILT_DOWNLOADED FALSE)
```
ipanfilo: Unused variable

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -32,21 +32,44 @@ message(STATUS "AITER V3_ASM_ARCHS: ${V3_ASM_ARCHS}")
...
+  include("${CMAKE_CURRENT_LIST_DIR}/aiter_prebuilt.cmake")
+
+  # If cached path already exists and is valid
+  if(DEFINED __AITER_MHA_PATH)
```
ipanfilo: Why is this condition needed? CACHE_ROOT is calculated based on build environment and tree state so it can be checked unconditionally. It will also allow to copy pre-built aiter from one local tree to another and make caching of __AITER_MHA_PATH unnecessary

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -32,21 +32,44 @@ message(STATUS "AITER V3_ASM_ARCHS: ${V3_ASM_ARCHS}")
...
+    is_aiter_cache_valid(AITER_CACHE_VALID)
+  endif()
+
+  if(NOT AITER_CACHE_VALID)
+    # Try downloading prebuilt files if NVTE_AITER_PREBUILT_BASE_URL is set.
+    download_aiter_prebuilt(AITER_PREBUILT_DOWNLOAD_SUCCESS)
```
ipanfilo, line 49: Is it supposed to be AITER_PREBUILT_DOWNLOADED ?

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -0,0 +1,167 @@
...
+endfunction()
+
+# Cache locally built libs
+function(cache_local_aiter_build SOURCE_DIR)
+  file(MAKE_DIRECTORY "${EXTRACT_DIR}")
+  message(STATUS "[AITER-PREBUILT] Caching locally built libs to ${EXTRACT_DIR}")
+  file(COPY "${SOURCE_DIR}/libmha_fwd.so" "${SOURCE_DIR}/libmha_bwd.so" DESTINATION "${EXTRACT_DIR}")
+endfunction()
+
+# Download prebuilt zip file
```
ipanfilo: tgz file?

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -0,0 +1,167 @@
...
+  else()
+    file(REMOVE_RECURSE "${CACHE_ROOT}")
+    message(STATUS "[AITER-PREBUILT] Download unsuccessfull.")
+  endif()
+endfunction()
+
+# Upload prebuilt
+# Requires: NVTE_AITER_PREBUILT_BASE_URL
+# Uses __AITER_MHA_PATH contents to create zip file and upload
+function(upload_aiter_prebuilt)
```
ipanfilo: This method is not completed - it does not create .sha256, it also seems made in assumption of using OAuth2.
It may be mainly kept to create all necessary files and composing URLs but uploading itself better to offload to user

ipanfilo (changes requested):

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
VeeraRajasekhar: Removed the unused init_aiter_base_url() helper since URL validation is now handled directly in the download & upload functions.

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
VeeraRajasekhar: But this will not work if we run upload via _**cmake -P transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake -DACTION=upload**_, cause CMAKE_BINARY_DIR will have different path for upload & download, either I need to maintain separate copies for upload and download, or else find TE_ROOT and use this every where.

transformer_engine/common/ck_fused_attn/CMakeLists.txt
VeeraRajasekhar: Updated with AITER_PREBUILT_DOWNLOAD_SUCCESS variable

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
ipanfilo: What is CMAKE_BINARY_DIR in this case and why CMAKE_SORUCE_DIR is available in script mode?
Anyway, like said in other comments, upload function does not seem having necessary functionality to make "downloadable" upload so it cannot be used as-is.
Even preparation of package for upload at minimum needs SHA-256 calculation. Or it can be fully offloaded to user for now and upload can be removed then

VeeraRajasekhar: Using CMAKE_CURRENT_LIST_DIR, which points to the same directory in both the upload and download mode.

Upload --> changed to just create upload files as below
```
cmake -DACTION=upload -P transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake 
-- [AITER-PREBUILT] Creating prebuilt files...
-- [AITER-PREBUILT] tgz and sha256 files generated successfully:
--   /tmp/rocm-7.0_aiter-74e71eb8ee8a663d5e33c0cfd8b4dad7708ae84b.tar.gz
--   /tmp/rocm-7.0_aiter-74e71eb8ee8a663d5e33c0cfd8b4dad7708ae84b.tar.gz.sha256
```

user can manually upload these files.

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
VeeraRajasekhar: Upload --> changed to just create upload files as below
```
cmake -DACTION=upload -P transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake 
-- [AITER-PREBUILT] Creating prebuilt files...
-- [AITER-PREBUILT] tgz and sha256 files generated successfully:
--   /tmp/rocm-7.0_aiter-74e71eb8ee8a663d5e33c0cfd8b4dad7708ae84b.tar.gz
--   /tmp/rocm-7.0_aiter-74e71eb8ee8a663d5e33c0cfd8b4dad7708ae84b.tar.gz.sha256
```

user can manually upload these files.

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -16,34 +16,26 @@ string(REGEX MATCH "^[0-9]+\\.[0-9]+" ROCM_VER "${ROCM_VER_CONTENT}")
...
+    set(${CACHE_VALID} TRUE PARENT_SCOPE)
+    message(STATUS "[AITER-PREBUILT] Found Cached build files at ${EXTRACT_DIR}")
+    return()
   endif()
 
   # Cache is invalid/outdated - clean it
-  file(REMOVE_RECURSE "${EXTRACT_DIR}")
-  unset(__AITER_MHA_PATH CACHE)
+  file(REMOVE_RECURSE "${CACHE_ROOT}")
+  file(REMOVE_RECURSE "${CMAKE_CURRENT_LIST_DIR}/../../../build/cmake/_deps")
```
ipanfilo: Why is this needed? While it is OK to use CMAKE_CURRENT_LIST_DIR as a base for CACHE_ROOT or to find AITER location, for CMake generated files CMAKE_BINARY_DIR must be used  because 'build' is not not hardcoded

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -53,16 +45,13 @@ function(cache_local_aiter_build SOURCE_DIR)
...
 function(download_aiter_prebuilt DOWNLOAD_SUCCESS)
   if(NOT DEFINED ENV{NVTE_AITER_PREBUILT_BASE_URL} OR "$ENV{NVTE_AITER_PREBUILT_BASE_URL}" STREQUAL "")
     return()
   endif()
 
-  set(BASE_URL "$ENV{NVTE_AITER_PREBUILT_BASE_URL}")
-  string(REGEX REPLACE "/$" "" BASE_URL "${BASE_URL}")
-  set(FILE_URL "${BASE_URL}/${KEY}.tar.gz")
-
+  string(REGEX REPLACE "/$" "" FILE_URL "$ENV{NVTE_AITER_PREBUILT_BASE_URL}/${KEY}.tar.gz")
```
ipanfilo: Doing this way, you have no trailing / so regex will not change anything.

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
VeeraRajasekhar: @ipanfilo 
As I wanted to keep common for both upload and download, I am using  CMAKE_CURRENT_LIST_DIR which points to the same path in both the cases.

If I were to use CMAKE_BINARY_DIR, it is pointed to different location in the two cases.

VeeraRajasekhar: Line 38 is to remove the cache of FETCH_CONTENT.
Updated Line 38 to use CMAKE_BINARY_DIR,

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
VeeraRajasekhar: Removed the regex

ipanfilo (approved):

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -0,0 +1,123 @@
...
+if("${ROCM_PATH}" STREQUAL "")
+  set(ROCM_PATH "/opt/rocm")
+endif()
+file(READ "${ROCM_PATH}/.info/version" ROCM_VER_CONTENT)
+string(STRIP "${ROCM_VER_CONTENT}" ROCM_VER_CONTENT)
+string(REGEX MATCH "^[0-9]+\\.[0-9]+" ROCM_VER "${ROCM_VER_CONTENT}")
+
+# AITER commit
+execute_process(
+  COMMAND git -C "${CMAKE_CURRENT_LIST_DIR}/../../../3rdparty/aiter" rev-parse HEAD
```
wangye805: If I git checked out the TE/aiter repo in bare-metal and then start a docker container, will this git command work as expected? 

Usually when I run git status in a docker container for repos checked out in bare metal, I'll run into this following error:
```
root@ctr-cx64-mi300x-13:/workspace/te_ifu_v2.6_oct23# git status
fatal: detected dubious ownership in repository at '/workspace/te_ifu_v2.6_oct23'
To add an exception for this directory, call:

        git config --global --add safe.directory /workspace/te_ifu_v2.6_oct23
```

wangye805 (changes requested):

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
VeeraRajasekhar: yes, ye I am getting the same issue, cause of ownership. should I add that directory as safe by default inside the cmake file?

wangye805: I think so, otherwise we cannot check the aiter commit number...

ipanfilo: setup.py also checks git revision so I think TE installation should fail anyway if permissions are wrong.
I hesitate changing git config from build script

ipanfilo (approved):

log.txt
```diff
@@ -0,0 +1,1639 @@
+Using pip 25.2 from /opt/venv/lib/python3.10/site-packages/pip (python 3.10)
```
wangye805: We don't need to include this file

wangye805 (changes requested):

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
wangye805: @ipanfilo can you elaborate on where does setup.py check the git revision? I didn't recall running into same problem when pip installing our current dev branch

log.txt
VeeraRajasekhar: removed

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
ipanfilo: https://github.com/ROCm/TransformerEngine/blob/91876593ff7f721e06f5d7197aac49d3804b3479/build_tools/te_version.py#L26

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -0,0 +1,126 @@
...
+if("${ROCM_PATH}" STREQUAL "")
+  set(ROCM_PATH "/opt/rocm")
+endif()
+file(READ "${ROCM_PATH}/.info/version" ROCM_VER_CONTENT)
+string(STRIP "${ROCM_VER_CONTENT}" ROCM_VER_CONTENT)
+string(REGEX MATCH "^[0-9]+\\.[0-9]+" ROCM_VER "${ROCM_VER_CONTENT}")
+
+# AITER commit
+execute_process(
+  COMMAND git -c safe.directory=${CMAKE_CURRENT_LIST_DIR}/../../../3rdparty/aiter
```
VeeraRajasekhar: @ipanfilo @wangye805 

Should i revert this ?

One thing to note is, when we build TE with aiter-prebuilt enabled, and if permissions is not right, it will just throw an exception and proceeds to build from source.

ipanfilo: As soon as it works it is good to have.

VeeraRajasekhar: okay

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
wangye805: Thanks for pointing out, I tried our current dev with some debug messages printed. Indeed this git command failed with TE checked out in bare-metal but we are trying to build in a docker container:
```
Using pip 25.1.1 from /opt/conda/envs/py_3.10/lib/python3.10/site-packages/pip (python 3.10)
Obtaining file:///workspace/te_aiter_native_padding_wo_runtime_segment_seqlen
  Preparing metadata (setup.py): started
  Running command python setup.py egg_info
  /workspace/te_aiter_native_padding_wo_runtime_segment_seqlen/3rdparty/hipify_torch/hipify_torch/hipify_python.py:44: FutureWarning: you are using hipify version 1.
  hipify's version 2 behavior will become the default in a later version of torch
    warnings.warn("""you are using hipify version 1.
  error running git rev-parse --short HEAD
  in setup.py, __version__:  2.4.0.dev0
```
But this error does not prevent the subsequent building/installation so previously I didn't notice it

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
wangye805: I checked out your branch with commit in bare-metal and try to install it in docker container, however it still shows the following error. Then it goes into local compilation...
```
    CMake Warning (dev) at /opt/conda/envs/py_3.10/share/cmake-3.31/Modules/ExternalProject/shared_internal_commands.cmake:1276 (message):
      The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
      not set.  The policy's OLD behavior will be used.  When using a URL
      download, the timestamps of extracted files should preferably be that of
      the time of extraction, otherwise code that depends on the extracted
      contents might not be rebuilt if the URL changes.  The OLD behavior
      preserves the timestamps from the archive instead, but this is usually not
      what you want.  Update your project to the NEW behavior or specify the
      DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
      robustness issue.
    Call Stack (most recent call first):
      /opt/conda/envs/py_3.10/share/cmake-3.31/Modules/ExternalProject.cmake:3041 (_ep_add_download_command)
      CMakeLists.txt:327 (ExternalProject_Add)
    This warning is for project developers.  Use -Wno-dev to suppress it.

    -- AITER V3_ASM_ARCHS: gfx942
    fatal: detected dubious ownership in repository at '/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter'
    To add an exception for this directory, call:

        git config --global --add safe.directory /workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter
    --  [AITER-PREBUILT] Building aiter from source.
    aiter is not installed.
    Detected CUDA files, patching ldflags
    Emitting ninja build file /workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/aiter/jit/build/libmha_fwd/build/build.ninja...
    Building extension module libmha_fwd...
    Using envvar MAX_JOBS (32) as the number of workers...
    [1/3954] /opt/rocm/bin/hipcc  -DWITH_HIP -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/aiter/jit/build/ck/include -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/aiter/jit/build/ck/library/include -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/csrc/include -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/aiter/jit/build/libmha_fwd/blob -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/aiter/jit/build/ck/example/ck_tile/01_fmha -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/csrc/include/torch -fPIC -std=c++20 -O3 -std=c++20 -fPIC -D__HIP_PLATFORM_AMD__=1 -DUSE_ROCM=1 -DHIPBLAS_V2 -DCUDA_HAS_FP16=1 -D__HIP_NO_HALF_OPERATORS__=1 -D__HIP_NO_HALF_CONVERSIONS__=1 -mcmodel=large -fno-unique-section-names -ffunction-sections -fdata-sections -fvisibility=hidden -fvisibility-inlines-hidden --offload-arch=gfx942 -DCK_TILE_FLOAT_TO_BFLOAT16_DEFAULT=3 -DCK_TILE_FMHA_FWD_FAST_EXP2=1 -DCK_TILE_FMHA_FWD_SPLITKV_API=1 -DLEGACY_HIPBLAS_DIRECT -DUSE_PROF_API=1 -D__HIP_PLATFORM_AMD__=1 -D__HIP_PLATFORM_HCC__=1 -U__HIP_NO_HALF_CONVERSIONS__ -U__HIP_NO_HALF_OPERATORS__ -Wno-macro-redefined -Wno-missing-template-arg-list-after-template-kw -Wno-switch-bool -Wno-undefined-func-template -Wno-unused-result -Wno-vla-cxx-extension -fgpu-flush-denormals-to-zero -fno-offload-uniform-block -mllvm --amdgpu-kernarg-preload-count=16 -mllvm --lsr-drop-solution=1 -mllvm -amdgpu-coerce-illegal-types=1 -mllvm -amdgpu-early-inline-all=true -mllvm -amdgpu-function-calls=false -mllvm -enable-post-misched=0 -fno-gpu-rdc -c /workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/aiter/jit/build/libmha_fwd/blob/fmha_batch_prefill_api.cpp -o fmha_batch_prefill_api.cuda.o
    [2/3954] /opt/rocm/bin/hipcc  -DWITH_HIP -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/aiter/jit/build/ck/include -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/aiter/jit/build/ck/library/include -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/csrc/include -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/aiter/jit/build/libmha_fwd/blob -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/aiter/jit/build/ck/example/ck_tile/01_fmha -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/csrc/include/torch -fPIC -std=c++20 -O3 -std=c++20 -fPIC -D__HIP_PLATFORM_AMD__=1 -DUSE_ROCM=1 -DHIPBLAS_V2 -DCUDA_HAS_FP16=1 -D__HIP_NO_HALF_OPERATORS__=1 -D__HIP_NO_HALF_CONVERSIONS__=1 -mcmodel=large -fno-unique-section-names -ffunction-sections -fdata-sections -fvisibility=hidden -fvisibility-inlines-hidden --offload-arch=gfx942 -DCK_TILE_FLOAT_TO_BFLOAT16_DEFAULT=3 -DCK_TILE_FMHA_FWD_FAST_EXP2=1 -DCK_TILE_FMHA_FWD_SPLITKV_API=1 -DLEGACY_HIPBLAS_DIRECT -DUSE_PROF_API=1 -D__HIP_PLATFORM_AMD__=1 -D__HIP_PLATFORM_HCC__=1 -U__HIP_NO_HALF_CONVERSIONS__ -U__HIP_NO_HALF_OPERATORS__ -Wno-macro-redefined -Wno-missing-template-arg-list-after-template-kw -Wno-switch-bool -Wno-undefined-func-template -Wno-unused-result -Wno-vla-cxx-extension -fgpu-flush-denormals-to-zero -fno-offload-uniform-block -mllvm --amdgpu-kernarg-preload-count=16 -mllvm --lsr-drop-solution=1 -mllvm -amdgpu-coerce-illegal-types=1 -mllvm -amdgpu-early-inline-all=true -mllvm -amdgpu-function-calls=false -mllvm -enable-post-misched=0 -fno-gpu-rdc -c /workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/aiter/jit/build/libmha_fwd/blob/asm_fmha_fwd_v3_gfx942.cpp -o asm_fmha_fwd_v3_gfx942.cuda.o
```

VeeraRajasekhar: @wangye805 can you try now

wangye805: After git pull --rebase, I still cannot use the aiter online prebuilt:
```
    CMake Warning (dev) at /opt/conda/envs/py_3.10/share/cmake-3.31/Modules/ExternalProject/shared_internal_commands.cmake:1276 (message):
      The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
      not set.  The policy's OLD behavior will be used.  When using a URL
      download, the timestamps of extracted files should preferably be that of
      the time of extraction, otherwise code that depends on the extracted
      contents might not be rebuilt if the URL changes.  The OLD behavior
      preserves the timestamps from the archive instead, but this is usually not
      what you want.  Update your project to the NEW behavior or specify the
      DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
      robustness issue.
    Call Stack (most recent call first):
      /opt/conda/envs/py_3.10/share/cmake-3.31/Modules/ExternalProject.cmake:3041 (_ep_add_download_command)
      CMakeLists.txt:327 (ExternalProject_Add)
    This warning is for project developers.  Use -Wno-dev to suppress it.

    -- AITER V3_ASM_ARCHS: gfx942
    --  [AITER-PREBUILT] Building aiter from source.
    aiter is not installed.
    Detected CUDA files, patching ldflags
    Emitting ninja build file /workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/aiter/jit/build/libmha_fwd/build/build.ninja...
    Building extension module libmha_fwd...
    Using envvar MAX_JOBS (32) as the number of workers...
    [1/3954] /opt/rocm/bin/hipcc  -DWITH_HIP -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/aiter/jit/build/ck/include -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/aiter/jit/build/ck/library/include -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/csrc/include -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/aiter/jit/build/libmha_fwd/blob -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/aiter/jit/build/ck/example/ck_tile/01_fmha -I/workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/csrc/include/torch -fPIC -std=c++20 -O3 -std=c++20 -fPIC -D__HIP_PLATFORM_AMD__=1 -DUSE_ROCM=1 -DHIPBLAS_V2 -DCUDA_HAS_FP16=1 -D__HIP_NO_HALF_OPERATORS__=1 -D__HIP_NO_HALF_CONVERSIONS__=1 -mcmodel=large -fno-unique-section-names -ffunction-sections -fdata-sections -fvisibility=hidden -fvisibility-inlines-hidden --offload-arch=gfx942 -DCK_TILE_FLOAT_TO_BFLOAT16_DEFAULT=3 -DCK_TILE_FMHA_FWD_FAST_EXP2=1 -DCK_TILE_FMHA_FWD_SPLITKV_API=1 -DLEGACY_HIPBLAS_DIRECT -DUSE_PROF_API=1 -D__HIP_PLATFORM_AMD__=1 -D__HIP_PLATFORM_HCC__=1 -U__HIP_NO_HALF_CONVERSIONS__ -U__HIP_NO_HALF_OPERATORS__ -Wno-macro-redefined -Wno-missing-template-arg-list-after-template-kw -Wno-switch-bool -Wno-undefined-func-template -Wno-unused-result -Wno-vla-cxx-extension -fgpu-flush-denormals-to-zero -fno-offload-uniform-block -mllvm --amdgpu-kernarg-preload-count=16 -mllvm --lsr-drop-solution=1 -mllvm -amdgpu-coerce-illegal-types=1 -mllvm -amdgpu-early-inline-all=true -mllvm -amdgpu-function-calls=false -mllvm -enable-post-misched=0 -fno-gpu-rdc -c /workspace/te_veergopu_aiter_prebuilt/3rdparty/aiter/aiter/jit/build/libmha_fwd/blob/fmha_batch_prefill_api.cpp -o fmha_batch_prefill_api.cuda.o

```

I was using base docker image compute-artifactory.amd.com:5000/rocm-plus-docker/framework/private/te-ci:rocm-6.4_0_ubuntu22_py310_torch25_jax0435qa_fa273 and running:
```
NVTE_ROCM_ARCH=gfx942 pip install -e . -v 2>&1| tee pip_install.log0
```

Is there something I'm missing?

VeeraRajasekhar: @wangye805 you need to set this env export NVTE_AITER_PREBUILT_BASE_URL=https://compute-artifactory.amd.com:5000/artifactory/rocm-generic-local/te-ci/aiter-prebuilts

ipanfilo: That location does not have binaries matching your config anyway

wangye805: Ah, I see. Thanks

wangye805 (approved):

PR #336, opened 2025-10-16
Fused Cross Entropy Triton - Loss Scaling and Vanishing Grads Bugfix

sarthak-amd:
# Description

The Fused Cross Entropy Triton Kernel currently has 2 bugs
1. if `ignore_idx` is not None`, the loss should be computed only over valid tokens and not all tokens (new fix)
2. gradient scaling when `reduce_loss=False`. (This is already fixed in upstream)
    - If `reduced loss=False`, we should compute per token loss and not reduce it else it would shrink the gradients by 1/N giving wrong (higher) loss. 
    - if `reduce_loss=False`, `grad_output` is a tensor, not a scalar. We need to load 1 value per row instead of just a scalar. 

This fix is validated on Llama3.1 8B model for Pre-training. 

## Type of change

### DIFF
--- a/transformer_engine/pytorch/cross_entropy.py
+++ b/transformer_engine/pytorch/cross_entropy.py
@@ -1,3 +1,5 @@
+# This file was modified for portability to AMDGPU
+# Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
 # Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # See LICENSE for license information.
--- a/transformer_engine/pytorch/cross_entropy.py
+++ b/transformer_engine/pytorch/cross_entropy.py
@@ -1,4 +1,3 @@
-# This file was modified for portability to AMDGPU
 # Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
 # Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
--- a/tests/pytorch/test_parallel_cross_entropy.py
+++ b/tests/pytorch/test_parallel_cross_entropy.py
@@ -3,9 +3,10 @@
 # See LICENSE for license information.
 
 import random
-import pytest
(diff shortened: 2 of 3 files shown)


wenchenvincent:
@sarthak-amd  Could you post the PR for the upstream fix?

transformer_engine/pytorch/cross_entropy.py
```diff
@@ -1,3 +1,5 @@
+# This file was modified for portability to AMDGPU
```
wenchenvincent: There is no real change in this file. Let's keep this file intact and then we don't need to add the AMD copyright statement.

sarthak-amd:
> @sarthak-amd Could you post the PR for the upstream fix?

https://github.com/NVIDIA/TransformerEngine/commit/e9a5fa4e368464f3b310b90ab7f670f35319344b @wenchenvincent

wenchenvincent:
> > @sarthak-amd Could you post the PR for the upstream fix?
> 
> [NVIDIA/TransformerEngine@e9a5fa4](https://github.com/NVIDIA/TransformerEngine/commit/e9a5fa4e368464f3b310b90ab7f670f35319344b) @wenchenvincent

Another fix came from the upstream PR https://github.com/NVIDIA/TransformerEngine/pull/1879. Is the change of test in that PR also reflected?

wenchenvincent:
For the fix for `ignore_idx`, is there a test for it (without the fix, the test would fail)?

wenchenvincent (changes requested):
@sarthak-amd Could you refactor the PR as 3 commits:
- 2 commits would be cherrypicking from the upstream PRs.
- 1 commit for the `ignore_idx` with a test to cover it.

This way the PR would be very clear and easy to understand.

wenchenvincent:
@sarthak-amd Could you address the comments? Also, please rebase upon latest dev so that hot fixes for sgpu tests could pass.

wenchenvincent:
@sarthak-amd Could you rebase the PR and update it per reviewer comments?

PR #338, opened 2025-10-16
MXFP8 test scale off by 1 fix

alextmagro:
# Description

Added a combined check for scale and values
If an scale difference is detected, values are checked for an off by one boundary condition
mismatch tolerance allows for a few scale differences (1% by default), before throwing an error.

Additionally, issues with dgelu CPU-side jitter have been resolved with data generation fixes.

### DIFF
--- a/tests/cpp/CMakeLists.txt
+++ b/tests/cpp/CMakeLists.txt
@@ -60,7 +60,7 @@ if(USE_CUDA)
   set(CMAKE_CUDA_STANDARD_REQUIRED ON)
   project(transformer_engine_tests LANGUAGES CUDA CXX)
 else()
-  set(CMAKE_CXX_STANDARD 17)
+  set(CMAKE_CXX_STANDARD 20)
   project(transformer_engine_tests LANGUAGES HIP CXX)
   # Ask hcc to generate device code during compilation so we can use
   # host linker to link.
--- a/tests/cpp/CMakeLists.txt
+++ b/tests/cpp/CMakeLists.txt
@@ -60,7 +60,7 @@ if(USE_CUDA)
   set(CMAKE_CUDA_STANDARD_REQUIRED ON)
   project(transformer_engine_tests LANGUAGES CUDA CXX)
 else()
-  set(CMAKE_CXX_STANDARD 20)
+  set(CMAKE_CXX_STANDARD 17)
   project(transformer_engine_tests LANGUAGES HIP CXX)
   # Ask hcc to generate device code during compilation so we can use
   # host linker to link.
--- a/tests/cpp/operator/test_cast_mxfp8.cu
+++ b/tests/cpp/operator/test_cast_mxfp8.cu
@@ -320,7 +320,7 @@ void performTest_x1(const ProcessingMethod processing_method,
                                 unpadded_blocks_Y, unpadded_blocks_X, scales_stride, 0.01, rowwise, mismatch_idx);
 
         if (mismatch_idx.size()) {
-            adjust_ref<OutputType>(mismatch_idx, ref_output_c.get(), unpadded_blocks_Y, unpadded_blocks_X, rows, cols);
+            adjust_ref(mismatch_idx, ref_output_c.get(), unpadded_blocks_Y, unpadded_blocks_X, rows, cols, otype);
         }
 
         auto [atol, rtol] = getTolerances(otype);
@@ -339,7 +339,7 @@ void performTest_x1(const ProcessingMethod processing_method,
     compare_e8m0_scaling_factors("scales", gpu_scales_ptr, ref_output_scales.get(),
                                  unpadded_blocks_Y, unpadded_blocks_X, scales_stride);
     }
-    
+
     if (processing_method == ProcessingMethod::CAST_DBIAS || processing_method == ProcessingMethod::CAST_DBIAS_DACT) {
         auto [atol_dbias, rtol_dbias] = getTolerances(itype);
         if (itype == DType::kFloat32) {
@@ -479,14 +479,14 @@ void performTest_x2(const ProcessingMethod processing_method,
                                 unpadded_blocks_Y_rowwise, unpadded_blocks_X_rowwise, scales_stride_rowwise, 0.01, true, mismatch_idx_r);
 
         if (mismatch_idx_r.size()) {
-            adjust_ref<OutputType>(mismatch_idx_r, ref_output_c_rowwise.get(), unpadded_blocks_Y_rowwise, unpadded_blocks_X_rowwise, rows, cols);
+            adjust_ref(mismatch_idx_r, ref_output_c_rowwise.get(), unpadded_blocks_Y_rowwise, unpadded_blocks_X_rowwise, rows, cols, otype);
         }
         std::vector<std::tuple<size_t, size_t, int>> mismatch_idx_c;
         compare_e8m0_scaling_factors("scales_colwise", output, ref_scales_colwise.get(),
                                 unpadded_blocks_Y_colwise, unpadded_blocks_X_colwise, scales_stride_colwise, 0.01, false, mismatch_idx_c);
 
         if (mismatch_idx_c.size()) {
-            adjust_ref<OutputType>(mismatch_idx_c, ref_output_c_colwise.get(), unpadded_blocks_Y_colwise, unpadded_blocks_X_colwise, rows, cols);
+            adjust_ref(mismatch_idx_c, ref_output_c_colwise.get(), unpadded_blocks_Y_colwise, unpadded_blocks_X_colwise, rows, cols, otype);
         }
 
         auto [atol, rtol] = getTolerances(otype);
--- a/tests/cpp/operator/test_cast_mxfp8.cu
+++ b/tests/cpp/operator/test_cast_mxfp8.cu
@@ -76,12 +76,12 @@ void scale_block(const ProcessingMethod processing_method,
                 continue;
             }
             amax = std::max(amax, std::abs(elt));
-#else
+#else // #ifdef __HIP_PLATFORM_AMD__
             if (std::isinf(elt) || std::isnan(elt)) {
                 continue;
             }
             amax = fmaxf(amax, fabsf(elt));
-#endif
+#endif // #ifdef __HIP_PLATFORM_AMD__
         }
     }
 
@@ -312,6 +312,12 @@ void performTest_x1(const ProcessingMethod processing_method,
                                               block_size_cols,
                                               scales_stride);
 
+    
+#ifdef __HIP_PLATFORM_AMD__
+compare_mxfp8_results("scales", ref_output_scales.get(), unpadded_blocks_Y, 
+                           unpadded_blocks_X, scales_stride, "output_c", 
+                           output_c, ref_output_c.get(), rowwise, rows, cols);
+#else // #ifdef __HIP_PLATFORM_AMD__
     auto [atol, rtol] = getTolerances(otype);
     compareResults("output_c", output_c, ref_output_c.get(), rowwise, atol, rtol);
 
@@ -321,6 +327,7 @@ void performTest_x1(const ProcessingMethod processing_method,
 
     compare_e8m0_scaling_factors("scales", gpu_scales_ptr, ref_output_scales.get(),
                                  unpadded_blocks_Y, unpadded_blocks_X, scales_stride);
+#endif // #ifdef __HIP_PLATFORM_AMD__
 
     if (processing_method == ProcessingMethod::CAST_DBIAS || processing_method == ProcessingMethod::CAST_DBIAS_DACT) {
         auto [atol_dbias, rtol_dbias] = getTolerances(itype);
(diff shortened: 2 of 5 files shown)


tests/cpp/CMakeLists.txt
```diff
@@ -60,7 +60,7 @@ if(USE_CUDA)
   set(CMAKE_CUDA_STANDARD_REQUIRED ON)
   project(transformer_engine_tests LANGUAGES CUDA CXX)
 else()
-  set(CMAKE_CXX_STANDARD 17)
+  set(CMAKE_CXX_STANDARD 20)
```
wenchenvincent: Is C++ 20 necessary?

tests/cpp/test_common.cu
```diff
@@ -711,6 +711,114 @@ void compare_e8m0_scaling_factors(const std::string &name, const uint8_t *test,
   }
 }
 
+#ifdef __HIP_PLATFORM_AMD__
+template <typename T>
```
ipanfilo: Why using such complicated scheme? Isn't it more straightforward to compare scales and dequantized values? And it should require less code adding

tests/cpp/CMakeLists.txt
alextmagro: The motivation for moving to C++20 was for templatized lambda functions and default comparisons. It is also likely that C++20 will come in handy for the parallelization of the reference code in a later ticket. 

We can refactor using only C++17, but C++20 does make things a bit easier.

tests/cpp/test_common.cu
alextmagro: Good question, we do our quantize then dequantize test in test_dequantize_mxfp8. For the MXFP8 cast tests, we don't dequantize any values, so would have to add code to dequantize both the reference and the source code. The motivation here is to do the scale and value check in the same way as upstream, but combine the checks so we can handle the off by one issue appropriately.

wenchenvincent: It does seem complicated. And I am wondering if there is anything we could miss by only adjusting the values.
Essentially, we want to check (1) the scale is at most off by 1 (2) The test and reference values should both fall in between two adjacent fp8 values (if scale is off by 1, then one value is larger than midpoint and the other smaller than midpoint).

alextmagro: If scale is off by one, we expect the test value to be exactly twice or half of the reference value. This is what the OP is checking for -- That if the scale is off, we have a value off by a single power of 2 as well.

If the scales match, we don't do anything to the value, and the compare function does the exact same thing as the compare_e8m0 and compare_results functions, just concurrently.

We also verify that there isn't something going on with the scales throughout the test by only allowing a few scales to be off, given this is an edge case that shouldn't occur frequently. I set the tol to 1% of the # of scales, but this can be adjusted too.

alextmagro: > I am wondering if there is anything we could miss by only adjusting the values.

This is one reason I am hesitant to move away from just checking the scales and dequantizing if a scale is off -- We may miss something else going on. I also have a feeling our code diff would end up larger.

ipanfilo: > It does seem complicated. And I am wondering if there is anything we could miss by only adjusting the values. Essentially, we want to check (1) the scale is at most off by 1 (2) The test and reference values should both fall in between two adjacent fp8 values (if scale is off by 1, then one value is larger than midpoint and the other smaller than midpoint).

Do we really need to check that the values are around midpoint? If we actually expect very few such different scales, instead of comparing dequantized tensors we can perform checking and adjustment of corresponding reference data. So it will be scales checking, if pass, for every different scale value fix it and corresponding output data block and do possible validation that data is around midpoint, and then do normal comparison of output data

alextmagro:
I have added a refactor in the 2nd commit -- I am unsure if it is cleaner this way around, but both solve the issue. @wenchenvincent and @ipanfilo , please have a look at both versions and let me know which one looks better

tests/cpp/operator/test_cast_mxfp8.cu
```diff
@@ -312,6 +312,12 @@ void performTest_x1(const ProcessingMethod processing_method,
                                               block_size_cols,
                                               scales_stride);
 
+    
+#ifdef __HIP_PLATFORM_AMD__
+compare_mxfp8_results("scales", ref_output_scales.get(), unpadded_blocks_Y,
```
wangye805: I understand that after complicated gated function, cpu and gpu results could be slightly different. Do we need to tolerate those off by 1 scales for just mxfp8 cast without any complicated math operation?

tests/cpp/CMakeLists.txt
wangye805: Looks like c++ 20 is used for spaceship operator "<=>" and templated lambda. I would suggest just use plain uint8 comparison and an int flag to denote whether we need to multiply /divide by 2. Then we can use c++17

tests/cpp/test_common.cu
```diff
@@ -711,6 +711,114 @@ void compare_e8m0_scaling_factors(const std::string &name, const uint8_t *test,
...
+std::pair<size_t, size_t> compare_mxfp8_block(const T *test_data, const T *ref_data, const uint8_t *scale_test, 
+                         const uint8_t *scale_ref, size_t row_blocks, size_t col_blocks, size_t stride,
+                         size_t rows, size_t cols, double atol, double rtol) {
+  size_t first_failure    = (size_t)-1;
+  size_t scale_mismatches = 0;
+
+  std::function<void(double&)> OP;
+
+#pragma omp parallel for collapse(2) schedule(dynamic, 8) nonmonotonic reduction(+:scale_mismatches) reduction(first_failure)
```
wangye805: What's the benefits we can bring using omp here? Will it result in building error on docker containers without omp installed?

wangye805 (changes requested):

tests/cpp/test_common.h
```diff
@@ -461,6 +462,34 @@ void compare_e8m0_scaling_factors(const std::string &name, const uint8_t *test,
...
+  for (const auto &[i, j, scale_diff] : mismatch_idx) {
+    size_t ii_min = i * col_blocks;
+    const size_t ii_max = std::min(ii_min + col_blocks, rows);
+    for (; ii_min < ii_max; ii_min++) {
+      size_t jj_min = j * row_blocks;
+      const size_t jj_max = std::min(jj_min + row_blocks, cols);
+      for (; jj_min < jj_max; jj_min++) {
+        const size_t data_idx = ii_min * cols + jj_min;
+        if (scale_diff == 1) {
+          ref_data[data_idx] = static_cast<T>(static_cast<float>(ref_data[data_idx])*2);
```
wangye805: I think this will create additional accuracy loss with the true quantized ref. 

In my mind, the true adjusted ref (quantized) should be
```
quantize(original high-precision output / 2^(adjusted scale)).
```
Currently your result will be adjust(quantize(original high precision output)/2^(scale)). The quantize is actually a nonlinear operation

wangye805 (changes requested):

tests/cpp/operator/test_cast_mxfp8.cu
alextmagro: The gated functions call this as well, so should be covered.

tests/cpp/test_common.cu
alextmagro: It's just speedup -- no build error without OMP flag -fopenmp as it is a #pragma  (although we require OpenMP in our cmake)

tests/cpp/test_common.h
alextmagro: I agree, but we don't have any way to detect when the scale is off by one while computing the quantized data, as far as I can tell.

tests/cpp/CMakeLists.txt
alextmagro: After separate discussion with Ilya, moved things back to C++17 until docker issues with C++20 are resolved

tests/cpp/test_common.cu
```diff
@@ -711,6 +711,48 @@ void compare_e8m0_scaling_factors(const std::string &name, const uint8_t *test,
   }
 }
 
+#ifdef __HIP_PLATFORM_AMD__
+void compare_e8m0_scaling_factors(const std::string &name, Tensor &output, const uint8_t *ref,
+                             const size_t row_blocks, const size_t col_blocks, const size_t stride, 
+                             double tol, bool rowwise, std::vector<std::tuple<size_t, size_t, int>> &mismatch_idx) {
+  constexpr bool on_gpus = true;
+  if (on_gpus) output.to_cpu();
```
ipanfilo: getting cpu_scale_inv_ptr() below performs to_cpu()

tests/cpp/test_common.cu
```diff
@@ -711,6 +711,48 @@ void compare_e8m0_scaling_factors(const std::string &name, const uint8_t *test,
...
+  for (int i = 0; i < row_blocks; i++) {
+    for (int j = 0; j < col_blocks; j++) {
+      const int idx = i * stride + j;
+      if (test[idx] != ref[idx]) {
+        int t_scale = static_cast<int>(test[idx]);
+        int r_scale = static_cast<int>(ref[idx]);
+        if (std::abs(t_scale - r_scale) == 1) {
+          mismatch_idx.emplace_back(i, j, r_scale-t_scale);
+        } else {
+          ASSERT_FALSE(1) << "Error in " << name << std::endl
```
ipanfilo: You can use GTEST_FAIL() instead of ASSERT_FALSE(1)

tests/cpp/test_common.h
```diff
@@ -461,6 +462,34 @@ void compare_e8m0_scaling_factors(const std::string &name, const uint8_t *test,
...
+void adjust_ref(std::vector<std::tuple<size_t, size_t, int>> mismatch_idx, void *ref, const size_t row_blocks,
+                const size_t col_blocks, const size_t rows, const size_t cols) {
+  T *ref_data = reinterpret_cast<T*>(ref);
+  for (const auto &[i, j, scale_diff] : mismatch_idx) {
+    size_t ii_min = i * col_blocks;
+    const size_t ii_max = std::min(ii_min + col_blocks, rows);
+    for (; ii_min < ii_max; ii_min++) {
+      size_t jj_min = j * row_blocks;
+      const size_t jj_max = std::min(jj_min + row_blocks, cols);
+      for (; jj_min < jj_max; jj_min++) {
```
ipanfilo: Why do we have ii and jj nested loops here? One scale value refers to 32 items either in row or in col but not in both

alextmagro: Either row_blocks or col_blocks is always 1, so we are doing the logic in one direction or the other.

tests/cpp/test_common.h
wangye805: Discussed with Alex offline and I agreed with his current plan

wangye805 (approved):
LGTM 

please merge after getting LGTM from Ilya as well

tests/cpp/test_common.h
ipanfilo: So one of loops is guaranteed to be 1, which means either of col_blocks or row_blocks is always 1, right?

tests/cpp/test_common.h
```diff
@@ -461,6 +462,34 @@ void compare_e8m0_scaling_factors(const std::string &name, const uint8_t *test,
...
+        const size_t data_idx = ii_min * cols + jj_min;
+        if (scale_diff == 1) {
```
ipanfilo: May be move IF out of the loops by making float scale_value 2.0 or 0.5

tests/cpp/test_common.h
```diff
@@ -461,6 +462,34 @@ void compare_e8m0_scaling_factors(const std::string &name, const uint8_t *test,
                                   const size_t row_blocks, const size_t col_blocks, const size_t stride);
 void compare_e8m0_scaling_factors(const std::string &name, const uint8_t *test, const uint8_t *ref,
                                   const size_t N);
+#ifdef USE_ROCM
+void compare_e8m0_scaling_factors(const std::string &name, Tensor &output, const uint8_t *ref,
+                             const size_t row_blocks, const size_t col_blocks, const size_t stride, 
+                             double tol, bool rowwise, std::vector<std::tuple<size_t, size_t, int>> &mismatch_idx);
+template <typename T>
+void adjust_ref(std::vector<std::tuple<size_t, size_t, int>> mismatch_idx, void *ref, const size_t row_blocks,
```
ipanfilo, line 470: As a possible polishing - move the method to test_common.cu and explicitly instantiate for FP8/BF8 - I think those are the only used T

tests/cpp/test_common.h
```diff
@@ -461,6 +462,34 @@ void compare_e8m0_scaling_factors(const std::string &name, const uint8_t *test,
...
+      size_t jj_min = j * row_blocks;
+      const size_t jj_max = std::min(jj_min + row_blocks, cols);
+      for (; jj_min < jj_max; jj_min++) {
+        const size_t data_idx = ii_min * cols + jj_min;
+        if (scale_diff == 1) {
+          ref_data[data_idx] = static_cast<T>(static_cast<float>(ref_data[data_idx])*2);
+        } else if (scale_diff == -1) {
+          ref_data[data_idx] = static_cast<T>(static_cast<float>(ref_data[data_idx])/2);
+        } else { // Shouldn't ever reach this
+          ASSERT_FALSE(1) << "Error in adjust_ref, |scale_diff| > 1";
```
ipanfilo: GTEST_FAIL() too?

tests/cpp/test_common.h
alextmagro: That's right.

ipanfilo (approved):

wenchenvincent (approved):
LGTM. Let's wait for CI before merging.

PR #341, opened 2025-10-20
[TE] Implement Triton current scaling

matthiasdiener:
# Description

Implements current scaling in Triton.

Fixes https://github.com/ROCm/frameworks-internal/issues/13252

Since there are no global barriers in Triton, there must be at least two kernel launches for each quantization:
- Calculate amax on device (kernel 1)
- Calculate scale factor (kernel 2)
- Quantize (kernel 3)

**TODOs:**
- [x] Performance comparison to TE HIP kernels (see https://github.com/ROCm/frameworks-internal/issues/13252#issuecomment-3499645181)  

- [x] Check if it is possible to make 1 "combined" Triton kernel that works correctly with recomputation/checkpointing, instead of the 2.5 kernels currently used in this PR (amax+scale+cast_transpose). 
  - I tried doing this, but could not find a solution.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Implements current-scaling support in the Triton kernel path
- Updates the `cast_transpose`  launcher and test infrastructure to support the new logic
- Adds a test to compare the current scaling implementations of TE and Triton 

# Checklist:

### DIFF
--- a/tests/pytorch/test_numerics.py
+++ b/tests/pytorch/test_numerics.py
@@ -1955,9 +1955,6 @@ def test_grouped_linear_accuracy(
     if IS_HIP_EXTENSION:
         if dtype not in (torch.float32,) and fuse_wgrad_accumulation and not fp8:
             pytest.skip(f"Rocm does not support fused wgrad accumulation for {dtype}.")
-        use_cast_transpose_triton =  bool( int(os.environ.get('NVTE_USE_CAST_TRANSPOSE_TRITON', '0')) )
-        if fp8 and recipe.float8_current_scaling() and use_cast_transpose_triton:
-            pytest.skip("Float8 Current Scaling unsupported for grouped linear accuracy.")
     if fp8 and not fp8_available:
         pytest.skip(reason_for_no_fp8)
     if fp8 and recipe.mxfp8() and not mxfp8_available:
--- a/tests/pytorch/test_numerics.py
+++ b/tests/pytorch/test_numerics.py
@@ -737,10 +737,6 @@ def test_gpt_full_activation_recompute(
         pytest.skip(reason_for_no_fp8)
     if recipe.mxfp8() and not mxfp8_available:
         pytest.skip(reason_for_no_mxfp8)
-    if IS_HIP_EXTENSION:
-        use_cast_transpose_triton =  bool( int(os.environ.get('NVTE_USE_CAST_TRANSPOSE_TRITON', '0')) )
-        if fp8 and recipe.float8_current_scaling() and use_cast_transpose_triton:
-            pytest.skip("Float8 Current Scaling unsupported for full recompute.")
     if recipe.float8_block_scaling() and not fp8_block_scaling_available:
         pytest.skip(reason_for_no_fp8_block_scaling)
 
--- a/tests/pytorch/triton_kernels/test_cast.py
+++ b/tests/pytorch/triton_kernels/test_cast.py
@@ -6,10 +6,12 @@ import pytest
 import torch
 
 from transformer_engine.pytorch.triton_kernels.cast import te_quantize_triton
-from transformer_engine.pytorch.tensor.float8_tensor import Float8Quantizer
+from transformer_engine.pytorch.tensor.float8_tensor import Float8Quantizer, Float8CurrentScalingQuantizer
 from transformer_engine.pytorch.triton_kernels.common import te_dtype_to_torch_dtype
 import transformer_engine_torch as tex
 from test_common import te_compare_results, fill_uniform, get_tolerances
+from transformer_engine.pytorch.fp8 import fp8_autocast
+from transformer_engine.common import recipe
 
 @pytest.mark.parametrize("shape", 
                          [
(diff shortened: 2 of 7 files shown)


tests/pytorch/triton_kernels/test_cast.py
```diff
@@ -112,3 +114,75 @@ def test_quantize_bad_transpose(t_shape, fp8_dtype):
...
+                             (65536, 128),
+                             (65536, 160),
+                             (16384, 1616),
+                             (1, 128),
+                             (1, 1296),
+                             (1, 16),
+                             (5, 160),
+                             (5, 4, 3, 160),
+                             (217, 256),
+                         ])
```
matthiasdiener: Not sure if all of these shapes are necessary for this test (they are copied from the test above).

transformer_engine/pytorch/triton_kernels/cast_transpose.py
```diff
@@ -254,6 +295,28 @@ def te_cast_transpose_noop_triton(input, noop_flag, input_scale, cast_out, trans
...
+        #   3) cast+transpose with that current scale (otherwise same as delayed)
+
+        # global amax
+        amax_out.fill_(-float("inf"))
+        _amax_reduce_triton[grid](
+            input_2d_view,
+            input_stride_M, input_stride_N,
+            num_rows, row_length,
+            amax_out,
+        )
```
matthiasdiener, line 391: Could also use `amax_out = torch.amax(torch.abs(input_2d_view))` here.

tests/pytorch/test_numerics.py
```diff
@@ -737,10 +737,6 @@ def test_gpt_full_activation_recompute(
         pytest.skip(reason_for_no_fp8)
     if recipe.mxfp8() and not mxfp8_available:
         pytest.skip(reason_for_no_mxfp8)
-    if IS_HIP_EXTENSION:
-        use_cast_transpose_triton =  bool( int(os.environ.get('NVTE_USE_CAST_TRANSPOSE_TRITON', '0')) )
```
alextmagro, line 741: I believe there is one other test that we would expect to pass with float8_current_scaling support, test_grouped_linear_accuracy.

Upstream also has the file test_float8_current_scaling_exact.py, do those tests pass as well? If so, we should add them to CI script.

matthiasdiener: I enabled the tests in `test_float8_current_scaling_exact.py`  in [c293b67](https://github.com/ROCm/TransformerEngine/pull/341/commits/c293b6743c4b5a5447f794e2fb64ae49e3f7f04c).

matthiasdiener: [6bcfec1](https://github.com/ROCm/TransformerEngine/pull/341/commits/6bcfec166d52a5fce61b950719b411c3a868d3a7) enables the GroupedLinear tests.

alextmagro: Thanks!

tests/pytorch/triton_kernels/test_cast.py
```diff
@@ -112,3 +114,75 @@ def test_quantize_bad_transpose(t_shape, fp8_dtype):
...
+@pytest.mark.parametrize("in_dtype", [torch.float32, torch.float16, torch.bfloat16])
+@pytest.mark.parametrize("out_dtype", [tex.DType.kFloat8E4M3, tex.DType.kFloat8E5M2])
+def test_quantize_current_scaling_with_autocast(shape, in_dtype, out_dtype):
```
ipanfilo: Maybe merge with test_quantize with adding extra parameter?

matthiasdiener: What do you think of [a57c249](https://github.com/ROCm/TransformerEngine/pull/341/commits/a57c2496a8483349218210d73d612787317302a7)?

matthiasdiener:
This is ready for another review @wenchenvincent @ipanfilo @wangye805

transformer_engine/pytorch/triton_kernels/cast_transpose.py
```diff
@@ -254,7 +341,32 @@ def te_cast_transpose_noop_triton(input, noop_flag, input_scale, cast_out, trans
...
+
+        # Compute scale = fp8_max / amax
+        fp8_max = get_fp8_max(otype)
+        s = (fp8_max / amax_out).to(input_scale.dtype)
```
wenchenvincent: In the cuda/hip kernel, there are some more checks for handling the corner cases.

matthiasdiener: I added the additional checks for corner cases (as well as a test) in https://github.com/ROCm/TransformerEngine/pull/341/commits/02c852c3a4775c8a19eb721978be303ba50b5ea9.

transformer_engine/pytorch/triton_kernels/cast.py
```diff
@@ -96,6 +96,10 @@ def te_quantize_triton(
                 cast_out = out._data
                 trans_out = out._transpose
                 scale_inv_out = out._scale_inv
+
+                from ..tensor.float8_tensor import Float8CurrentScalingQuantizer
```
ipanfilo, line 100: Can it be imported from top of the file or it causes circular import issues?

matthiasdiener: Yes, importing at the top didn't work due to a circular import issue.

matthiasdiener:
Based on the updated performance numbers in https://github.com/ROCm/frameworks-internal/issues/13252#issuecomment-3504217622, I think this is ready for another review.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -125,9 +125,14 @@ def forward(
             recipe = FP8GlobalStateManager.get_fp8_recipe()
             if hasattr(recipe, "fp8_gemm_fprop"):
                 fprop_gemm_use_split_accumulator = recipe.fp8_gemm_fprop.use_split_accumulator
-            inputmats = tex.fused_multi_quantize(
-                inputmats_no_fp8, None, input_quantizers, TE_DType[activation_dtype]
-            )
+
+            inputmats = []
+            for i, x in enumerate(inputmats_no_fp8):
```
wangye805: Does this for-loop introduce performance degradation?

transformer_engine/pytorch/triton_kernels/cast_transpose.py
```diff
@@ -254,7 +374,35 @@ def te_cast_transpose_noop_triton(input, noop_flag, input_scale, cast_out, trans
...
+        amax_out.fill_(-float("inf"))
+        _amax_reduce_triton[grid](
+            input_2d_view,
+            input_stride_M, input_stride_N,
+            num_rows, row_length,
+            amax_out,
+        )
+
+        # Compute scale
+        fp8_max = get_fp8_max(otype)
```
wangye805, line 394: I'm not quite familiar with triton but have a concern: where were this fp8_max stored, in host side or device side? If that returns a host side variable, then we could introduce an unnecessary host/device synchronization which may degrade our performance

wangye805 (changes requested):

transformer_engine/pytorch/triton_kernels/cast_transpose.py
wangye805: I misread this section of codes

transformer_engine/pytorch/module/grouped_linear.py
matthiasdiener: Performance seems to be essentially identical across both HIP implementations (fused/loop):

| Kernel                                                            | Calls (Fused) | Calls (Loop) | Total Ns (Fused) | Total Ns (Loop)
| ----------------------------------------------------------------- | ---------- | ---------- | ----------: | ----------: 
| `cast_transpose_general_kernel<..., te_hip_fp8_e4m3>`             | 4          | 4          |      64,363 |      64,723 
| `amax_kernel<8, true, float>`                                     | 14         | 14         |      60,281 |      61,004 
| `compute_scale_from_amax_kernel`                                  | 14         | 14         |      27,800 |      27,601 
| `cast_transpose_fused_kernel_notaligned<..., e5m2>`               | 4          | 4          |      58,161 |      57,402
| `cast_transpose_optimized_kernel(..., e4m3*, e4m3*, ...)`         | 6          | 6          |      31,122 |      32,042
| `__amd_rocclr_copyBuffer`                                         | 20         | 20         |      78,322 |      77,122
| `__amd_rocclr_fillBufferAligned`                                  | 27         | 27         |      95,203 |      90,881
| `distribution_elementwise_grid_stride_kernel<float,...normal...>` | 8          | 8          |      93,525 |      93,964

If you prefer, I could change this so that the loop is only used when NVTE_USE_CAST_TRANSPOSE_TRITON is enabled.

wangye805: Emm, for your specific test case with current scaling probably it's okay to replace the tex.fused_multi_quantize with a for loop since nvte_multi_cast_transpose (https://github.com/ROCm/TransformerEngine/blob/32e2d1d4d4af5976970dc3ca21bb17c97821a239/transformer_engine/pytorch/csrc/extensions/transpose.cpp#L71) will not be called because with_fused_kernel is set false https://github.com/ROCm/TransformerEngine/blob/32e2d1d4d4af5976970dc3ca21bb17c97821a239/transformer_engine/pytorch/csrc/extensions/transpose.cpp#L38

But for the traditional float8quantizer (delayed scaling), nvte_multi_cast_transpose will be called to dispatch multi_cast_transpose_kernel: https://github.com/ROCm/TransformerEngine/blob/32e2d1d4d4af5976970dc3ca21bb17c97821a239/transformer_engine/common/transpose/multi_cast_transpose.cu#L258. Then maybe the performance will differ. 

I agree with you that we should guard this by NVTE_USE_CAST_TRANSPOSE_TRITON for two reasons:
1). users can choose the original fused multi-tensor cast if using the hipified cuda kernel
2). it keeps the original flow for NV upstream if we are running on NV machines

matthiasdiener: How does https://github.com/ROCm/TransformerEngine/pull/341/commits/242591f7304ced418c79a28da9426748db4ec840 look?

transformer_engine/pytorch/module/grouped_linear.py
ipanfilo, line 1: Please add AMD copyright to this file

ipanfilo (changes requested):

transformer_engine/pytorch/module/grouped_linear.py
matthiasdiener: Added in [0e1e1dd](https://github.com/ROCm/TransformerEngine/pull/341/commits/0e1e1ddea312aa0ec957cc5194ad67dddf8d236d), thanks.

wangye805 (approved):
LGTM

ipanfilo (approved):

PR #344, opened 2025-10-21
Add envvar to cleanup hipified files

alextmagro:
Added NVTE_CLEAN_HIP logic that removes hipified files before building. Useful for devs when switching between TE versions before incrementally building.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -92,11 +92,7 @@ After a successful Transformer Engine installation via `pip install`, execute th
 .. code-block:: bash
 
   cd tests/cpp
-  mkdir build
-  cd build
-  cmake ../
-  make
-  make test
+  cmake -GNinja -Bbuild . && cmake --build build
 
 Note that some of operator unit tests fail in hipBLASLt config due to limited input data configurations support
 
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -92,11 +92,7 @@ After a successful Transformer Engine installation via `pip install`, execute th
 .. code-block:: bash
 
   cd tests/cpp
-  mkdir build
-  cd build
-  cmake ../
-  make
-  make test
+  cmake -GNinja -Bbuild . && cmake --build build
```
ipanfilo, line 95: This command does not perform tests execution like the paragraph suggest. Please add run test command

build_tools/utils.py
```diff
@@ -232,7 +232,20 @@ def rocm_path() -> Tuple[str, str]:
         hipcc_bin = rocm_home / "bin" / "hipcc"
     return rocm_home, hipcc_bin
 
-
+def clean_hipified_files() -> None:
+    """Removes hipified files and regenerates them -- useful when switching between TE versions"""
```
ipanfilo, line 236: This should be doable by git clean that has advantage of not cleaning hip files that are part of repository

ipanfilo (changes requested):

README.rst
alextmagro: Will do

build_tools/utils.py
alextmagro: Good point -- wasn't aware you could pass substrings to git clean. Will revert these changes and update PR.

PR #349, opened 2025-10-23
Te2.4 fsdp2 fp8 allgather autocast

sudhu2k:
# Description

This PR introduces the following improvements on FSDP2 pytorch implementation on TE.
- Performing all gather in uint8 instead of higher precision data for fp8 autocast path to reduce bandwidth utilization.
- Discarding the fp8 quantized tensor after used in forward and backward passes to help reduce the memory more.

Fixes [#13926](https://github.com/ROCm/frameworks-internal/issues/13926)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Introduces a custom tensor subclass to wrap the parameters with when use_fsdp2 flag is set
- The custom tensor subclass implements fsdp_pre_all_gather and fsdp_pre_all_gather functions appropriately
- The hooks are written so it works for higher precision and also fp8 scaling methods like mxfp8, delayed and current scaling.
- Moves all-reduce and update of scales from forward to backward passes in modules when use_fsdp2 is enabled.

# Results
https://github.com/ROCm/frameworks-internal/issues/13926#issuecomment-3429151572
https://github.com/ROCm/frameworks-internal/issues/13926#issuecomment-3433479561

# Feature Checklist

Checklist:
- [x] Implement all gather hooks with Float8Tensors (delayed) scaling
- [x] Unit test for sanity check
- [x] Perform initial analysis on memory reduction
- [x] Extend this to layernorm mlp and layernorm linear.
- [x] Verify with FP8 current scaling recipe
- [x] Enable MXFP8
- [ ] Finish up the documentation (https://amd.atlassian.net/wiki/spaces/~7120207a46f504e5fc41ffa10a6da189d8402d/pages/1152247003/FSDP2+Memory+Reduction+with+Hooks+-+Design+Documentation+In-progress...)
- [x] Perform analysis with Megatron-LM using the llama3 model.
- [x] Perform analysis on bandwidth and memory utilization.
- [x] Create PR

# Benchmarking Results:
image

# Checklist:


qa/L1_pytorch_distributed_unittest/test.sh
```diff
@@ -25,6 +25,7 @@ pip3 install pytest==8.2.1 || error_exit "Failed to install pytest"
 python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_numerics.xml $TE_PATH/tests/pytorch/distributed/test_numerics.py || test_fail "test_numerics.py"
 python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_fusible_ops.xml $TE_PATH/tests/pytorch/distributed/test_fusible_ops.py || test_fail "test_fusible_ops.py"
 python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_torch_fsdp2.xml $TE_PATH/tests/pytorch/distributed/test_torch_fsdp2.py || test_fail "test_torch_fsdp2.py"
+python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_torch_fsdp2_fp8.xml $TE_PATH/tests/pytorch/distributed/test_torch_fsdp2_fp8.py || test_fail "test_torch_fsdp2_fp8.py"
```
wangye805: This qa directory is for NV upstream, let's not modify it right now

tests/pytorch/distributed/test_torch_fsdp2_fp8.py
wangye805, line 1: Can we merge these two files into the original test_torch_fsdp2.py and run_fsdp2_model.py since they share a lot of common codes? It will save us a lot of maintaining trouble in the future

transformer_engine/pytorch/module/base.py
```diff
@@ -43,6 +43,7 @@
 from ..tensor.float8_tensor import Float8Quantizer, Float8CurrentScalingQuantizer
 from ..tensor.mxfp8_tensor import MXFP8Quantizer
 from ..tensor.float8_blockwise_tensor import Float8BlockQuantizer
+from ..tensor.fsdp2_allgather_tensor import FSDPAGTensor
```
wangye805: Guard by IS_HIP_EXTENSION?

transformer_engine/pytorch/module/base.py
```diff
@@ -1153,6 +1160,14 @@ def reset_parameters(self, defer_init: Optional[bool] = False) -> None:
                 if IS_HIP_EXTENSION and not self.keep_fp8_weight_transpose_cache:
                     quantizer.columnwise_usage=False
                 param = quantizer(param)
+            if self.use_fsdp2 and not self.primary_weights_in_fp8 and fp8_meta_index is not None:
+                self.keep_fp8_weight_transpose_cache = False
```
wangye805, line 1165: In line 945, we forced keep_fp8_weight_transpose_cache to true in mxfp8. But here we force it to false. Do we need to make it to true for mxfp8 cases?

transformer_engine/pytorch/optimizers/multi_tensor_apply.py
```diff
@@ -16,7 +16,7 @@ def __call__(self, op, noop_flag_buffer, tensor_lists, *args):
         for i, ts in enumerate(tensor_lists):
             for j, t in enumerate(ts):
                 if isinstance(t, DTensor):
-                    tensor_lists[i][j] = t._local_tensor
+                    tensor_lists[i][j] = t._local_tensor.data
```
wangye805: Is this a bug or required by our new PR (this one)? If this is required by our new PR, let's guard this by IS_HIP_EXTENSION

transformer_engine/pytorch/tensor/fsdp2_allgather_tensor.py
```diff
@@ -0,0 +1,199 @@
...
+        # If metadata isn't initialized yet, we can't access the quantizers
+        if not self._module.fp8:
+            module_class_name = self._module.__class__.__name__  
+            if "LayerNormMLP" in module_class_name:  
+                num_gemms = 2  
+            else:  # Linear, LayerNormLinear, etc.  
+                num_gemms = 1  
+
+            self._module.init_fp8_metadata(num_gemms=num_gemms)
+        if not self._module.fp8:
```
wangye805, line 144: Is line 144 redundant? I mean will sef._module.fp8 be changed from line137 to 141?

wangye805 (changes requested):

transformer_engine/pytorch/tensor/fsdp2_allgather_tensor.py
sudhu2k: It's not redundant,  self._module.init_fp8_metadata(num_gemms=num_gemms) line sets the fp8 metadata and is responsible for setting the fp8 boolean value. The fp8 boolean value is set based on FP8GlobalStateManager.is_fp8_enabled().
https://github.com/ROCm/TransformerEngine/blob/ba7f3d57bcc081903e02dab93d8eff0a10207c3a/transformer_engine/pytorch/module/base.py#L902-L908
Once the initialization is done, Line 144 checks if fp8 is enabled or not, if not, it does all gather on higher precision data.

https://github.com/ROCm/TransformerEngine/blob/ba7f3d57bcc081903e02dab93d8eff0a10207c3a/transformer_engine/pytorch/tensor/fsdp2_allgather_tensor.py#L136-L143

If the data is higher precision, then it does go into thIS first "if block" over and over again since module.fp8 won't be true, but it won't be much of an overhead since it just runs the below 4 assignment operations.
https://github.com/ROCm/TransformerEngine/blob/ba7f3d57bcc081903e02dab93d8eff0a10207c3a/transformer_engine/pytorch/module/base.py#L904-L908

transformer_engine/pytorch/module/base.py
sudhu2k: The reset_parameters runs during initialization of model parameters but fp8 autocast will be run during forward pass. The forward pass runs init_fp8_metadata which sets all the fp8 metadata for that module. Hence it would be good to force set the keep_fp8_weight_transpose_cache to True once we know what recipe is being used during the forward pass. Otherwise, we wouldn't know what recipe is being used during initialization. 

So, I set that flag to False during initialization since it'll be false for delayed scaling and current scaling and just force set it to True during fp8 metadata initialization for mxfp8.

tests/pytorch/distributed/test_torch_fsdp2_fp8.py
sudhu2k: Yes, that's how I had it before, but since it'll make the IFUs a little bit more complicated (if upstream changes these files) I chose to separate them. Let me know what you think.

transformer_engine/pytorch/optimizers/multi_tensor_apply.py
sudhu2k: The TE fused optimizers call this function to extract data from custom tensor subclasses, since the parameters will be of type FSDPAGTensor, we had to extract only the tensor data part of the custom subclass. Also, .data is redundant for native tensors so it doesn't change anything.

https://github.com/ROCm/TransformerEngine/blob/ba7f3d57bcc081903e02dab93d8eff0a10207c3a/transformer_engine/pytorch/tensor/fsdp2_allgather_tensor.py#L60-L63

tests/pytorch/distributed/test_torch_fsdp2_fp8.py
wangye805: I checked the diffs: 
run_fsdp2_model.py vs run_fsdp2_fp8_model.py: 61 removal and 187 addition for 189 lines
test_torch_fsdp2.py vs test_torch_fsdp2_fp8.py: 21 removal and 74 addition for 111 lines. 

I think the diff is quite significant. Let's keep those files separate

tests/pytorch/distributed/test_torch_fsdp2_fp8.py
```diff
@@ -0,0 +1,111 @@
...
+    if fp_init:
+        test_cmd += ["--fp8-init"]
+    test_cmd += ["--recipe", recipe]
+    
+    subprocess.run(test_cmd + ['--use-fsdp2','--gradients-save-file', 'all_iters_fsdp2.pt'], env=os.environ, check=True)
+    subprocess.run(test_cmd + ['--gradients-save-file', 'all_iters_dp.pt'], env=os.environ, check=True)
+        
+    # Load outputs
+    output_fsdp = torch.load("all_iters_fsdp2.pt", map_location="cpu")
+    output_dp = torch.load("all_iters_dp.pt", map_location="cpu")
```
wangye805, line 62: Don't forget to delete those files generated in this test

tests/pytorch/distributed/test_torch_fsdp2_fp8.py
```diff
@@ -0,0 +1,111 @@
...
+    input_path = Path("shared_input.pt")
+    if input_path.exists():
+        input_data = torch.load(input_path).to('cuda')
+    else:
+        input_data = torch.randn(batch_size, input_size, requires_grad=True).to('cuda')
+        torch.save(input_data.cpu(), input_path)
+        print("Generated and saved shared input tensor.")
+
+    model = SimpleNet(input_size, 2048, 2048)
+    torch.save(model.state_dict(), 'fsdp_model.pth')
```
wangye805, line 99: Same here. Don't forget to clean intermediate files up

transformer_engine/pytorch/tensor/fsdp2_allgather_tensor.py
wangye805: I see. Got it. Thanks

transformer_engine/pytorch/fp8.py
```diff
@@ -494,7 +495,7 @@ def fp8_autocast_exit(cls, enabled: bool, _graph: bool) -> None:
         # Reduce only the non-FP8 weight modules here.
         # FP8 weight modules are reduced at the end of the optimizer
         # step after the weight amax is populated.
-        if enabled and cls.FP8_AUTOCAST_DEPTH == 0 and not _graph and torch.is_grad_enabled():
+        if not cls.SKIP_FP8_REDUCTION_FOR_FSDP2 and enabled and cls.FP8_AUTOCAST_DEPTH == 0 and not _graph and torch.is_grad_enabled():
```
wangye805, line 498: I checked the activation recompute. It seems that your changes will cause some issues.

Use our TE activation recompute testcase (https://github.com/ROCm/TransformerEngine/blob/a037a09c0c27a7435aec85a72c862708dda7df10/tests/pytorch/test_numerics.py#L764) an example: 
1). In this example, fwd pass is inside a fp8_autocast context manager: https://github.com/ROCm/TransformerEngine/blob/a037a09c0c27a7435aec85a72c862708dda7df10/tests/pytorch/test_numerics.py#L691-L707. Upon exiting, our dev branch will update the fp8 amax but your PR branch will not update the amax
2). In the fwd computation inside activation recompute, both your PR branch and our dev branch will try to retrieve the previous fp8 meta data (scale, amax) for quantizers: https://github.com/ROCm/TransformerEngine/blob/a037a09c0c27a7435aec85a72c862708dda7df10/transformer_engine/pytorch/module/base.py#L955 
3). In the bwd, there is another fp8_autocast context manager: https://github.com/ROCm/TransformerEngine/blob/a037a09c0c27a7435aec85a72c862708dda7df10/transformer_engine/pytorch/distributed.py#L417-L420. Upon existing this context manager in fwd, our dev branch will do the quantize again, and update the fp8 amax again. However your branch will do the amax update at the end of bwd

To conclude for this activation recompute case:
dev branch without activation recompute: quantize --> update fp8 meta 
dev branch:        quantize --> update fp8 meta --> get to previous fp8 meta --> quantize --> update fp8 meta
your PR branch: quantize --> get to previous fp8 meta --> quantize --> quantize --> update fp8 meta

wangye805, line 498: By the way, this condition also contain _graph in the cuda graph mode, which means that we may need to modify https://github.com/ROCm/TransformerEngine/blob/a037a09c0c27a7435aec85a72c862708dda7df10/transformer_engine/pytorch/graph.py#L466 using your current idea (move the update from fwd to bwd)

So perhaps saving a scale copy somewhere in fwd , then modify the quantizer's scale in newly added fsdp_pre_all_gather can help us align with upstream better? 

in the activation recompute case: 
quantize (use saved scale) --> update fp8 meta --> get to previous fp8 meta --> quantize (use the saved scale) --> update fp8 meta
in non activation recompute case:
quantize (use saved scale) --> update fp8 meta --> quantize (use the cached scale)

wangye805 (changes requested):

transformer_engine/pytorch/fp8.py
sudhu2k: ## In the current dev branch:

### Forward
1) **Before forward**: When entering prepare_forward() during the initial forward pass (recompute_phase=False), the system calls FP8GlobalStateManager.copy_forward_fp8_meta_tensors_for_recompute() to save a copy of the current FP8 metadata (amax_history and scale tensors).
2) **During forward**: The forward pass executes with FP8 quantization, accumulating new amax values.
3) **After forward (autocast exit)**: When fp8_autocast exits and FP8_AUTOCAST_DEPTH reaches 0, it calls reduce_and_update_fp8_tensors(forward=True) which updates the amax_history and scale tensors based on the accumulated amax values from the forward pass.

### Recomputation During Backward
1) **Before recompute:** When entering prepare_forward() during the recompute phase (recompute_phase=True), the system calls FP8GlobalStateManager.get_old_fp8_meta_tensors_for_recompute() which:
- Saves the updated metadata from phase 1 to temporary storage
- Restores the original metadata that was copied before phase 1.

2) **During Recompute:** The forward pass executes again with the same FP8 scaling factors as the initial forward, ensuring numerical consistency.

3) **After recompute:** When exiting prepare_forward(), the system calls restore_fp8_meta_tensors() restores the updated metadata from phase 1, which effectively discards any amax values that were accumulated during the recompute.

## With the new solution:

### Forward
1) **Before forward**: 
- pre all gather quantizes the input sharded tensor and is all gathered. So, the quantizers would have its scale that was used to quantize and amax that was calculated on each of the sharded tensor. Which means each GPU would have the same quantizer with different amax values.
- Again, when entering prepare_forward() during the initial forward pass (recompute_phase=False), the system calls FP8GlobalStateManager.copy_forward_fp8_meta_tensors_for_recompute() to save a copy of the current FP8 metadata (amax_history and scale tensors). The difference between the current dev branch and the new implementation is that, quantization has already happened and fp8_metadata has already been initialized. Even though the quantization has happened already, it doesn't change the amax_history or the scale tensors until after the all-reduce which mean the fp8 meta tensors which are copied for recompute in this case would be the same as the current dev.
2) **During forward**: The forward pass executes with already FP8 quantized parameters.
3) **After forward (autocast exit)**: Nothing happens here

### Recomputation During Backward
1) **Before recompute**: When entering prepare_forward() during the recompute phase (recompute_phase=True), the system calls FP8GlobalStateManager.get_old_fp8_meta_tensors_for_recompute() which:
- Saves the updated metadata from phase 1 to temporary storage
- Restores the original metadata that was copied before phase 1.

2) **During recompute:** The forward pass executes again with the same FP8 scaling factors as the initial forward, ensuring numerical consistency.

3) **After recompute:** When exiting prepare_forward(), the system calls restore_fp8_meta_tensors() restores the updated metadata from phase 1, which effectively discards any amax values that were accumulated during the recompute.


In conclusion:
Since the storing of fp8 meta tensors happen before forward, which means both the dev/current PR would be using the fp8 meta tensors which were calculated until the earlier iteration. The fp8 meta tensors which were calculated until the earlier iteration, would be the same.

wangye805: Thanks for your investigation and detailed explanation. Now since get_old_fp8_meta_tensors_for_recompute() is retrieving the meta from a dedicated recompute fp8 buffer, not the general fp8 meta history, then your branch should be good. 

How about the cudagraph mode? Will your PR also be compatible?

sudhu2k: For cudagraph mode, it won't be compatible. I'll add a commit now and request your review. Thanks for catching that!

wangye805 (approved):
LGTM. 

Don't forget to include those activation recompute and cuda graph related discussions into the design doc so future maintainers can learn about this

PR #355, opened 2025-10-28
FIX Occasional import error when only building for a single framework

Micky774:
# Description

Sometimes, during import, both frameworks will _try_ to get imported but in the event that their relevant `.so` is not found, will throw a `RuntimeError` that currently **is not caught** by the try/catch clauses, breaking an otherwise viable build. This PR fixes that.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/__init__.py
+++ b/transformer_engine/__init__.py
@@ -30,12 +30,12 @@ if os.getenv("NVTE_FRAMEWORK"):
 
 try:
     if _use_pytorch: from . import pytorch
-except (ImportError, FileNotFoundError, RuntimeError):
+except (ImportError, FileNotFoundError):
     pass
 
 try:
     if _use_jax: from . import jax
-except (ImportError, FileNotFoundError, RuntimeError):
+except (ImportError, FileNotFoundError):
     pass
 
 try:
--- a/transformer_engine/__init__.py
+++ b/transformer_engine/__init__.py
@@ -30,12 +30,12 @@ if os.getenv("NVTE_FRAMEWORK"):
 
 try:
     if _use_pytorch: from . import pytorch
-except (ImportError, FileNotFoundError):
+except (ImportError, FileNotFoundError, RuntimeError):
     pass
 
 try:
     if _use_jax: from . import jax
-except (ImportError, FileNotFoundError):
+except (ImportError, FileNotFoundError, RuntimeError):
     pass
 
 try:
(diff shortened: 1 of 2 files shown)


transformer_engine/__init__.py
```diff
@@ -30,12 +30,12 @@
 
 try:
     if _use_pytorch: from . import pytorch
-except (ImportError, FileNotFoundError):
+except (ImportError, FileNotFoundError, RuntimeError):
```
ipanfilo: RuntimeError is too wide exception and supposedly wrong  - if .so is not found AssertionError is rather thrown. Better wrap/replace assertion in _get_shared_object_file() to throw more specific FileNotFoundError

Micky774: Done!

transformer_engine/common/__init__.py
ipanfilo: AssertionError raises here in test that I performed

ipanfilo (changes requested):

wangye805:
I'm trying to understand the motivation of this PR. By "their relevant .so", are you referring to transformer_engine_pytorch.so and transformer_engine_jax.so? Why do we try to import both? Should we first check whether pytorch or jax are included in the docker container and whether we built pytorch/jax TE extension before we import them? 

Maybe you can post more details like which docker image you are using, how you build TE, and what's the error message you saw. It will help future developers and maintainers to understand :-)

transformer_engine/common/__init__.py
Micky774: I've now converted this to a `FileNotFoundError` so that it is caught correctly.

Micky774:
> I'm trying to understand the motivation of this PR. By "their relevant .so", are you referring to transformer_engine_pytorch.so and transformer_engine_jax.so?

Yes
> Why do we try to import both? Should we first check whether pytorch or jax are included in the docker container and whether we built pytorch/jax TE extension before we import them?

That is currently _how we check_. We try to import, and if it fails in expected ways then we deem it not installed/supported. Specifically, during import of an extension, it attempts to load the extension's `.so`. The overall import is guarded by a `try/catch` block. The problem is that the current implementation doesn't produce the correct errors (namely `FileNotFoundError`) and instead produces `AssertionError`/`RuntimeError` which are not caught via the `try/catch` block and break the runtime import of TE.

So this PR is mainly corrected how our implementation handles certain cases of not encountering a particular framework extension `.so`

ipanfilo (approved):

wangye805 (approved):

PR #356, opened 2025-10-29
Experimental rocSHMEM support

alextmagro:
Requires support for rocSHMEM within pytorch for full functionality -- Able to bootstrap with PMI codechanges but not safe or recommended.

rocSHMEM also lacks stream specific functionality, and signaling. Given that Pytorch communications tend to be stream focused, this is a feature we need from rocSHMEM team.

### DIFF
--- a/transformer_engine/common/CMakeLists.txt
+++ b/transformer_engine/common/CMakeLists.txt
@@ -382,6 +382,26 @@ if(USE_CUDA)
 
 # Hack to enable dynamic loading in cuDNN frontend
 target_compile_definitions(transformer_engine PUBLIC NV_CUDNN_FRONTEND_USE_DYNAMIC_LOADING)
+else()
+  option(NVTE_ENABLE_ROCSHMEM "Compile with ROCSHMEM library" OFF)
+  if (NVTE_ENABLE_ROCSHMEM)
+    add_subdirectory(rocshmem_api)
+    if(DEFINED ENV{ROCSHMEM_HOME})
+      set(ROCSHMEM_HOME "$ENV{ROCSHMEM_HOME}" CACHE STRING "Location of ROCSHMEM installation")
+    else()
+      set(ROCSHMEM_HOME "/opt/rocm" CACHE STRING "Location of ROCSHMEM installation (default)")
+    endif()
+    target_link_options(transformer_engine PRIVATE
+        -fgpu-rdc
+    )
+    target_link_libraries(transformer_engine PUBLIC 
+        -Wl,--whole-archive
+        rocshmemapi
+        "${ROCSHMEM_HOME}/lib/librocshmem.a"
+        -Wl,--no-whole-archive
+    )
+    target_include_directories(transformer_engine PUBLIC ${ROCSHMEMAPI_INCLUDE_DIR})
+  endif()
 endif()
 
 # Helper functions to make header files with C++ strings
--- a/transformer_engine/common/CMakeLists.txt
+++ b/transformer_engine/common/CMakeLists.txt
@@ -386,32 +386,13 @@ target_compile_definitions(transformer_engine PUBLIC NV_CUDNN_FRONTEND_USE_DYNAM
 else()
   option(NVTE_ENABLE_ROCSHMEM "Compile with ROCSHMEM library" OFF)
   if (NVTE_ENABLE_ROCSHMEM)
-    find_package(MPI REQUIRED)
-    target_link_libraries(transformer_engine PUBLIC MPI::MPI_CXX)
-    target_include_directories(transformer_engine PRIVATE ${MPI_CXX_INCLUDES})
     add_subdirectory(rocshmem_api)
-    if(DEFINED ENV{ROCSHMEM_HOME})
-      set(ROCSHMEM_HOME "$ENV{ROCSHMEM_HOME}" CACHE STRING "Location of ROCSHMEM installation")
-    else()
-      set(ROCSHMEM_HOME "/opt/rocm" CACHE STRING "Location of ROCSHMEM installation (default)")
-    endif()
-    set(ROCSHMEM_LIBRARY_PATH "${ROCSHMEM_HOME}/lib/librocshmem.a")
-    if (EXISTS ${ROCSHMEM_LIBRARY_PATH})
-      add_library(rocshmem STATIC IMPORTED)
-      set_target_properties(rocshmem PROPERTIES
-        IMPORTED_LOCATION "${ROCSHMEM_LIBRARY_PATH}"
-        IMPORTED_LINK_INTERFACE_LANGUAGES "CXX"
-      )
-    else()
-      message(FATAL_ERROR "ROCSHMEM library not found at ${ROCSHMEM_LIBRARY_PATH}. PLease set ROCSHMEM_HOME.")
-    endif()
     target_link_options(transformer_engine PRIVATE
         -fgpu-rdc
     )
     target_link_libraries(transformer_engine PUBLIC 
         -Wl,--whole-archive
         rocshmemapi
-        rocshmem
         -Wl,--no-whole-archive
     )
   endif()
--- a/transformer_engine/common/CMakeLists.txt
+++ b/transformer_engine/common/CMakeLists.txt
@@ -238,6 +238,7 @@ else()
       IGNORES "*/amd_detail/*"
       IGNORES "*/aotriton/*"
       IGNORES "*/ck_fused_attn/*"
+      IGNORES "*/rocshmem_api/*"
       CUSTOM_MAP_FILE "${TE}/hipify_custom_map.json"
       NO_MATH_REPLACE
   )
@@ -385,6 +386,9 @@ target_compile_definitions(transformer_engine PUBLIC NV_CUDNN_FRONTEND_USE_DYNAM
 else()
   option(NVTE_ENABLE_ROCSHMEM "Compile with ROCSHMEM library" OFF)
   if (NVTE_ENABLE_ROCSHMEM)
+    find_package(MPI REQUIRED)
+    target_link_libraries(transformer_engine PUBLIC MPI::MPI_CXX)
+    target_include_directories(transformer_engine PRIVATE ${MPI_CXX_INCLUDES})
     add_subdirectory(rocshmem_api)
     if(DEFINED ENV{ROCSHMEM_HOME})
       set(ROCSHMEM_HOME "$ENV{ROCSHMEM_HOME}" CACHE STRING "Location of ROCSHMEM installation")
--- a/transformer_engine/common/CMakeLists.txt
+++ b/transformer_engine/common/CMakeLists.txt
@@ -395,13 +395,23 @@ else()
     else()
       set(ROCSHMEM_HOME "/opt/rocm" CACHE STRING "Location of ROCSHMEM installation (default)")
     endif()
+    set(ROCSHMEM_LIBRARY_PATH "${ROCSHMEM_HOME}/lib/librocshmem.a")
+    if (EXISTS ${ROCSHMEM_LIBRARY_PATH})
+      add_library(rocshmem STATIC IMPORTED)
+      set_target_properties(rocshmem PROPERTIES
+        IMPORTED_LOCATION "${ROCSHMEM_LIBRARY_PATH}"
+        IMPORTED_LINK_INTERFACE_LANGUAGES "CXX"
+      )
+    else()
+      message(FATAL_ERROR "ROCSHMEM library not found at ${ROCSHMEM_LIBRARY_PATH}. PLease set ROCSHMEM_HOME.")
+    endif()
     target_link_options(transformer_engine PRIVATE
         -fgpu-rdc
     )
     target_link_libraries(transformer_engine PUBLIC 
         -Wl,--whole-archive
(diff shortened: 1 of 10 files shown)


transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -303,6 +303,44 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
   m.def("nvshmem_finalize", &transformer_engine::pytorch::nvshmem_finalize,
         "Clean up and finalize the NVSHMEM communication backend and free associated resources",
         py::call_guard<py::gil_scoped_release>());
+#else
+  // rocshmem functions
```
wenchenvincent: I think we don't need these function names at python level. We can just use the nvshmem wrappers at python level.

transformer_engine/common/rocshmem_api/rocshmem_waitkernel.hip
```diff
@@ -0,0 +1,114 @@
...
+              wait_kind <= WaitKind::STREAM_WAIT,
+              "Invalid wait kind");
+
+  switch (wait_kind) {
+// ### wait_until_on_stream not yet implemented for rocshmem  ###
+// ### KernelWait is robust but slightly slower due to launch ###
+    case WaitKind::ROCSHMEM_WAIT:
+    //   rocshmem__ulonglong_wait_until_on_stream(sig_addr,
+    //                                 ROCSHMEM_CMP_EQ,
+    //                                 wait_value,
```
wenchenvincent, line 60: Let's add a warning here saying that rocshmem_wait is not supported on ROCm.

wenchenvincent:
@alextmagro Could you please address the comments?
Also, please rebase/merge latest dev to incorporate fixes for sgpu tests.

alextmagro:
> @alextmagro Could you please address the comments? Also, please rebase/merge latest dev to incorporate fixes for sgpu tests.

I have made the changes requested and rebased. A few issues with submodules that were added accidentally but things should be clean now -- will run level 3.

wenchenvincent (approved):
LGTM. Was the CI failure due to networking issue?

transformer_engine/common/rocshmem_api/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+# License for AMD contributions = MIT. See LICENSE for more information
+cmake_minimum_required (VERSION 3.21)
+project(rocshmem LANGUAGES HIP)
+
+find_package(hipblaslt REQUIRED)
+find_package(hiprtc REQUIRED)
+find_package(hip REQUIRED)
+find_package(MPI REQUIRED)
+
+if(NOT DEFINED ENV{NVTE_ROCM_ARCH})
```
ipanfilo: It should be inherited from parent CMakeLists

transformer_engine/common/libtransformer_engine.version
```diff
@@ -18,7 +18,8 @@
 			*transformer_engine::CommOverlapP2PBase*;
 			*transformer_engine::CommOverlapCore*;
 			*nvshmem_wait_on_stream*;
-			*nvshmemi_init_thread*
+			*nvshmemi_init_thread*;
+			*rocshmem*
```
ipanfilo: should it be te_rocshmem to be more specific?

transformer_engine/common/rocshmem_api/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
+# License for AMD contributions = MIT. See LICENSE for more information
+cmake_minimum_required (VERSION 3.21)
+project(rocshmem LANGUAGES HIP)
+
+find_package(hipblaslt REQUIRED)
+find_package(hiprtc REQUIRED)
+find_package(hip REQUIRED)
+find_package(MPI REQUIRED)
```
ipanfilo, line 8: Does rocshmemapi really use hipblasLt, hipRTC and MPI?

transformer_engine/common/rocshmem_api/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+else()
+  set(ROCSHMEM_HOME "/opt/rocm" CACHE STRING "Location of ROCSHMEM installation (default)")
+endif()
+
+set(ROCSHMEM_INCLUDE_DIR "${ROCSHMEM_HOME}/include/rocshmem")
+if(NOT EXISTS "${ROCSHMEM_INCLUDE_DIR}")
+  set(ROCSHMEM_INCLUDE_DIR "${ROCSHMEM_HOME}/include")
+endif()
+
+add_library(rocshmemapi OBJECT rocshmem_waitkernel.hip)
```
ipanfilo: Why not making it a static library so it can include lib/librocshmem.a?

transformer_engine/common/rocshmem_api/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+
+target_compile_options(rocshmemapi PRIVATE
```
ipanfilo, line 12: Is it because HIP_SEPARABLE_COMPILATION is not actually supported? Move it closer to target properties then

transformer_engine/common/rocshmem_api/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+)
+
+set_target_properties(rocshmemapi PROPERTIES
+    CXX_STANDARD 17
+    HIP_STANDARD 17
+    POSITION_INDEPENDENT_CODE ON
+    HIP_SEPARABLE_COMPILATION ON
+)
+
+set(ROCSHMEMAPI_INCLUDE_DIR "${ROCSHMEM_INCLUDE_DIR}" PARENT_SCOPE)
```
ipanfilo: Seems unused in parent scope

transformer_engine/common/CMakeLists.txt
```diff
@@ -382,6 +381,26 @@ if(USE_CUDA)
...
+    target_link_options(transformer_engine PRIVATE
+        -fgpu-rdc
+    )
+    target_link_libraries(transformer_engine PUBLIC 
+        -Wl,--whole-archive
+        rocshmemapi
+        "${ROCSHMEM_HOME}/lib/librocshmem.a"
+        -Wl,--no-whole-archive
+    )
+    target_include_directories(transformer_engine PUBLIC ${ROCSHMEMAPI_INCLUDE_DIR})
```
ipanfilo: Seems unused?

transformer_engine/common/CMakeLists.txt
```diff
@@ -382,6 +381,26 @@ if(USE_CUDA)
...
+    add_subdirectory(rocshmem_api)
+    if(DEFINED ENV{ROCSHMEM_HOME})
+      set(ROCSHMEM_HOME "$ENV{ROCSHMEM_HOME}" CACHE STRING "Location of ROCSHMEM installation")
+    else()
+      set(ROCSHMEM_HOME "/opt/rocm" CACHE STRING "Location of ROCSHMEM installation (default)")
+    endif()
+    target_link_options(transformer_engine PRIVATE
+        -fgpu-rdc
+    )
+    target_link_libraries(transformer_engine PUBLIC
```
ipanfilo, line 393: Can it be specified in target_link_libraries for rocshmem_api?

alextmagro: I don't think so. My understanding is as rocshmem provides a static library, it needs to be linked directly to the parent for symbols to be defined in libtransformer_engine.so.

transformer_engine/common/rocshmem_api/CMakeLists.txt
alextmagro: rocshmem needs MPI to initialize. Don't need hipblaslt & hipRTC; was following the style of nvshmemapi/CMakeLists.txt, and editing for us. Will remove.

transformer_engine/common/rocshmem_api/CMakeLists.txt
alextmagro: I had this here so we could build rocshmem_api without building the full TE, but think its ok to remove now.

transformer_engine/common/rocshmem_api/rocshmem_waitkernel.hip
ipanfilo, line 1: Since this code is not hipified, lets add rocshmem_api explicitly to IGNORES in hipify call in tensformer_engine/common/CMakeLists.txt

transformer_engine/common/CMakeLists.txt
ipanfilo: It might propagate from there if you make rocshmemapi STATIC library instead of OBJECT. And how does it work now w/o MPI linking here?

transformer_engine/common/CMakeLists.txt
```diff
@@ -382,6 +382,25 @@ if(USE_CUDA)
 
 # Hack to enable dynamic loading in cuDNN frontend
 target_compile_definitions(transformer_engine PUBLIC NV_CUDNN_FRONTEND_USE_DYNAMIC_LOADING)
+else()
+  option(NVTE_ENABLE_ROCSHMEM "Compile with ROCSHMEM library" OFF)
+  if (NVTE_ENABLE_ROCSHMEM)
+    add_subdirectory(rocshmem_api)
+    if(DEFINED ENV{ROCSHMEM_HOME})
```
ipanfilo: Can ROCSHMEM_HOME be defined inside rocshmem_api CMAke? 
Or better declare rocshmem library as IMPORTED and set IMPORTED_LOCATION there, so you'll be able to use just rocshmem in target_link_libraries below instead of "${ROCSHMEM_HOME}/lib/librocshmem.a"

transformer_engine/pytorch/csrc/extensions/rocshmem_comm.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+#include "../extensions.h"
+
+#ifdef NVTE_ENABLE_ROCSHMEM
+#include <mpi.h>
+#include <rocshmem_api/rocshmem_waitkernel.hpp>
+#endif
+
+#include <cuda.h>
+#include <cuda_fp8.h>
```
ipanfilo: FP8 seems unused here. 
And it looks like all headers can be guarded with ROCSHMEM not only mpi and rocshmem_api, that also leads to that use hip/hip_runtime.h can be used instead of cuda.h

ipanfilo (changes requested):

transformer_engine/common/CMakeLists.txt
alextmagro: I had a lot of linking issues when trying to make rocshmem_api static. I ended up going with an object as we can link the MPI into the object.

transformer_engine/common/CMakeLists.txt
alextmagro: I tried declaring rocshmem library as IMPORTED -- please have a look at 77158d to see if this is what you meant. Thanks!

ipanfilo: But why all of this is made here? The idea was to move it to rocshmem_api/CMakeLists.txt so here there are only add_subdirectory() and target_link_options() and target_link_libraries() - bare minimum. 
When you define rocshmem library or find MPI package in child you can use MPI:MPI_CXX or rocshmem in target_link_libraries after that

transformer_engine/common/rocshmem_api/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+    POSITION_INDEPENDENT_CODE ON
+)
+
+target_include_directories(rocshmemapi PUBLIC
+        "${ROCSHMEM_INCLUDE_DIR}"
+        "${CMAKE_CURRENT_SOURCE_DIR}"
+        "${MPI_INCLUDE_PATH}"
+)
+
+target_link_libraries(rocshmemapi PUBLIC
```
ipanfilo, line 49: This seems having no effect since it is OBJECT library and actual linking dependencies are set in parent CMakeLists.txt

transformer_engine/common/CMakeLists.txt
```diff
@@ -382,6 +383,38 @@ if(USE_CUDA)
 
 # Hack to enable dynamic loading in cuDNN frontend
 target_compile_definitions(transformer_engine PUBLIC NV_CUDNN_FRONTEND_USE_DYNAMIC_LOADING)
+else()
+  option(NVTE_ENABLE_ROCSHMEM "Compile with ROCSHMEM library" OFF)
+  if (NVTE_ENABLE_ROCSHMEM)
+    find_package(MPI REQUIRED)
+    target_link_libraries(transformer_engine PUBLIC MPI::MPI_CXX)
+    target_include_directories(transformer_engine PRIVATE ${MPI_CXX_INCLUDES})
```
ipanfilo: these include directories seem unused in current scope

transformer_engine/common/CMakeLists.txt
alextmagro: Ah, understood. I have moved everything to the rocshmem_api project.

ipanfilo (approved):

PR #361, opened 2025-11-04
std::max type mismatch hotfix

alextmagro:
hotfix for cpp unit test build failure. This issue occurred when merging the hipify_transpilation ticket at the same time as the mxfp8 cast off by one ticket.

Rebase resolved correctly, however, in the cast off by 1 ticket std::max is mapped to ::max,  where the integer function of max is used for this, causing scale_tol calculation to be cast to an int then back to float:

const float scale_tol = ::max(1.f, row_blocks * col_blocks * tol);.

In rocm 7.0, there is no function template for std::max(float, double), so build fails. Making all variables/values doubles fixes the issue.


wangye805:
Do you mind have more details about the story like which tests failed and why, and why this new commit can fix it? Therefore we can understand it in the future. Thanks

wangye805 (approved):

ipanfilo (approved):

PR #367, opened 2025-11-11
Userbuffer epic

alextmagro:
This is the userbuffer_epic branch, to be merged only once all epic tasks have been completed. PRs for epic tasks will be onto this branch.


build_tools/pytorch.py
```diff
@@ -85,14 +85,6 @@ def setup_pytorch_extension(
             if version < (12, 0):
                 raise RuntimeError("Transformer Engine requires CUDA 12.0 or newer")
 
-    if bool(int(os.getenv("NVTE_UB_WITH_MPI", "0"))):
```
wangye805, line 84: Guard via ROCm specifc guards?

examples/pytorch/comm_gemm_overlap/te_layer_with_overlap.py
```diff
@@ -68,7 +68,7 @@ def _parse_args(argv=None, namespace=None):
     )
     parser.add_argument("--seed", type=int, default=1234, help="RNG seed.")
     parser.add_argument(
-        "--fp8", action="store_true", default=False, help="Enables the te.fp8_autocast() context."
+        "--fp8", action="store_true", default=False, help="Enables the te.autocast() context."
```
wangye805, line 77: Up to TE v2.8, I think it's still fp8_autocast. Were you targeting at higher versions?

examples/pytorch/comm_gemm_overlap/te_layer_with_overlap_profile.py
```diff
@@ -0,0 +1,504 @@
+#!/usr/bin/python3
+
+# This file was modified for portability to AMDGPU
+# Copyright (c) 2025-2026, Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
wangye805: Was this file sharing a lot of codes with examples/pytorch/comm_gemm_overlap/te_layer_with_overlap.py? Is it possible to consolidate those two files

examples/pytorch/comm_gemm_overlap/ub_config.json
```diff
@@ -0,0 +1,15 @@
+{
```
wangye805: Why do we put this file here? Should it be under <TE>/transformer_engine/common or pytorch

tests/pytorch/distributed/test_comm_gemm_overlap.py
```diff
@@ -11,6 +13,9 @@
 import transformer_engine.pytorch.cpp_extensions as tex
 from transformer_engine.pytorch.fp8 import FP8GlobalStateManager
 
+from transformer_engine.jax.cpp_extensions.misc import is_hip_extension
```
wangye805: Let's not import jax specific code into pytorch side. Use this instead: https://github.com/ROCm/TransformerEngine/blob/0dfee560c7ad795fe554ebde7e0c9d9d30116d2e/tests/pytorch/test_numerics.py#L17

tests/pytorch/distributed/run_layer_with_overlap.py
```diff
@@ -30,6 +32,11 @@
 warnings.filterwarnings("ignore", category=FutureWarning)
 warnings.filterwarnings("ignore", category=UserWarning)
 
+import transformer_engine.pytorch.cpp_extensions as tex
```
wangye805: Is this rocm specific?

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers-host.cpp
```diff
@@ -357,12 +359,12 @@ int create_communicator_grouped2(communicator **comm, int myrank, int numranks,
   NVTE_CHECK_CUDA(cudaDeviceSynchronize());
   register_user_buffer_collective(&((*comm)->gpu_ptrs), LOCALSIZE, *comm, true);
   NVTE_CHECK_CUDA(
-      cudaMalloc(reinterpret_cast<void **>(&(*comm)->send_id), (*comm)->nranks * sizeof(int)));
+      cudaMalloc(reinterpret_cast<void **>(&(*comm)->send_id), (*comm)->nranks * NVTE_MAX_RINGS * sizeof(int)));
```
wangye805: Is this a bug on NV upstream?

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers-host.cpp
```diff
@@ -693,4 +725,5 @@ int register_user_buffer_collective(void **gpubuff, size_t bytes, communicator *
   comm->mem_ptr[hndl] = *gpubuff;
 
   return comm->free_region++;
+  printf("***** Returning *****\n");
```
wangye805: Debug printing?

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers-host.cpp
```diff
@@ -670,9 +675,36 @@ int register_user_buffer_collective(void **gpubuff, size_t bytes, communicator *
                      reinterpret_cast<void *>(&memhndl), sizeof(cudaIpcMemHandle_t),
                      comm->comm_intra);
 
+    // Check for NVLINK support before attempting IPC operations
```
wangye805, line 678: Put a comment if this is an NV bug. Or guard this by __HIP_PLATFORM_AMD__ if this is needed by us

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
```diff
@@ -1,10 +1,21 @@
...
+
+#ifdef __HIP_PLATFORM_AMD__
+#include <hip/hip_runtime.h>
+#include <hip/hip_bfloat16.h>
+#include "amd_detail/hip_float8.h"
+#define half_dtype hip_bfloat16 
+#define __nv_fp8_e5m2 te_hip_fp8_e5m2
+#define __nv_fp8_e4m3 te_hip_fp8_e4m3
+#else
 #include <cuda_fp8.h>
```
wangye805: Just curious, is our hipify script not able to convert cuda_fp8.h to amd_detail/hip_float8.h directly?

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
```diff
@@ -24,6 +36,7 @@
 
 #define MAX_THREADS 1024
 
+#if !defined(__HIP_PLATFORM_AMD__) && defined(__HIP_PLATFORM_NVIDIA__)
```
wangye805: Is there a __HIP_PLATFORM_NVIDIA__? I mean, we can simply use ifndef HIP_PLATFORM_AMD instead

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
```diff
@@ -62,7 +87,7 @@
   printf("[%s:%s:%d] " message "\n", FILENAME(__FILE__), __FUNCTION__, __LINE__, __VA_ARGS__)
 
 // Report and error on timeout
-#define CHECK_TIMEOUT(t, timeout) ((clock64() - (t)) > timeout)
+#define CHECK_TIMEOUT(t, timeout) (((uint64_t)clock64() - (t)) > timeout)
```
wangye805: clock64() returns int64_t? Why do we need convert it to uint64_t?

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
```diff
@@ -132,7 +157,7 @@ __global__ void __launch_bounds__(MAX_THREADS)
   }
 
   __syncthreads();
-  if (threadIdx.x == 0) __threadfence_system();
+  if (threadIdx.x == 0) __threadfence();
```
wangye805: Is it safe to relax from __threadfence_system() to just __threadfence()?

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
```diff
@@ -1116,7 +1145,11 @@ __global__ void __launch_bounds__(MAX_THREADS)
 
         // reset counter for next producer.
         ((unsigned int *)counters)[chunk_i] = 1;
+#ifndef __HIP_PLATFORM_AMD__
         asm volatile("fence.sc.gpu;\n");
```
wangye805, line 1149: Is the sequential consistency required here: 

Image

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
```diff
@@ -1329,7 +1362,7 @@ __global__ void __launch_bounds__(MAX_THREADS)
   }
 
   __syncthreads();
-  if (threadIdx.x == 0) __threadfence_system();
+  if (threadIdx.x == 0) __threadfence();
```
wangye805: Same question about the safety for the relaxation

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
```diff
@@ -2303,12 +2361,13 @@ __global__ void __launch_bounds__(MAX_THREADS) kuserbuffers_pushsendrecv_multiat
    ((NVTE_REG0_OFFSET(comm) + NVTE_REG0_RECV + (recv_peer) * NVTE_MAX_REGIONS + (dsth) + \
      (index) * NVTE_MAX_NVLINK * NVTE_MAX_REGIONS) *                                     \
     sizeof(int)))
+#endif // #ifndef __HIP_PLATFORM_AMD__
 
 void userbuffers_send(const int srchandler, const size_t srcoffset, const int dsthandler,
                       const size_t dstoffset, const size_t bytes, communicator *comm,
-                      const int peer, cudaStream_t stream) {
+                      const int peer, cudaStream_t stream, int ring_id) {
```
wangye805, line 2368: Why do we need ring_id. Is this to work-around some hardware feature difference?

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.h
```diff
@@ -35,6 +37,7 @@ using ExtBarrierOp = std::function<void(ExtComm)>;
 #define NVTE_LAUNCH_GPU 1
 #define NVTE_LAUNCH_CPU 2
 #define NVTE_MAX_NVLINK 32
+#define NVTE_MAX_RINGS 7
```
wangye805: Can you put a comment on what NVTE_MAX_RINGS mean?

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -278,6 +295,11 @@ CommOverlapBase::CommOverlapBase(const std::vector<size_t> &buffer_shape, DType
                       allgather_handle, barrier_handle, num_splits, num_max_streams, comm_cga_size,
                       gemm_priority, comm_priority, num_comm_sm, set_sm_margin, false,
                       atomic_gemm) {
+  initialize(buffer_shape, buffer_dtype, rs_overlap_first_gemm);
+}
+
+void CommOverlapBase::initialize(const std::vector<size_t> &buffer_shape, DType buffer_dtype,
```
wangye805, line 297: Is this initialize function used somewhere else? Or just to make NV upstream codes look cleaner?

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -288,7 +310,9 @@ CommOverlapBase::CommOverlapBase(const std::vector<size_t> &buffer_shape, DType
   size_t buffer_bytes = get_buffer_size_bytes(buffer_shape[0], buffer_shape[1], buffer_dtype);
   void *buffer_ptr;
   _ub_reg = register_user_buffer_collective(&buffer_ptr, buffer_bytes, _ub_comm, true);
-  if (_ub_comm->myrank == 0) printf("!!! [UB] Register UBuf %d\n", _ub_reg);
+  if (_ub_comm->myrank == 0) {
+    printf("!!! [UB] Register UBuf %d\n", _ub_reg);
+  }
```
wangye805, line 311: I would prefer aligning the coding style with NV upstream so it's easier for us to maintain/IFU later

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -640,35 +665,40 @@ CommOverlapP2PBase::CommOverlapP2PBase(const std::vector<size_t> &buffer_shape,
                       allgather_handle, barrier_handle, tp_size, num_max_streams, comm_cga_size,
                       gemm_priority, comm_priority, num_comm_sm, set_sm_margin, use_ce,
                       atomic_gemm) {
+  initialize(buffer_shape, buffer_dtype, comm_type, aggregate);
```
wangye805, line 664: Same question here for the motivation of this initialize function in the constructor

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -640,35 +665,40 @@ CommOverlapP2PBase::CommOverlapP2PBase(const std::vector<size_t> &buffer_shape,
...
   _is_p2p = true;
   _is_reduce_scatter = comm_type == CommOverlapType::RS;
   _aggregate = aggregate;
 
   // Create workspace tensor with userbuffer
   NVTE_CHECK(buffer_shape.size() == 2, "Userbuffer shape must be 2-dimensional!");
   size_t buffer_bytes = get_buffer_size_bytes(buffer_shape[0], buffer_shape[1], buffer_dtype);
-  int buffer_chunk_bytes = buffer_bytes / tp_size;
-  _num_ubuf_chunks = tp_size;
+  int buffer_chunk_bytes = buffer_bytes / _tp_size;
```
wangye805, line 676: Does NV's original code compile successfully? I mean tp_size -> _tp_size sounds like a typo in their original code :-)

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -691,15 +721,28 @@ CommOverlapP2PBase::CommOverlapP2PBase(const std::vector<size_t> &buffer_shape,
     NVTE_CHECK_CUDA(cudaMemset(_counter.dptr(), 0, sizeof(int32_t)));
   }
 
-  for (int i = 0; i < std::min(num_max_streams, _tp_size); i++) {
+  for (int i = 0; i < _stream_compute.size(); i++) {
     cudaStream_t stream;
     NVTE_CHECK_CUDA(cudaStreamCreateWithPriority(&stream, cudaStreamNonBlocking, _comm_priority));
     _stream_send.push_back(std::move(stream));
   }
+  for (int i = 0; i < 7; i++) {
```
wangye805: Why do we need more streams than NV upstream and where does the constant 7 comes out?

wangye805 (changes requested):

examples/pytorch/comm_gemm_overlap/te_layer_with_overlap.py
alextmagro: I think you had a few comments on this, so will address it here quickly. I moved the UB code up to release 2.10, as there were a few bugs and inefficiencies that NV fixed. Most of the changes that aren't guarded in the files are NV upstream changes.

I am fixing up the te_layer_with_overlap differences, and working on integrating the benchmark script into the file directly.

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
alextmagro: Good question. We require more streams as NV goes through a single NVLink, whereas we have the # of GPUs - 1 connections with xGMI. The solid 7 comes from a 8 GPU max on a system, but this should probably be replaced with a macro definition like MAX_TP_SIZE

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.h
alextmagro: Sure, will do. I renamed it NVTE_ROCM_MAX_RINGS as well, and it defines a limit on how many rings we are able to use for communications on a system.

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
alextmagro: GPUs write a flag to global memory when they send data, and a GPU looks for the write flag before it assumes that the data has arrived. With more than 1 message being sent per step (up to max TP size) we need to properly id the location of the flag.

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
alextmagro: Looking into this, but I don't believe we need sequential as we generally check for the flag written by the sending GPU in a while loop.

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
alextmagro: We need to avoid wraparound with t, which can cause us to immediately timeout -- t comes from a different SM, so may not be on the same clock as the clock64() call here.

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
alextmagro: There is actually! But this is an artifact from Yeonsoo's work, and can be standardized to the rest of TE. Will make that fix.

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
alextmagro: Good question. I don't think it was when this project was started, but it may work now.

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers-host.cpp
alextmagro: Yes, will remove. Thanks!

tests/pytorch/distributed/test_comm_gemm_overlap.py
alextmagro: Good catch, this is an mistake. Will fix.

examples/pytorch/comm_gemm_overlap/ub_config.json
alextmagro: This file is used for the examples scripts specifically to allow a user to change up algorithms used for each overlap scenario.

tests/pytorch/distributed/run_layer_with_overlap.py
alextmagro: Yes, I have guarded this.

build_tools/pytorch.py
alextmagro: I have unified this section with upstream

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
alextmagro: _tp_size is the proper private variable. In C++, it's standard practice to name private variables with a leading underscore for readability outside of the class definition.

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers-host.cpp
```diff
@@ -375,8 +377,11 @@ int create_communicator_grouped2(communicator **comm, int myrank, int numranks,
       cudaMalloc(reinterpret_cast<void **>(&(*comm)->flags_baseptr), 2 * GPU_PAGE_SIZE));
   NVTE_CHECK_CUDA(cudaMemset((*comm)->flags_baseptr, 0, 2 * GPU_PAGE_SIZE));
   (*comm)->flags = reinterpret_cast<int *>(
+#ifdef __HIP_PLATFORM_AMD__
+      (reinterpret_cast<uintptr_t>((*comm)->flags) + GPU_PAGE_SIZE - 1) & GPU_PAGE_MASK);
```
wangye805: Should it be (*comm)->flags_baseptr as the nv upstream below? (*comm)->flags is not allocated/assigned above

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
wangye805: Looks like __threadfence_system() is now supported in rocm 7.2: https://rocm.docs.amd.com/projects/HIP/en/latest/how-to/hip_cpp_language_extensions.html#memory-fence-instructions

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
wangye805: Emm, I guess my question then would be why NV upstream does not need a ring_id? Is it because of we have different implementation? The NVTE_ROCM_MAX_RINGS?

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -64,6 +72,15 @@ CommOverlapCore::CommOverlapCore(int myrank, int numranks, int mylocal, int numl
...
   }
+
+  initialize(tp_size, num_splits, num_max_streams, comm_cga_size, gemm_priority, comm_priority,
+             num_comm_sm, set_sm_margin, use_ce, atomic_gemm);
+}
+
+void CommOverlapCore::initialize(int tp_size, int num_splits, int num_max_streams,
+                                 int comm_cga_size, int gemm_priority, int comm_priority,
+                                 int num_comm_sm, bool set_sm_margin, bool use_ce,
+                                 bool atomic_gemm) {
```
wangye805, line 77: Just curious, what's the motivation for creating the CommOverlapCore::initialize in addition to refactor NV upstream coding format?

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -74,7 +91,7 @@ CommOverlapCore::CommOverlapCore(int myrank, int numranks, int mylocal, int numl
     _gemm_priority = gemm_priority;
     _comm_priority = comm_priority;
   }
-  for (int i = 0; i < std::min(num_max_streams, num_splits); i++) {
+  for (int i = 0; i < std::max(num_max_streams, num_splits); i++) {
```
wangye805, line 88: In fact, do we need stream numbers more than the min of max_stream and num_splits?

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -189,6 +206,7 @@ TensorWrapper CommOverlapCore::get_tensor_chunk(const TensorWrapper &source, siz
   NVTE_DIM_CHECK(chunk_height > 0 && chunk_width > 0, "Attempted to get empty tensor chunk");
   NVTE_DIM_CHECK(chunk_height <= height && chunk_width <= width,
                  "Attempted to get out-of-bounds tensor chunk");
+#ifndef __HIP_PLATFORM_AMD__
```
wangye805, line 203: Since we already support mxfp8. Add a to-do comment so that we won't forget to turn it on later

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -709,6 +754,43 @@ CommOverlapP2PBase::~CommOverlapP2PBase() {
...
+  // Check element size
+  const size_t element_size = source.element_size();
+  NVTE_CHECK(_ubuf.element_size() == element_size,
+             "Tried to copy data into a Userbuffers buffer but dtypes are not compatible ",
+             "(source dtype has ", element_size, " bytes, UB dtype has ", _ubuf.element_size(),
+             " bytes)");
+
+  // Input data
+  const size_t source_size = source.numel();
+  const void *src_ptr = (rowwise) ? source.dptr() : source.columnwise_dptr();
```
wangye805, line 773: Well, what if we need both row-wise and colwise? How about other fields of a tensor, for example, scale inv?

transformer_engine/pytorch/module/base.py
```diff
@@ -346,7 +366,7 @@ def get_default_config(name):
             "is_reduce_scatter": is_reduce_scatter,
             "num_sm": 1 if method == "ring_exchange" else 16,
             "cga_size": 1 if method == "ring_exchange" else 2,
-            "set_sm_margin": not method == "ring_exchange",
+            "set_sm_margin": not method == "ring_exchange" and not IS_HIP_EXTENSION,
```
wangye805: Ilya already had the sm_margin feature supported on rocm

transformer_engine/pytorch/module/base.py
```diff
@@ -474,13 +495,19 @@ def add_ub(
...
                     new_method = user_ub_cfg[name]["method"]
                     methods[new_method].append(name)
 
-        for name in (
-            methods["ring_exchange"] + methods["pipeline"] + methods["bulk"] + methods["external"]
-        ):
+        if IS_HIP_EXTENSION and user_ub_cfg is not None:
+            for name, cfg in user_ub_cfg.items():
+                assert cfg.get("method") != "bulk", (
+                    f"Bulk overlap method for '{name}' is not supported on HIP/ROCm. "
```
wangye805, line 520: I recall we supported bulk overlap but the performance is not great?

build_tools/hipify/custom_map.json
```diff
@@ -6,7 +6,14 @@
         "ATen/cudnn/Handle.h" : "ATen/miopen/Handle.h",
         "CUfunc_cache" : "hipFuncCache_t", 
         "<nvtx3/nvToolsExt.h>" : "<roctracer/roctx.h>",
-        "cudaFuncSetAttribute(" : "hipFuncSetAttribute((const void*)"
+        "cudaFuncSetAttribute(" : "hipFuncSetAttribute((const void*)",
+        "cudaLaunchKernel": "hipLaunchKernel",
```
wangye805: cudaLaunchKernel cannot be hipified?

wangye805 (changes requested):

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
wangye805: Okay, I saw the same refactor in release v2.10

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
wangye805: I saw the same refactor on v2.10 release

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers-host.cpp
alextmagro: Yes, I have fixed that. Thanks!

transformer_engine/pytorch/module/base.py
alextmagro: Yeonsoo was seeing some race conditions and weird hangs, and submitted a PR request to upstream. I am still seeing failures after rebasing to IFU 2.10, so I think that the issue is still there. I think if we want to enable this, we should consider it for a different PR as it will require a new implementation.

build_tools/hipify/custom_map.json
alextmagro: Looks like this was very recently added to hipify_torch, so we can probably remove this after we update our hipify_torch commit. I would recommend we do that separately, however.

transformer_engine/pytorch/module/base.py
alextmagro: This was a performance decision, not a functionality one. set_sm_margin seems to slow down UB on ROCm, probably because we have dedicated SDMA engines being used that don't require CU blocking?

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
alextmagro: Within these functions we are working with only colwise or rowwise data. When we call hipblaslt we are passing in only one or the other for the GEMM, so no need for both within a single overlap call.

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
alextmagro: This is a ifndef, so is enabled for us since we don't have the padding issues.

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
alextmagro: We do. I am convinced that this is an upstream bug, as num_splits has a default value of 0, which seems off. Either way, we need at least 1 stream for each tp peer, which is what num_splits is supposed to be.

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
alextmagro: NV only has one ring, so ring_id would always be 0, and would not affect the index.

build_tools/hipify/custom_map.json
```diff
@@ -1,14 +1,21 @@
 {
  "custom_map" : {
         "<cuda_bf16.h>" : "<hip/hip_bfloat16.h>",
+        "util/cuda_runtime.h" : "util/hip_runtime.h",
```
ipanfilo: it should be covered by line 5

build_tools/hipify/custom_map.json
ipanfilo: I can actually see it in hipify_torch maps. I think custom map is not needed after recent TE hipification changes

tests/pytorch/distributed/test_comm_gemm_overlap.py
```diff
@@ -66,6 +71,8 @@ def _run_gemm_with_overlap(comm_type, bulk, p2p, atomic, aggregate, quantization
     if bulk:
         test_cmd.append("--bulk-overlap")
     else:
+        if IS_HIP_EXTENSION and not p2p:
+            pytest.skip("HIP only supports P2P operations.")
```
ipanfilo, line 76: Is it some difference in HIP API vs CUDA API?

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
```diff
@@ -1025,7 +1048,11 @@ __global__ void __launch_bounds__(MAX_THREADS)
 
       // reset counter for next producer.
       ((unsigned int *)counters)[0] = 1;
+#ifndef __HIP_PLATFORM_AMD__
       asm volatile("fence.sc.gpu;\n");
+#else
+      __threadfence_system();
```
ipanfilo, line 1056: Isn't _threadfence a close analogogue of fence.sc.gpu?

transformer_engine/pytorch/module/base.py
ipanfilo: Please put comment that it is disabled by performance reason then

build_tools/hipify/custom_map.json
alextmagro: Looking back I had had this since we need to hipify cudaLaunchKernelExC as well as cudaLaunchKernel. The former is still not in the map, so I have updated the custom map to specifically pick up the ExC variation.

tests/pytorch/distributed/test_comm_gemm_overlap.py
alextmagro: P2P and not P2P lead to different algorithms -- ring exchanges vs pipelined communication overlap. We have only implemented P2P at this point, as theoretically it should be much faster for us.

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
alextmagro: Technically yes, however,  fence.sc.gpu works as a fence for all gpus in a NVSwitch fabric. Since TP is only available within a NVSwitch fabric for Nvidia, they can get away with the weaker barrier. Our GPUs are not memory coherent across IB connections, so the systemwide fence is necessary.

build_tools/hipify/custom_map.json
alextmagro: Thanks, have removed.

alextmagro:
L3 CI
https://github.com/ROCm/TransformerEngine/actions/runs/23252049008

-- missing distributed/test_cast_master_weights_to_fp8.py hotfix that is now in dev.

tests/pytorch/distributed/run_layer_with_overlap.py
```diff
@@ -552,8 +563,8 @@ def run_fwd_bwd(model, x):
         # Now validate accuracy
         if not bool(numerics_failed.item()):
             for i, (test_g, ref_g) in enumerate(zip(test_grads, ref_grads)):
-                rtol = 0.125 if opts.fp8 else 0.025
-                atol = 0.0625 if opts.fp8 else 0.00125
+                rtol = 0.125 if opts.fp8 else 0.025 if not IS_HIP_EXTENSION else 5e-2
```
ipanfilo: Why does ROCm have such a huge difference in accuracy with CUDA

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.h
```diff
@@ -36,6 +38,10 @@ using ExtBarrierOp = std::function<void(ExtComm)>;
 #define NVTE_LAUNCH_CPU 2
 #define NVTE_MAX_NVLINK 32
 
+#define NVTE_ROCM_MAX_TP_SIZE 8
+// Maximum # of rings possible for ring_exchange
+#define NVTE_ROCM_MAX_RINGS (NVTE_ROCM_MAX_TP_SIZE - 1)
```
ipanfilo, line 44: It should be 1 on CUDA to make usebuffers.cu working as before

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.h
```diff
@@ -63,6 +69,28 @@ using ExtBarrierOp = std::function<void(ExtComm)>;
 #define NVTE_HF_NVREDUCEDONE (userbuffers_op_types + 3)
 #define NVTE_MAX_SHARP 16
 
+#ifdef __HIP_PLATFORM_AMD__ // Moved to header for visibility
```
ipanfilo: Better say where is it moved from

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -21,6 +23,12 @@
 #define HALF_BYTES 2
 #define UB_MAX_SM 32
 
+#ifdef __HIP_PLATFORM_AMD__
+#define half_dtype hip_bfloat16
```
ipanfilo: Those types seem unused in the code

transformer_engine/common/include/transformer_engine/comm_gemm_overlap.h
```diff
@@ -172,7 +196,7 @@ class CommOverlapBase : public CommOverlapCore {
   CommOverlapBase(const std::vector<size_t> &buffer_shape, DType buffer_dtype, int myrank,
                   int numranks, int mylocal, int numlocal, int mynode, int numnodes, int tp_size,
                   ExtAllgatherOp allgather_handle, ExtBarrierOp barrier_handle, int num_splits = 3,
-                  int num_max_streams = NVTE_COMM_OVERLAP_MAX_STREAMS, int comm_cga_size = 2,
```
ipanfilo, line 175: Should NVTE_COMM_OVERLAP_MAX_STREAMS be redefined on ROCm?

transformer_engine/common/util/pybind_helper.h
```diff
@@ -35,9 +32,6 @@
         .value("NVTE_No_Backend", NVTE_Fused_Attn_Backend::NVTE_No_Backend); 
 #endif
 
-// Define comm overlap handles if not using ROCm
-#ifndef USE_ROCM
-
 #define NVTE_DECLARE_COMM_OVERLAP_HANDLES(m)                                                       \
```
ipanfilo: This define is not needed then. It can be inlined to NVTE_DECLARE_COMMON_PYBIND11_HANDLES like in upstream

transformer_engine/pytorch/module/base.py
```diff
@@ -288,15 +289,18 @@ def initialize_ub(
                 flush=True,
             )
 
-    # Allocate cuBLAS workspace with expanded size for chunking in overlapping GEMM calls
+    # Allocate cuBLAS workspace with expanded size for chunking in overlapping GEMM calls.
+    # The workspace must have enough copies for max(num_max_streams, tp_size) compute streams,
+    # since CommOverlapCore creates that many streams and divides the workspace among them.
+    num_workspace_copies = max(_NUM_MAX_UB_STREAMS, tp_size)
```
ipanfilo: If you make num_workspace_copies == _NUM_MAX_UB_STREAMS on CUDE, you don't need multiple 'if IS_HIP_EXTENSION' below

transformer_engine/pytorch/ops/fused/__init__.py
```diff
@@ -34,13 +34,12 @@
     ForwardLinearScaleAdd,
     fuse_forward_linear_scale_add,
 )
-from torch.utils.cpp_extension import IS_HIP_EXTENSION
```
ipanfilo, line 37: Is the file still different from upstream?

ipanfilo (changes requested):

tests/pytorch/distributed/run_layer_with_overlap.py
alextmagro: Userbuffer code introduces numerical differences due to different GEMM sizes and different order of operations for reductions. NV already accounts for this by limiting to 4 GPUs, and These small errors compound through different layers, but ROCm version has NUM_RINGS times more GEMMs than NV, so a bit more error is expected. These errors tend to become more exaggerated around 0 with BF16.

I have reduced the error tolerances to better match what we actually need, however.

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.h
alextmagro: Sure, updated comments. It came from userbuffers.cu, but is needed in rocm_comm_gemm_overlap.cpp

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
alextmagro: I removed them, and moved the conversion to the hipify map

transformer_engine/common/include/transformer_engine/comm_gemm_overlap.h
alextmagro: I am not sure, we currently don't use that variable -- its usage is replaced by NVTE_ROCM_MAX_RINGS and I 
like that this new variable name explains how num_max_streams is being determined.

transformer_engine/common/util/pybind_helper.h
alextmagro: OK, I have inlined like upstream does.

transformer_engine/pytorch/ops/fused/__init__.py
alextmagro: It is the same, have removed AMD copyright

build_tools/hipify/custom_map.json
```diff
@@ -7,8 +7,16 @@
         "CUfunc_cache" : "hipFuncCache_t", 
         "<nvtx3/nvToolsExt.h>" : "<roctracer/roctx.h>",
         "cudaFuncSetAttribute(" : "hipFuncSetAttribute((const void*)",
-        "__nv_bfloat162":"__hip_bfloat162",
-        "cuda::getCurrentCUDAStream" : "hip::getCurrentHIPStreamMasqueradingAsCUDA"
+        "__nv_bfloat162" : "__hip_bfloat162",
```
ipanfilo: nit: this actually exists in hipify_torch map

examples/pytorch/comm_gemm_overlap/rocm_te_layer_with_overlap.py
```diff
@@ -0,0 +1,453 @@
...
+    dist_print("Destroying all process groups...", debug=True)
+    dist.destroy_process_group()
+    if opts.debug and WORLD_RANK == 0:
+        print("Exiting...\n", end="", flush=True)
+
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(_train(_parse_args()))
```
ipanfilo: EOL

tests/pytorch/distributed/run_layer_with_overlap.py
```diff
@@ -552,8 +567,8 @@ def run_fwd_bwd(model, x):
         # Now validate accuracy
         if not bool(numerics_failed.item()):
             for i, (test_g, ref_g) in enumerate(zip(test_grads, ref_grads)):
-                rtol = 0.125 if opts.fp8 else 0.025
-                atol = 0.0625 if opts.fp8 else 0.00125
+                rtol = 0.125 if opts.fp8 else 0.025 if not IS_HIP_EXTENSION else 3e-2
```
ipanfilo: nit: please use 0.03 so it is easier to visually compare with other values in the expression

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
```diff
@@ -212,7 +233,7 @@ __global__ void __launch_bounds__(MAX_THREADS)
     userptr[myrank][lineoffset + line] = sum;
   }
   __syncthreads();
-  if (threadIdx.x == 0) __threadfence();
+  if (threadIdx.x == 0) __threadfence_system();
```
ipanfilo, line 237: Put guard here

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -353,7 +357,7 @@ void CommOverlapBase::bulk_overlap(const TensorWrapper &A, bool transa, const Te
       char *rs_output_ptr = reinterpret_cast<char *>(rs_output.dptr());
       reducescatter2_userbuff_fp8<__nv_fp8_e5m2>(rs_output_ptr, _ubuf.scale_inv(), _ub_reg, 0,
                                                  comm_elements, _ub_comm, _stream_comm,
-                                                 (cudaEvent_t)_comm_launch_event);
+                                                 (cudaEvent_t)_comm_launch_event);
```
ipanfilo: nit: not needed formatting change

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -717,10 +722,23 @@ void CommOverlapP2PBase::initialize(const std::vector<size_t> &buffer_shape, DTy
     NVTE_CHECK_CUDA(cudaStreamCreateWithPriority(&stream, cudaStreamNonBlocking, _comm_priority));
     _stream_send.push_back(std::move(stream));
   }
+  for (int i = 0; i < NVTE_ROCM_MAX_RINGS; i++) {
+    cudaStream_t stream;
```
ipanfilo: Does CUDA use default stream w/o those changes?

transformer_engine/common/comm_gemm_overlap/rocm_comm_gemm_overlap.cpp
```diff
@@ -0,0 +1,584 @@
...
+
+#include "common/common.h"
+#include "common/util/cuda_driver.h"
+#include "common/util/cuda_runtime.h"
+#include "common/util/logging.h"
+#include "common/util/system.h"
+#include "userbuffers/userbuffers.h"
+
+namespace transformer_engine {
+#if 0
```
ipanfilo, line 19: Is it WIP or temporary change?

transformer_engine/common/include/transformer_engine/comm_gemm_overlap.h
ipanfilo: (1) Current change breaks NV path. (2) Variable is num_max_streams so having it assigned with NVTE_COMM_OVERLAP_MAX_STREAMS is quite normal

transformer_engine/common/include/transformer_engine/comm_gemm_overlap.h
```diff
@@ -238,9 +278,12 @@ class CommOverlapP2PBase : public CommOverlapCore {
   int _num_ubuf_chunks;
   int _self_chunk_id;
   std::vector<TensorWrapper> _ubufs;
-  std::vector<cudaStream_t> _stream_send;
+  std::vector<cudaStream_t> _stream_send, l_stream_send, l_stream_recv;
   cudaStream_t _stream_recv;
-  cudaEvent_t _stop_send, _stop_recv;
+  cudaEvent_t _stop_send, _stop_recv, l_stop_recv[7];
```
ipanfilo: Why 7?

transformer_engine/common/include/transformer_engine/comm_gemm_overlap.h
```diff
@@ -252,7 +295,7 @@ class CommOverlapP2PBase : public CommOverlapCore {
   CommOverlapP2PBase(const std::vector<size_t> &buffer_shape, DType buffer_dtype, int myrank,
                      int numranks, int mylocal, int numlocal, int mynode, int numnodes, int tp_size,
                      ExtAllgatherOp allgather_handle, ExtBarrierOp barrier_handle,
-                     CommOverlapType comm_type, int num_max_streams = NVTE_COMM_OVERLAP_MAX_STREAMS,
+                     CommOverlapType comm_type, int num_max_streams = NVTE_ROCM_MAX_RINGS,
```
ipanfilo: The same as in CommOverlapBase

transformer_engine/common/CMakeLists.txt
```diff
@@ -173,7 +173,10 @@ set(transformer_engine_cuda_arch_specific_sources)
 
 # Source files in both cuda and rocm
 list(APPEND transformer_engine_cpp_sources
-     transformer_engine.cpp
+     transformer_engine.cpp
```
ipanfilo: What is a diff in this line?

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -503,7 +502,7 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
                     int, int, int, int, bool, bool, bool>(),
            py::call_guard<py::gil_scoped_release>(), py::arg("buffer_shape"),
            py::arg("buffer_dtype"), py::arg("helper"), py::arg("tp_size"),
-           py::arg("num_splits") = 3, py::arg("num_max_streams") = NVTE_COMM_OVERLAP_MAX_STREAMS,
+           py::arg("num_splits") = 3, py::arg("num_max_streams") = NVTE_ROCM_MAX_RINGS,
```
ipanfilo: The same like in CommOverlap* classes declaration

transformer_engine/pytorch/csrc/extensions.h
```diff
@@ -553,7 +544,7 @@ class CommOverlap : torch::CustomClassHolder, public transformer_engine::CommOve
  public:
   CommOverlap(const std::vector<size_t> &buffer_shape, at::ScalarType buffer_dtype,
               CommOverlapHelper *helper, int tp_size, int num_splits = 3,
-              int num_max_streams = NVTE_COMM_OVERLAP_MAX_STREAMS, int comm_cga_size = 2,
+              int num_max_streams = NVTE_ROCM_MAX_RINGS, int comm_cga_size = 2,
```
ipanfilo: The same like in core CommOverlap* classes

ipanfilo (changes requested):

build_tools/hipify/custom_map.json
alextmagro: Good catch. Now we are on v2 we can remove this.

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
alextmagro: I have guarded these.

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
alextmagro: CUDA uses the _stream_send vector of streams, we use l_stream_send in the rocm_ functions.

transformer_engine/common/comm_gemm_overlap/rocm_comm_gemm_overlap.cpp
alextmagro: This is WIP -- something we should revisit for 450. Comment was left as is to avoid making that too explicit

transformer_engine/common/include/transformer_engine/comm_gemm_overlap.h
alextmagro: Understood. I have defined it as NVTE_ROCM_MAX_RINGS and reverted these var changes.

transformer_engine/common/include/transformer_engine/comm_gemm_overlap.h
alextmagro: This should be defined by NVTE_ROCM_MAX_RINGS, have fixed it.

transformer_engine/common/CMakeLists.txt
alextmagro: There appears to be a trailing space after transformer_engine.cpp. I have removed it.

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
ipanfilo: Upstream uses __threadfence_system(). The changes here are either not needed or there should be ifndef instead

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
```diff
@@ -477,7 +506,11 @@ __global__ void __launch_bounds__(MAX_THREADS)
   }
 
   __syncthreads();
+#ifdef __HIP_PLATFORM_AMD__
   if (threadIdx.x == 0) __threadfence_system();
+#else
```
ipanfilo: The same as line 159

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
```diff
@@ -708,7 +741,11 @@ __global__ void __launch_bounds__(MAX_THREADS)
...
```
ipanfilo: The same as line 159

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
ipanfilo: The same as line 159

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
ipanfilo: If l_stream_* vectors are not used on CUDA, the code should be guarded. The same is for _stream_send

ipanfilo (changes requested):
Please also review newly enabled code for FP8 FNUZ/OCP data type selection: torch.float8_e4m3fn ones should be replaced with get_torch_float8_e4m3_type() and the same is for e5m2.
run_comm_gemm_overlap.py - is one of such modules

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers.cu
alextmagro: I have removed those changes, and unified with upstream.

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
alextmagro: I have guarded their definition and declaration, _stream_send is used by upstream so was left as is.

alextmagro (commented):
Merge conflicts addressed, rerunning L3 just in case.

transformer_engine/pytorch/csrc/extensions.h
```diff
@@ -17,14 +17,6 @@ class CommOverlapHelper;
 class CommOverlap;
 class CommOverlapP2P;
 
-#ifdef USE_ROCM
```
ipanfilo, line 20: Is it still different from upstream?

alextmagro: This file still contains some other rocm specific changes such as cublaslt helper functions, nvshmem/rocshmem, and some still unsupported comm gemm features.

ipanfilo (approved):

wangye805 (approved):

PR #369, opened 2025-11-12
Current scaling: two-stage HIP amax kernel

matthiasdiener:
# Description

Implements a two-stage HIP kernel for the amax operation, as an alternative to the original implementation that uses atomic reductions. Make the two-stage kernel the default implementation. Users can use `export NVTE_USE_ATOMIC_AMAX=1` to use the atomic amax kernel.

The corresponding Triton kernel is at #385.

Fixes https://github.com/ROCm/frameworks-internal/issues/14303.

See https://github.com/ROCm/frameworks-internal/issues/14303#issuecomment-3554900809 for a performance analysis.

**TODO:**
- [x] Fix other call sites of `nvte_compute_amax`
- [x] Address `FIXME`s in the code

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/common/include/transformer_engine/recipe.h
+++ b/transformer_engine/common/include/transformer_engine/recipe.h
@@ -1,4 +1,6 @@
 /*************************************************************************
+ * This file was modified for portability to AMDGPU
+ * Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
  * Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  *
  * See LICENSE for license information.
--- a/transformer_engine/common/include/transformer_engine/recipe.h
+++ b/transformer_engine/common/include/transformer_engine/recipe.h
@@ -97,6 +97,9 @@ void nvte_compute_amax_with_workspace(const NVTETensor input_, const NVTETensor 
  *  \param[in]     config           Quantization configuration.
  *  \param[in]     stream           CUDA stream used for the operation.
  */
+
+constexpr int amax_kernel_threads = 512;
+
 void nvte_compute_scale_from_amax(NVTETensor output, const NVTEQuantizationConfig config,
                                   cudaStream_t stream);
 
--- a/transformer_engine/common/include/transformer_engine/recipe.h
+++ b/transformer_engine/common/include/transformer_engine/recipe.h
@@ -75,6 +75,8 @@ void nvte_delayed_scaling_recipe_amax_and_scale_update_after_reduction(
     std::vector<NVTETensor> scales, const char* amax_compute_algo, NVTEDType fp8_dtype,
     float margin, cudaStream_t stream);
 
+#ifdef __HIP_PLATFORM_AMD__
+
 constexpr int amax_kernel_threads = 512;
 
 inline bool nvte_use_atomic_amax() {
@@ -89,6 +91,8 @@ inline bool nvte_use_atomic_amax() {
   return cached == 1;
 }
 
+#endif
+
 /*! \brief Compute an FP8 tensor's amax.
  *
  *  The amax (maximum absolute value) of the input tensor is computed
@@ -100,7 +104,21 @@ inline bool nvte_use_atomic_amax() {
  */
 void nvte_compute_amax(const NVTETensor input, NVTETensor output, cudaStream_t stream);
 
-void nvte_compute_amax_with_workspace(const NVTETensor input_, const NVTETensor output_, const NVTETensor workspace_, cudaStream_t stream);
+#ifdef __HIP_PLATFORM_AMD__
+
+/*! \brief Compute an FP8 tensor's amax.
+ *
+ *  The amax (maximum absolute value) of the input tensor is computed
+ *  and written to the amax buffer of the output tensor.
+ *
+ *  \param[in]     input            Input tensor. Must be unquantized.
+ *  \param[in,out] output           Output tensor. Must be an FP8 tensor with per-tensor scaling.
+ *  \param[out]    workspace        Output tensor. Must be FP32.
+ *  \param[in]     stream           CUDA stream used for the operation.
+ */
+void nvte_compute_amax_with_workspace(const NVTETensor input, NVTETensor output, NVTETensor workspace, cudaStream_t stream);
+
+#endif
 
 /*! \brief Update an FP8 tensor's scale based on its amax.
  *
@@ -111,7 +129,6 @@ void nvte_compute_amax_with_workspace(const NVTETensor input_, const NVTETensor 
  *  \param[in]     config           Quantization configuration.
  *  \param[in]     stream           CUDA stream used for the operation.
  */
-
 void nvte_compute_scale_from_amax(NVTETensor output, const NVTEQuantizationConfig config,
                                   cudaStream_t stream);
 
--- a/transformer_engine/common/include/transformer_engine/recipe.h
+++ b/transformer_engine/common/include/transformer_engine/recipe.h
@@ -94,7 +94,7 @@ void nvte_compute_amax(const NVTETensor input, NVTETensor output, cudaStream_t s
 
 #ifdef __HIP_PLATFORM_AMD__
 
-size_t nvte_amax_workspace_size(size_t N);
+size_t nvte_amax_workspace_num_blocks(size_t N);
 
 /*! \brief Compute an FP8 tensor's amax.
  *
--- a/transformer_engine/common/include/transformer_engine/recipe.h
+++ b/transformer_engine/common/include/transformer_engine/recipe.h
@@ -79,18 +79,6 @@ void nvte_delayed_scaling_recipe_amax_and_scale_update_after_reduction(
 
 constexpr int amax_kernel_threads = 512;
 
-inline bool nvte_use_atomic_amax() {
-  static int cached = -1;
-  if (cached == -1) {
-    cached = 0;
-    const char *env_p = std::getenv("NVTE_USE_ATOMIC_AMAX");
-    if (env_p && std::string(env_p) == "1") {
-      cached = 1;
-    }
-  }
-  return cached == 1;
-}
-
 #endif
 
 /*! \brief Compute an FP8 tensor's amax.
--- a/transformer_engine/common/include/transformer_engine/recipe.h
+++ b/transformer_engine/common/include/transformer_engine/recipe.h
@@ -94,6 +94,8 @@ void nvte_compute_amax(const NVTETensor input, NVTETensor output, cudaStream_t s
 
 #ifdef __HIP_PLATFORM_AMD__
 
+size_t nvte_amax_workspace_size(size_t N);
+
 /*! \brief Compute an FP8 tensor's amax.
  *
  *  The amax (maximum absolute value) of the input tensor is computed
--- a/transformer_engine/common/include/transformer_engine/recipe.h
+++ b/transformer_engine/common/include/transformer_engine/recipe.h
@@ -84,6 +84,8 @@ void nvte_delayed_scaling_recipe_amax_and_scale_update_after_reduction(
  */
 void nvte_compute_amax(const NVTETensor input, NVTETensor output, cudaStream_t stream);
 
+void nvte_compute_amax_with_workspace(const NVTETensor input_, const NVTETensor output_, const NVTETensor workspace_, cudaStream_t stream);
+
 /*! \brief Update an FP8 tensor's scale based on its amax.
  *
  *  This is only supported for FP8 tensors with per-tensor scaling.
--- a/transformer_engine/common/include/transformer_engine/recipe.h
+++ b/transformer_engine/common/include/transformer_engine/recipe.h
@@ -75,6 +75,20 @@ void nvte_delayed_scaling_recipe_amax_and_scale_update_after_reduction(
     std::vector<NVTETensor> scales, const char* amax_compute_algo, NVTEDType fp8_dtype,
     float margin, cudaStream_t stream);
 
+constexpr int amax_kernel_threads = 512;
+
+inline bool nvte_use_atomic_amax() {
+  static int cached = -1;
+  if (cached == -1) {
+    cached = 0;
+    const char *env_p = std::getenv("NVTE_USE_ATOMIC_AMAX");
+    if (env_p && std::string(env_p) == "1") {
+      cached = 1;
+    }
+  }
+  return cached == 1;
+}
+
 /*! \brief Compute an FP8 tensor's amax.
  *
  *  The amax (maximum absolute value) of the input tensor is computed
@@ -97,8 +111,6 @@ void nvte_compute_amax_with_workspace(const NVTETensor input_, const NVTETensor 
  *  \param[in]     config           Quantization configuration.
  *  \param[in]     stream           CUDA stream used for the operation.
  */
-
-constexpr int amax_kernel_threads = 512;
 
 void nvte_compute_scale_from_amax(NVTETensor output, const NVTEQuantizationConfig config,
                                   cudaStream_t stream);
--- a/transformer_engine/common/recipe/current_scaling.cu
+++ b/transformer_engine/common/recipe/current_scaling.cu
@@ -114,11 +114,7 @@ void launch_amax_kernel(const InputType *input, float *amax, const size_t N, flo
   constexpr size_t max_blocks = 65535;
   num_blocks = std::min(num_blocks, max_blocks);
 
-  const bool UseBlockAmax = (block_amax != nullptr);
-
-  if (UseBlockAmax) {
-    NVTE_CHECK(block_capacity >= num_blocks);
-  }
+  const bool UseBlockAmax = (block_amax != nullptr) && (block_capacity >= num_blocks);
 
   // Launch kernel
   switch (align) {
@@ -167,6 +163,10 @@ void launch_amax_kernel(const InputType *input, float *amax, const size_t N, flo
 }  // namespace transformer_engine
 
 void nvte_compute_amax(const NVTETensor input_, const NVTETensor output_, cudaStream_t stream) {
+  nvte_compute_amax_with_workspace(input_, output_, /*workspace=*/nullptr, stream);
+}
+
+void nvte_compute_amax_with_workspace(const NVTETensor input_, const NVTETensor output_, const NVTETensor workspace_, cudaStream_t stream) {
   NVTE_API_CALL(nvte_compute_amax);
   using namespace transformer_engine;
 
@@ -202,12 +202,19 @@ void nvte_compute_amax(const NVTETensor input_, const NVTETensor output_, cudaSt
              to_string(output.amax.dtype), ")");
   CheckOutputTensor(output, "output_compute_amax", true);
 
-  // Interpret output.data as workspace if present
-  float *block_amax = nullptr;
+  // Optional workspace
+  float* block_amax = nullptr;
   size_t block_capacity = 0;
-  if (output.data.dptr != nullptr) {
-    block_amax     = reinterpret_cast<float*>(output.data.dptr);
-    block_capacity = output.data.numel();     // #floats in workspace
+
+  if (workspace_ != nullptr) {
+    auto &workspace = *reinterpret_cast<Tensor *>(workspace_);
+    NVTE_CHECK(workspace.data.dptr != nullptr,
+               "Workspace tensor for amax computation has no data");
+    NVTE_CHECK(workspace.data.dtype == DType::kFloat32,
+               "Workspace tensor for amax computation must be FP32, got dtype=",
+               to_string(workspace.data.dtype));
+    block_amax     = reinterpret_cast<float*>(workspace.data.dptr);
+    block_capacity = workspace.data.numel();
   }
 
   // Compute amax
--- a/transformer_engine/common/recipe/current_scaling.cu
+++ b/transformer_engine/common/recipe/current_scaling.cu
@@ -28,9 +28,28 @@ using bf16__ = __hip_bfloat16;
 
 constexpr int amax_kernel_threads = 512;
 
+template <int BLOCK_THREADS>
+__global__ void amax_final_reduce(const float* __restrict__ block_amax,
+                                  float* __restrict__ global_amax,
+                                  int num_blocks) {
+  float val = 0.f;
+
+  for (int i = threadIdx.x; i < num_blocks; i += BLOCK_THREADS) {
+    val = fmaxf(val, block_amax[i]);
+  }
+
+  const int warp_id = threadIdx.x / THREADS_PER_WARP;
+  const float block_max =
+      reduce_max<BLOCK_THREADS / THREADS_PER_WARP>(val, warp_id);
+
+  if (threadIdx.x == 0) {
+    *global_amax = block_max;
+  }
+}
+
 template <int nvec, bool aligned, typename InputType>
 __launch_bounds__(amax_kernel_threads) __global__
-    void amax_kernel(const InputType *input, float *amax, const size_t N,
+    void amax_kernel(const InputType *input, float* __restrict__ block_amax, const size_t N,
                      const size_t num_aligned_elements) {
   VectorizedLoader<InputType, nvec, aligned> loader(input, N);
   InputType max{0.f};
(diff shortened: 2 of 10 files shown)


transformer_engine/common/recipe/current_scaling.cu
```diff
@@ -65,12 +96,17 @@ __launch_bounds__(amax_kernel_threads) __global__
   // Reduce amax over block
   max = reduce_max<amax_kernel_threads / THREADS_PER_WARP>(max, warp_id);
   if (threadIdx.x == 0) {
-    atomicMaxFloat(amax, max);
+    if (use_block_amax) {
```
ipanfilo: Does branch here have measurable performance impact? If so, may be make use_block_amax as template arg.

transformer_engine/pytorch/csrc/extensions/activation.cpp
```diff
@@ -36,10 +38,23 @@ py::object activation_helper(const at::Tensor& input, py::handle quantizer, int
     auto [te_output_act, out_act] =
         my_quantizer_none->create_tensor(input_shape, GetTransformerEngineDType(fake_tensor_type));
 
+    // Workspace for nvte_compute_amax_with_workspace
```
ipanfilo: Can it be encapsulated inside nvte_compute_amax()? Moreover, of atomic paths is selected, no need to allocate WS

transformer_engine/common/recipe/current_scaling.cu
matthiasdiener: I tried both approaches, `use_block_amax` as a normal argument and as a template argument; both had the same performance. I chose this approach since it makes calling the kernel slightly simpler.

transformer_engine/pytorch/csrc/extensions/activation.cpp
matthiasdiener: > Can it be encapsulated inside nvte_compute_amax()? 

Unfortunately, I did not manage to encapsulate this inside `nvte_compute_amax`. The main issue is that there would be a need to allocate/deallocate memory for the workspace in that function, which appears to be fragile when running with CUDA graph capture, leading to random crashes in that function when capture is enabled.

> Moreover, of atomic paths is selected, no need to allocate WS

Good catch, thanks. https://github.com/ROCm/TransformerEngine/pull/369/commits/16d3bf9010d2ff6ddabce24d2a54e1627a77d067 should address this.

ipanfilo: > Unfortunately, I did not manage to encapsulate this inside `nvte_compute_amax`. The main issue is that there would be a need to allocate/deallocate memory for the workspace in that function, which appears to be fragile when running with CUDA graph capture, leading to random crashes in that function when capture is enabled.

Is it because memory allocation/freeing should be outside of NVTE_SCOPED_GIL_RELEASE()? If so repeating code still can be encapsulated in separate function inside pytorch extension.
Common TE code on the other hand should not worry about environment variable and choose the code path based on workspace presence. It actually does not need use_block_amax too, because block_amax itself may be nullptr or not

transformer_engine/common/include/transformer_engine/recipe.h
```diff
@@ -84,6 +100,8 @@ void nvte_delayed_scaling_recipe_amax_and_scale_update_after_reduction(
  */
 void nvte_compute_amax(const NVTETensor input, NVTETensor output, cudaStream_t stream);
 
+void nvte_compute_amax_with_workspace(const NVTETensor input_, const NVTETensor output_, const NVTETensor workspace_, cudaStream_t stream);
```
wangye805: output and workspace should be writable (drop the const)?

wangye805: By the way, let's align our naming style with NV upstream:
input_ --> input, 
output_ --> output, 
workspace_ --> workspace

transformer_engine/common/recipe/current_scaling.cu
```diff
@@ -26,11 +26,29 @@ using bf16__ = __nv_bfloat16;
 using bf16__ = __hip_bfloat16;
 #endif //__HIP_PLATFORM_AMD__
 
-constexpr int amax_kernel_threads = 512;
+
```
wangye805, line 30: Let's guard our rocm specific code changes by macro __HIP_PLATFORM_AMD__

transformer_engine/common/recipe/current_scaling.cu
```diff
@@ -26,11 +26,29 @@ using bf16__ = __nv_bfloat16;
...
+  if (threadIdx.x == 0) {
+    *global_amax = block_max;
+  }
+}
 
 template <int nvec, bool aligned, typename InputType>
 __launch_bounds__(amax_kernel_threads) __global__
-    void amax_kernel(const InputType *input, float *amax, const size_t N,
+    void amax_kernel(const InputType *input, float *amax, float* __restrict__ block_amax, const size_t N,
```
wangye805, line 57: Guard the api change so NV upstream can remain their flow

transformer_engine/pytorch/csrc/extensions/activation.cpp
```diff
@@ -36,11 +38,33 @@ py::object activation_helper(const at::Tensor& input, py::handle quantizer, int
...
+    } else {
+      // Workspace for nvte_compute_amax_with_workspace
+      const auto N = static_cast<size_t>(input_tensor.numel());
+      constexpr size_t max_blocks_hw = 65535;
+
+      // Worst-case (nvec = 1) upper bound on number of blocks
+      size_t max_blocks = std::min(DIVUP(N, static_cast<size_t>(amax_kernel_threads)), max_blocks_hw);
+
+      // Allocate FP32 workspace for block-wise amax
+      auto ws = at::empty({static_cast<long>(max_blocks)}, at::CUDA(at::kFloat));
```
wangye805: Do we need to cast max_blocks to long where the maximum block is 65535

transformer_engine/common/include/transformer_engine/recipe.h
```diff
@@ -73,6 +75,20 @@ void nvte_delayed_scaling_recipe_amax_and_scale_update_after_reduction(
     std::vector<NVTETensor> scales, const char* amax_compute_algo, NVTEDType fp8_dtype,
     float margin, cudaStream_t stream);
 
+constexpr int amax_kernel_threads = 512;
+
+inline bool nvte_use_atomic_amax() {
```
wangye805: Do we need to cache the env evaluation? Usually those host side operations are pretty cheap and are ahead of gpu kernels in e2e training

transformer_engine/common/include/transformer_engine/recipe.h
```diff
@@ -84,6 +100,8 @@ void nvte_delayed_scaling_recipe_amax_and_scale_update_after_reduction(
  */
 void nvte_compute_amax(const NVTETensor input, NVTETensor output, cudaStream_t stream);
```
wangye805, line 94: Let's have a brief doc just like the nvte_compute_amax above

transformer_engine/pytorch/csrc/extensions/activation.cpp
```diff
@@ -36,11 +38,33 @@ py::object activation_helper(const at::Tensor& input, py::handle quantizer, int
     auto [te_output_act, out_act] =
         my_quantizer_none->create_tensor(input_shape, GetTransformerEngineDType(fake_tensor_type));
 
-    NVTE_SCOPED_GIL_RELEASE({
-      act_func(te_input.data(), te_output_act.data(), at::cuda::getCurrentCUDAStream());
-      // use te_output_act as input to the compute amax and find the amax of activated tensor
-      nvte_compute_amax(te_output_act.data(), te_output.data(), at::cuda::getCurrentCUDAStream());
-    });
+    if (nvte_use_atomic_amax()) {
```
wangye805: Guard our rocm specific behavior

transformer_engine/pytorch/csrc/extensions/bias.cpp
```diff
@@ -48,9 +50,30 @@ std::vector<py::object> bgrad_quantize(const at::Tensor& input, py::handle py_qu
   if (detail::IsFloat8CurrentScalingQuantizers(py_quantizer.ptr())) {
     // my_quantizer here has to be a Float8CurrentScalingQuantizer
     auto my_quantizer_cs = static_cast<Float8CurrentScalingQuantizer*>(quantizer.get());
-    NVTE_SCOPED_GIL_RELEASE({
-      nvte_compute_amax(input_tensor.data(), out_tensor.data(), at::cuda::getCurrentCUDAStream());
-    });
+
+    if (nvte_use_atomic_amax()) {
```
wangye805: Guard the rocm specific code changes

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -52,9 +54,33 @@ py::object quantize(const at::Tensor& tensor, py::handle quantizer, const py::ob
   if (detail::IsFloat8CurrentScalingQuantizers(quantizer.ptr())) {
     // my_quantizer here has to be a Float8CurrentScalingQuantizer
     auto my_quantizer_cs = static_cast<Float8CurrentScalingQuantizer*>(my_quantizer.get());
-    NVTE_SCOPED_GIL_RELEASE({
-      nvte_compute_amax(te_input.data(), te_output.data(), at::cuda::getCurrentCUDAStream());
-    });
+
```
wangye805: Same here, needs guarding

wangye805 (changes requested):

transformer_engine/common/include/transformer_engine/recipe.h
matthiasdiener: Upstream also uses `const` in the actual function implementation for the output (as well as naming the arguments `input_` etc.):

https://github.com/NVIDIA/TransformerEngine/blob/b3c25057405fc35d10be8109b635696e341ccf86/transformer_engine/common/recipe/current_scaling.cu#L180

I think the `const`  only refers to the pointer itself, not the content of what it points to (which is what gets modified). Not sure which way is better.

matthiasdiener: See above regarding the naming. I'm happy to change it, but not sure what the best way is. It seems to be somewhat inconsistent either way.

transformer_engine/pytorch/csrc/extensions/activation.cpp
matthiasdiener: I think https://github.com/ROCm/TransformerEngine/pull/369/commits/f933ef379584ed7acae9bff6593104633898f60c covers what we discussed today? cc @wangye805 @Micky774 

https://github.com/ROCm/TransformerEngine/pull/369/commits/ef532b19e4ba56521242421c03b947cb8096e00e removes the `use_block_amax`  parameter.

transformer_engine/common/include/transformer_engine/recipe.h
ipanfilo: This method is not needed here but can be moved to pytorch extension

transformer_engine/common/recipe/current_scaling.cu
```diff
@@ -89,24 +127,56 @@ void launch_amax_kernel(const InputType *input, float *amax, const size_t N, cud
   constexpr size_t max_blocks = 65535;
   num_blocks = std::min(num_blocks, max_blocks);
 
+#ifdef __HIP_PLATFORM_AMD__
+  const bool UseBlockAmax =
+      (block_amax != nullptr) &&
+      (block_capacity >= num_blocks) &&
+      !nvte_use_atomic_amax();
```
ipanfilo: block_amax is expected to be nullptr if nvte_use_atomic_amax() is True so it is redundant

transformer_engine/common/recipe/current_scaling.cu
matthiasdiener: Done in https://github.com/ROCm/TransformerEngine/pull/369/commits/c7d44a72ff22fe34836b537908ec7fc6edbac683

transformer_engine/common/recipe/current_scaling.cu
matthiasdiener: Done in https://github.com/ROCm/TransformerEngine/pull/369/commits/c7d44a72ff22fe34836b537908ec7fc6edbac683

transformer_engine/common/include/transformer_engine/recipe.h
matthiasdiener: Done in https://github.com/ROCm/TransformerEngine/pull/369/commits/c7d44a72ff22fe34836b537908ec7fc6edbac683

transformer_engine/common/include/transformer_engine/recipe.h
matthiasdiener: const dropped in https://github.com/ROCm/TransformerEngine/pull/369/commits/c7d44a72ff22fe34836b537908ec7fc6edbac683

matthiasdiener: Naming aligned in https://github.com/ROCm/TransformerEngine/pull/369/commits/c7d44a72ff22fe34836b537908ec7fc6edbac683

transformer_engine/pytorch/csrc/extensions/activation.cpp
matthiasdiener: Removed the cast in 8eda427b.

transformer_engine/common/include/transformer_engine/recipe.h
matthiasdiener: Removed caching and moved to pytorch extension in https://github.com/ROCm/TransformerEngine/pull/369/commits/eba552e8fdfd06a447b77089977df1c6ecaae121. I did not notice a performance difference.

transformer_engine/pytorch/csrc/extensions/activation.cpp
matthiasdiener: Done in https://github.com/ROCm/TransformerEngine/commit/c7d44a72ff22fe34836b537908ec7fc6edbac683

transformer_engine/pytorch/csrc/extensions/bias.cpp
matthiasdiener: Done in https://github.com/ROCm/TransformerEngine/commit/c7d44a72ff22fe34836b537908ec7fc6edbac683

transformer_engine/pytorch/csrc/extensions/cast.cpp
matthiasdiener: Done in https://github.com/ROCm/TransformerEngine/commit/c7d44a72ff22fe34836b537908ec7fc6edbac683

transformer_engine/common/recipe/current_scaling.cu
matthiasdiener: Changed the logic in https://github.com/ROCm/TransformerEngine/pull/369/commits/eba552e8fdfd06a447b77089977df1c6ecaae121.

matthiasdiener:
See https://github.com/ROCm/TransformerEngine/pull/384 for the GH actions CI.

ipanfilo (approved):

transformer_engine/pytorch/csrc/common.cpp
```diff
@@ -10,6 +12,10 @@
 #include "pybind.h"
 #include "transformer_engine/transformer_engine.h"
 
+#ifdef __HIP_PLATFORM_AMD__
+#include "common/common.h"
```
wangye805, line 16: Generally it's not recommended to include transformer_engine/common/common.h since only header files inside transformer_engine/common/include/transformer_engine will be exported/installed together with libtransformer_engine.so. 

Which declaration do we need inside "common/common.h" here?

transformer_engine/pytorch/csrc/common.cpp
```diff
@@ -277,4 +283,32 @@ int roundup(const int value, const int multiple) {
...
+  if (nvte_use_atomic_amax() || input_tensor.numel() == 0) {
+    // User chose atomic path, or empty tensor -> no need for workspace
+    return TensorWrapper{};
+  }
+
+  const auto N = input_tensor.numel();
+  constexpr size_t max_blocks_hw = 65535;
+
+  size_t max_blocks = DIVUP(N, static_cast<size_t>(amax_kernel_threads));
+  size_t workspace_blocks = std::min(max_blocks, max_blocks_hw);
```
wangye805: Basically this section is a logic copy/duplicate of https://github.com/matthiasdiener/TransformerEngine/blob/77b1bc32b0fc17cc735e02f713f179e578d2547e/transformer_engine/common/recipe/current_scaling.cu#L125-L128. Since this workspace size calculation is straight forward so this is okay for now. 

A better way is to define a amax workspace calculation api in libtransformer_engine.so, then call this amax workspace calculation api here to avoid having a duplicate logic

transformer_engine/common/recipe/current_scaling.cu
```diff
@@ -26,12 +26,39 @@ using bf16__ = __nv_bfloat16;
 using bf16__ = __hip_bfloat16;
 #endif //__HIP_PLATFORM_AMD__
 
-constexpr int amax_kernel_threads = 512;
```
wangye805, line 29: This amax_kernel_threads removal is unconditional but https://github.com/matthiasdiener/TransformerEngine/blob/77b1bc32b0fc17cc735e02f713f179e578d2547e/transformer_engine/common/include/transformer_engine/recipe.h#L78-L82 is guarded by __HIP_PLATFORM_AMD__. So for cuda platform, this variable is not defined.

transformer_engine/common/include/transformer_engine/recipe.h
```diff
@@ -73,6 +75,12 @@ void nvte_delayed_scaling_recipe_amax_and_scale_update_after_reduction(
     std::vector<NVTETensor> scales, const char* amax_compute_algo, NVTEDType fp8_dtype,
     float margin, cudaStream_t stream);
 
+#ifdef __HIP_PLATFORM_AMD__
+
+constexpr int amax_kernel_threads = 512;
```
wangye805, line 80: I presume this constexpr int moved to <transformer_engine/recipe.h> since it will be used externally. If so, let's check whether this naming align with other constexpr/const exported from libtransformer_engine.so? 

If this amax_kernel_threads are only used in workspace calculation/allocation outside, we can have an dedicated api to calculate the workspace size so that we won't need to export this constexpr int

wangye805 (changes requested):

transformer_engine/pytorch/csrc/common.cpp
matthiasdiener: We only need the `DIVUP` declaration; I couldn't find an equivalent within `transformer_engine/pytorch/csrc/`. Perhaps it's better just to copy `DIVUP` here?

wangye805: Yeah, I agree to copy DIVUP here instead of including the common/common.h

matthiasdiener: Done in 1357d4ba

transformer_engine/common/recipe/current_scaling.cu
matthiasdiener: Good catch, thanks. What do you think of 01b61b5a?

transformer_engine/common/include/transformer_engine/recipe.h
matthiasdiener: As far as I can tell, `amax_kernel_threads` isn't exported from `libtransformer_engine.so` (I think it is similar to a `#define`  in that regard). `nm -C libtransformer_engine.so` does not show it at all.

It is used in `common/recipe/current_scaling.cu` and `pytorch/csrc/common.cpp` which is why I moved its declaration. It is used in the workspace size calculation, as well as for the kernel's launch bounds.

In terms of names of symbols defined in the headers, `common/include/transformer_engine/` has a handful of `constexpr` without any obvious naming pattern:

```console
$ rg constexpr transformer_engine/common/include/transformer_engine/
transformer_engine/common/include/transformer_engine/gemm.h
122:constexpr int num_streams = 4;

transformer_engine/common/include/transformer_engine/recipe.h
80:constexpr int amax_kernel_threads = 512;

transformer_engine/common/include/transformer_engine/transformer_engine.h
709:  static constexpr size_t defaultData = 1;
710:  static constexpr NVTEShape defaultShape = {
```
I did not see any `const`  variables.

I think we could perhaps make the symbol `static` to further reduce the likelyhood it will be exported.

**Edit:** Or perhaps move it to the `transformer_engine::` namespace (outside the `extern "C"`)?

transformer_engine/pytorch/csrc/common.cpp
ipanfilo: While having private DIVUP is OK,please note that there is no problem in including "common/common.h" by extensions since they don't rely in on installing TE headed. Moreover, "common/common.h" is already included by both frameworks extensions.

wangye805: > While having private DIVUP is OK,please note that there is no problem in including "common/common.h" by extensions since they don't rely in on installing TE headed. Moreover, "common/common.h" is already included by both frameworks extensions.

You are right. I just searched for "common/common.h" in nvte upstream: 
```
yewang12@ctr-cx64-mi300x-13:~/workspace/NVTE_ifu_v2.4$ grep -nr "common/common.h" transformer_engine/
transformer_engine/jax/csrc/extensions.h:27:#include "common/common.h"
transformer_engine/pytorch/csrc/extensions/recipe.cpp:12:#include "common/common.h"
```

So @matthiasdiener your original way should be fine

matthiasdiener: Thanks, I reverted the change in ed16f8f1.

transformer_engine/common/include/transformer_engine/recipe.h
wangye805: Thanks for your investigation on other constexpr naming conventions. Then I'm fine with the current naming style.

BTW, is this amax_kernel_threads only used for workspace size calculation in pytorch/csrc/common.cpp? If so, it makes more sense to define a workspace size calculation api inside libtransformer_engine.so (maybe inside transformer_engine/common/recipe/current_scaling.cu). Therefore we won't need to move this constexpr into header file

matthiasdiener: > BTW, is this amax_kernel_threads only used for workspace size calculation in pytorch/csrc/common.cpp? If so, it makes more sense to define a workspace size calculation api inside libtransformer_engine.so (maybe inside transformer_engine/common/recipe/current_scaling.cu). Therefore we won't need to move this constexpr into header file

I think that would work. `amax_kernel_threads` is only used to calculate the workspace size in `pytorch/csrc/common.cpp`. ~Should this new API (`nvte_allocate_amax_workspace`?) be exported in this `recipe.h` file?~

---

~**Edit:** A downside here is that if we do not use `at::empty` but `cudaMalloc` (which I think is necessary if we move the function to `transformer_engine/common`), whoever calls `allocate_amax_workspace` must remember to deallocate the buffer `allocate_amax_workspace` returns.~

ipanfilo: What is benefit of defining API vs having const in header file?

wangye805: > What is benefit of defining API vs having const in header file?

To unify the duplicated logic of workspace size calculation logic

ipanfilo: I see. Won't it divert too much from upstream code just not to duplicate few lines? TE https://github.com/matthiasdiener/TransformerEngine/blob/77b1bc32b0fc17cc735e02f713f179e578d2547e/transformer_engine/common/recipe/current_scaling.cu#L125-L128 that you're referring to is generic.

matthiasdiener: 95dcbdff is what I came up with, and it seems to work fine (bonus: I learned about `libtransformer_engine.version` 😢). I'm not sure it is an improvement though, what do you think?

wangye805 (approved):
LGTM

transformer_engine/common/recipe/current_scaling.cu
```diff
@@ -186,6 +186,19 @@ void launch_amax_kernel(const InputType *input, float *amax, const size_t N, flo
 }  // namespace
 }  // namespace transformer_engine
 
+
+#ifdef __HIP_PLATFORM_AMD__
+
+size_t nvte_amax_workspace_size(size_t N) {
```
ipanfilo: Ambiguous workspace_size - in TE it is usually byte size but here number of float32 elements is returned.
It should either return bytes and cast to float only when launch kernels, or method should be renamed to indicate it is float elements number.

ipanfilo (changes requested):

transformer_engine/common/recipe/current_scaling.cu
matthiasdiener: Thanks, I changed it to `nvte_amax_workspace_num_blocks` in b07edf62.

ipanfilo (approved):

PR #372, opened 2025-11-16
Enable AITER V3 kernels by default

ipanfilo:
# Description

Enable FA AITER V3 kernels by default.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/ci/jax.sh
+++ b/ci/jax.sh
@@ -50,11 +50,18 @@ run_lbl() {
     _test_label=""
 }
 
+run_default_fa_lbl() {
+    if [ $_fus_attn = "$_DEFAULT_FUSED_ATTN" ]; then
+        run_lbl "$@"
+    fi
+}
+
 run_test_config() {
     echo ==== Run with Fused attention backend: $_fus_attn ====
     run_default_fa 1 test_custom_call_compute.py
     run_default_fa 1 test_functions.py
     run 1 test_fused_attn.py
+    NVTE_CK_USES_FWD_V3=0 NVTE_CK_USES_BWD_V3=0 run_default_fa_lbl "v2" 3 test_fused_attn.py # Using FAv2 for forward and backward pass
     run_default_fa 1 test_helper.py
     run_default_fa 1 test_layer.py #it effectevly always uses unfused attention
     run_default_fa 1 test_sanity_import.py
--- a/ci/jax.sh
+++ b/ci/jax.sh
@@ -55,7 +55,6 @@ run_test_config() {
     run_default_fa 1 test_custom_call_compute.py
     run_default_fa 1 test_functions.py
     run 1 test_fused_attn.py
-    NVTE_CK_USES_FWD_V3=1 NVTE_CK_USES_BWD_V3=1 run_lbl "v3" 1 test_fused_attn.py # Using FAv3 for forward and backward pass
     run_default_fa 1 test_helper.py
     run_default_fa 1 test_layer.py #it effectevly always uses unfused attention
     run_default_fa 1 test_sanity_import.py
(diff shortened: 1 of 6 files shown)


ci/jax.sh
```diff
@@ -55,7 +55,6 @@ run_test_config() {
     run_default_fa 1 test_custom_call_compute.py
     run_default_fa 1 test_functions.py
     run 1 test_fused_attn.py
-    NVTE_CK_USES_FWD_V3=1 NVTE_CK_USES_BWD_V3=1 run_lbl "v3" 1 test_fused_attn.py # Using FAv3 for forward and backward pass
```
wangye805, line 58: Let's add a dedicated v2 fused-attn check in level 3 CI like what we did for v3 before?

ci/pytorch.sh
```diff
@@ -72,7 +72,6 @@ run_test_config(){
     run 1 test_sanity.py
     run_default_fa 1 test_sanity_import.py
     run_default_fa 1 fused_attn/test_fused_attn.py # Backend selection is controlled by the test
-    NVTE_CK_USES_FWD_V3=1 NVTE_CK_USES_BWD_V3=1 run_default_fa_lbl "v3" 1 fused_attn/test_fused_attn.py # Using FAv3 for forward and backward pass
```
wangye805, line 75: Similar here. Change NVTE_CK_USES_FWD_V3/NVTE_CK_USES_BWD_V3 from 1 to 0 to test v2 ck kernels

wangye805 (changes requested):

ci/pytorch.sh
ipanfilo: Will it be better to add V2 inside test_fused_attn, so it can compare all backends including V3 vs V2.

wangye805: Functionally I think adding v2 inside test_fused_attn as a new backend is good. But it will increase our pytest runtime and violate our previous discussion that level 1 should be using all default settings?

ipanfilo: sGPU fused_attn test is quite fast one so runtime increasing with any approach is not critical. And it changes backend settings internally so using default settings is moot point for this test anyway. Let me try and see

ci/jax.sh
```diff
@@ -50,12 +50,18 @@ run_lbl() {
...
+    fi
+}
+
 run_test_config() {
     echo ==== Run with Fused attention backend: $_fus_attn ====
     run_default_fa 1 test_custom_call_compute.py
     run_default_fa 1 test_functions.py
     run 1 test_fused_attn.py
-    NVTE_CK_USES_FWD_V3=1 NVTE_CK_USES_BWD_V3=1 run_lbl "v3" 1 test_fused_attn.py # Using FAv3 for forward and backward pass
+    NVTE_CK_USES_FWD_V3=0 NVTE_CK_USES_BWD_V3=0 run_default_fa_lbl "v2" 3 test_fused_attn.py # Using FAv2 for forward and backward pass
```
wangye805, line 64: In pytorch fused-attn pytests, your current code change tests both aiter/ck v2 and v3 kernels in level 1 CI but why in jax you propose to run v2 kernels in level 3 CI?

wangye805 (changes requested):

ci/jax.sh
ipanfilo: In pytorch test it is run on level 1 because it adds just seconds to execution and it is not worth of implementing extra logic to skip it. Here it adds long running test. Level 3 is because V2 is just a fallback - V3 are expected to be always used unless some problems are encountered.

wangye805 (approved):

PR #394, opened 2025-12-05
Add instructions to download and install from wheels.

wenchenvincent:
# Description

Add instructions to download and install from wheels.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring


README.rst
```diff
@@ -28,38 +28,32 @@ Feature Support Status
 Installation
 ============
 
-Execute the following commands to install ROCm Transformer Engine from source on AMDGPUs:
-
-Known Issue with ROCm 6.4 PyTorch Release
+Install from manylinux wheels
```
wangye805, line 31: Do we want to keep both wheel installation and building from source?

wangye805 (changes requested):

README.rst
```diff
@@ -72,15 +66,40 @@ Install TE
   # Build Platform Selection (optional)
   # Note: Useful when both ROCm and CUDA platforms are present in the Docker
   export NVTE_USE_ROCM=1  #Use 1 for ROCm, or set to 0 to use CUDA; If not set will try to detect installed platform, prioritizing ROCm
-  # If you are building for gfx942 variants, also specify the number of Compute Units
-  export CU_NUM=304
 
   # Note: If the following fails with messages about missing pip packages that are installed, add "--no-build-isolation" to the command below
```
alextmagro: Can we remove the comment about --no-build-isolation now that we require it?

README.rst
wenchenvincent: yeah, let's keep it.

wangye805 (approved):

README.rst
```diff
@@ -28,38 +28,32 @@ Feature Support Status
...
-Known Issue with ROCm 6.4 PyTorch Release
+Install from manylinux wheels
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Using the docker image ``rocm/pytorch:rocm6.4_ubuntu22.04_py3.10_pytorch_release_2.5.1`` triggers a failure in the unit-test ``tests/pytorch/test_permutation.py`` (tracked in Jira ticket SWDEV-534311).
+Starting from ROCm 7.0, we provide manylinux wheels for Transformer Engine releases on `https://repo.radeon.com/rocm/manylinux`. For example, the wheels for ROCm 7.1.1 are at `https://repo.radeon.com/rocm/manylinux/rocm-rel-7.1.1/`. From the page, you can find four files related to Transformer Engine:
 
-Rebuilding PyTorch at commit ``f929e0d602a71aa393ca2e6097674b210bdf321c`` resolves the issue.
+* transformer_engine_rocm-*-py3-none-manylinux_2_28_x86_64.whl - This is the wheel file for installing the common library.
+* transformer_engine-*-py3-none-any.whl - This is the wheel file for installing the (Pytorch and JAX) extensions.
```
ipanfilo: This one is for installing common TE python package and the _rocm one is to install common ROCm library. Transfornmer_engine_rocm* should not be installed by itself

README.rst
```diff
@@ -72,15 +66,40 @@ Install TE
...
-  export CU_NUM=304
 
   # Note: If the following fails with messages about missing pip packages that are installed, add "--no-build-isolation" to the command below
-  pip install .
+  pip install . --no-build-isolation
 
 It is also possible to build wheels for later installation with "pip wheel ." although those wheels will not be portable to systems with
 different libraries installed. This build may also require "--no-build-isolation" and if the build still fails with this flag try installing setuptools<80.0.0
 
+Known Issue with ROCm 6.4 PyTorch Release
```
ipanfilo, line 85: Do we still need this ROCm 6.4 section?

README.rst
wenchenvincent: Removed.

README.rst
```diff
@@ -28,38 +28,32 @@ Feature Support Status
...
 
-Rebuilding PyTorch at commit ``f929e0d602a71aa393ca2e6097674b210bdf321c`` resolves the issue.
+* transformer_engine_rocm-*-py3-none-manylinux_2_28_x86_64.whl - This is the wheel file for installing the common library.
+* transformer_engine-*-py3-none-any.whl - This is the wheel file for installing the (Pytorch and JAX) extensions.
+* transformer_engine_jax-*.tar.gz - This is the source tar ball for the JAX extension.
+* transformer_engine_torch-*.tar.gz - This is the source tar ball for the Pytorch extension.
 
-Re-install PyTorch
-^^^^^^^^^^^^^^^^^^
+Below are the example commands to download and install the wheels:
```
ipanfilo: Let's add that they install both Pytorch and JAX integration on system with both frameworks installed

README.rst
```diff
@@ -28,8 +28,50 @@ Feature Support Status
...
 Execute the following commands to install ROCm Transformer Engine from source on AMDGPUs:
 
+.. code-block:: bash
+
+  # Clone TE repo and submodules
+  git clone --recursive https://github.com/ROCm/TransformerEngine.git
+
+  cd TransformerEngine
+  export NVTE_FRAMEWORK=pytorch,jax #optionally set framework, currently only support pytorch and jax; if not set will try to detect installed frameworks
+  export NVTE_ROCM_ARCH=gfx942 # CK fused attn only support MI200 and MI300 and fp8 features are only supported on MI300
```
ipanfilo: This comment is outdated.

ipanfilo (changes requested):

ipanfilo:
Would it also make sense to explicitly mention that it is safe way to uninstall TE first when switching between wheel and source install and give sample command how to clearly uninstall TE installed from wheels: pip list | grep transformer.engine | xargs pip uninstall -y

README.rst
wenchenvincent: Some customers still use ROCm 6.4...I am thinking of removing this section in 2 months.

wenchenvincent: Actually, do we know if this is fixed in 6.4.x? If so, we can just point to the right version.

README.rst
wenchenvincent: transformer_engine-*-py3-none-any.whl is for also for extracting from the pytorch and jax extension tar balls and installing the extensions, isn't it?

Agreed that Transfornmer_engine_rocm* should not be installed by itself

README.rst
wenchenvincent: You mean, we should add MI350, right?

wenchenvincent:
> Would it also make sense to explicitly mention that it is safe way to uninstall TE first when switching between wheel and source install and give sample command how to clearly uninstall TE installed from wheels: pip list | grep transformer.engine | xargs pip uninstall -y

Did you see issues if we didn't uninstall TE first?

README.rst
ipanfilo: Maybe comment here should just contain info that for example gfx942 is for MI300, gfx950 for MI350.
Fused attention backend or other features  support matrix, if we want to put it, should be separate topic.

README.rst
ipanfilo: It is not used for extracting but it defines dependencies on _jax and _torch packages as extras: transformer_engine[jax] transformer_engine[torch]

ipanfilo:
> Did you see issues if we didn't uninstall TE first?

Yes. It is partially https://github.com/ROCm/frameworks-internal/issues/13350 but recent TE versions also made changes and if install from source on top of wheels TE loading fails complaining it found two versions of library

alextmagro (approved):

README.rst
wenchenvincent: Updated.

README.rst
wenchenvincent: Updated.

README.rst
wenchenvincent: Updated.

wenchenvincent:
> > Did you see issues if we didn't uninstall TE first?
> 
> Yes. It is partially [ROCm/frameworks-internal#13350](https://github.com/ROCm/frameworks-internal/issues/13350) but recent TE versions also made changes and if install from source on top of wheels TE loading fails complaining it found two versions of library

I added a note on that.

README.rst
wenchenvincent: Let's update that in a later PR.

README.rst
```diff
@@ -28,8 +28,59 @@ Feature Support Status
...
+different libraries installed. If the build still fails with the "--no-build-isolation" flag try installing setuptools<80.0.0
+
+Note on Switching between Installation from Source and Installation from Wheels
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Sometimes, issues might occur when installing from source on a system where a previous installation with wheels, or vice versa. It is safe to uninstall TE first before 
+switching between installing from source and installing from wheels. Here is the example command:
+
+.. code-block:: bash
+
+  pip list | grep transformer_engine | xargs pip uninstall -y
```
ipanfilo: Nit: old setuptools may generate transformer-engine name so it might be better to do 'grep transformer.engine'

ipanfilo (approved):

README.rst
wenchenvincent: Good catch. Updated.

PR #417, opened 2026-01-16
GEMMTestSuite: perform input data generation on GPU (incl. hiprand)

matthiasdiener:
# Description

Use `hiprand` random number generation (instead of CPU/OpenMP).

Partly addresses https://github.com/ROCm/frameworks-internal/issues/14746.

Notes:
- The generated numbers are deterministic across executions
- The generated numbers are *not* the same as the ones generated via the previous CPU implementation
- Performance: https://github.com/ROCm/frameworks-internal/issues/14746#issuecomment-3761360289

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Use `hiprand` random number generation (instead of CPU/OpenMP).

# Checklist:


tests/cpp/test_common.cu
```diff
@@ -783,7 +783,6 @@ std::pair<double, double> getTolerances(const DType type) {
 template <typename T>
 void generate_data_uniformly(T* data, const size_t size, std::mt19937* gen) {
 #ifdef __HIP_PLATFORM_AMD__
-  // TODO: Introduce a parallel RNG library (Random123, PCG, rocRAND)
```
alextmagro, line 792: We can probably remove this entire #ifdef guarded section, right?

tests/cpp/test_common.cu
```diff
@@ -822,21 +821,71 @@ void generate_data_uniformly(T* data, const size_t size, std::mt19937* gen) {
...
   } else {
     const size_t size = product(t->columnwise_shape());
     TRANSFORMER_ENGINE_TYPE_SWITCH_ALL(t->dtype(), T,
       {
+#ifdef __HIP_PLATFORM_AMD__
+        fillUniformDevice(t);
+#else
         T *data = t->columnwise_cpu_dptr<T>();
         generate_data_uniformly(data, size, &(t->gen()));
+#endif
```
alextmagro, line 938: Are we able to use rocRAND for fillCase_special as well? Also, I think there were a few tests that for some reason generate their own data... I might be wrong about that, or it may have been updated.

alextmagro (commented):
Awesome work! Do you have any rough estimates of how much faster the entire cpp test suite is?

tests/cpp/test_common.cu
matthiasdiener: I think so, unless we want to keep it around for future reference?

matthiasdiener:
> Awesome work! Do you have any rough estimates of how much faster the entire cpp test suite is?

Thanks! I left some performance numbers here: https://github.com/ROCm/frameworks-internal/issues/14746#issuecomment-3761360289

tests/cpp/test_common.cu
alextmagro: Let's remove, we can always revert if we need it again.

ipanfilo:
Update copyright date of modified files

tests/cpp/test_common.cu
```diff
@@ -822,21 +821,71 @@ void generate_data_uniformly(T* data, const size_t size, std::mt19937* gen) {
 #endif
 }
 
+#ifdef __HIP_PLATFORM_AMD__
+#include <rocrand/rocrand.h>
```
ipanfilo: Even though it does not cause errors, better move #include to the top of the file, out of test namespace

tests/cpp/test_common.cu
```diff
@@ -822,21 +821,71 @@ void generate_data_uniformly(T* data, const size_t size, std::mt19937* gen) {
...
+}
+#endif
+
 void fillUniform(Tensor *t) {
   if (t->rowwise()) {
     const size_t size = product(t->rowwise_shape());
     TRANSFORMER_ENGINE_TYPE_SWITCH_ALL(t->dtype(), T,
       {
+#ifdef __HIP_PLATFORM_AMD__
+        fillUniformDevice(t);
```
ipanfilo: Is there any test that tests this generation? I think using GPU generation here does not produce correct result because of using t->from_cpu() below in this method

tests/cpp/test_common.cu
ipanfilo: The whole method seems unused on ROCm and can be guarded

tests/cpp/test_common.cu
```diff
@@ -822,21 +821,71 @@ void generate_data_uniformly(T* data, const size_t size, std::mt19937* gen) {
...
+  // per-tensor deterministic seed
+  const unsigned long long seed = static_cast<unsigned long long>(t->gen()());
+  rocrand_generator gen;
+  rocrand_create_generator(&gen, ROCRAND_RNG_PSEUDO_PHILOX4_32_10);
+  rocrand_set_seed(gen, seed);
+
+  rocrand_generate_uniform(gen, tmp, N);
+
+  // map to [-2.0, 1.0] (like generate_data_uniformly) and cast into tensor dtype
+  TRANSFORMER_ENGINE_TYPE_SWITCH_ALL(t->dtype(), T, {
```
ipanfilo: T should either be template parameter and no TRANSFORMER_ENGINE_TYPE_SWITCH_ALL here, or the method calling should be moved out of TRANSFORMER_ENGINE_TYPE_SWITCH_ALL in fillUniform

tests/cpp/test_common.cu
```diff
@@ -828,25 +831,81 @@ void generate_data_uniformly(T* data, const size_t size, std::mt19937* gen) {
...
+  // Clamp values in *in* to [lo, hi] and cast to type *T* for *out*.
+  size_t idx = blockIdx.x * blockDim.x + threadIdx.x;
+  if (idx < n) {
+    in[idx] = lo + (hi - lo) * in[idx];
+    out[idx] = static_cast<T>(in[idx]);
+  }
+}
+
+void fillUniformDevice(Tensor* t) {
+  void* dst = t->rowwise() ? t->rowwise_dptr() : t->columnwise_dptr();
```
wangye805: I think we should have another wrapper just to fill an linear buffer with random values, regardless of whether this tensor contains valid rowwise or columnwise data. We can have a fillUniformTensor calling such an fill uniform linear buffer function.

tests/cpp/test_common.cu
```diff
@@ -828,25 +831,81 @@ void generate_data_uniformly(T* data, const size_t size, std::mt19937* gen) {
...
+  TRANSFORMER_ENGINE_TYPE_SWITCH_ALL(t->dtype(), T, {
+    dim3 block(256);
+    dim3 grid((N + block.x - 1) / block.x);
+    affine_transform_and_cast<T><<<grid, block, 0, 0>>>(
+      tmp, reinterpret_cast<T*>(dst), N, -2.0f, 1.0f);
+
+    // Copy into the CPU mirror. We could use Tensor::to_cpu() here,
+    // but that does more than just copying the data.
+    T* cpu_dst = t->rowwise() ? t->rowwise_cpu_dptr<T>() : t->columnwise_cpu_dptr<T>();
+    cudaMemcpy(cpu_dst, dst, N * sizeof(T), hipMemcpyDeviceToHost);
```
wangye805: I recall we have to_cpu and from_cpu methods in test tensor class

wangye805 (changes requested):

tests/cpp/test_common.cu
matthiasdiener: c8e5c733 removes the extra AMD code and guards this method completely.

tests/cpp/test_common.cu
matthiasdiener: My worry was that since `Tensor::to_cpu()` does more than just copy the raw tensor data (e.g., it also copies amax and scale data), it may be somewhat dangerous to use `to_cpu()` here, which is why I used `cudaMemcpy` here instead. 

What do you think?

tests/cpp/test_common.cu
matthiasdiener: What do you think of bdb8349a?

tests/cpp/test_common.cu
matthiasdiener: > Are we able to use rocRAND for fillCase_special as well?

I added this in b6eee816.

> Also, I think there were a few tests that for some reason generate their own data... I might be wrong about that, or it may have been updated.

At least test_dequantize_mxfp8.cu seems to have its own data generation, but I haven't had a deeper look.

tests/cpp/test_common.cu
matthiasdiener: Done in https://github.com/ROCm/TransformerEngine/pull/417/commits/0f008e9859e385ad0b40510ffc8dedc962b1526d

tests/cpp/test_common.cu
matthiasdiener: With the restructuring in https://github.com/ROCm/TransformerEngine/commit/bdb8349ac65a1edcd86ada78f77bdd4edfc04181, I believe this comment is now addressed?

tests/cpp/test_common.cu
matthiasdiener: Good catch, thanks. I disabled the `from_cpu`  call here, and added an explicit copy of the data to CPU in https://github.com/ROCm/TransformerEngine/pull/417/changes/0f008e9859e385ad0b40510ffc8dedc962b1526d, as there are other places that copy from CPU->GPU (like `set_scale_inv`). I also added a new test to check that GPU and CPU copies are the same in https://github.com/ROCm/TransformerEngine/pull/417/commits/097ecd4e8dd9e41c86a0095733e482afb9ab185b.

tests/cpp/test_common.cu
```diff
@@ -23,6 +23,10 @@
 #include <transformer_engine/transformer_engine.h>
 #include "util/logging.h"
 
+#ifdef __HIP_PLATFORM_AMD__
```
alextmagro: I don't think we need to guard this, since it is hipified.

tests/cpp/test_common.cu
```diff
@@ -825,28 +822,134 @@ void generate_data_uniformly(T* data, const size_t size, std::mt19937* gen) {
...
+  }
+}
+
+template <typename T>
+__global__ void apply_random_sign(T* __restrict__ data,
+                                 const float* __restrict__ signs,
+                                 size_t n) {
+  size_t idx = blockIdx.x * blockDim.x + threadIdx.x;
+  if (idx < n) {
+    if (signs[idx] < 0.5f) {
```
alextmagro: Before, we would generate data uniformly from -2 to 1 -- I don't know if there was a specific reason for that, but for consistency could we change the if statement to < .66f?

alextmagro (approved):
LGTM! A few small comments, but otherwise excited to see the benefits for our CI.

tests/cpp/test_common.cu
matthiasdiener: Maybe I'm overlooking something, but this kernel (`apply_random_sign`) is only used for the `fillCase_special` case, where the previous code did:

https://github.com/ROCm/TransformerEngine/blob/2bc74c8281037b8ff9ffd77a568a9002fc2cb94e/tests/cpp/test_common.cu#L873-L881

, which seems to indicate a 50% chance of a ~negative~ flipped sign.

Edit:
The original CPU code here is weird, it creates random values between -2 and +1, and then randomly flips the sign. My changes reproduce this behavior, but perhaps it all could be a bit clearer....

tests/cpp/test_common.cu
matthiasdiener: Thanks, I removed it in dfd51e14.

tests/cpp/test_common.cu
alextmagro: Ah, understood!

tests/cpp/test_common.cu
wangye805: Looks great. Thanks

tests/cpp/test_common.cu
wangye805: You are right. Here we just want to copy the generated rowwise/colwise data ptr

tests/cpp/test_common.cu
```diff
@@ -825,28 +820,134 @@ void generate_data_uniformly(T* data, const size_t size, std::mt19937* gen) {
...
+  // Fill a linear device buffer with uniform randoms in [*lo*, *hi*] and cast them to *T*.
+  // Optionally mirror the result into a provided CPU pointer.
+  if (N == 0)
+    return;
+
+  float* tmp = nullptr;
+  NVTE_CHECK_CUDA(cudaMalloc(&tmp, N * sizeof(float)));
+
+  float* tmp_sign = nullptr;
+  if (random_sign) {
```
wangye805, line 863: I checked the usage of fillCaseSpecial in our current rocm dev, it's not used anywhere:

Image

tests/cpp/test_common.cu
```diff
@@ -825,28 +820,134 @@ void generate_data_uniformly(T* data, const size_t size, std::mt19937* gen) {
...
+  dim3 block(256);
+  dim3 grid((N + block.x - 1) / block.x);
+
+  affine_transform_and_cast<T><<<grid, block, 0, 0>>>(
+      tmp, reinterpret_cast<T*>(dst_dev), N, lo, hi);
+
+  if (random_sign) {
+    apply_random_sign<T><<<grid, block, 0, 0>>>(
+          reinterpret_cast<T*>(dst_dev), tmp_sign, N);
+  }
```
wangye805: If we really need this random_sign feature, can we fuse this apply_random_sign kernel to its previous affine_transform_and_cast before to save some IO?

wangye805 (changes requested):

tests/cpp/test_common.cu
matthiasdiener: As far as I can see, it is used via the `fillCase()` function:

https://github.com/ROCm/TransformerEngine/blob/2bc74c8281037b8ff9ffd77a568a9002fc2cb94e/tests/cpp/test_common.cu#L890-L904

, e.g. in https://github.com/ROCm/TransformerEngine/blob/2bc74c8281037b8ff9ffd77a568a9002fc2cb94e/tests/cpp/operator/test_cast_mxfp8.cu#L220

tests/cpp/test_common.cu
matthiasdiener: Good idea, thanks. What do you think of ddfcf2d0?

tests/cpp/test_common.cu
wangye805: Oh, thanks for pointing our

wangye805 (approved):
LGTM

tests/cpp/test_common.cu
ipanfilo: Seems not. fillUniformTensorDevice is still called from t->dtype() switch and in turn calls fillUniformLinearBuferDevice from t->dtype() switch

ipanfilo (changes requested):

tests/cpp/test_common.cu
matthiasdiener: You're right. What do you think of [33f6124](https://github.com/ROCm/TransformerEngine/pull/417/commits/33f6124511e4e85af536ba40854a3711b2f7e5aa)?

ipanfilo (approved):

PR #434, opened 2026-01-28
Grouped GEMM with ck_tile

matthiasdiener:
# Description

See https://github.com/ROCm/frameworks-internal/issues/15185 and https://github.com/ROCm/frameworks-internal/issues/13792 for context.

Primus-Turbo implementation: https://github.com/AMD-AGI/Primus-Turbo/blob/5bcd13785ef380fec0eec0911b7d6db5e606143e/csrc/kernels/grouped_gemm

TODOs:
- [x] Enable tests in test_numerics.py
- [x] Make kernels selectable & tunable
- [x] Handle gelu/bias (or make sure these are not passed in)
- [x] Performance analysis and improvements: https://github.com/ROCm/frameworks-internal/issues/15185#issuecomment-3863052452
- [x] More tests

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Implement ck_tile-based group GEMM, similar to Cutlass

# Checklist:


tests/pytorch/test_numerics.py
```diff
@@ -2133,6 +2133,9 @@ def test_grouped_linear_accuracy_cutlass(
     delay_wgrad_compute,
 ):
     os.environ["NVTE_USE_CUTLASS_GROUPED_GEMM"] = "1"
+    if IS_HIP_EXTENSION:
+        os.environ["NVTE_USE_CK_GROUPED_GEMM"] = "1"
+        os.environ["NVTE_CK_GROUPED_GEMM_WARN_FALLBACK"] = "1"
```
Micky774: Do we want to use Ilya's `EnvVarCleaner` manager?

tests/pytorch/test_numerics.py
```diff
@@ -2133,6 +2133,9 @@ def test_grouped_linear_accuracy_cutlass(
     delay_wgrad_compute,
 ):
     os.environ["NVTE_USE_CUTLASS_GROUPED_GEMM"] = "1"
+    if IS_HIP_EXTENSION:
```
wangye805: Is our CK grouped gemm a drop-in replacement with NV upstream CUTLASS grouped gemm? If so, we can share the same env. It's like cublaslt vs hipblaslt...

transformer_engine/common/gemm/ck_grouped_gemm.cpp
```diff
@@ -0,0 +1,387 @@
...
+          AType, BType, ck_tile::tuple<>, AccType,
+          CType, ck_tile::tuple<>, CLayout,
+          ck_tile::element_wise::PassThrough,
+          Partitioner::MPerBlock, Partitioner::NPerBlock,
+          TileCfg::M_Warp, TileCfg::N_Warp,
+          TileCfg::M_Warp_Tile, TileCfg::N_Warp_Tile, TileCfg::K_Warp_Tile,
+          Problem::TransposeC, MemOp>>;
+
+  using Kernel = ck_tile::GroupedGemmKernel<Partitioner, Pipeline, Epilogue>;
+};
```
wangye805: Are these codes from CK repo? If so, can you add a comment to point to the reference?

transformer_engine/common/gemm/ck_grouped_gemm.cpp
```diff
@@ -0,0 +1,387 @@
...
+  using Kernel = typename R::Kernel;
+
+  const size_t needed = Kernel::GetWorkSpaceSize(group_num);
+  if (!workspace || workspace_bytes < needed) {
+    NVTE_ERROR("grouped_gemm_ck_tile: insufficient workspace. Needed bytes=", needed);
+    return false;
+  }
+
+  std::vector<ck_tile::GroupedGemmHostArgs<0>> descs;
+  descs.reserve(group_num);
```
wangye805: Why not put group_num inside the desc vector definition?

transformer_engine/common/gemm/ck_grouped_gemm.cpp
```diff
@@ -0,0 +1,387 @@
...
+                             transformer_engine::Tensor* const* D,
+                             int group_num,
+                             bool transA_use,
+                             bool transB_use,
+                             void* workspace,
+                             size_t workspace_bytes,
+                             hipStream_t stream)
+{
+  using R = Runner<T, T, T, ALayout, BLayout, CLayout, TileCfg_basic, MemOp>;
+  using Kernel = typename R::Kernel;
```
wangye805: This R is not used anywhere else

transformer_engine/common/gemm/ck_grouped_gemm.cpp
```diff
@@ -0,0 +1,387 @@
...
+
+  for (int i = 0; i < group_num; ++i) {
+    const auto& a = data_view(*A_use[i]);
+    const auto& b = data_view(*B_use[i]);
+    const auto& d = data_view(*D[i]);
+
+    if (a.shape.size() != 2 || b.shape.size() != 2 || d.shape.size() != 2) {
+      NVTE_ERROR("grouped_gemm_ck_tile: expected all groups to be 2D.");
+      return false;
+    }
```
wangye805: Does grouped gemm support generalized matrices from high-dimensional tensors? Regular gemm supports that. And TE treat the last dim as col with other dimensions as row: https://github.com/ROCm/TransformerEngine/blob/9d6b0e55d5affbdad34352d697032dd67419ee19/transformer_engine/common/common.h#L238-L262

transformer_engine/common/gemm/ck_grouped_gemm.cuh
```diff
@@ -0,0 +1,387 @@
...
+                          bool accumulate,
+                          hipStream_t stream)
+{
+  const transformer_engine::Tensor* const* A_use = A;
+  const transformer_engine::Tensor* const* B_use = B;
+  bool transA_use = transA;
+  bool transB_use = transB;
+
+  // If TE's flags disagree with storage, infer the correct mode from shapes.
+  if (!infer_gemm_mode_group0(A, B, D, group_num, A_use, B_use, transA_use, transB_use)) {
```
wangye805: I'm confused about this step. As far as I know, NV upstream does not need this. Is this needed because of the difference between our ck groupedgemm and cutlass groupedgemm?

transformer_engine/common/gemm/ck_grouped_gemm.cuh
```diff
@@ -0,0 +1,387 @@
...
+            A_use, B_use, D, group_num, true, false, workspace, workspace_bytes, stream)
+        : run_grouped_impl<T, ColMajor, RowMajor, RowMajor, ck_tile::memory_operation_enum::atomic_add>(
+            A_use, B_use, D, group_num, true, false, workspace, workspace_bytes, stream);
+
+    return (memop == ck_tile::memory_operation_enum::set)
+      ? run_grouped_impl<T, ColMajor, ColMajor, RowMajor, ck_tile::memory_operation_enum::set>(
+          A_use, B_use, D, group_num, true, true, workspace, workspace_bytes, stream)
+      : run_grouped_impl<T, ColMajor, ColMajor, RowMajor, ck_tile::memory_operation_enum::atomic_add>(
+          A_use, B_use, D, group_num, true, true, workspace, workspace_bytes, stream);
+  }
```
wangye805: We can create some macros like https://github.com/ROCm/TransformerEngine/blob/9d6b0e55d5affbdad34352d697032dd67419ee19/transformer_engine/common/common.h#L482 to make this section more concise

transformer_engine/common/gemm/ck_grouped_gemm.cuh
```diff
@@ -0,0 +1,387 @@
...
+}
+
+bool grouped_gemm_ck_tile(const NVTETensor* A,
```
wangye805: Why do we overload this function? In cublaslt_gemm.cu, it's only called by this signature. Perhaps we can rename the grouped_gemm_ck_tile in line 255

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -788,7 +791,35 @@ void nvte_multi_tensor_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor
   NVTE_API_CALL(nvte_multi_tensor_gemm);
 
 #ifdef __HIP_PLATFORM_AMD__
-    multi_stream_cublas_gemm(A, B, D, bias, pre_gelu_out, num_gemms, transa, transb, grad,
+  const bool use_ck = transformer_engine::getenv<bool>("NVTE_USE_CK_GROUPED_GEMM", false);
+  const bool warn_fallback =
+      transformer_engine::getenv<bool>("NVTE_CK_GROUPED_GEMM_WARN_FALLBACK", false);
+
+  auto is_supported_dtype = [&]() -> bool {
+    auto *inputA = transformer_engine::convertNVTETensorCheck(A[0]);
```
wangye805: Is it possible that num_group=0 so A[0] access not valid?

transformer_engine/common/CMakeLists.txt
```diff
@@ -241,6 +241,14 @@ endif()
 target_include_directories(transformer_engine PUBLIC
                            "${CMAKE_CURRENT_SOURCE_DIR}/include")
 
+set(CK_ROOT ${CMAKE_SOURCE_DIR}/../../3rdparty/aiter/3rdparty/composable_kernel)
+
+target_include_directories(transformer_engine
+  BEFORE PRIVATE
```
wangye805: Why using keyword BEFORE in this target_include_directories? Is it because cmake will not be able to find the correct header files without prioritizing the ck include dirs?

transformer_engine/common/CMakeLists.txt
```diff
@@ -241,6 +241,14 @@ endif()
 target_include_directories(transformer_engine PUBLIC
                            "${CMAKE_CURRENT_SOURCE_DIR}/include")
 
+set(CK_ROOT ${CMAKE_SOURCE_DIR}/../../3rdparty/aiter/3rdparty/composable_kernel)
```
wangye805: CMAKE_SOURCE_DIR --> CMAKE_CURRENT_SOURCE_DIR? Not sure whether other upstream libs will depend on us but let's make it future proof

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -24,8 +24,11 @@
 #include "../util/logging.h"
 #include "../util/multi_stream.h"
 #include "common/util/cuda_runtime.h"
+#include "common/util/system.h"
 #ifndef __HIP_PLATFORM_AMD__
 #include "cutlass_grouped_gemm.cuh"
```
wangye805, line 29: NV upstream made another .cu file for their cutlass_grouped_gemm and compiled it separately. Maybe we can follow their structure for better isolation (avoid CK defining some macros contaminating our cublaslt_gemm.cu)

wangye805 (changes requested):

transformer_engine/common/gemm/ck_grouped_gemm.cpp
matthiasdiener: I used `reserve()` here instead of `std::vector<ck_tile::GroupedGemmHostArgs<0>> descs(group_num);` to avoid default-constructing GroupedGemmHostArgs objects that are immediately overwritten, to reduce construction overhead.

transformer_engine/common/gemm/cublaslt_gemm.cu
matthiasdiener: I restructured this to a cpp file and a header file in https://github.com/ROCm/TransformerEngine/pull/434/changes/259645cdbee80c336acaa1b47420a4db2873c47a.

transformer_engine/common/gemm/ck_grouped_gemm.cpp
matthiasdiener: I added a reference in fac7c111.

transformer_engine/common/gemm/ck_grouped_gemm.cpp
matthiasdiener: I merged R into the next line in fac7c111.

transformer_engine/common/gemm/ck_grouped_gemm.cpp
matthiasdiener: I added (untested) support for higher-dim tensors in dd3ed2f3

transformer_engine/common/gemm/ck_grouped_gemm.cpp
aris134: I can see the comment with the reference to CK repo, so I am resolving this.

aris134 (approved):

transformer_engine/common/gemm/ck_grouped_gemm.cuh
matthiasdiener: What do you think of the current version? (changed mostly in https://github.com/ROCm/TransformerEngine/pull/434/changes/9986bd4a684a2ab6a7e61e398323ed22e8070d6a)

transformer_engine/common/gemm/ck_grouped_gemm.cuh
matthiasdiener: I simplified this in https://github.com/ROCm/TransformerEngine/pull/434/changes/259645cdbee80c336acaa1b47420a4db2873c47a so that there is no more overload (only this signature remains).

transformer_engine/common/gemm/cublaslt_gemm.cu
matthiasdiener: https://github.com/ROCm/TransformerEngine/pull/434/changes/a42f7ca9245d6ffc40e3408c239bbc4593f3d2a8 removed the separate implementation of is_supported_dtype (among others), changing the code to use the upstream version of that function for the most part. I believe the upstream code has the same issue, so I added an early exit in 0b162874. What do you think?

transformer_engine/common/CMakeLists.txt
matthiasdiener: I removed BEFORE in https://github.com/ROCm/TransformerEngine/commit/259645cdbee80c336acaa1b47420a4db2873c47a, compilation still seems to work fine.

transformer_engine/common/CMakeLists.txt
matthiasdiener: Changed to in CMAKE_CURRENT_SOURCE_DIR in https://github.com/ROCm/TransformerEngine/commit/259645cdbee80c336acaa1b47420a4db2873c47a.

tests/pytorch/test_numerics.py
matthiasdiener: It mostly is a drop-in replacement for upstream, so I changed the envs to the upstream versions in https://github.com/ROCm/TransformerEngine/commit/259645cdbee80c336acaa1b47420a4db2873c47a

transformer_engine/common/gemm/ck_grouped_gemm.cuh
matthiasdiener: There is perhaps an easier way to do it - in [58b34e7](https://github.com/ROCm/TransformerEngine/pull/434/commits/58b34e7ba39dfdfcae58bad689ecfd2850eea384), I'm just imitating what upstream does. What do you think?

tests/pytorch/test_numerics.py
```diff
@@ -28,7 +28,7 @@
     is_bf16_compatible,
 )
 if IS_HIP_EXTENSION:
-    from transformer_engine.pytorch.utils import is_mi200, is_mi308
+    from transformer_engine.pytorch.utils import is_mi200, is_mi308, is_mi300_class
```
ipanfilo: is_mi300_class methods is not needed, it is just 9.4 gfx family

transformer_engine/common/gemm/ck_grouped_gemm.cpp
```diff
@@ -0,0 +1,276 @@
+/* Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved. */
```
ipanfilo: Add proper copyright header

transformer_engine/common/gemm/ck_grouped_gemm.h
```diff
@@ -0,0 +1,11 @@
+/* Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved. */
```
ipanfilo: Put proper copyright header

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -788,9 +791,10 @@ void nvte_multi_tensor_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor
...
 #ifdef __HIP_PLATFORM_AMD__
-    multi_stream_cublas_gemm(A, B, D, bias, pre_gelu_out, num_gemms, transa, transb, grad,
-                             workspace, accumulate, use_split_accumulator, math_sm_count, stream);
-#else
+  if (num_gemms <= 0)
+    return;
+#endif
+
   const int current_device = transformer_engine::cuda::current_device();
   const bool is_hopper = (transformer_engine::cuda::sm_arch(current_device) == 90);
```
ipanfilo, line 798: These constants are not used on ROCm

transformer_engine/common/gemm/cublaslt_gemm.cu
ipanfilo, line 829: Unused on ROCm

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -837,12 +846,21 @@ void nvte_multi_tensor_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor
     auto *inputA = transformer_engine::convertNVTETensorCheck(A[0]);
     auto *inputB = transformer_engine::convertNVTETensorCheck(B[0]);
     auto *OutputD = transformer_engine::convertNVTETensorCheck(D[0]);
+#ifdef __HIP_PLATFORM_AMD__
+    auto A_dt = inputA->data.dtype;
+    auto B_dt = inputB->data.dtype;
+    auto D_dt = OutputD->data.dtype;
+    return (A_dt == B_dt) && (A_dt == D_dt) &&
```
ipanfilo, line 855: Are CK tile constraints the same as CUTLASS?

transformer_engine/common/gemm/ck_grouped_gemm.cpp
```diff
@@ -0,0 +1,276 @@
...
+  size_t ws_bytes = 0;
+  if (workspace) {
+    auto* ws_te = transformer_engine::convertNVTETensorCheck(*workspace);
+    ws_ptr   = ws_te->data.dptr;
+    ws_bytes = ws_te->data.numel() *
+               transformer_engine::typeToSize(ws_te->data.dtype);
+  }
+
+  // Normalize similar to upstream
+  // See https://github.com/NVIDIA/TransformerEngine/blob/59f6f3876767d07045152bfae07b5dd4c54e1725/transformer_engine/common/gemm/cutlass_grouped_gemm.cu#L54-L68
```
ipanfilo, line 251: There is no similar code in referenced upstream file. And can you explain transA_use = transB and vice versa

transformer_engine/common/gemm/ck_grouped_gemm.cpp
```diff
@@ -0,0 +1,276 @@
...
+                                    bool transB_use,
+                                    const transformer_engine::Tensor* const* A_use,
+                                    const transformer_engine::Tensor* const* B_use,
+                                    transformer_engine::Tensor* const* D,
+                                    int group_num,
+                                    void* workspace,
+                                    size_t workspace_bytes,
+                                    hipStream_t stream) {
+
+// FIXME: This could be a templated lambda function in C++20.
```
ipanfilo: As an alternative dispatch_grouped can be incorporated to ck_tile_grouped_gemm with using of nested TRANSFORMER_ENGINE_SWITCH_CONDITION

ipanfilo (changes requested):

transformer_engine/common/gemm/ck_grouped_gemm.cpp
wangye805: I see. That makes sense. Thanks

transformer_engine/common/gemm/ck_grouped_gemm.cuh
wangye805: Yeah, the codes look cleaner

transformer_engine/common/gemm/cublaslt_gemm.cu
wangye805: Yeah, not sure about how upstream pytorch/jax is going to use this API. So adding this guard/early exit is a good idea

tests/pytorch/test_numerics.py
```diff
@@ -2121,6 +2121,8 @@ def test_grouped_linear_accuracy(
     atol, rtol = 0, 0
     if use_cutlass:
         atol, rtol = 1e-3, 1e-3
+        if IS_HIP_EXTENSION and is_mi300_class():
+            atol, rtol = 3e-2, 3e-2
```
wangye805: Just curious, why does our MI300 class need such a higher atol/rtol? Is this expected?

tests/pytorch/test_numerics.py
```diff
@@ -2936,7 +2938,10 @@ def test_grouped_gemm(shape, dtype, layout, accumulate, use_cutlass):
             # cublas implementation should be bit-wise match
             torch.testing.assert_close(o, o_ref, rtol=0, atol=0)
         else:
-            torch.testing.assert_close(o, o_ref, rtol=1.5e-2, atol=1.5e-2)
+            if IS_HIP_EXTENSION and is_mi300_class():
+                torch.testing.assert_close(o, o_ref, rtol=2.0e-2, atol=3.0e-2)
```
wangye805: Same question here

transformer_engine/common/gemm/cublaslt_gemm.cu
matthiasdiener: Right, would you like me to #ifdef this function out (it is coming from upstream)?

transformer_engine/common/gemm/ck_grouped_gemm.cpp
matthiasdiener: Thanks, done in f680d6a8

transformer_engine/common/gemm/ck_grouped_gemm.h
matthiasdiener: Thanks, done in f680d6a8

transformer_engine/common/gemm/cublaslt_gemm.cu
matthiasdiener: In terms of supported data types (which this function handles), yes - only bf16/fp16 are supported.

transformer_engine/common/gemm/ck_grouped_gemm.cpp
matthiasdiener: What do you think of 6d85088f?

tests/pytorch/test_numerics.py
matthiasdiener: Removed in 79100384

transformer_engine/common/gemm/cublaslt_gemm.cu
matthiasdiener: I disabled them in e8ebb0ea.

transformer_engine/common/gemm/ck_grouped_gemm.cpp
matthiasdiener: In the [referenced upstream code](https://github.com/NVIDIA/TransformerEngine/blob/59f6f3876767d07045152bfae07b5dd4c54e1725/transformer_engine/common/gemm/cutlass_grouped_gemm.cu#L54-L68), the same swap is performed. For example, consider the following case in the upstream code:

```python
} else if (!transb && transa) {
      grouped_gemm::CutlassGroupedGemm<false, true, T>(B, A, D, workspace, alpha, beta, num_gemms,
                                                       stream, device, math_sm_count);
```

Here, `transa==true` and `transb==false`, but they get passed into the template as  `transa==false` and `transb==true`, and A and B are swapped in the function call itself.

My best understanding regarding _why_ the swap needs to be performed is that it matches the BLAS semantics regarding column-major storage (see e.g. https://rocm.docs.amd.com/projects/rocBLAS/en/latest/conceptual/rocblas-design-notes.html#column-major-storage-and-1-based-indexing).

transformer_engine/common/gemm/ck_grouped_gemm.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+#include "ck_tile/core.hpp"
+#include "ck_tile/ops/epilogue.hpp"
+#include "ck_tile/ops/gemm.hpp"
+
+using RowMajor = ck_tile::tensor_layout::gemm::RowMajor;
+using ColMajor = ck_tile::tensor_layout::gemm::ColumnMajor;
+
+template <typename TeScalar> struct TeTypeToCkType;
+template <> struct TeTypeToCkType<transformer_engine::fp16> { using type = ck_tile::half_t; };
```
wangye805: We can add namespace transformer_engine and grouped_gemm like NV upstream https://github.com/ROCm/TransformerEngine/blob/179af4e428a7a7585145d851fc6405c43795663c/transformer_engine/common/gemm/cutlass_grouped_gemm.cu#L11-L12
Therefore, we can 1) better align with nv upstream in group gemm api call and 2) skip the transformer_engine:: when referring to bf16 and fp16

transformer_engine/common/gemm/ck_grouped_gemm.cpp
```diff
@@ -0,0 +1,276 @@
...
+#include "../common.h"
+
+#include "ck_tile/core.hpp"
+#include "ck_tile/ops/epilogue.hpp"
+#include "ck_tile/ops/gemm.hpp"
+
+using RowMajor = ck_tile::tensor_layout::gemm::RowMajor;
+using ColMajor = ck_tile::tensor_layout::gemm::ColumnMajor;
+
+template <typename TeScalar> struct TeTypeToCkType;
```
wangye805: Te --> TE and Ck --> CK?

transformer_engine/common/gemm/ck_grouped_gemm.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+// FIXME: This could be a templated lambda function in C++20.
+#define CALL(ALayout_, BLayout_, ta_, tb_)                                    \
+  return run_grouped_impl<T, ALayout_, BLayout_, CLayout, MemOp>(             \
+      A_use, B_use, D, group_num, (ta_), (tb_), workspace, workspace_bytes, stream)
+  
+  if (!transA_use && !transB_use) { CALL(RowMajor, RowMajor, false, false); }
+  if (!transA_use &&  transB_use) { CALL(RowMajor, ColMajor, false, true ); }
+  if ( transA_use && !transB_use) { CALL(ColMajor, RowMajor, true,  false); }
+  /* transA_use && transB_use */  { CALL(ColMajor, ColMajor, true,  true ); }
```
wangye805: NV upstream does not support TT, do we support TT?

transformer_engine/common/gemm/ck_grouped_gemm.cpp
```diff
@@ -0,0 +1,276 @@
...
+{
+  if (group_num <= 0)
+    return true;
+
+  // Convert A/B/D arrays into TE Tensor arrays
+  std::vector<const transformer_engine::Tensor*> A_te(group_num);
+  std::vector<const transformer_engine::Tensor*> B_te(group_num);
+  std::vector<transformer_engine::Tensor*>       D_te(group_num);
+
+  for (int i = 0; i < group_num; ++i) {
```
wangye805: Actually, you can save the convertNVTETensorCheck into dispatch_grouped. Then we don't need to allocate the vector for A_te, B_te, and D_te

transformer_engine/common/gemm/ck_grouped_gemm.cpp
```diff
@@ -0,0 +1,276 @@
...
+  
+  if (!transA_use && !transB_use) { CALL(RowMajor, RowMajor, false, false); }
+  if (!transA_use &&  transB_use) { CALL(RowMajor, ColMajor, false, true ); }
+  if ( transA_use && !transB_use) { CALL(ColMajor, RowMajor, true,  false); }
+  /* transA_use && transB_use */  { CALL(ColMajor, ColMajor, true,  true ); }
+
+#undef CALL
+}
+
+bool ck_tile_grouped_gemm(const NVTETensor* A,
```
wangye805, line 225: This function logic looks quite similar to NV upstream one: https://github.com/ROCm/TransformerEngine/blob/179af4e428a7a7585145d851fc6405c43795663c/transformer_engine/common/gemm/cutlass_grouped_gemm.cu#L42. 

Not sure whether we can consolidate them to further reduce our code diff with respect to upstream

wangye805 (changes requested):

transformer_engine/common/gemm/ck_grouped_gemm.cpp
ipanfilo: I didn't notice that A_use is actually B_te and vice versa. Please explicitly say that you're swapping A and B. Also does it produces matching result to hipBlasLt implementation with that swap?

transformer_engine/common/gemm/ck_grouped_gemm.cpp
```diff
@@ -0,0 +1,279 @@
...
+      s,
+      ck_tile::make_kernel<1>(
+          Kernel{}, grids, blocks, 0,
+          ck_tile::cast_pointer_to_constant_address_space(workspace),
+          group_num));
+  return true;
+}
+
+template <typename T, typename CLayout, ck_tile::memory_operation_enum MemOp>
+static inline bool dispatch_grouped(bool transA_use,
```
ipanfilo: Why separate function is needed?

transformer_engine/common/gemm/cublaslt_gemm.cu
ipanfilo: Yes, please

transformer_engine/common/CMakeLists.txt
ipanfilo, line 1: Copyright date

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -788,11 +791,12 @@ void nvte_multi_tensor_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor
...
 
 #ifdef __HIP_PLATFORM_AMD__
-    multi_stream_cublas_gemm(A, B, D, bias, pre_gelu_out, num_gemms, transa, transb, grad,
-                             workspace, accumulate, use_split_accumulator, math_sm_count, stream);
+  if (num_gemms <= 0)
+    return;
 #else
-  const int current_device = transformer_engine::cuda::current_device();
-  const bool is_hopper = (transformer_engine::cuda::sm_arch(current_device) == 90);
+    const int current_device = transformer_engine::cuda::current_device();
```
ipanfilo: Please restore indent

tests/pytorch/test_numerics.py
matthiasdiener: I wish I could answer this better, but I'm really not sure. I was able to tighten the tolerances in https://github.com/ROCm/TransformerEngine/pull/434/changes/0cbf1cddfc844cd7102c38e1256ffbca287807fc (perhaps this was possible due to a rocm or aiter version update?). Note that the tolerances in this test are much tighter than the ones in `test_grouped_gemm` below, for both MI300 and MI350.

tests/pytorch/test_numerics.py
matthiasdiener: I was able to remove these higher tolerances in https://github.com/ROCm/TransformerEngine/commit/0cbf1cddfc844cd7102c38e1256ffbca287807fc, they match upstream now for both MI300 and MI350.

transformer_engine/common/gemm/cublaslt_gemm.cu
matthiasdiener: Done in https://github.com/ROCm/TransformerEngine/pull/434/commits/98e0c66844bbd4a7c936e2e3831910ca15d09358

transformer_engine/common/CMakeLists.txt
matthiasdiener: Done in https://github.com/ROCm/TransformerEngine/pull/434/commits/98e0c66844bbd4a7c936e2e3831910ca15d09358

transformer_engine/common/gemm/ck_grouped_gemm.cpp
matthiasdiener: Done in https://github.com/ROCm/TransformerEngine/pull/434/commits/98e0c66844bbd4a7c936e2e3831910ca15d09358

transformer_engine/common/gemm/ck_grouped_gemm.cpp
matthiasdiener: I added the namespaces in https://github.com/ROCm/TransformerEngine/pull/434/commits/98e0c66844bbd4a7c936e2e3831910ca15d09358. Note that the actual front-end APIs (i.e., `ck_tile_grouped_gemm` , `cutlass_grouped_gemm`) are outside the namespaces.

transformer_engine/common/gemm/ck_grouped_gemm.cpp
matthiasdiener: > I didn't notice that A_use is actually B_te and vice versa. Please explicitly say that you're swapping A and B. 

I added a mention of this swap in 98e0c668.

> Also does it produces matching result to hipBlasLt implementation with that swap?

Yes, results match.

transformer_engine/common/gemm/cublaslt_gemm.cu
matthiasdiener: ~Commented~ #ifdefd out in 98e0c668

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -875,8 +877,12 @@ void nvte_multi_tensor_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor
   if (is_empty_arr(bias) && is_empty_arr(pre_gelu_out) && is_supported_dtype() &&
 #ifdef __HIP_PLATFORM_AMD__
       true)                               {
-    if (!ck_tile_grouped_gemm(A, B, D, num_gemms, transa, transb, workspace, accumulate, stream))
+    if (!ck_tile_grouped_gemm(A, B, D, num_gemms, transa, transb, workspace, accumulate, stream)) {
+        if (warn_fallback) {
+          NVTE_WARN("Fallback to cuBLAS grouped GEMM.");
```
ipanfilo, line 882: hipBlasLt?

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -794,8 +794,8 @@ void nvte_multi_tensor_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor
   if (num_gemms <= 0)
     return;
 #else
-    const int current_device = transformer_engine::cuda::current_device();
-    const bool is_hopper = (transformer_engine::cuda::sm_arch(current_device) == 90);
+  const int current_device = transformer_engine::cuda::current_device();
+  const bool is_hopper = (transformer_engine::cuda::sm_arch(current_device) == 90);
 #endif
   const bool use_cutlass = transformer_engine::getenv<bool>("NVTE_USE_CUTLASS_GROUPED_GEMM", false);
```
ipanfilo, line 800: I wonder, should we use different env name on ROCm? Or it should be well documented - what does CUTLASS mean on ROCm

transformer_engine/common/gemm/ck_grouped_gemm.cpp
matthiasdiener: Not strictly needed, but merging `dispatch_grouped` and `run_grouped_impl`  makes the resulting function very complex, and this complexity will likely increase when we add FP8 support. 

Here is how it would look like:

<details>

```c++
template <typename T, typename CLayout, ck_tile::memory_operation_enum MemOp>
static inline bool dispatch_grouped(bool transA_use,
                                    bool transB_use,
                                    const transformer_engine::Tensor* const* A_use,
                                    const transformer_engine::Tensor* const* B_use,
                                    transformer_engine::Tensor* const* D,
                                    int group_num,
                                    void* workspace,
                                    size_t workspace_bytes,
                                    hipStream_t stream) {

  int64_t ref_d0 = 0, ref_d1 = 0;
  if (!get_flat_2d_dims(*D[0], ref_d0, ref_d1)) {
    NVTE_ERROR("ck_tile_grouped_gemm: expected rank>=2 for D[0]");
    return false;
  }
  const ck_tile::index_t N = static_cast<ck_tile::index_t>(ref_d1);

  auto run_with_tilecfg = [&](auto tile_tag) -> bool {
    using TileCfgSel = decltype(tile_tag);
    TRANSFORMER_ENGINE_SWITCH_CONDITION(transA_use, kTransA, {
      using ALayout = std::conditional_t<kTransA, ColMajor, RowMajor>;

      TRANSFORMER_ENGINE_SWITCH_CONDITION(transB_use, kTransB, {
        using BLayout = std::conditional_t<kTransB, ColMajor, RowMajor>;

        using Kernel = typename Runner<T, T, T, ALayout, BLayout, CLayout, TileCfgSel, MemOp>::Kernel;

        const size_t needed = Kernel::GetWorkSpaceSize(group_num);
        if (!workspace || workspace_bytes < needed) {
          NVTE_ERROR("ck_tile_grouped_gemm: insufficient workspace. Needed bytes=", needed);
          return false;
        }

        thread_local std::vector<ck_tile::GroupedGemmHostArgs<0>> descs;
        descs.clear();
        descs.reserve(group_num);

        for (int i = 0; i < group_num; ++i) {
          const auto& a = data_view(*A_use[i]);
          const auto& b = data_view(*B_use[i]);
          const auto& d = data_view(*D[i]);

          int64_t Ad0 = 0, Ad1 = 0, Bd0 = 0, Bd1 = 0, Dd0 = 0, Dd1 = 0;
          if (!get_flat_2d_dims(*A_use[i], Ad0, Ad1) ||
              !get_flat_2d_dims(*B_use[i], Bd0, Bd1) ||
              !get_flat_2d_dims(*D[i],     Dd0, Dd1)) {
            NVTE_ERROR("ck_tile_grouped_gemm: expected all groups to be rank>=2 (2D or higher).");
            return false;
          }

          const int64_t M  = transA_use ? Ad1 : Ad0;
          const int64_t K  = transA_use ? Ad0 : Ad1;
          const int64_t N  = transB_use ? Bd0 : Bd1;
          const int64_t Kb = transB_use ? Bd1 : Bd0;

          if (Kb != K) {
            NVTE_ERROR("ck_tile_grouped_gemm: K mismatch between A and B in group ", i);
            return false;
          }

          if (Dd0 != M || Dd1 != N) {
            NVTE_ERROR("ck_tile_grouped_gemm: D shape mismatch in group ", i);
            return false;
          }

          // Leading dimensions under the flattened-contiguous interpretation
          const ck_tile::index_t stride_A = Ad1;
          const ck_tile::index_t stride_B = Bd1;
          const ck_tile::index_t stride_E = Dd1;

          descs.emplace_back(
              a.dptr,
              b.dptr,
              std::array<const void*, 0>{},
              d.dptr,
              1,
              M,
              N,
              K,
              stride_A,
              stride_B,
              std::array<ck_tile::index_t, 0>{},
              stride_E);
        }

        const dim3 grids = Kernel::GridSize(descs);
        auto kargs = Kernel::MakeKargs(descs);
        if (!Kernel::IsSupportedArgument(kargs)) {
          NVTE_ERROR("ck_tile_grouped_gemm: CK_Tile kernel arguments not supported for this config.");
          return false;
        }

        HIP_CHECK_ERROR(hipMemcpyAsync(workspace,
                                      kargs.data(),
                                      kargs.size() * sizeof(typename decltype(kargs)::value_type),
                                      hipMemcpyHostToDevice,
                                      stream));

        const ck_tile::stream_config s{stream};
        const dim3 blocks = Kernel::BlockSize();

        ck_tile::launch_kernel(
            s,
            ck_tile::make_kernel<1>(
                Kernel{}, grids, blocks, 0,
                ck_tile::cast_pointer_to_constant_address_space(workspace),
                group_num));
        return true;
      });
    });
  };

  // Select tile config like Primus-Turbo for FP16/BF16:
  //   N%256 -> 256x256x64
  //   N%128 -> 256x128x64
  //   else  -> 256x128x64 padding
  // NOTE: We assume N is uniform across groups.
  if ((N % 256) == 0) {
    return run_with_tilecfg(TileCfg_256x256x64{});
  } else if ((N % 128) == 0) {
    return run_with_tilecfg(TileCfg_256x128x64{});
  } else {
    return run_with_tilecfg(TileCfg_256x128x64_padding{});
  }
}
```

</details>

Which one do you prefer?

transformer_engine/common/gemm/ck_grouped_gemm.cpp
matthiasdiener: Good point, thanks. What do you think of 36bd68e9?

tests/pytorch/test_numerics.py
wangye805: Emm, I see. Test grouped gemm assumes exact bitwise match btw the reference (cublaslt gemm) and multistream grouped gemm and much larger tolerance btw cublaslt gemm and cutlass gemm. Unless we have shape difference between this test_grouped_linear_accuracy and test_grouped_gemm, I think it's okay for us to relax the tolerence

transformer_engine/common/gemm/ck_grouped_gemm.cpp
wangye805: Looks good. Thanks

transformer_engine/common/gemm/ck_grouped_gemm.cpp
wangye805: I guess it's okay to not consolidate. So I'm resolving this conversation

transformer_engine/common/gemm/ck_grouped_gemm.cpp
matthiasdiener: I found a way to merge `dispatch_grouped`  and `ck_tile_grouped_gemm` instead. Implemented in https://github.com/ROCm/TransformerEngine/pull/434/commits/c5d83a426d49c448aabc26def95b5cb49c2e172d

transformer_engine/common/gemm/cublaslt_gemm.cu
wangye805: Previously Matthias has another env. I left the comment to suggest use the same env as NV upstream since I recall CK is meant to be a drop-in replacement for cutlass?

Maybe we can explain this in README?

transformer_engine/common/gemm/ck_grouped_gemm.cpp
matthiasdiener: I misread your initial comment, https://github.com/ROCm/TransformerEngine/pull/434/commits/c5d83a426d49c448aabc26def95b5cb49c2e172d merges `dispatch_grouped` and `ck_tile_grouped_gemm`.

wangye805 (approved):

transformer_engine/common/gemm/cublaslt_gemm.cu
matthiasdiener: I don't have a strong opinion either way; my only concern would be that it may look strange if we had warnings that say 'Fallback to cublas...' (from the upstream code below this one) in some cases, and 'Fallback to hipblasLt...'  in other cases.

transformer_engine/common/gemm/ck_grouped_gemm.cpp
matthiasdiener: We do, yes.

transformer_engine/common/gemm/cublaslt_gemm.cu
matthiasdiener: I added a paragraph in the README in https://github.com/ROCm/TransformerEngine/commit/7b1dbfa2d4614a168c8d9847bb3bbf64a6a3280c, what do you think?

matthiasdiener:
Level 3 run: https://github.com/ROCm/TransformerEngine/actions/runs/22582355671

Edit: The run passed.

ipanfilo (approved):

PR #438, opened 2026-01-30
[CI] Use default checkout depth

matthiasdiener:
# Description

This is intended to reduce the time spent on the `git checkout`  step in CI.
Not sure if we need the full checkout depth. Tests seem to pass with the default depth (=1) + tags.

Timings (min:sec) for checkout (not necessarily representative, timings vary quite a bit):

| CI run                                                                 | Checkout time on MI325 | Checkout time on MI355 | Data transferred | Disk size |
|-------------------------------------------------------------------------|-------|-------|-----|-----|
| [This PR](https://github.com/ROCm/TransformerEngine/actions/runs/21531700231)                                                                 | 3:03  | 0:53  | 370 MByte | 880 MByte |
| [CI run w/o this PR](https://github.com/ROCm/TransformerEngine/actions/runs/21530142179/) | 5:55  | 13:36 | 1540 MByte | 2100 MByte| 

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

## Changes

- Do not set checkout depth, ~but require git tags~ 
- Update checkout version to v6

# Checklist:

### DIFF
--- a/.github/workflows/rocm-ci.yml
+++ b/.github/workflows/rocm-ci.yml
@@ -49,10 +49,10 @@ jobs:
         runner: [linux-mi325-8, linux-mi355-8]
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
           submodules: 'recursive'
-          fetch-depth: 0
+          fetch-tags: 'true'
         
       - name: Host Diagnostics & Environment Setup
         id: host-setup
@@ -78,7 +78,7 @@ jobs:
           # Default to input (or '1' if input is missing/null)
           CALC_LEVEL="${{ inputs.test_level || '1' }}"
 
-          # COnly force Level 3 if this is a direct PUSH to dev or a release branch
+          # Only force Level 3 if this is a direct PUSH to dev or a release branch
           if [[ "${{ github.event_name }}" == "push" ]]; then
              if [[ "${{ github.ref_name }}" == "dev" || "${{ github.ref_name }}" =~ ^release_v.*_rocm$ ]]; then
                 echo "::notice::Push to monitored branch (${{ github.ref_name }}) detected. Forcing Level 3."
--- a/.github/workflows/rocm-ci.yml
+++ b/.github/workflows/rocm-ci.yml
@@ -21,10 +21,6 @@ on:
         description: 'Test Level (1-3)'
         required: true
         default: '1'
-      skip_dev_merge:
-        description: 'Skip merging dev branch'
-        type: boolean
-        default: false
       docker_image_override:
         description: 'Manual Docker Image (Leave empty to use config file value)'
         required: false
--- a/.github/workflows/rocm-ci.yml
+++ b/.github/workflows/rocm-ci.yml
@@ -52,7 +52,6 @@ jobs:
         uses: actions/checkout@v6
         with:
           submodules: 'recursive'
-          fetch-tags: 'true'
         
       - name: Host Diagnostics & Environment Setup
         id: host-setup


.github/workflows/rocm-ci.yml
```diff
@@ -49,10 +49,10 @@ jobs:
         runner: [linux-mi325-8, linux-mi355-8]
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
           submodules: 'recursive'
-          fetch-depth: 0
+          fetch-tags: 'true'
```
wenchenvincent: I think removal of fetch-depth of all levels should be fine and `fetch-tags` might not be necessary. I will let @ipanfilo and @leo-amd  comment

ipanfilo: What is desired result of removing fetch-depth:0?

matthiasdiener: > What is desired result of removing fetch-depth:0?

I noticed that the `checkout` stage in CI runs takes a considerable amount of time (highest I've seen is 36+ minutes, e.g. in https://github.com/ROCm/TransformerEngine/actions/runs/21598826002/job/62238786074). Not fetching the full git history (and only fetching the GitHub Actions default, i.e. the latest commit in each repo) helps reduce that (see e.g. the numbers in the initial message in this PR).

ipanfilo: My concern is that to run CI on a PR, GHA makes merge commit to the destination branch. I think to make proper merge commit git should have history at least to fork point. How does checking out w/o history work in this case?

matthiasdiener: > My concern is that to run CI on a PR, GHA makes merge commit to the destination branch. I think to make proper merge commit git should have history at least to fork point. How does checking out w/o history work in this case?

Can you clarify, is there some custom merge commit creation in our CI, or is it the default GHA behavior? In the latter case, GitHub creates the merge commit internally, before the checkout action runs, so the fetch-depth setting does not affect the merge.

matthiasdiener: [1f93d39](https://github.com/ROCm/TransformerEngine/pull/438/commits/1f93d39a4c409d597bdd4194e008524d8bcc7ac6) removes the `fetch-tags`, and CI seems to work fine.

ipanfilo: There is still [skip_]dev_merge option in manual run but it seems currently unused. So having no depth should be OK.
Can skip_dev_merge be removed to make sure we don't start using it

matthiasdiener: I removed the skip_dev_merge  option in https://github.com/ROCm/TransformerEngine/pull/438/commits/01fa0d5d4e9f84e639e4ff0b8ee2da0a325f560b.

wenchenvincent: @ipanfilo Is there a problem with skip_dev_merge with having no depth? If we only want to run a branch, does github action just checkout the branch with no depth? If so, then it seems Okay.

ipanfilo: skip_dev_merge was not actually used, it was an artifact from an original implementation that did merge in script. I asked to remove it not to accidentally start using it. I presume that our implementation of merging might have problems with no-depth

ipanfilo (approved):

.github/workflows/rocm-ci.yml
wenchenvincent: Also, it seems that `skip_dev_merge` is not really used?

wenchenvincent: Another scenario where no-depth could be a problems is that if the project uses setuptools-scm or a similar mechanism that derives the package version from git tags. But in our case, TE hardcodes the versions so it is fine.

wenchenvincent (approved):

PR #442, opened 2026-02-04
 Remove padding from scales for hipBLASlt calls

alextmagro:
Removes padding for scale vectors that are used mainly for MXFP8.


tests/cpp/operator/test_cublaslt_gemm.cu
```diff
@@ -345,8 +347,11 @@ void performTest(const TestParams& params) {
     if (!has_fp8) {
       GTEST_SKIP() << "MXFP8 scaling mode requires Float8 types";
     }
-    if (params.m % 32 != 0 || params.n % 32 != 0 || params.k % 32 != 0) {
-      GTEST_SKIP() << "MXFP8 requires M, N, K to be multiples of 32";
+    if (params.m % 16 || params.n % 16) {
+      GTEST_SKIP() << "MXFP8 requires M & N to be multiples of 16";
+    }
+    if (params.k % 128) {
```
ipanfilo, line 352: Is it hipblasLt limitation?

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -189,6 +189,7 @@ TensorWrapper CommOverlapCore::get_tensor_chunk(const TensorWrapper &source, siz
   NVTE_DIM_CHECK(chunk_height > 0 && chunk_width > 0, "Attempted to get empty tensor chunk");
   NVTE_DIM_CHECK(chunk_height <= height && chunk_width <= width,
                  "Attempted to get out-of-bounds tensor chunk");
+#ifndef __HIP_PLATFORM_AMD__
```
ipanfilo: I think this file is not currently compiled for ROCm - it is for UB

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -197,9 +197,59 @@ struct GemmParam {
...
+    float *scale_dptr = (float*)(scale_tensor.dptr);
+
+    const size_t total_elements = unpadded_rows * unpadded_cols;
+    const size_t threads        = 256;
+    const size_t blocks         = (total_elements + threads - 1) / threads;
+
+    unpad_mxfp8_scales_kernel<<<blocks, threads, 0, stream>>>
+                (scale_dptr, unpadded_rows, unpadded_cols, padded_cols);
+
+    NVTE_CHECK_CUDA(hipStreamSynchronize(stream));
```
ipanfilo: If we're using the same stream for GEMM, synchronize creates unnecessary bubble

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -235,6 +285,9 @@ GemmParam CanonicalizeGemmInput(const transformer_engine::Tensor &A, const cubla
     // MXFP8
     // Note: Row-wise and column-wise data are scaled along different
     // dimensions (with matrix interpreted in row-major order).
+
+    unpad_mxfp8_checkpoint(A, is_A_transposed, m, n, k, stream);
+
```
ipanfilo: It is probably not the best place for this call - it does not update input parameters.
Also, what will be the code behaviour if we call GEMM twice with the same input - the first call will updad scale tensor and the second will do the same, right?

tests/cpp/operator/test_cublaslt_gemm.cu
alextmagro: Yes, these are the values that hipblastlt team provided to us. I tested just in case, but nothing smaller that 128 works for k.

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
alextmagro: Yes, I can move it to the UB PR if you prefer?

transformer_engine/common/gemm/rocm_gemm.cu
alextmagro: We talked about this during standup and the issues it would create in Python, I am currently working on moving the logic over there.

alextmagro: I have moved the check and unpadding to generic_gemm in pytorch. This provides a more robust way to check before calling the gemms, and directly modifies the pytorch tensors.

tests/cpp/operator/test_cublaslt_gemm.cu
ipanfilo: Is 32x128x32 config needed with 16x128x16 then?

tests/cpp/test_common.cu
```diff
@@ -147,7 +147,11 @@ std::pair<scale_inv_meta, scale_inv_meta> get_scales(const NVTEShape& shape,
 
     scale_inv_meta ret_rowwise, ret_colwise;
 
+#ifndef __HIP_PLATFORM_AMD__
```
ipanfilo: Nit: swap to #ifdef to match other changes in the file

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
ipanfilo: Yes, better move to UB because this file wold require more changes than those ifdefs, right?

tests/pytorch/test_sanity.py
```diff
@@ -913,6 +913,78 @@ def test_sanity_fp8_gemm_with_unalignment(N, datatype):
     )
     torch.cuda.synchronize()
 
+@pytest.mark.skipif(not mxfp8_available, reason=reason_for_no_mxfp8)
+@pytest.mark.parametrize("N", [32])
+@pytest.mark.parametrize("K", [128])
+@pytest.mark.parametrize("M", [32])
```
ipanfilo: Better use non multiple of 32 to test this path is unpadding

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -30,6 +36,42 @@ def validate_gemm_scale(scale: Optional[float], required: bool) -> float:
     return 0.0
 
 
+def unpad_scales(tensor: torch.Tensor, transpose: bool) -> torch.Tensor:
```
ipanfilo: I wonder if it can be called once when tensors are created?

alextmagro: When a user creates a tensor for the first time, we don't do padding to begin with -- This logic is for loading NV checkpoints only. I was thinking that when we load a pytorch checkpoint, the tensors are filled without calling the initializers, potentially missing the logic. Is there a way to guarantee the unpadding function is called when a Tensor is loaded?

tests/pytorch/test_sanity.py
alextmagro: We require block sizes of 32 at the python level, so not possible to do a non-multiple. We are padding scales, so we will see a rowwise scale of (1,4) padded to (128,4), and a colwise scale of (4,1) being padded to (4,128).

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
alextmagro: Yes, makes sense.

tests/cpp/operator/test_cublaslt_gemm.cu
alextmagro: I would say it makes sense to keep. This allows us to test a TE acceptable size with 32 while also ensuring unpadding and hipBLASlt is working with 16.

ipanfilo: In this case I'd change 32x128x32 to 32x128x16 to test they work together

transformer_engine/pytorch/cpp_extensions/gemm.py
ipanfilo: I think __init__ should be called when checkpointing is loading. Or loading can be intercepted be overriding load_from_state_dict

ipanfilo (approved):

tests/pytorch/references/blockwise_quantizer_reference.py
```diff
@@ -1,3 +1,5 @@
+# This file was modified for portability to AMDGPU
+# Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.
```
wangye805, line 2: This is for blockwise fp8 recipe which we does not support yet

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
```diff
@@ -1,3 +1,5 @@
+# This file was modified for portability to AMDGPU
+# Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.
```
wangye805, line 2: I don't think we support this blockwise fp8 yet

PR #446, opened 2026-02-09
Update AITER subcommit and refactor internal AITER/CK FA API usage

Micky774:
# Description

Updates AITER subcommit as well as refactors our internal usage of the API for greater clarity and explicitness. TODO: Update AITER commit upon merger of https://github.com/ROCm/aiter/pull/2055

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Updates AITER subcommit
- Refactors to central implementation functions for batch vs group mode fwd/bwd functions
- Refactors to explicit instantiation of arguments (rather than relying on positional)
- Includes new AITER mask type conversion as part of necessary arguments for the new API
- Streamlines logging
- Adds helper python script to explicitly compare our application of the AITER API vs the AITER API's declaration through its header files.
- Uses helper script during install as sanity-check on appropriate usage of AITER API
- WIP pending tests

# Checklist:


transformer_engine/__init__.py
```diff
@@ -83,4 +83,11 @@
                 category=RuntimeWarning,
             )
 
+# Set AITER_ASM_DIR to point to the asm dir in the installed package
+local_asm_dir = os.path.join(
+    os.path.dirname(os.path.abspath(__file__)),
+    "..", "3rdparty", "aiter", "hsa",
+)
+if "AITER_ASM_DIR" not in os.environ:
+    os.environ["AITER_ASM_DIR"] = local_asm_dir
```
Micky774: This is temporary and needs to be replaced -- pending discussion

tools/check_aiter_mha_args_usage.py
Micky774, line 1: This helper script scans semi-hard-coded files wrt TE source-code in order to directly compare AITER's internal API and our attempt at utilizing it. This script is run during setup through `setup.py`

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -16,6 +16,17 @@
...
+int ck_to_aiter_mask_type(mask_enum mask_type, ck_tile::index_t left, ck_tile::index_t right){
+  if(
+    mask_type == mask_enum::no_mask ||
+    mask_type == mask_enum::window_generic
+  ) return 0;
+  if(left == -1 && right == 0){
+    return mask_type == mask_enum::mask_top_left ? 1 : 2;
+  }
+  return 3;
+}
```
Micky774: This is based on their `op_tests/cpp/mha/*` which isn't the most stable or reliable source, but it's the only concrete location where such a mapping is used or documented.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -333,125 +344,118 @@ __global__ void dbias_reduce_b1ss(
...
-    std::cout<<"dropout_seed_ptr: "<<std::get<0>(std::get<std::pair<const void*, const void*>>(fmha_args.drop_seed_offset))<<std::endl;
-    std::cout<<"dropout_offset_ptr: "<<std::get<1>(std::get<std::pair<const void*, const void*>>(fmha_args.drop_seed_offset))<<std::endl;
+void log_bwd_config(const char* func_name, const aiter::mha_bwd_args& fmha_args, bool ck_log_config){
+  if (!ck_log_config) {
+    return;
   }
 
+  auto log_value = [](const char* label, const auto& value) {
+    std::cout << label << ": " << value << "\n";
+  };
```
Micky774: This merely standardizes the logging to make it a bit easier to parse through it at a glance while guaranteeing uniformity. This is similarly implemented across both files. Note the signature has been reduced to account for only what isn't stored in `fmha_args`.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -499,10 +508,13 @@ hipError_t ck_attn_bwd(
   bool uses_bwd_v3,
   bool is_v3_atomic_fp32,
   int how_v3_bf16_cvt,
+  bool is_group_mode,
+  const char* func_name,
+  bool ck_log_config,
   hipStream_t stream){
```
Micky774, line 491: This abstracted implementation function has a signature that is a superset of the other higher-level API functions in the file, so that both can route directly to this function without affecting the API outside of this file.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -518,196 +530,276 @@ hipError_t ck_attn_bwd(
...
-                         split_stride_dq_acc,
-                         left,
-                         right,
-                         static_cast<ck_tile::index_t>(mask_type),
-                         p_drop,
-                         p_undrop,
-                         std::pair<const void*, const void*>{philox_seed_ptr, philox_offset_ptr}};
-  }();
+  aiter::mha_bwd_args fmha_args{};
+  fmha_args.mask_type = ck_to_aiter_mask_type(mask_type, left, right);
```
Micky774: This is the **AITER** mask type, despite the same argument referring to the CK mask type in the FWD pass

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -518,196 +530,276 @@ hipError_t ck_attn_bwd(
...
   if(dump_path){
     dump_bwd_timings(dump_path, average_runtime);
   }
   if(average_runtime < 0){
     //TODO: better error out system
     throw std::runtime_error("fused attn configs not supported in ck_fused_attn bwd pass.");
   }
+  return hipSuccess;
+}
+hipError_t ck_attn_bwd(
```
Micky774, line 674: Both `ck_attn_{varlen_}bwd` are wrappers around `_ck_attn_bwd_impl` + TE-side post-processing kernels. Those kernels can probably be refactored similarly but that's a bit outside the scope here and of dubious value.

Micky774 (commented):
A few comments to hopefully help reviewers.

setup.py
```diff
@@ -56,6 +59,25 @@ def run(self):
...
+            # by creating a symlink to them.
+            project_dir = Path(__file__).parent
+            asm_src_dir = project_dir / '3rdparty' / 'aiter' / 'hsa'
+            # Must be synced with
+            # TransformerEngine/transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
+            asm_target_dir = project_dir / 'transformer_engine' / 'lib' / 'aiter'
+            if asm_src_dir.is_dir() and not asm_target_dir.is_dir():
+                print(f"Setting up symlink for AITER ASM kernels: {asm_target_dir} -> {asm_src_dir}")
+                asm_target_dir.symlink_to(asm_src_dir)
+
```
Micky774: This is necessary to not break AITER ASM on editable installs -- we ran into this issue before.

setup.py
```diff
@@ -88,6 +110,14 @@ def setup_common_extension() -> CMakeExtension:
...
         elif os.getenv("NVTE_FUSED_ATTN_CK") or os.getenv("NVTE_FUSED_ATTN"):
             cmake_flags.append("-DUSE_FUSED_ATTN_CK=ON")
+            try:
+                subprocess.run(
+                    sys.executable + " tools/check_aiter_mha_args_usage.py --mode both",
+                    shell=True, check=True
+                )
+            except subprocess.CalledProcessError:
+                print("Error checking AITER mha_args usage.")
+                sys.exit(1)
```
Micky774: Explicitly checks the AITER API usage

setup.py
```diff
@@ -241,7 +271,7 @@ def setup_requirements() -> Tuple[List[str], List[str]]:
         long_description=long_description,
         long_description_content_type="text/x-rst",
         ext_modules=ext_modules,
-        cmdclass={"egg_info": HipifyMeta, "build_ext": CMakeBuildExtension, "bdist_wheel": TimedBdist},
+        cmdclass=cmdclass,
```
Micky774: There is no change when `NVTE_BUILD_METAPACKAGE=1` -- this only affects the `NVTE_BUILD_METAPACKAGE=0` case.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: Note that there is a PR for [removing this argument](https://github.com/ROCm/aiter/pull/2026) which is still pending.

tools/check_aiter_mha_args_usage.py
```diff
@@ -0,0 +1,97 @@
+import argparse
```
wangye805, line 12: Consider adding this script running into our CI?

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -333,125 +333,117 @@ __global__ void dbias_reduce_b1ss(
...
+    "dropout_seed_ptr",
+    std::get<0>(std::get<std::pair<const void*, const void*>>(fmha_args.drop_seed_offset))
+  );
+  log_value(
+    "dropout_offset_ptr",
+    std::get<1>(std::get<std::pair<const void*, const void*>>(fmha_args.drop_seed_offset))
+  );
 }
 
 void dump_bwd_timings(const char* dump_path, float average_runtime){
```
wangye805, line 438: Does this aiter commit update break your previous jax benchmarking functionalities?

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -518,196 +518,275 @@ hipError_t ck_attn_bwd(
...
+                                  : 0;
+  fmha_args.nhead_stride_o = stride_h_o;
+  fmha_args.nhead_stride_randval = is_group_mode ? 0 : seqlen_q * max_seqlen_k;
+  fmha_args.nhead_stride_do = stride_h_do;
+  fmha_args.nhead_stride_lsed = is_group_mode ? max_tokens_q : max_seqlen_q;
+  fmha_args.nhead_stride_dq_acc = static_cast<int64_t>((is_group_mode ? max_tokens_q : s_q) * d_qk);
+  fmha_args.nhead_stride_dq = stride_h_dq;
+  fmha_args.nhead_stride_dk = is_mqa_gqa? stride_h_dk_expanded:stride_h_dk;
+  fmha_args.nhead_stride_dv = is_mqa_gqa? stride_h_dv_expanded:stride_h_dv;
+  fmha_args.nhead_stride_dbias = (!is_group_mode) ? max_seqlen_q * max_seqlen_k : 0;
```
wangye805: is_group_mode? 0: max_seqlen_q * max_seqlen_k?

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_fwd.cpp
```diff
@@ -162,134 +150,137 @@ hipError_t ck_attn_fwd(
...
-                                         is_group_mode,
-                                         mask_type,
-                                         bias_type,
-                                         has_lse,
-                                         quant_scale_enum::no_scale,
-                                         uses_fwd_v3, 
-                                         false,//has_sink
-                                         how_v3_bf16_cvt);
+  // Handle bias
+  ck_tile::index_t _nhead_stride_bias = 0;
```
wangye805: Why do we need variable name start with _? Is it consistent with other parts in TE or transformer_engine/common/ck_fused_attn?

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -518,196 +518,275 @@ hipError_t ck_attn_bwd(
...
+
+  fmha_args.window_size_left = left;
+  fmha_args.window_size_right = right;
+  fmha_args.p_drop = p_drop;
+  fmha_args.p_undrop = p_undrop;
+  fmha_args.drop_seed_offset = std::pair<const void*, const void*>{philox_seed_ptr, philox_offset_ptr};
+
+  // modify the max_seqlen_q for better performance in 0-length cases
+  // lse_thd_ptr used as buffer
+  if(const char* env_p = std::getenv("NVTE_CK_RUNTIME_MAX_SEQLEN")) {
```
wangye805, line 649: Runtime max seqlen is only needed in varlen mode

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_fwd.cpp
```diff
@@ -318,174 +363,40 @@ hipError_t ck_attn_varlen_fwd(
...
-                         std::pair<const void*, const void*>{philox_seed_ptr, philox_offset_ptr}};
-  }();
-  // modify the max_seqlen_q for better performance in 0-length cases
-  // lse_thd_ptr used as buffer
-  if(const char* env_p = std::getenv("NVTE_CK_RUNTIME_MAX_SEQLEN")){
-    if(std::string(env_p) == "1"){
-      if(ck_fused_attn_log_config){
-        std::cout << "attn_fwd(ck): Enabling runtime max_seqlen calculation for small seqlen optimization.";
-      }
-      fmha_args.max_seqlen_q = get_runtime_max_seqlen(b, cu_seqlen_q_ptr, cu_seqlen_q_padded_ptr, lse_thd_ptr, stream);
```
wangye805, line 451: I didn't see runtime_max_seqlen_q in your new _impl fwd function.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
```diff
@@ -13,6 +19,32 @@
...
 namespace ck_fused_attn{
 
+void set_aiter_asm_dir() {
+  static std::once_flag aiter_asm_dir_once;
+  std::call_once(aiter_asm_dir_once, []() {
+    Dl_info info;
+    dladdr((void*)set_aiter_asm_dir, &info);
+    auto install_lib_path = std::filesystem::path(info.dli_fname).parent_path() / "aiter";
+    const char* log_ck_config = std::getenv("NVTE_LOG_CK_CONFIG");
+    auto editable_install_path = std::filesystem::path(info.dli_fname).parent_path().parent_path().parent_path() / "3rdparty" / "aiter" / "hsa";
```
wangye805: I recall we also copy hsa in editable installation

transformer_engine/common/ck_fused_attn/aiter_build.sh
```diff
@@ -37,7 +37,7 @@ if [[ -z "${AITER_DIR}" || -z "${AITER_TEST_DIR}" || -z "${GPU_ARCHS_VAL}" ]]; t
 fi
 
 rm -rf "${AITER_DIR}/aiter/jit/build"
-AITER_LOG_MORE=1 \
+AITER_LOG_MORE=0 \
```
wangye805: Is it possible to link pip install -v to AITER_LOG_MORE?

setup.py
wangye805: Actually you can put your comment in PR to the comments in the source codes

wangye805 (changes requested):

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_fwd.cpp
Micky774: It's meant to be a temp variable to help dynamically set the const variable in a readable way. I'm not sure if we have used this pattern elsewhere, but I don't think it _violates_ existing style.

tools/check_aiter_mha_args_usage.py
Micky774: It runs through `setup.py` on every build.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
Micky774: I don't think that we do

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: I don't think so, but I haven't explicitly checked yet

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: Done.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: Done.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_fwd.cpp
Micky774: Done.

transformer_engine/common/ck_fused_attn/aiter_build.sh
Micky774: Done.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_fwd.cpp
wangye805: I tried to search for this two variables starting with _ but didn't find it used anywhere else?

tools/check_aiter_mha_args_usage.py
wangye805: You can put this in the comment of this file. 

Also don't forget to add the copyright

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
wangye805: Same here. Put those helpful description into the comments

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
wangye805: Can you help try it out. Thanks

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -518,196 +518,300 @@ hipError_t ck_attn_bwd(
...
-                         static_cast<ck_tile::index_t>(mask_type),
-                         p_drop,
-                         p_undrop,
-                         std::pair<const void*, const void*>{philox_seed_ptr, philox_offset_ptr}};
+  bias_enum bias_type = bias_enum::no_bias;
+  BiasShape bias_shape = BiasShape::k11SS;
+  if (!is_group_mode) {
+    std::tie(bias_type, bias_shape) = get_ck_bias_type_shape(attn_bias_type, b, h, bias_b, bias_h);
+  }
+  const ck_tile::index_t batch_stride_bias = [&]() -> ck_tile::index_t {
```
wangye805: This overlaps with the lambda defined in fwd: https://github.com/ROCm/TransformerEngine/blob/29878cf0769f2b1eee11d7afcc59a43ed2598583/transformer_engine/common/ck_fused_attn/src/ck_fused_attn_fwd.cpp#L187-L217
Try to consolidate them into util.cpp? 

Also it's better to avoid using lambda with catching every reference. Previously I used it because I was copying from ck example.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_fwd.cpp
```diff
@@ -162,153 +150,246 @@ hipError_t ck_attn_fwd(
...
-                         std::pair<const void*, const void*>{philox_seed_ptr, philox_offset_ptr}};
+  // Handle bias
+  ck_tile::index_t _nhead_stride_bias = 0;
+  ck_tile::index_t _batch_stride_bias = 0;
+  const void* seqstart_q_ptr = nullptr;
+  const void* seqstart_k_ptr = nullptr;
+  bias_enum bias_type = bias_enum::no_bias;
+  BiasShape bias_shape = BiasShape::k11SS;
+  if(is_group_mode){
+    seqstart_q_ptr = cu_seqlen_q_padded_ptr==nullptr? cu_seqlen_q_ptr: cu_seqlen_q_padded_ptr;
```
wangye805: Make it consistent with bwd: https://github.com/ROCm/TransformerEngine/blob/29878cf0769f2b1eee11d7afcc59a43ed2598583/transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp#L606-L620

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -518,196 +518,300 @@ hipError_t ck_attn_bwd(
...
+  ck_tile::stream_config stream_config{stream, dump_path!=nullptr, ck_log_config};
 
   std::string data_type_str = get_data_type_str(dtype);
 
-  auto fmha_args = [&]() {
-    // setup stride_* arguments
-    const ck_tile::index_t stride_q = stride_s_q;
-    const ck_tile::index_t stride_k = stride_s_k;
-    const ck_tile::index_t stride_v = stride_s_v;
-    // bias of shape (bias_b, bias_h, s_q, s_kv)
```
wangye805, line 540: Can you keep this kind of comments so that later we can understand why we set stride_bias/dbias to max_seqlen_k

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -518,196 +518,300 @@ hipError_t ck_attn_bwd(
...
-    const ck_tile::index_t stride_randval = max_seqlen_k;
-    const ck_tile::index_t stride_do = stride_s_do;
-    const ck_tile::index_t stride_dq = stride_s_dq;
-    const ck_tile::index_t stride_dk = stride_s_dk;
-    const ck_tile::index_t stride_dv = stride_s_dv;
-    const ck_tile::index_t stride_dk_expanded = stride_s_dk_expanded;
-    const ck_tile::index_t stride_dv_expanded = stride_s_dv_expanded;
-    const ck_tile::index_t stride_dq_acc = d_qk; //dq_acc of shape (nsplits, B, H, S, D)
-    // dbias is of the same shape as bias
-    // but ck only take dbias with BHSS
```
wangye805, line 552: Also this comment is crucial for dbias reduction

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_fwd.cpp
```diff
@@ -162,153 +150,246 @@ hipError_t ck_attn_fwd(
...
+    dtype,
+    b, h, hg, s_q, s_kv, d_qk, d_v, bias_b, bias_h,
+    0,
+    q_ptr, stride_b_q, stride_h_q, stride_s_q,
+    k_ptr, stride_b_k, stride_h_k, stride_s_k,
+    v_ptr, stride_b_v, stride_h_v, stride_s_v,
+    bias_ptr,
+    alibi_slope_ptr,
+    nullptr, nullptr,
+    nullptr, nullptr,
```
wangye805: Add some comments so code readers can know quickly what's those 4 nullptrs are for?

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
wangye805: I don't have a dev TE sandbox at hand but here is one with v2.6 release editable installation: 

Image

wangye805 (changes requested):

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
Micky774: Maybe there's a difference in building from source vs downloading the pre-built AITER lib? My current editable install doesn't have the copied `aiter` directory under `transformer_engine/lib`.

tools/check_aiter_mha_args_usage.py
```diff
@@ -0,0 +1,105 @@
...
+    lines = text.splitlines()
+    buffer = [""]
+    for line in lines:
+        parse_with_skip_comments(buffer, line, assign_re, assignments)
+    return set(assignments)
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser(description="Check aiter args usage vs header definition")
+    parser.add_argument("--mode", choices=["fwd", "bwd", "both"], required=True, help="Mode: fwd, bwd, or both")
```
ipanfilo: Nit: should 'both' be default mode?

tools/check_aiter_mha_args_usage.py
```diff
@@ -0,0 +1,105 @@
...
+    args = parser.parse_args()
+    modes = ["fwd", "bwd"] if args.mode == "both" else [args.mode]
+    mismatch = 0
+    for mode in modes:
+        header_path = Path(f"3rdparty/aiter/csrc/include/mha_{mode}.h")
```
ipanfilo: It expects that the process CWD is TE root, which is not always true

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -518,196 +513,277 @@ hipError_t ck_attn_bwd(
...
   if(average_runtime < 0){
     //TODO: better error out system
     throw std::runtime_error("fused attn configs not supported in ck_fused_attn bwd pass.");
   }
+  if(dump_path){
+    dump_bwd_timings(dump_path, average_runtime);
+  }
+  return hipSuccess;
+}
+hipError_t ck_attn_bwd(
```
ipanfilo, line 674: Please add empty lines between methods

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -333,125 +332,113 @@ __global__ void dbias_reduce_b1ss(
...
-    std::cout<<"batch_stride_dbias: "<<fmha_args.batch_stride_dbias<<std::endl;
-    std::cout<<"window_size_left: "<<fmha_args.window_size_left<<std::endl;
-    std::cout<<"window_size_right: "<<fmha_args.window_size_right<<std::endl;
-    std::cout<<"mask_type: "<<fmha_args.mask_type<<std::endl;
-    std::cout<<"p_drop: "<<fmha_args.p_drop<<std::endl;
-    std::cout<<"p_undrop: "<<fmha_args.p_undrop<<std::endl;
-    std::cout<<"dropout_seed_ptr: "<<std::get<0>(std::get<std::pair<const void*, const void*>>(fmha_args.drop_seed_offset))<<std::endl;
-    std::cout<<"dropout_offset_ptr: "<<std::get<1>(std::get<std::pair<const void*, const void*>>(fmha_args.drop_seed_offset))<<std::endl;
+void log_bwd_config(const char* func_name, const aiter::mha_bwd_args& fmha_args, bool ck_log_config){
+  if (!ck_log_config) {
```
ipanfilo: May be not to call the function when ck_log_config=false, instead of passing the arg to it?

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_fwd.cpp
```diff
@@ -16,108 +16,87 @@
...
-    std::cout<<"min_seqlen_q: "<<fmha_args.min_seqlen_q<<std::endl;
-
-    std::cout<<"p_drop: "<<fmha_args.p_drop<<std::endl;
-    std::cout<<"s_randval: "<<fmha_args.s_randval<<std::endl;
-
-    std::cout<<"dropout_seed_ptr: "<<std::get<0>(std::get<std::pair<const void*, const void*>>(fmha_args.drop_seed_offset))<<std::endl;
-    std::cout<<"dropout_offset_ptr: "<<std::get<1>(std::get<std::pair<const void*, const void*>>(fmha_args.drop_seed_offset))<<std::endl;
+// print the fmha traits and fmha_args when calling ck apis
+void log_fwd_config(const char* func_name, bool has_dropout, const aiter::mha_fwd_args& fmha_args, bool ck_log_config){
+  if (!ck_log_config) {
```
ipanfilo: The same like in bwd: just not call the method if ck_log_config is false

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.hpp
```diff
@@ -56,5 +57,27 @@ std::pair<bias_enum, BiasShape> get_ck_bias_type_shape(BiasType attn_bias_type,
 
 uint64_t get_runtime_max_seqlen(uint64_t b, const void* cu_seqlen_ptr, const void* cu_seqlen_padded_ptr, void* workspace, hipStream_t stream);
 
+// This helper merely standardizes the logging to make it a bit easier to parse
+// through it at a glance while guaranteeing uniformity.
+template<typename T>
```
ipanfilo, line 61: I wonder about benefit of printing pointer values in CK parameters. Drawback is clear - it make harder to compare parameters between different runs because pointers may differe

.gitignore
```diff
@@ -55,3 +55,4 @@ artifacts/
 **/times.csv
 transformer_engine/build_info.txt
 transformer_engine/common/util/hip_nvml.*
+transformer_engine/lib/aiter
```
ipanfilo: transformer_engine/lib is already ignored

setup.py
```diff
@@ -88,6 +89,16 @@ def setup_common_extension() -> CMakeExtension:
             cmake_flags.append("-DUSE_FUSED_ATTN_CK=OFF")
         elif os.getenv("NVTE_FUSED_ATTN_CK") or os.getenv("NVTE_FUSED_ATTN"):
             cmake_flags.append("-DUSE_FUSED_ATTN_CK=ON")
+            # This helper script scans semi-hard-coded files wrt TE source-code in order to directly compare
+            # AITER's internal API and our attempt at utilizing it.
+            try:
+                subprocess.run(
```
ipanfilo: Since it is called from subprocess, can it be called from transformer_engine/common/cf_sued_attn/CMakeLisats.txt?

ipanfilo (commented):
Please updatre copyright date on modified files

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.hpp
Micky774: @wangye805 wondering what your thoughts are on this. So far, I've never really benefited from raw pointer prints during debug processes.

setup.py
Micky774: I think so, but it's just a bit easier to catch the appropriate error at the python level. If you think it's important enough, I can look into transfering it to the CMakeLists.txt -- I'm just not super familiar with cmake and don't know how to easily determine whether the script passed or failed without altering the stdout of the script.

tools/check_aiter_mha_args_usage.py
Micky774: Done

tools/check_aiter_mha_args_usage.py
Micky774: Updated

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: Done

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: Done

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_fwd.cpp
Micky774: Done

.gitignore
Micky774: Done

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
wangye805: Okay, I think now I understand the root cause. Your aiter commit update need to bring back the hsa dir copy which was removed in Dragan's PR: https://github.com/ROCm/TransformerEngine/pull/402/changes#diff-e0641ac80f912d730377ecbf45edc9573b4ad4a07ee875e474a49f3d57605f1f

Image

Otherwise, the non-editable installation will not have hsa dir in transformer_engine installation dir either

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.hpp
wangye805: @ipanfilo Yeah, it's hard to compare parameters btw runs but it will be helpful to compare pointers within the same run. For example, we hope the fwd/bwd share the same q/k/v/o/cu_seqlen. We can also use the raw pointers to calculate whether our workspace allocation is correct or not. Previously we had such bugs before

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: Tested and confirm it still works as expected

wangye805 (changes requested):

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
wangye805: With the hsa dir copy functionality brought back, we don't need to distinguish the install and editable install path, I believe?

setup.py
wangye805: In fact, if we think of the building libfuse_attn_ck.so as a standalone/self-contained component, we probably should put the checking script into transformer_engine/common/ck_fused_attn/CMakeLists.txt

wangye805 (changes requested):

setup.py
ipanfilo: execute_process in CMake has RESULT_VARIABLE argument to get subprocess exit code. So it is pretty easy to incorporate this to CMakeLists.txt

tools/check_aiter_mha_args_usage.py
```diff
@@ -0,0 +1,106 @@
+# Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.
+
```
ipanfilo, line 5: Please put proper copyright header with
`#`
`# See LICENSE for license information.`

setup.py
```diff
@@ -8,6 +8,7 @@
 
 from importlib import metadata
 import os
+import sys
```
ipanfilo: Likely not needed

transformer_engine/common/gemm/ck_grouped_gemm.cpp
```diff
@@ -110,7 +110,7 @@ struct Runner{
           Partitioner::MPerBlock, Partitioner::NPerBlock,
           TileCfg::M_Warp, TileCfg::N_Warp,
           TileCfg::M_Warp_Tile, TileCfg::N_Warp_Tile, TileCfg::K_Warp_Tile,
-          Problem::TransposeC, MemOp>>;
```
ipanfilo, line 113: Incorrect ident?

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -105,6 +105,27 @@ if(NOT EXISTS "${AITER_INCLUDE_DIR}")
           "within the Transformer Engine source.")
 endif()
 
+if(NOT Python_EXECUTABLE)
+  find_package(Python COMPONENTS Interpreter REQUIRED)
```
ipanfilo: Should it be mandatory or skip checking if python is not found?
Also, better move the check before AITER build to fail earlier

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -117,3 +138,13 @@ set_target_properties(ck_fused_attn PROPERTIES INSTALL_RPATH "$ORIGIN")
...
+# copy v3 kernels to destination
+foreach(ARCH IN LISTS V3_ASM_ARCHS)
+  foreach(KERNEL_TYPE IN ITEMS fmha_v3_fwd fmha_v3_bwd)
+    file(REMOVE_RECURSE ${CMAKE_INSTALL_PREFIX}/${AITER_MHA_INSTALL_PREFIX}/lib/aiter/${ARCH}/${KERNEL_TYPE})
+    install(DIRECTORY
+          ${__AITER_SOURCE_DIR}/hsa/${ARCH}/${KERNEL_TYPE}
+          DESTINATION ${CMAKE_INSTALL_PREFIX}/${AITER_MHA_INSTALL_PREFIX}/lib/aiter/${ARCH}/
+          PATTERN "codegen.py" EXCLUDE)
+  endforeach()
+endforeach()
```
ipanfilo: EOL

Micky774: Updated

transformer_engine/common/gemm/ck_grouped_gemm.cpp
Micky774: Updated

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -50,6 +50,31 @@ else()
     # If not downloaded, Fallback: Build from source
     if(NOT AITER_PREBUILT_DOWNLOAD_SUCCESS)
       message(STATUS " [AITER-PREBUILT] Building aiter from source.")
+
+      if(NOT Python_EXECUTABLE)
```
ipanfilo: As far as I understood this API validation is not to validate AITER itself but validate that TE calls AITER correctly. I.e. it should not be bound to AITER library rebuilding but also applying to download pre-built case.
The desire to move validation before AITER lib building/obtaining was to fail earlier which is especially important if not pre-built is available, not to couple validation with AITER lib re-complilation

Micky774:
level 3 run for tracking: https://github.com/ROCm/TransformerEngine/actions/runs/22781278349

transformer_engine/common/ck_fused_attn/CMakeLists.txt
Micky774: Right, good point, this should always be run. I've moved it up as the first part of the build.

transformer_engine/common/gemm/ck_grouped_gemm.cpp
```diff
@@ -110,7 +110,7 @@ struct Runner{
...
+        Problem::TransposeC>>;
```
matthiasdiener: Just a note: MemOp is removed because of https://github.com/ROCm/composable_kernel/pull/3359.

wangye805 (approved):

Micky774:
Level 3 run for tracking: https://github.com/ROCm/TransformerEngine/actions/runs/23265047818 (updated for commit [`638c9e6`](https://github.com/ROCm/TransformerEngine/pull/446/commits/638c9e60786a1c5be58c577c71663b09327993ee)

Micky774:
Okay so after patching the race condition w/ Xinya's mutex solution, another bug was revealed through the JAX MGPU tests which is actually a hip launch error due to device ordinal mismatch. This is essentially because the kernel map is keyed by the kernel name only, NOT by device ID, meaning that multiple device threads will map different kernels to the same key, and when loading we end up with the wrong kernels on the wrong devices. I've added a patch to address this issue, with a corresponding AITER PR here: https://github.com/ROCm/aiter/pull/2401

wangye805:
> Okay so after patching the race condition w/ Xinya's mutex solution, another bug was revealed through the JAX MGPU tests which is actually a hip launch error due to device ordinal mismatch. This is essentially because the kernel map is keyed by the kernel name only, NOT by device ID, meaning that multiple device threads will map different kernels to the same key, and when loading we end up with the wrong kernels on the wrong devices. I've added a patch to address this issue, with a corresponding AITER PR here: [ROCm/aiter#2401](https://github.com/ROCm/aiter/pull/2401)

Okay, Xinya's solution may not be that good. 

I have a quick question: we need the (id, kernel name) -> kernel mapping, is it because in our jax mgpu pytest, there is one thread trying to launch the same kernel on two different devices? Or Just due to the inherent issue in Xinya's patch that other threads found the kernel in name->kernel mapping but not loaded into the correct device? A quick way to verify is to change the bwd map to just thread_local without the (device id, kernel name)->kernel map

Micky774:
Level 3 run: https://github.com/ROCm/TransformerEngine/actions/runs/23456502776

PR #453, opened 2026-02-19
Update ck_fused_attn logging to direct to thread-specific files

Micky774:
# Description

Fixes https://github.com/ROCm/frameworks-internal/issues/15080

The main caveat here is that the kernel-level logging provided by AITER/CK is still pointing to `std::cout` and we can't redirect it since redirection is process-wide, not thread-local.

Implements logging using cached `thread_local` streams to minimize cost, and guarantee thread safety.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
+++ b/transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
@@ -16,34 +16,6 @@
 
 namespace ck_fused_attn{
 
-// We want to cache and reuse the log stream so we use thread_local here.
-namespace {
-std::ostream* get_bwd_log_stream() {
-  thread_local std::ofstream log_file;
-  thread_local bool attempted = false;
-  thread_local bool opened = false;
-  thread_local bool requested = false;
-  thread_local std::string log_dir_str;
-  if (!attempted) {
-    attempted = true;
-    if (const char* env_p = std::getenv("CK_FUSED_ATTN_LOG_CONFIG")) {
-      log_dir_str = std::string(env_p);
-      requested = !log_dir_str.empty() && log_dir_str != "0";
-    }
-    if (requested) {
-      opened = open_ck_fused_attn_log_file(log_file, "ck_fused_attn_bwd", log_dir_str);
-    }
-  }
-  if (!requested) {
-    return nullptr;
-  }
-  if (!opened) {
-    return &std::cout;
-  }
-  return &log_file;
-}
-}  // namespace
-
 // TODO: unify with binary search in TE/common/fused_attn(rocm)/util
 // no device std::upper_bound
 // in an increasing array with given size len, search for the index that:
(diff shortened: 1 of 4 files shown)


wangye805 (approved):

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -531,7 +541,7 @@ hipError_t ck_attn_bwd(
   mask_enum mask_type = static_cast<mask_enum>(attn_mask_type);
   bool ck_fused_attn_log_config = false;
   if (const char* env_p = std::getenv("CK_FUSED_ATTN_LOG_CONFIG") ) {
```
ipanfilo: Calling open_ck_fused_attn_log_file() here and using log_file != nullptr will hell avoiding interpreting env variable here and will also enable logging only if ENV variable is set to non "0"

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.hpp
```diff
@@ -56,5 +56,7 @@ std::pair<bias_enum, BiasShape> get_ck_bias_type_shape(BiasType attn_bias_type,
 
 uint64_t get_runtime_max_seqlen(uint64_t b, const void* cu_seqlen_ptr, const void* cu_seqlen_padded_ptr, void* workspace, hipStream_t stream);
 
+bool open_ck_fused_attn_log_file(std::ofstream& log_file, const char* file_prefix);
```
ipanfilo: The result of the method is not used.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
```diff
@@ -13,6 +17,22 @@
 
 namespace ck_fused_attn{
 
+bool open_ck_fused_attn_log_file(std::ofstream& log_file, const char* file_prefix) {
+  const char* env_p = std::getenv("CK_FUSED_ATTN_LOG_CONFIG");
+  if (env_p == nullptr) {
+    return false;
+  }
+  const std::string log_dir_str(env_p);
+  if (log_dir_str.empty() || log_dir_str == "0") {
```
ipanfilo: Please leave option to fallback to stdout. Since "0" is used as DISABLE, "1" can be use for this.
Or if use any non empty string to ENABLE, "-" may indicate stdout

ipanfilo (changes requested):
Copyright update

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: Done in a variation

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.hpp
Micky774: Fixed

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
Micky774: Updated so that `std::cout` is used whenever the log-file creation fails, or when the env variable is set to `"1"` explicitly.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
```diff
@@ -17,20 +17,20 @@
...
     return false;
   }
   std::filesystem::path log_dir(log_dir_str);
   std::ostringstream filename;
   filename << file_prefix << "_" << getpid() << "_" << std::this_thread::get_id() << ".log";
   log_file.open(log_dir / filename.str(), std::ios_base::app);
-  return log_file.is_open();
+  if (!log_file.is_open()) {
+    std::cerr << "Failed to open log file: " << (log_dir / filename.str()) << "\n";
+    return false;
```
ipanfilo, line 28: So, if user specifies invalid log file it will be forced to stdout?

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_fwd.cpp
```diff
@@ -17,16 +17,28 @@
 namespace ck_fused_attn{
 
 namespace {
-std::ofstream* get_fwd_log_stream() {
+std::ostream* get_fwd_log_stream() {
   thread_local std::ofstream log_file;
   thread_local bool attempted = false;
+  thread_local bool opened = false;
```
ipanfilo: Why so many thread_local variables are needed? log_dir_str  is not used by further calls at all. requested duplicates log_file.is_open()

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
Micky774: Currently yes, implemented as such to address your earlier feedback of "Please leave option to fallback to stdout". Did you mean fallback in that sense, or did you mean just as another option that could be explicitly specified?

ipanfilo: Sorry for confusion. I did not mean fall back as fault tolerance option but as an option for user to switch to old behaviour. I.e. now they are able to write logs to per thread files but still can do it old way to stdout.
BTW, what is benefit of splitting FWD and BWD logs of the same thread? I'd say it is more nice to see them together

Micky774: Thanks for the clarification -- updated behavior to remove it as a fallback and keep it only as an additional option to explicitly enable. I've also merged the fwd/bwd logging. I initially thought it would be helpful to keep them separated, but I agree it probably does make more sense to merge them (while maintaining thread-local files).

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_fwd.cpp
Micky774: I've streamlined the implementation a bit now

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.hpp
```diff
@@ -1,14 +1,15 @@
...
  *
  * License for AMD contributions = MIT. See LICENSE for more information
  ************************************************************************/
 
 #ifndef CK_FUSED_ATTN_UTILS_H
 #define CK_FUSED_ATTN_UTILS_H
 
-#include<iostream>
-#include<cstdint>
+#include<cstdlib>
```
ipanfilo: Looks like only iostream is needed

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
```diff
@@ -13,6 +17,41 @@
...
+  thread_local std::ofstream log_file;
+  thread_local std::ostream* log_stream = nullptr;
+  thread_local bool initialized = false;
+  if (!initialized) {
+    initialized = true;
+    if (const char* env_p = std::getenv("CK_FUSED_ATTN_LOG_CONFIG")) {
+      std::string log_dir_str(env_p);
+      if (!log_dir_str.empty() && log_dir_str != "0") {
+        if (log_dir_str == "1") {
+          log_stream = static_cast<std::ostream*>(&std::cout);
```
ipanfilo: Static_cast should not be needed here and when assigning log_file too

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.hpp
```diff
@@ -1,14 +1,13 @@
...
  *
  * License for AMD contributions = MIT. See LICENSE for more information
  ************************************************************************/
 
 #ifndef CK_FUSED_ATTN_UTILS_H
 #define CK_FUSED_ATTN_UTILS_H
 
-#include<iostream>
-#include<cstdint>
+#include<fstream>
```
ipanfilo, line 10: why does it need fstream here?

Micky774: All the `ck_fused_attn_*.cpp` files require it, so just to factor it out into the common header.

ipanfilo (approved):

PR #454, opened 2026-02-20
Updated test to include CK/AITER V2/V3 test in single backend case

Micky774:
# Description

Modifies pytorch FA tests to essentially treat the CK backend as two virtual backends, thus directly comparing V2/V3 implementations even when no other backend is available.

Note that sometimes V3 will fallback to V2, so it's not a rigorous test and doesn't substitute having another backend to compare against, but it enables more tests that would otherwise be skipped.

Fixes https://github.com/ROCm/frameworks-internal/issues/15114

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/tests/pytorch/attention/test_attention.py
+++ b/tests/pytorch/attention/test_attention.py
@@ -1301,8 +1301,7 @@ def test_transformer_layer(
                 RoPE,
                 is_training,
             )
-        elif len(fused_attn_backends) == 2:
-            os.environ["NVTE_FUSED_ATTN_BACKEND"] = "0"
+        elif IS_HIP_EXTENSION and len(fused_attn_backends) == 2:
             os.environ["NVTE_FUSED_ATTN_CK"] = "1"
             os.environ["NVTE_FUSED_ATTN_AOTRITON"] = "0"
             fused_attn_fwd, fused_attn_bwd = _run_transformer_layer(
@@ -1316,7 +1315,6 @@ def test_transformer_layer(
                 RoPE,
                 is_training,
             )
-            os.environ["NVTE_FUSED_ATTN_BACKEND"] = "1"
             os.environ["NVTE_FUSED_ATTN_CK"] = "0"
             os.environ["NVTE_FUSED_ATTN_AOTRITON"] = "1"
             fused_attn_fwd_1, fused_attn_bwd_1 = _run_transformer_layer(
--- a/tests/pytorch/attention/test_attention.py
+++ b/tests/pytorch/attention/test_attention.py
@@ -258,7 +258,7 @@ def test_dot_product_attention(
 
     # FusedAttention backend
     if fused_attn_supported:
-        if len(fused_attn_backends) == 1 and FusedAttnBackend["CK"] not in fused_attn_backends:
+        if len(fused_attn_backends) == 1:
             fused_attn_fwd, fused_attn_bwd = _run_dot_product_attention(
                 dtype,
                 config,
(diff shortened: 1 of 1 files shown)


tests/pytorch/attention/test_attention.py
```diff
@@ -258,7 +258,7 @@ def test_dot_product_attention(
 
     # FusedAttention backend
     if fused_attn_supported:
-        if len(fused_attn_backends) == 1:
+        if len(fused_attn_backends) == 1 and FusedAttnBackend["CK"] not in fused_attn_backends:
```
ipanfilo: FusedAnntBackend["CK"] should be guarded - it is invalid key on NV platform.

tests/pytorch/attention/test_attention.py
```diff
@@ -269,7 +269,33 @@ def test_dot_product_attention(
                 pad_between_seqs,
                 is_training,
             )
-        if len(fused_attn_backends) == 2:
+        # We can consider the CK backend as being two, since we have V2/V3 kernels
+        elif len(fused_attn_backends) == 1:
```
ipanfilo: To avoid code duplication it should rather be outside of len(fused_atn_backends) but under separate 'if IS_HIP_EXTENSION and FusedAnndBackends["CK"] in fused_attention_backends'

ipanfilo (changes requested):

ipanfilo:
How many previously skipped test cases are run with that change? It only allows running configs that are not supported by AOTriton but supported by Unfused or FA.
If the goal is to make CK V3/V2 comparison in all configs, skip condition for total number of supported backends should also be updated

tests/pytorch/attention/test_attention.py
Micky774: Done

tests/pytorch/attention/test_attention.py
Micky774: Done

Micky774:
> How many previously skipped test cases are run with that change? It only allows running configs that are not supported by AOTriton but supported by Unfused or FA. If the goal is to make CK V3/V2 comparison in all configs, skip condition for total number of supported backends should also be updated

Good point, I've updated the skip conditions to double-count CK directly.

tests/pytorch/attention/test_attention.py
```diff
@@ -352,6 +369,11 @@ def test_dot_product_attention(
             torch.testing.assert_close(fused_attn_fwd, fused_attn_fwd_2, **tols)
             for i, _ in enumerate(fused_attn_bwd):
                 torch.testing.assert_close(fused_attn_bwd[i], fused_attn_bwd_2[i], **tols)
+    if fused_attn_supported and len(fused_attn_backends) == 1 and FusedAttnBackend["CK"] in fused_attn_backends:
```
wangye805: On NVTE upstream, will querying key "CK" in dict FusedAttnBackend get us a crash/failure?

tests/pytorch/attention/test_attention.py
wangye805, line 1303: How about len(fused_attn_backends) ==1 and IS_HIP_EXTENSION? We need similiar v2 vs v3 check

wangye805 (changes requested):

tests/pytorch/attention/test_attention.py
ipanfilo: Doesn't seem so. The code is still duplicated for len(fused_attn_backends) of 1 and 2

tests/pytorch/attention/test_attention.py
Micky774: Done

tests/pytorch/attention/test_attention.py
Micky774: Yeah I missed this unguarded usage, good catch

wangye805 (approved):

tests/pytorch/attention/test_attention.py
```diff
@@ -269,7 +274,21 @@ def test_dot_product_attention(
                 pad_between_seqs,
                 is_training,
             )
-        if len(fused_attn_backends) == 2:
+            # We can consider the CK backend as being two, since we have V2/V3 kernels
```
ipanfilo: This code duplicates the same addition under len(fused_attn_backend) == 2: https://github.com/ROCm/TransformerEngine/blob/7ac6eaba7ddc6b45e69cb4349a50ff906c348f84/tests/pytorch/attention/test_attention.py#L318
Running of V2 is independent from len(fised_attn_backend), but is controlled by FusedAttnBackend["CK"] presence so it should be outside of that if.

tests/pytorch/attention/test_attention.py
```diff
@@ -240,7 +240,12 @@ def test_dot_product_attention(
         flash_attn_supported = True
 
     # Skip if only unfused backend is supported
-    if (len(fused_attn_backends) + flash_attn_supported + unfused_attn_supported) < 2:
+    # Double-count the CK backend since we want to compare V2/V3 kernels
+    if (
+        len(fused_attn_backends) +
+        int(IS_HIP_EXTENSION and FusedAttnBackend["CK"] in fused_attn_backends) +
```
ipanfilo: Why using int? Because FusedAttnBackend gives two variants its presence is enough to continue so add 'not (IS_HIP_EXTENSION and FusedAttnBackend["CK"] in fused_attn_backends)' to original condition

Micky774: The int expression double-counts CK as a backend, so if `fused_attn_backends=FusedAttnBackend["CK"]` then the `int(...) == 1`. It seemed like applying this to the condition explained why the skip was being avoided in a more literal way (i.e. adjusting the count of the number of backends directly, rather than an excluding condition).

I'm happy to adjust to your approach though.

tests/pytorch/attention/test_attention.py
Micky774: I've reorganized the test a bit now, thanks.

tests/pytorch/attention/test_attention.py
```diff
@@ -347,11 +351,20 @@ def test_dot_product_attention(
...
         for i, _ in enumerate(fused_attn_bwd):
             torch.testing.assert_close(fused_attn_bwd[i], fused_attn_bwd_1[i], **tols)
-        if IS_HIP_EXTENSION:
-            logging.info("[test_dot_product_attention]: fused attn backend 0 vs 2")
+        if has_ck_backend:  # Compare CK V2/V3 if both are available
+            logging.info("[test_dot_product_attention]: CK fused attn V2 vs V3")
             torch.testing.assert_close(fused_attn_fwd, fused_attn_fwd_2, **tols)
             for i, _ in enumerate(fused_attn_bwd):
                 torch.testing.assert_close(fused_attn_bwd[i], fused_attn_bwd_2[i], **tols)
+    if (
```
ipanfilo: just "if has_ck_backend" and you can remove V/2/V3 comparison from len(fused_attn_backend) == 2 section

tests/pytorch/attention/test_attention.py
```diff
@@ -1293,6 +1310,20 @@ def test_transformer_layer(
                 RoPE,
                 is_training,
             )
+            if has_ck_backend:
```
ipanfilo: apply the same logic as in test_dot_product_attention

ipanfilo (approved):

PR #459, opened 2026-02-24
Always use V2 hipify. Make all hipify results consistent

ipanfilo:
# Description

Fixes https://github.com/ROCm/frameworks-internal/issues/15641

## Type of change

- [x] Infra/Build change
- [x] Code refactoring

## Changes

- Force using V2 hipify
- Make single unified hipification library used everywhere across build process

# Checklist:


build_tools/hipify/hipify.cmake
```diff
@@ -0,0 +1,59 @@
...
+    # Read result from output file
+    file(READ "${list_file}" result_content)
+    string(REPLACE "\n" ";" result_content "${result_content}")
+    
+    # Clean up temp files
+    file(REMOVE "${list_file}")
+    
+    # Set output variable in parent scope
+    set(${OUTPUT_VARIABLE} "${result_content}" PARENT_SCOPE)
+endfunction()
```
alextmagro: nit: EOF

build_tools/hipify/hipify.py
```diff
@@ -0,0 +1,200 @@
...
+        output_directory=src_dir,
+        includes=["*/common/*", str(Path(src_dir)/"*")],
+        ignores=["*/amd_detail/*", "*/aotriton/*", "*/ck_fused_attn/*", "*/rocshmem_api/*"],
+        header_include_dirs=include_dirs,
+        custom_map_list= te_root / "build_tools" / "hipify" / "custom_map.json",
+        extra_files=[],
+        is_pytorch_extension=True,
+        hipify_extra_files_only=False,
+        show_detailed=False,
+        no_math_replace=True)
```
alextmagro, line 62: Before, we only did no_math_replace for our cpp tests. I had limited it there as Jeff had mentioned that the performance of the std::math functions was not guaranteed to be as performant as the global namespace versions for GPUs. Not sure if that is actually the case, but we might want to double check for heavy ops like exp and log. I remember running a script that verified the accuracy of the std:: functions, but didn't check their performance. I'll send the file to you in case you think it's a good idea to double check.

ipanfilo: I reviewed the code and it seems that the flag only makes a difference for tests. Among math methods used by TE only max() and min() for double and float types produce different GPU code and they are not used by TE GPU kernels. On host side max()/min() with float and double types are used in tests only. 
Thus, while ::max/::min produce more optimized code than std::max/std::min for some datatypes, it should not affect TE performance.
Below are examples of std:max and ::max with float on Gfx950:

     33         v_cmp_lt_f32_e32 vcc, v0, v1                               // 000000001954: 7C820300
     34         s_nop 1                                                    // 000000001958: BF800001
     35         v_cndmask_b32_e32 v0, v0, v1, vcc                          // 00000000195C: 00000300
     36         s_setpc_b64 s[30:31]                                       // 000000001960: BE801D1E

     40         v_max_f32_e32 v1, v1, v1                                   // 000000001968: 16020301
     41         v_max_f32_e32 v0, v0, v0                                   // 00000000196C: 16000100
     42         v_max_f32_e32 v0, v0, v1                                   // 000000001970: 16000300
     43         s_setpc_b64 s[30:31]                                       // 000000001974: BE801D1E

alextmagro: Sounds good! LGTM.

alextmagro (approved):

PR #466, opened 2026-02-28
Add ROCm mismatch bypass env flag

jiagaoxiang:
## Summary
- Make the ROCm runtime/build version check configurable at import time.
- Keep strict validation as the default behavior.
- Allow an explicit bypass via `NVTE_ALLOW_ROCM_MISMATCH=1` for environments that accept mismatch risk.

## Test plan
- [ ] Build/install TransformerEngine in an environment where ROCm runtime and build versions match; verify import succeeds.
- [ ] Simulate a mismatch and verify import fails by default.
- [ ] Set `NVTE_ALLOW_ROCM_MISMATCH=1` and verify import succeeds despite mismatch.

Made with [Cursor](https://cursor.com)


ipanfilo:
Close in favor of #467

PR #472, opened 2026-03-04
NVFP4 cast/transpose without TMA

matthiasdiener:
# Description

Fixes https://github.com/ROCm/frameworks-internal/issues/15731

TODO:
- [x] Implement other cases, not just fwd 1D
- [x] tests for other cases (2D, SR)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


tests/cpp/operator/CMakeLists.txt
```diff
@@ -37,6 +37,7 @@ if(USE_CUDA)
               test_swizzle.cu)
 else()
   list(APPEND test_cuda_sources
+              test_cast_nvfp4_transpose.cu
```
ipanfilo: It should rather go to common section, where all other cast tests are.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
```diff
@@ -189,7 +213,11 @@ __device__ __forceinline__ float ComputeOutputFP4(IType input, float encode_scal
 }
 
 __device__ __forceinline__ float ComputeGlobalEncodeScaleFP4(const float global_amax) {
+#ifdef __HIP_PLATFORM_AMD__
+  const float fp8_max = TypeExtrema<fp8e4m3>::max;
```
ipanfilo, line 201: For device code constexpr should still work

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
```diff
@@ -270,17 +299,34 @@ __device__ __forceinline__ __nv_fp4x4_e2m1 cvt_fp32_to_fp4_4x_with_stochastic_ro
         : "f"(in01.y), "f"(in01.x), "f"(in23.y), "f"(in23.x), "r"(rbits));
     return *reinterpret_cast<__nv_fp4x4_e2m1*>(&out_4x);
   } else {
+#endif
     NVTE_DEVICE_ERROR(
```
ipanfilo, line 286: This error message is not valid on ROCm

transformer_engine/common/CMakeLists.txt
```diff
@@ -259,7 +259,8 @@ else()
     list(APPEND transformer_engine_cuda_sources
         fused_attn_rocm/fused_attn_aotriton.cpp
         fused_attn_rocm/fused_attn_ck.cpp
-        fused_attn_rocm/utils.cpp)
+        fused_attn_rocm/utils.cpp
+        transpose/quantize_transpose_vector_blockwise_fp4.cu)
```
ipanfilo: It is not ROCm specific source code, it should be added out of IS_ROCM/IS_CUDA if

ipanfilo:
The code contains number of ifdefs for just substitution of cuda_fp4.h __nv_fp4_e2m1, etc, with HIP counterparts. I suggest to use custom hipification map (build_tools/hipify/custom_map.json) and remove ifdefs from code. It can also be used for #include <cudaTypedefs.h>

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -117,10 +117,20 @@ void quantize_fwd_helper(const NVTETensor input, NVTETensor output,
               *input_tensor, noop_tensor, output_tensor, &quant_config_cpp, stream);
         }
       } else {
+#endif
         auto &global_amax = (output_tensor->amax.dptr != nullptr) ? output_tensor->amax
                                                                   : output_tensor->columnwise_amax;
+
+        // If amax was not explicitly set, fall back to the scale field which
```
ipanfilo: Guard it.

tests/cpp/operator/CMakeLists.txt
matthiasdiener: Fixed in 95d0c9fd

transformer_engine/common/CMakeLists.txt
matthiasdiener: Fixed in 95d0c9fd

transformer_engine/common/cast/dispatch/quantize.cuh
matthiasdiener: Added a guard in 95d0c9fd

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
matthiasdiener: I changed it to `#if defined(__HIP_PLATFORM_AMD__) && !defined(__HIP_DEVICE_COMPILE__)` in 95d0c9fd

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
matthiasdiener: I added a new message in https://github.com/ROCm/TransformerEngine/commit/95d0c9fd42987f87302d506443c5a3a3c5d6418d.

matthiasdiener:
> The code contains number of ifdefs for just substitution of cuda_fp4.h __nv_fp4_e2m1, etc, with HIP counterparts. I suggest to use custom hipification map (build_tools/hipify/custom_map.json) and remove ifdefs from code. It can also be used for #include <cudaTypedefs.h>

Thank you for the suggestion. I changed to using the hipify map in https://github.com/ROCm/TransformerEngine/pull/472/commits/55a8c849ee2c598487389d6a680c11754e18dc59

build_tools/hipify/custom_map.json
```diff
@@ -2,13 +2,20 @@
...
         "cudaFuncSetAttribute(" : "hipFuncSetAttribute((const void*)",
         "__nv_bfloat162":"__hip_bfloat162",
-        "cuda::getCurrentCUDAStream" : "hip::getCurrentHIPStreamMasqueradingAsCUDA"
+        "cuda::getCurrentCUDAStream" : "hip::getCurrentHIPStreamMasqueradingAsCUDA",
+        "__nv_fp4_e2m1" : "__hip_fp4_e2m1",
+        "__nv_fp4x2_e2m1" : "__hip_fp4x2_e2m1",
+        "__nv_fp4x4_e2m1" : "__hip_fp4x4_e2m1",
+        "__nv_fp4x2_storage_t" : "__hip_fp4x2_storage_t",
+        "<cudaTypedefs.h>" : "<hip/hip_version.h>",
+        "<cuda/barrier>" : "<hip/hip_version.h>"
```
wangye805: Why cudaTypedefs.h and cuda/barrier are both translated to hip/hip_version.h?

tests/cpp/operator/test_cast_nvfp4_transpose.cu
```diff
@@ -30,14 +32,27 @@ enum ActivationType {
     SReLU
 };
 
+static constexpr float E2M1_LUT[16] = {
```
wangye805, line 36: Guard by our rocm specific macro?

tests/cpp/test_common.cu
```diff
@@ -1219,12 +1228,20 @@ std::array<size_t, 4> get_scale_tensor_dims(const size_t rows,
     const bool is_rowwise = (block_size_rows == 1)
                             && ((block_size_cols == 32) || (block_size_cols == 16));
 
+    // On AMD, MXFP8 scales (block_size=32) are passed unpadded to hipBLASlt.
+    // NVFP4 scales (block_size=16) still require [128,4] padding for kernel indexing.
+#ifdef __HIP_PLATFORM_AMD__
+    const bool needs_padding = (block_size_cols == 16 || block_size_rows == 16);
+    const size_t alignment_Y = needs_padding ? (is_rowwise ? 128 : 4) : 1;
```
wangye805: Try to encode this 128, 4 by some const, const expr, or macro, instead of hardcoding
perhaps nvfp4_scale_tensor_lignment_X/Y_rowwise?

tests/cpp/test_common.h
```diff
@@ -19,12 +19,13 @@
...
 #include <cuda_fp4.h>
-#endif
+#endif //FP4_TYPE_SUPPORTED
 #else
-#define FP4_TYPE_SUPPORTED (false)
+#define FP4_TYPE_SUPPORTED (true)
 #include <hip/hip_bfloat16.h>
 #include "amd_detail/hip_float8.h"
-#endif
+#include <hip/hip_fp4.h>
```
wangye805: I recall you put cuda_fp4.h translation rule into our hipify json file already?

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -117,10 +117,22 @@ void quantize_fwd_helper(const NVTETensor input, NVTETensor output,
               *input_tensor, noop_tensor, output_tensor, &quant_config_cpp, stream);
         }
       } else {
+#endif
         auto &global_amax = (output_tensor->amax.dptr != nullptr) ? output_tensor->amax
                                                                   : output_tensor->columnwise_amax;
+#ifdef __HIP_PLATFORM_AMD__
+        // If amax was not explicitly set, fall back to the scale field which
+        // holds the same value when set via set_scale().
+        NVTE_CHECK(global_amax.dptr != nullptr || output_tensor->scale.dptr != nullptr,
```
wangye805: Is it a bug fix for upstream? If not, why do we need this specific treatment for global amax?

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
```diff
@@ -273,14 +294,34 @@ __device__ __forceinline__ __nv_fp4x4_e2m1 cvt_fp32_to_fp4_4x_with_stochastic_ro
...
         "FP4 cvt.rs PTX instructions are architecture-specific. "
         "Try recompiling with sm_XXXa instead of sm_XXX.");
+#else
+    NVTE_DEVICE_ERROR(
+        "cvt_fp32_to_fp4_4x_with_stochastic_rounding is not supported on AMDGPU.");
```
wangye805: I recall we have some stochastic rounding ability since MI350 @wenchenvincent ?

tests/cpp/operator/test_cast_nvfp4_transpose.cu
wangye805, line 588: There was a question in NV upstream PR about why they only test 1D quantization. Do you think we can try both true and false to see if we already support 2d quantization or not?

wangye805 (changes requested):
So currently we don't have any walkaround for the stochastic rounding path?

build_tools/hipify/custom_map.json
matthiasdiener: It is just to "disable" those particular includes for which there seems to be no equivalent in ROCm. I changed this slightly in b4caf6f7, let me know if you prefer that (or just restoring the previous version that uses #ifndef guards around the nonexisting headers).

tests/cpp/test_common.h
matthiasdiener: I just adapted it to the surrounding code, which already includes the HIP headers directly: `#include <hip/hip_bfloat16.h>`. Let me know which one you prefer.

transformer_engine/common/cast/dispatch/quantize.cuh
matthiasdiener: Yes, I believe this is a bug in upstream.

tests/cpp/operator/test_cast_nvfp4_transpose.cu
matthiasdiener: Yes, 2D works as well (the test is still disabled upstream). I enabled 2D testing in [511db61](https://github.com/ROCm/TransformerEngine/pull/472/commits/511db6171b356817a3a855728645860bb8480ea7), this way was the simplest I found that avoids restructuring the whole test setup.

tests/cpp/operator/test_cast_nvfp4_transpose.cu
matthiasdiener: Done in [511db61](https://github.com/ROCm/TransformerEngine/pull/472/commits/511db6171b356817a3a855728645860bb8480ea7)

tests/cpp/test_common.cu
matthiasdiener: Changed in [511db61](https://github.com/ROCm/TransformerEngine/pull/472/commits/511db6171b356817a3a855728645860bb8480ea7)

matthiasdiener:
> So currently we don't have any walkaround for the stochastic rounding path?

I was able to implement SR via intrinsics on gfx950 in https://github.com/ROCm/TransformerEngine/pull/472/commits/36cf73a67ddaa650466eb33ea99b965ac48e6049. I also expanded the test to use it.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
matthiasdiener: See https://github.com/ROCm/TransformerEngine/commit/36cf73a67ddaa650466eb33ea99b965ac48e604 for the gfx950 implementation.

tests/cpp/test_common.h
ipanfilo: cuda_bf16 and cuda_fp8 headers are also hipified with cusom map

transformer_engine/common/cast/dispatch/quantize.cuh
ipanfilo: Maybe put comment then.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
ipanfilo: Yes, for host it is not a constexpr. However, host code translation results should be eliminated from final binary so value is not important. Even whole method could be guarded with the same final result so defined(__HIP_PLATFORM_AMD__) && !defined(__HIP_DEVICE_COMPILE__) is good.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
```diff
@@ -257,8 +284,13 @@ __device__ __forceinline__ size_t scale_factor_swizzled_offset(size_t row_idx, s
   return ((rb * cbg_cnt + cbg) * kRowsPerBaseBlockCol + d3) * 16 + d4 * kColsPerBaseBlockCol + d5;
 }
 
+#ifdef __HIP_PLATFORM_AMD__
+__device__ __forceinline__ fp4x4_storage_t cvt_fp32_to_fp4_4x_with_stochastic_rounding(
```
ipanfilo: fp4x4_storage_t is already correctly redefined for HIP and CUDA so no need ifdef here. Or if, you want to keep original declaration unchanged, you can use 'using __nv_fp4x4_e2m1= __hip_fp4x4_storage_t' on AMD.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
```diff
@@ -273,14 +305,51 @@ __device__ __forceinline__ __nv_fp4x4_e2m1 cvt_fp32_to_fp4_4x_with_stochastic_ro
     NVTE_DEVICE_ERROR(
         "FP4 cvt.rs PTX instructions are architecture-specific. "
         "Try recompiling with sm_XXXa instead of sm_XXX.");
+#else
+#ifdef __gfx950__
```
ipanfilo: It may make sense to have analogue of ARCH_HAS_STOCHASTIC_ROUNDING define if such guarding is used in multiple places - we'll later add more platforms with FP4 support.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
```diff
@@ -155,12 +164,25 @@ static_assert(kNumThreadsLoad <= kThreadsPerWarp, "kNumThreadsLoad must be <= kT
 static_assert(kNumThreadsStore <= kThreadsPerWarp, "kNumThreadsStore must be <= kThreadsPerWarp");
 
 // for 2D block scaling, we need to reduce amax in warp
+#ifdef __HIP_PLATFORM_AMD__
+static __device__ constexpr uint64_t WARP_REDUCE_AMAX_GROUP_MASKS[8] = {
```
ipanfilo: I think with 32 threads per wavefront actively used the high half of mask should be 0

build_tools/hipify/custom_map.json
wangye805: Oh, I see. Then I prefer just guard those header files via "#ifndef __HIP_PLATFORM_AMD__". Maybe several months later, hipify supports those two header files but we still map it to null or hip_version.h

tests/cpp/test_common.cu
wangye805: Cool. THanks

transformer_engine/common/cast/dispatch/quantize.cuh
wangye805: I see. Thanks! 

Also, check if upstream already had an fix. If not, I think it's okay to drop the rocm specific guard. What do you think @ipanfilo?

tests/cpp/test_common.h
wangye805: I mean we can simplify your line 16-28 by just one USE_ROCM guard on FP4_TYPE_SUPPORTED with other lines unchanged as NV upstream? This will make our code cleaner?

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
wangye805: Thanks

wangye805:
> > So currently we don't have any walkaround for the stochastic rounding path?
> 
> I was able to implement SR via intrinsics on gfx950 in [36cf73a](https://github.com/ROCm/TransformerEngine/commit/36cf73a67ddaa650466eb33ea99b965ac48e6049). I also expanded the test to use it.

Great. Thanks

build_tools/hipify/custom_map.json
ipanfilo: Well, hipify changes won't affect us until we update our submodule, so custom map can be reviewed. Also, if they are guarded with ifndef hipify support will not affect TE code too
Moreover, if those headers are not needed for ROCm path, no reason to hipify them to anything else except void
Thus, depending on how often those headers appear here than there and whether there are HIP ifdefs there already, custom map may be a good alternative to adding multiple guards

matthiasdiener: Disabling the #ifdefs in the custom map file reduces the number of files that we need to touch in this PR by 3. Are you ok with the way it is currently implemented @wangye805 ?

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
matthiasdiener: I'm not sure this can be simplified further.
Both sides need different return types and `fp4x4_storage_t` can only be used on AMD.
The map file has `"__nv_fp4x2_e2m1" : "__hip_fp4x2_e2m1",` , so using `__nv_fp4x4_e2m1 = __hip_fp4x4_storage_t` would become `using __hip_fp4x4_e2m1 = __hip_fp4x4_storage_t` after hipification, which is a redefinition of the existing struct in `amd_hip_fp4.h`.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
matthiasdiener: Isn't `kThreadsPerWarp=32` just a logical grouping value here, not the hardware wavefront width?

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
matthiasdiener: Added in https://github.com/ROCm/TransformerEngine/pull/472/commits/a607feb45c39632d9b79aafefd478839a3b56df5

transformer_engine/common/cast/dispatch/quantize.cuh
matthiasdiener: I don't think this is fixed in upstream yet. I added a comment in https://github.com/ROCm/TransformerEngine/pull/472/commits/a607feb45c39632d9b79aafefd478839a3b56df5

tests/cpp/test_common.h
matthiasdiener: I simplified this like you suggested in https://github.com/ROCm/TransformerEngine/pull/472/commits/a607feb45c39632d9b79aafefd478839a3b56df5.

tests/cpp/test_common.cu
```diff
@@ -181,12 +181,21 @@ std::pair<scale_inv_meta, scale_inv_meta> get_scales(const NVTEShape& shape,
 
     scale_inv_meta ret_rowwise, ret_colwise;
 
-    size_t scale_dim_Y = DIVUP_TO_MULTIPLE(first_dim, scale_tensor_alignment_Y_rowwise);
-    size_t scale_dim_X = DIVUP_TO_MULTIPLE(DIVUP(last_dim, 16lu), scale_tensor_alignment_X_rowwise);
+#ifdef __HIP_PLATFORM_AMD__
+    // NVFP4 requires [128,4] padding on AMD regardless of MXFP8 alignment constants
+    constexpr size_t nvfp4_align_Y = 128;
```
ipanfilo: Use nvfp4 constants from test_common.h

build_tools/hipify/custom_map.json
wangye805: Yeah. Thanks

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
ipanfilo: fp4x4_storage_t is declared not to have ifdef later in code but use fp4x4storage_t everywhere. If you use ifdef here, you don't need fp4x4storage_t but can use __hip_fp4x4_storage_t directly.
Also, why do you need to use __hip_fp4x4_storage_t, not __hip_fp4x4_e2m1 here which would let just using hipification for resulting type?

transformer_engine/common/cast/dispatch/quantize.cuh
wangye805: Thanks. I would like to understand more about this fix. Probably get a B200 to test NV upstream behavior is hard. Which cpp gtest failed due to this bug? According to NV upstream design, should the output_amax be set with the correct value? If so, we should fix the bug in the place where this setting was missed

wangye805 (changes requested):

tests/cpp/test_common.cu
matthiasdiener: Thanks, done in https://github.com/ROCm/TransformerEngine/pull/472/commits/5a5803c66ef742c9d06c1fad52a699b2cd3efc30.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
matthiasdiener: > fp4x4_storage_t is declared not to have ifdef later in code but use fp4x4storage_t everywhere. If you use ifdef here, you don't need fp4x4storage_t but can use __hip_fp4x4_storage_t directly. 

I did this simplification in fc5af655.

> Also, why do you need to use __hip_fp4x4_storage_t, not __hip_fp4x4_e2m1 here which would let just using hipification for resulting type?

Thanks, I was able to find another way to do the bit fiddling in this function and the SR function, that does not need `__hip_fp4x4_storage_t` in 94a4e5e1.

transformer_engine/common/cast/dispatch/quantize.cuh
matthiasdiener: The tests in `test_cast_nvfp4_transpose.cu` fail - I suspect this isn't tested upstream for the fallback path (which is the only path we currently have implemented). Looking at the upstream optimized kernel (in quantize_transpose_nvfp4.cuh), `amax_rowwise_ptr` is explicitly allowed to be null; the kernel falls back to 1.0f in that case. `amax.dptr` is never allocated for NVFP4 tensors in the upstream test Tensor class, and the optimized path never needs it. I changed the fix to match this null-handling behavior in the fallback kernel in 82af544f. I believe this is still incorrect in upstream's main branch too.

tests/cpp/operator/test_cast_nvfp4_transpose.cu
```diff
@@ -567,7 +584,18 @@ void performTest(float (*OP)(const float),
...
     output.set_scale(amax);
 
+#ifndef __HIP_PLATFORM_AMD__
     bool use_2d_quantization = false;
+#else
+    // Test both 1D and 2D quantization paths on AMDGPU,
+    // as well as stochastic rounding.
+    hipDeviceProp_t prop;
+    hipGetDeviceProperties(&prop, 0);
+    const bool is_gfx950 = std::string(prop.gcnArchName).find("gfx950") != std::string::npos;
```
ipanfilo: prop.major == 9 && prop.minor == 5

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
```diff
@@ -273,14 +294,45 @@ __device__ __forceinline__ __nv_fp4x4_e2m1 cvt_fp32_to_fp4_4x_with_stochastic_ro
     NVTE_DEVICE_ERROR(
         "FP4 cvt.rs PTX instructions are architecture-specific. "
         "Try recompiling with sm_XXXa instead of sm_XXX.");
+#else
+#if ARCH_HAS_STOCHASTIC_ROUNDING
+  // opsel=1 always writes to byte 1, result read from fp4x2[1]
```
ipanfilo, line 291: Is it about some ISA magic that you use opsel=1 for both lo and hi?

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
ipanfilo: Probably correct replacement for ROCm is __shfl_down(val, offset * shfl_down_stride, 8); TE code used to avoid using __shfl_down_sync directly. Does it behave like CUDA one?

transformer_engine/common/cast/dispatch/quantize.cuh
wangye805: Thanks. 

So currently in the gtest, we set the global encode/decode factor to be 1.0f, which is effectively having a global_amax=1. I understand that this should workaround the bug that global_amax is not set. But later this gtest is going to compare the quantized fp4 and scale tensors separately with the reference flow. The reference flow has the correct global amax set. Then how can we match the target with amax=1 with the reference with the true amax? 

Unless the amax in this gtest happen to be 1?

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
```diff
@@ -155,12 +157,25 @@ static_assert(kNumThreadsLoad <= kThreadsPerWarp, "kNumThreadsLoad must be <= kT
...
 template <int group_size, int shfl_down_stride>
-__device__ __forceinline__ float groupMax(float val, unsigned int groupMask) {
+__device__ __forceinline__ float groupMax(float val,
+#ifdef __HIP_PLATFORM_AMD__
+                                          uint64_t groupMask) {
+#else
+                                          unsigned int groupMask) {
+#endif
   for (int offset = group_size / 2; offset > 0; offset /= 2) {
     val = max(val, __shfl_down_sync(groupMask, val, offset * shfl_down_stride));
```
wangye805, line 171: If the unsigned int vs uint64_t code change agove is to resolve the __shfl_down_sync, TE previously use __shfl_down instead. See https://github.com/ROCm/TransformerEngine/blob/779f3fdbb208a9df373d1977fb9d1ad105946889/transformer_engine/common/utils.cuh#L113-L117 for more details.

wangye805 (changes requested):

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
wenchenvincent: Seems that we don't have the equivalent to `__shfl_down_sync` in HIP. We might need to use the cross-lane primitives at the assembly level: https://gpuopen.com/learn/amd-gcn-assembly-cross-lane-operations/

wenchenvincent: Looking back at it, it seems that while we don't support the mask, for the specific grouped mask used in this kernel, we can still do it with `__shfl_down`.

transformer_engine/common/cast/dispatch/quantize.cuh
wangye805: Synced with Matthias offline. Now understand that the ref flow is not setting the global amax either

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
wangye805: We have __shfl_down_sync in hip: https://rocm.docs.amd.com/projects/HIP/en/latest/how-to/hip_cpp_language_extensions.html

ipanfilo: We have, but contrary to CUDA, mask is not used to active lanes selection on ROCm. I think that calling __shfl_down with the same offsets and width=32 will have the same results - T0..T7 will contain max(amax) for corresponding groups.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
wangye805: Actually, after looking more into __shfl_down_sync vs __shfl_down, and amd lock-step vs NV architecture, I think  Matthias' current implementation is correct. 
1). in our amd platform, __shfl_down_sync provides us the mask which __shfl_down does not. And since amd platform runs wavefronts in the lock-step style, so it provides the "sync" for free. 
2). in the current tiling design, group size = 4, shfl_down_stride =8 are hardcoded. Only the results from threads (0-7, 32-39, 64-71, 96-103, ...) are needed (threadIdx.x/8)%4==0, according to https://github.com/ROCm/TransformerEngine/blob/56fefafdb6a1b9f8a7d7964e84feac96c564753f/transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu#L550. 
Take the first group into consideration, extending the 32bit mask 0x0101010101 to 64bit mask 0x01010101_01010101 does not change the results for threads 0-7 and 32-39
3). In fact, I think the mask is actually not needed at all with the shfl_down_stride as 8 already

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
wangye805: Oh, I didn't notice you comment here as well, @ipanfilo . In fact, I would say we should keep the same offsets but not changing the width from 64 to 32. Refer to my other comment for more details. I think we cannot force our hardware to run just 32 threads per warp (or wavefront)

wangye805 (approved):
LGTM

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
matthiasdiener: Per our discussion, I changed this to use `__shfl_down` in https://github.com/ROCm/TransformerEngine/pull/472/commits/83f9fa512e623ef8f69906dbbc678656f6857065.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
matthiasdiener: This matches how this builtin is used in e.g. https://github.com/ROCm/clr/blob/3dbb5f1c5e0734d21dd2424a38255e61ee0a73e0/hipamd/include/hip/amd_detail/amd_hip_ocp_fp.hpp#L1858-L1890. We could use opsel=0 then opsel=1 to pack both results into the same register directly, but this relies on the second call not overwriting the first byte, which seems more fragile.

I expanded the comment slightly in [71f36b6](https://github.com/ROCm/TransformerEngine/pull/472/commits/71f36b675f0969f451f3ad76498e989f3f775e51)

tests/cpp/operator/test_cast_nvfp4_transpose.cu
matthiasdiener: Done in [71f36b6](https://github.com/ROCm/TransformerEngine/pull/472/commits/71f36b675f0969f451f3ad76498e989f3f775e51)

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
```diff
@@ -162,7 +164,11 @@ static __device__ constexpr unsigned int WARP_REDUCE_AMAX_GROUP_MASKS[8] = {
 template <int group_size, int shfl_down_stride>
 __device__ __forceinline__ float groupMax(float val, unsigned int groupMask) {
   for (int offset = group_size / 2; offset > 0; offset /= 2) {
+#ifdef __HIP_PLATFORM_AMD__
+    val = max(val, __shfl_down(val, offset * shfl_down_stride, kThreadsPerWarp));
```
ipanfilo, line 169: nit: this code should produce unused argument warning for groupMask, please suppress

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
```diff
@@ -162,7 +164,11 @@ static __device__ constexpr unsigned int WARP_REDUCE_AMAX_GROUP_MASKS[8] = {
...
+#else
     val = max(val, __shfl_down_sync(groupMask, val, offset * shfl_down_stride));
```
ipanfilo, line 171: There are other occurrences of __shfl_sync and __shfl_down_sync in this code. Please elaborate why they do not throw assertion dues to mask size mismatch and update them is needed

ipanfilo (changes requested):

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
matthiasdiener: The other `__shfl_down_sync` and `__shfl_sync` calls are inside `if constexpr (kIsE8Scaling)` blocks. `kIsE8Scaling` corresponds to `kPow2Scale` at the call site (in `quantize_transpose_vector_blockwise_fp4()`), which is always false for NVFP4. The dispatch code asserts `NVTE_CHECK(!pow2_scale, "No support for pow2_scale for MXFP4 for now")`. Since this branch is eliminated at compile time and the calls never execute on AMD, there is no mask mismatch.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
matthiasdiener: Added in 13f53dda

ipanfilo (approved):

PR #476, opened 2026-03-09
Use major ROCm version to tag pre-built AITER library

ipanfilo:
# Description

Tag AITER binaries with ROCm major version only when uploading to cash. Download checks major.minor first and then major only. Thus, while using single pre-built binary, it still allows to make and use specific one for specific ROCm release.

Fixes https://github.com/ROCm/frameworks-internal/issues/15768

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

## Changes

- Rework AITER pre-built managing code to use ROCm major 
- Remove verbosity from aiter build by default to avoid 500Mb logs generation
- Make aiter_build.sh installing binaries to a location where TE expect them

# Checklist:


transformer_engine/common/ck_fused_attn/aiter_build.sh
```diff
@@ -6,39 +6,59 @@
...
 #   --gpu-archs <list>            GPU arches (required)
 #   --ck-tile-bf16 <val>          CK_TILE_FLOAT_TO_BFLOAT16_DEFAULT, default: 3
+#   --verbose                     Enable verbose logging
 set -euo pipefail
 
 AITER_DIR=""
 AITER_TEST_DIR=""
 GPU_ARCHS_VAL=""
 CK_TILE_BF16_DEFAULT="${CK_TILE_FLOAT_TO_BFLOAT16_DEFAULT:-3}"
+AITER_LOG_MORE=0
```
Micky774: `AITER_LOG_MORE=${AITER_LOG_MORE:-0}`?

transformer_engine/common/ck_fused_attn/aiter_build.sh
```diff
@@ -6,39 +6,59 @@
...
+  AITER_TEST_DIR="${AITER_DIR}/op_tests/cpp/mha"
+fi
+
+if [ ${AITER_LOG_MORE} -eq 1 ]; then
+  echo "[AITER-BUILD] AITER_DIR: ${AITER_DIR} TEST_DIR: ${AITER_TEST_DIR} GPU_ARCHS: ${GPU_ARCHS_VAL} CK_TILE_BF16_DEFAULT: ${CK_TILE_BF16_DEFAULT} INSTALL_DIR: ${INSTALL_DIR}"
+fi
+
 rm -rf "${AITER_DIR}/aiter/jit/build"
-AITER_LOG_MORE=1 \
+AITER_LOG_MORE=${AITER_LOG_MORE} \
```
Micky774: This wouldn't need to be set separately.

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -18,55 +18,83 @@ endif()
...
+    is_aiter_cache_valid("${ROCM_VER_PARAM}" RESULT)
+    if(RESULT)
+      get_aiter_cache_key("${ROCM_VER_PARAM}" _UNUSED CACHE_DIR)
+      set(${PREBUILT_DIR_VAR} "${CACHE_DIR}" PARENT_SCOPE)
+      return()
+    endif()
+  endforeach()
+  
+  # Cache is invalid/outdated - clean it and some build files
+  file(REMOVE_RECURSE "${AITER_CACHE_ROOT}")
```
Micky774, line 59: Rather than clearing out *all* local caches, how about we only clear out this single one deemed invalid?

Micky774 (commented):
Looks good, just a couple small comments. This should work fine with the static lib changes coming up as well. My main question is, how does one locally cache or upload with minor version as well? Currently it looks like it forces the use of major version only for local caching and uploads, but I might just be missing something.

transformer_engine/common/ck_fused_attn/aiter_build.sh
ipanfilo: Do you mean not setting it with --verbose but let caller set env variable directly?

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
ipanfilo: This is just copy from the original logic. Invalid in this case means rare condition when directory exists but does not contain full set of libraries, right?  If working with different AITER/ROCm versions it may be convenient to keep other versions binaries untouched by a cost of infinitely growing cache though. And with embedded HSA libs the size of cache will be big.

ipanfilo:
> Looks good, just a couple small comments. This should work fine with the static lib changes coming up as well. My main question is, how does one locally cache or upload with minor version as well? Currently it looks like it forces the use of major version only for local caching and uploads, but I might just be missing something.

That is right. Downloading of minor versions is left for (1) compatibility with already uploaded versions, (2) unlike condition when custom build for some major.minor version is needed. 
Because the 2nd is unlike, there is no built-in path to upload such binaries and it should rather been considered to revert versions list to download

transformer_engine/common/ck_fused_attn/aiter_build.sh
Micky774: I missed the verbose arg controlling this the first time around -- though we don't provide for a way to call the script w/ verbose enabled in the actual `CMakeLists.txt` right?

Either way would be fine. I slightly prefer letting the caller set the env variable directly simply since it is less indirection, but it is a weak preference.

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
Micky774: Since it's not a functional change of this PR, I'll resolve this comment and we can discuss later if it becomes relevant.

transformer_engine/common/ck_fused_attn/aiter_build.sh
```diff
@@ -31,14 +36,21 @@ while [[ $# -gt 0 ]]; do
...
+  echo "[AITER-BUILD] --aiter-dir, --aiter-test-dir, and --gpu-archs are required." >&2
   exit 1
 fi
 
+AITER_TEST_DIR="${AITER_TEST_DIR:-${AITER_DIR}/op_tests/cpp/mha}"
+
+echo "[AITER-BUILD] AITER_DIR: ${AITER_DIR} TEST_DIR: ${AITER_TEST_DIR} GPU_ARCHS: ${GPU_ARCHS_VAL} CK_TILE_BF16_DEFAULT: ${CK_TILE_BF16_DEFAULT} INSTALL_DIR: ${INSTALL_DIR}"
+
 rm -rf "${AITER_DIR}/aiter/jit/build"
-AITER_LOG_MORE=1 \
```
wangye805, line 40: Can we control this by passing "pip install -v" down? This AITER_LOG_MORE can be useful when we want to see which kernel failed the compilation

wangye805 (changes requested):

transformer_engine/common/ck_fused_attn/aiter_build.sh
ipanfilo: It it removed by two reasons: (1) it is too verbose even for -v - it generates 500Mb of log; and (2) based on conversation with Meekhail here - it is env variable and you can always set it before running building if needed;
Please note that this script is run not only during pip install so having AITER_LOG_MORE decoupled from pip -v lets set it in all scenarios

wangye805 (approved):

PR #478, opened 2026-03-10
Microbenchmarking, Torch+CSV-based

matthiasdiener:
# Description

See also #487.

Pytorch benchmark timing: https://docs.pytorch.org/tutorials/recipes/recipes/benchmark.html

**Open questions:**
- [x] What to do with performance-related env variables? (e.g., using Triton kernels for some operations, ck_tile for grouped GEMM, ...) **Decision: use defaults for now**
- ~Do we need to rebuild the PR branch after perf testing is done?~

---

Partly addresses https://github.com/ROCm/frameworks-internal/issues/15863

Microbenchmarking (not just) for CI.

- Implemented 5 microbenchmarks, (fp16 gemm, fp8 gemm, fp16 grouped gemm, normalization, casting)
- Implemented performance regression test for CI: 
  1. Run benchmarks for PR branch
  2. Checkout base branch, re-build that branch, re-run benchmarks
  3. Compare results between base branch and PR branch
  4. Report comparison results as PR comment, but don't fail CI if performance regresses
     - See https://github.com/ROCm/TransformerEngine/pull/478#issuecomment-4042289032 for the comment it creates
- Could be expanded to non-CI use cases (like nightly performance regression tests)
  - If additional CI time (currently, ~5 minutes) is an issue, could be part of just a Level 3 CI run

TODOs:
- [x] add attention, normalization, casting
- [x] print commit in header of comment

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

## Changes

- Added a set of 6 microbenchmarks (attention, fp16 gemm, fp8 gemm, fp16 grouped gemm, normalization, casting)
- Create a test within CI that compares microbenchmark performance between PR branch and base branch

# Checklist:


benchmarks/microbenchmarks/compare_results.py
```diff
@@ -0,0 +1,135 @@
...
+        if c not in metric_cols and c not in SKIP_COLS
+        and "Time" not in c
+    ]
+    return key_cols, metric_cols
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Compare benchmark CSVs")
+    parser.add_argument("base_csv", help="Base branch CSV")
+    parser.add_argument("pr_csv", help="PR branch CSV")
```
ipanfilo: It is not necessary PR. The script can compare results of two different branches, or two different machines or even run-to-run variations

.github/workflows/rocm-ci.yml
```diff
@@ -368,6 +371,168 @@ jobs:
           EOF
           )"
 
+      - name: Performance regression check
```
ipanfilo: Microbenhmarking hardly needs to be run on every PR commit. And it is not necessary to run on PR. IT may be push to dev branch.
Also, can the action store results in GHA cache and read them for compare instead of making two runs?

.github/workflows/rocm-ci.yml
```diff
@@ -368,6 +371,168 @@ jobs:
...
+          done
+          OUTER
+          )"
+
+          # Assemble this runner's section
+          SUMMARY="perf_results/reports/summary.md"
+          DETAILS="perf_results/reports/details.md"
+          [ -f "$SUMMARY" ] || exit 0
+
+          SECTION_START="<!-- perf-section:${DISPLAY_NAME} -->"
```
ipanfilo: There going to be one one section - it compares the same different runs on the same machine

matthiasdiener: There are two sections, one per runner (MI325 and MI355), both runners post to the same PR comment. The section markers let the second runner to finish insert its results without overwriting the first runner's section. Without them, whichever runner finishes last would clobber the other's report.

If you prefer, we could use two separate comments, one for each runner; I implemented the single-comment approach to keep the PR thread a bit cleaner.

benchmarks/microbenchmarks/benchmark_attention.py
```diff
@@ -0,0 +1,188 @@
...
+Output: benchmark_attention.csv (written to cwd)
+"""
+
+import torch
+import torch.utils.benchmark as benchmark
+
+import transformer_engine.pytorch as te
+
+# Sweep parameters
+BATCH_SIZE = 2
```
alextmagro: Do we need an additional BS to try or is 2 sufficient for attention?

.github/workflows/rocm-ci.yml
alextmagro: Are we okay with comparing across machines? Especially when our CI pool changes, we might find some differences that aren't as easy to attribute to code change or machine.

benchmarks/microbenchmarks/benchmark_attention.py
```diff
@@ -0,0 +1,188 @@
...
+SEQ_LEN_LIST = [1024, 2048, 4096, 8192]
+
+# (name, num_q_heads, num_kv_heads, head_dim, tp)
+MODEL_CONFIGS = [
+    ("Llama3-8B/TP1",   32,  8, 128,  1),
+    ("Llama3-8B/TP8",   32,  8, 128,  8),
+    ("Llama3-70B/TP8",  64,  8, 128,  8),
+    ("Llama3-405B/TP8", 128, 8, 128,  8),
```
alextmagro: Nit: Let's rename these 3.1 for accuracy

alextmagro (commented):
Hi Matthias, these look good! Are we able to expand this yaml to run benchmarks in C++ or are we limited to python with this setup?

.github/workflows/rocm-ci.yml
ipanfilo: It is very good question. CI machines are not ideal for benchmarking anyway because even if runner has exclusive access to GPU, overall system load may vary significantly. On the other hand, hosts of the same type are expected to be identical

.github/workflows/rocm-ci.yml
ipanfilo: I missed that it writes to PR comment. Why is it needed there?

Micky774:
@ipanfilo @alextmagro I wanted to also mention I have a proposal [implementation of ASV](https://github.com/ROCm/TransformerEngine/pull/487) as an alternative benchmarking approach that might be worth taking a look at as a point of comparison. Matthias and I have talked a bit about our two different approaches and the pros-and-cons of each. I've left a small write-up of some of the benefits in the PR body, and I'll let Matthias present his own opinion for posterity.

In short, I prefer ASV since it provides a lot of important/critical functionality for free (e.g. machine indexing, commit-based regression tracking, static visualization/dashboards, statistical sampling, data storage/parsing), and ensures we don't have a strong maintenance or development burden since it is a widely-adopted OSS project.

benchmarks/microbenchmarks/benchmark_grouped_gemm.py
```diff
@@ -0,0 +1,274 @@
+#!/usr/bin/env python
```
matthiasdiener, line 1: This benchmark fails with:

```console
$ time python bench_grouped_gemm.py

BenchGroupedGemm  (48 combos x 2 methods, 3 warmup, 7 timed)
----------------------------------------------------------------------------------------------------------------------------------------
      median        mean       stdev         q25         q75         min         max      TFLOPS  method                          params
----------------------------------------------------------------------------------------------------------------------------------------
  SKIP  M=512, config=DSV2-Lite_EP32-GateUp  setup failed: GroupedLinear.forward() missing 1 required positional argument: 'm_splits'
  SKIP  M=512, config=DSV2-Lite_EP32-Down  setup failed: GroupedLinear.forward() missing 1 required positional argument: 'm_splits'
  SKIP  M=512, config=DSV2-Lite_EP16-GateUp  setup failed: GroupedLinear.forward() missing 1 required positional argument: 'm_splits'
  SKIP  M=512, config=DSV2-Lite_EP16-Down  setup failed: GroupedLinear.forward() missing 1 required positional argument: 'm_splits'
[...]
```

benchmarks/microbenchmarks/benchmark_gemm.py
```diff
@@ -0,0 +1,82 @@
...
+    for M in M_SIZE_LIST:
+        for case_name, (N, K) in ACTIVE_SHAPES.items():
+            test_cases.append({
+                "Case": case_name,
+                "M": M,
+                "N": N,
+                "K": K,
+                "dtype": torch.bfloat16,
+            })
+    return test_cases
```
Micky774: Can we abstract this to `utils.py` since it is shared with `benchmark_gemm_fp8`

benchmarks/microbenchmarks/benchmark_grouped_gemm.py
```diff
@@ -0,0 +1,215 @@
...
+        return torch.full((b,), m, dtype=torch.int64)
+    else:
+        dist = 0.2 + 0.8 * torch.rand(b)
+        dist /= dist.sum()
+        group_lens = (dist * b * m).to(torch.int64)
+        error = b * m - group_lens.sum()
+        group_lens[-1] += error
+        return group_lens
+
+M_SIZE_LIST = [512, 1024, 2048, 4096]
```
Micky774: Can we either use the same `M_SIZE_LIST` in `utils.py` or rename it to `GROUPED_GEMM_M_SIZE_LIST` and document why it diverges?

benchmarks/microbenchmarks/compare_results.py
```diff
@@ -0,0 +1,135 @@
...
+    for idx in merged.index:
+        row_keys = {k: merged.loc[idx, k] for k in key_cols}
+        row_metrics = {}
+
+        for metric in metric_cols:
+            bc, pc = f"{metric}_base", f"{metric}_pr"
+            bv = merged.loc[idx, bc]
+            pv = merged.loc[idx, pc]
+
+            if pd.isna(bv) or pd.isna(pv) or bv < 0.5:
```
Micky774: What is `bv < 0.5` protecting?

benchmarks/microbenchmarks/benchmark_gemm_fp8.py
Micky774, line 1: Do we really need this to be a separate benchmark entirely, or can we combine with the `bencmhark_gemm.py` and include as e.g. a parameterization option?

benchmarks/microbenchmarks/utils.py
```diff
@@ -0,0 +1,152 @@
...
+def time_func(fn, method="adaptive"):
+    """Time *fn* and return elapsed milliseconds.
+
+    method: "adaptive" uses adaptive_autorange (good for compute-bound),
+            "blocked"  uses blocked_autorange  (good for memory-bound).
+    """
+    timer = benchmark.Timer(stmt="fn()", globals={"fn": fn})
+    if method == "blocked":
+        return timer.blocked_autorange().mean * 1e3
+    return timer.adaptive_autorange().mean * 1e3
```
Micky774: Perhaps we should set `min_run_time` explicitly like NV does?

benchmarks/microbenchmarks/utils.py
```diff
@@ -0,0 +1,152 @@
...
+        Each dict has at least the keys in *param_columns* plus any extra
+        keys the bench_fn needs (passed as **case).
+    bench_fn : callable
+        Called as ``bench_fn(**case)`` and must return a dict whose keys
+        match *metric_columns*.
+    param_columns : list[str]
+        Column names to pull from each test case into the output row.
+    metric_columns : list[str]
+        Column names to pull from the bench_fn return value.
+    default_csv : str or None
```
Micky774, line 288: Can we automatically derive this from the file name of the caller? That way we don't have to remember to manually set the default_csv to what is essentially just the file name.

benchmarks/microbenchmarks/benchmark_grouped_gemm.py
```diff
@@ -0,0 +1,215 @@
...
+
+    return fwd_func_te, bwd_func_te
+
+
+def bench_grouped_gemm(Case, B, M, N, K, dtype):
+    device = "cuda"
+
+    x = torch.randn((B * M, K), dtype=dtype, device=device, requires_grad=True)
+    w = torch.randn((B, N, K), dtype=dtype, device=device, requires_grad=True)
+    group_lens = generate_grouped_gemm_group_lens(B, M, balance=True).to(device)
```
Micky774, line 182: We currently hard-code `balance=True` but I think you've got infrastructure for general `balance={True, False}` so do we want to properly parameterize?

benchmarks/microbenchmarks/compare_results.py
```diff
@@ -0,0 +1,135 @@
...
+
+    if not metric_cols:
+        print("No metric columns found.")
+        return 0
+
+    for col in metric_cols:
+        base_df[col] = pd.to_numeric(base_df[col], errors="coerce")
+        pr_df[col] = pd.to_numeric(pr_df[col], errors="coerce")
+
+    merged = base_df.merge(pr_df, on=key_cols, suffixes=("_base", "_pr"), how="inner")
```
Micky774: `how="inner"` silently drops PR-only rows (new shapes added mid-run). Perhaps we should use `how="outer"` with explicit new/deleted-row handling?

Micky774 (changes requested):
A few general comments in addition to the inline:

1. Regarding copyright, some spots are 2026 only while others are 2025-2026 -- is there a specific reason, or can we be specific and only set 2026?
2. It seems that `dtype=torch.bfloat16` is hard-coded -- can we generalize to allow for e.g. fp16 benchmarks?
3. Can we document the `bench_fn` contract so that it's easier for new developers to contribute additional benchmarks?
4. Can we have a more general `RECIPES` dict similar to NV (https://github.com/ROCm/TransformerEngine/blob/a0b88f4efb1f5076653afb08b412571dfd352013/benchmarks/linear/benchmark_grouped_linear.py#L60-L65)
5. Can we add a `README.md` to document?

benchmarks/microbenchmarks/compare_results.py
matthiasdiener: Renamed to baseline/candidate in d4e116a

.github/workflows/rocm-ci.yml
matthiasdiener: Running the benchmarks in CI has been removed.

.github/workflows/rocm-ci.yml
matthiasdiener: Running the benchmarks in CI has been removed.

benchmarks/microbenchmarks/benchmark_attention.py
matthiasdiener: The attention benchmark has been removed.

benchmarks/microbenchmarks/benchmark_attention.py
matthiasdiener: Renamed in 372e6df

benchmarks/microbenchmarks/utils.py
matthiasdiener: Added in 284addab

benchmarks/microbenchmarks/compare_results.py
matthiasdiener: That threshold is there to avoid reporting noisy speedups when the baseline metric is near zero. I made that more explicit and configurable in 284addab

benchmarks/microbenchmarks/benchmark_grouped_gemm.py
matthiasdiener: Changed in https://github.com/ROCm/TransformerEngine/commit/284addab3df9761fe76620c94b55201c7ce14b51

benchmarks/microbenchmarks/benchmark_gemm_fp8.py
matthiasdiener: I kept it separate for now because it has FP8-specific recipe/autocast setup and a separate output result.

benchmarks/microbenchmarks/utils.py
matthiasdiener: Done in https://github.com/ROCm/TransformerEngine/commit/284addab3df9761fe76620c94b55201c7ce14b51

benchmarks/microbenchmarks/benchmark_grouped_gemm.py
matthiasdiener: I think we should not enable the balance=False support in this PR yet. In the current form, it is not really viable; we would need more data how that imbalance should look like.

benchmarks/microbenchmarks/compare_results.py
matthiasdiener: Done in 284addab

matthiasdiener:
> 1. Regarding copyright, some spots are 2026 only while others are 2025-2026 -- is there a specific reason, or can we be specific and only set 2026?

Changed to 2026 in 284addab.

> 2. It seems that `dtype=torch.bfloat16` is hard-coded -- can we generalize to allow for e.g. fp16 benchmarks?

Done in 284addab.

> 3. Can we document the `bench_fn` contract so that it's easier for new developers to contribute additional benchmarks?

Added documentation in utils.py and README.

> 4. Can we have a more general `RECIPES` dict similar to NV (https://github.com/ROCm/TransformerEngine/blob/a0b88f4efb1f5076653afb08b412571dfd352013/benchmarks/linear/benchmark_grouped_linear.py#L60-L65
>    )

Added in 284addab.

> 5. Can we add a `README.md` to document?

Added in ca1f442e.

benchmarks/microbenchmarks/benchmark_gemm.py
matthiasdiener: Done in https://github.com/ROCm/TransformerEngine/commit/284addab3df9761fe76620c94b55201c7ce14b51

benchmarks/microbenchmarks/benchmark_casting.py
```diff
@@ -0,0 +1,86 @@
...
+CAST_CONFIGS = [
+    # (name, direction, fp8_dtype)
+    ("BF16-to-FP8-E4M3", "quantize",   TE_FP8_E4M3),
+    ("FP8-E4M3-to-BF16", "dequantize", TE_FP8_E4M3),
+    ("BF16-to-FP8-E5M2", "quantize",   TE_FP8_E5M2),
+    ("FP8-E5M2-to-BF16", "dequantize", TE_FP8_E5M2),
+]
+
+
+def _generate_cast_test_cases():
```
wangye805: Emm, we are already in benchmark_casting.py so _generate_test_cases() should suffice? More generally, I saw each benchmark scripts have different functions or names for test case setups. Is it possible to unify them or just follow a unified pattern?

wangye805 (approved):

benchmarks/microbenchmarks/benchmark_casting.py
matthiasdiener: Renamed the cast and norm functions to _generate_test_cases in abefa95f.

benchmarks/microbenchmarks/utils.py
```diff
@@ -0,0 +1,280 @@
...
+def time_func(fn, method="adaptive", min_run_time=DEFAULT_MIN_RUN_TIME_SECONDS):
+    """Time *fn* and return elapsed milliseconds.
+
+    method: "adaptive" uses adaptive_autorange (good for compute-bound),
+            "blocked"  uses blocked_autorange  (good for memory-bound).
+    """
+    timer = benchmark.Timer(stmt="fn()", globals={"fn": fn})
+    if method == "blocked":
+        return timer.blocked_autorange(min_run_time=min_run_time).mean * 1e3
+    return timer.adaptive_autorange(min_run_time=min_run_time).mean * 1e3
```
Micky774: For the csv outputs you record only the means -- I think it would be useful to be able to save the underlying samples and individual runtimes for downstream analysis.

Micky774 (commented):
One final comment, otherwise LGTM.

benchmarks/microbenchmarks/utils.py
matthiasdiener: Thanks for the suggestion. What do you think of https://github.com/ROCm/TransformerEngine/pull/478/commits/39b5720c85790340deb4b215b7192214629eab50, which adds a new argument (--csv-samples), and stores the samples into a separate csv file?

Micky774 (approved):

PR #487, opened 2026-03-16
Microbenchmark suite

Micky774:
# Description

This PR is a port of https://github.com/ROCm/TransformerEngine/pull/478.

This PR uses a central driver to parse and run individual benchmark-defining scripts. The driver provides a function that can be imported and used by the individual scripts to make them self-sufficient and runnable. The benchmarks themselves, and the driver, have no hard ASV dependency. Instead, they simply produce results in an ASV-compatible format for later consumption.

ASV is only used for result tracking, visualization, and publishing. A helper bash script is provided to wrap the ASV commands for convenience (as well as offering a wrapper on the main driver script).

## Follow-up Work
In future PRs we will:
- extend benchmarking to new ops
- re-evaluate bench configs and scope
- update attention benchmarking to reach parity with the JAX FA benchmarking tool (mainly so we have persistent regression tracking)

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

## Changes

- Adds benchmarks
- Adds README.md for documentation
- Adds driver script
- Adds helper bash script to wrap driver and ASV

# Checklist:


Micky774:
Note the CI failure is unrelated

Micky774:
I've added a helper script like @alextmagro had suggested, as well as corresponding documentation to the `README.md`.

.github/workflows/rocm-ci.yml
```diff
@@ -368,6 +368,79 @@ jobs:
           EOF
           )"
 
+      - name: Restore previous ASV results
```
ipanfilo: I think benchmarks should go separate workflow from CI. I.e. these microbenchmarks and ones that are already run with CI

benchmarks/asv/asv.conf.json
```diff
@@ -0,0 +1,16 @@
+{
```
ipanfilo: Does it need to be in root of TE?

.github/workflows/rocm-ci.yml
```diff
@@ -368,6 +368,79 @@ jobs:
...
+        if: github.event_name == 'push' && github.ref_name == 'dev'
+        continue-on-error: true
+        env:
+          RUNNER_NAME: ${{ matrix.runner }}
+        run: |
+          set -ex
+
+          # Derive a stable machine name from the runner label
+          case "${RUNNER_NAME}" in
+            linux-te-mi325*) MACHINE_NAME="mi325" ;;
```
ipanfilo: Why do we need it if results are uploaded with just matrix.runner name?

.github/workflows/rocm-ci.yml
```diff
@@ -368,6 +368,79 @@ jobs:
...
+            linux-te-mi355*) MACHINE_NAME="mi355" ;;
+            *)               MACHINE_NAME="${RUNNER_NAME}" ;;
+          esac
+
+          docker exec -e MACHINE_NAME="$MACHINE_NAME" te-runner bash -c "$(cat <<'OUTER'
+          set -ex
+          pip install asv
+          cd /workspace
+          asv machine --yes --machine "$MACHINE_NAME"
```
ipanfilo: Will it re-register machine if it exists already?

.github/workflows/rocm-ci.yml
Micky774: So, my understanding is that the matrix.runner name is not 1-1 with the underlying system, i.e. different systems with different machine names can be part of a pool with the same runner name. ASV by default stores results by machine name. Here, we are manually specifying a **generic** machine name indexed by gpu arch so that each e.g. mi325 runner will store its results in a compatible way.

Ideally, we have dedicated machines for benchmarking (since this would likely be every commit or nightly even), but that's a constraint we'll need to discuss.

.github/workflows/rocm-ci.yml
Micky774: Yes, but it's registered in the container so it's transient

benchmarks/asv/asv.conf.json
Micky774: No, I've updated it

.github/workflows/rocm-ci.yml
Micky774: Will doing so require a separate TE build and setup? I added it here so that we'd piggy-back off of already running CI.

benchmarks/microbench/run_benchmarks.sh
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bash
+# Helper script for common ASV benchmark tasks.
+set -euo pipefail
+
+cd "$(git rev-parse --show-toplevel)"
```
ipanfilo: (1) On CI it may fail with dubious ownership error. (2) If current directory is not a part of git tree, it fails too. So it is better to determine BENCH_DIR as directory where current script is located

benchmarks/asv/run_benchmarks.sh
```diff
@@ -0,0 +1,69 @@
...
+  compare [REF] [NEW]   Compare two commits (default: HEAD~1 vs HEAD)
+  view                  Generate HTML dashboard and open preview server
+  list                  List available benchmark suites
+
+EOF
+}
+
+case "${1:-}" in
+    setup)
+        MACHINE="${2:-$(hostname)}"
```
ipanfilo: Are hostnames used for CI runners persistent?

benchmarks/asv/driver.py
```diff
@@ -0,0 +1,386 @@
...
+        parser.add_argument("suite", nargs="?", default=None,
+                            help="Benchmark module name (e.g. bench_casting)")
+        parser.add_argument("--all", action="store_true",
+                            help="Run all bench_*.py suites in the directory")
+    parser.add_argument("method_filter", nargs="?", default=None,
+                        help="Only run time_* methods containing this string")
+    parser.add_argument("-w", "--warmup", type=int, default=3,
+                        help="Number of warmup iterations (default: 3)")
+    parser.add_argument("-n", "--iters", type=int, default=7,
+                        help="Number of timed iterations (default: 7)")
```
matthiasdiener: 3/7 iterations is quite low, for microbenchmarks we could probably use much higher numbers (like 50/50) by default.

As an example, for the second shape in bench_gemm I sometimes get `0.127ms` as the median value with 3/7, while 50/50 is pretty consistently at `0.111ms`.

benchmarks/asv/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+```bash
+bash benchmarks/asv/run_benchmarks.sh <command> [options]
+```
+
+| Command | Description |
+|---|---|
+| `setup [name]` | Register machine with ASV (defaults to `hostname`) |
+| `run [suite] [method]` | Run benchmarks in-process (fast, saves ASV-compatible results) |
+| `run --asv [suite]` | Run via ASV subprocess isolation (for CI or statistical rigor) |
```
matthiasdiener: If we do keep the functionality to _run_ benchmarks with asv, can we make this a driver.py option?

benchmarks/asv/driver.py
```diff
@@ -0,0 +1,386 @@
...
+            ci_los.append(ci_lo)
+            ci_his.append(ci_hi)
+            q25s.append(q25)
+            q75s.append(q75)
+            numbers.append(1)
+            repeats.append(iters)
+
+            # Derive throughput from work_* companion
+            work = {}
+            wfn = getattr(instance, "work_" + method_name[5:], None)
```
matthiasdiener: Are the "work"/tflops values stored anywhere? I could only find them printed on stdout via the direct run, but not stored. The work_ methods seems to be unused for asv.

benchmarks/asv/run_benchmarks.sh
```diff
@@ -0,0 +1,69 @@
...
+
+usage() {
+    cat <<EOF
+Usage: bash benchmarks/asv/run_benchmarks.sh <command> [options]
+
+Commands:
+  setup                 Register this machine with ASV
+  run [-w W] [-n N] [SUITE] [METHOD]
+                        Run benchmarks in-process (fast, saves ASV-compatible results)
+  run --asv [SUITE]     Run benchmarks via ASV (subprocess isolation per benchmark)
```
matthiasdiener: The results seem to differ significantly between running with asv and running directly. E.g., the first three shapes in bench_gemm.py return:
```
[62.50%] ··· bench_gemm.BenchGemm.time_forward                                                                          ok
[62.50%] ··· ====== ========================= =============
               M              shape
             ------ ------------------------- -------------
              1024      Llama3-8B_TP1-QKV       160±0.4μs
              1024    Llama3-8B_TP1-AttnOut      129±1μs
              1024     Llama3-8B_TP1-GateUp      460±1μs
```
with asv, and 

```
----------------------------------------------------------------------------------------------------------------------------------------
      median        mean       stdev         q25         q75         min         max      TFLOPS  method                          params
----------------------------------------------------------------------------------------------------------------------------------------
     0.147ms     0.147ms     0.002ms     0.146ms     0.148ms     0.145ms     0.151ms       350.2  time_forward                    M=1024, shape=Llama3-8B_TP1-QKV
     0.120ms     0.120ms     0.007ms     0.112ms     0.130ms     0.110ms     0.130ms       287.1  time_forward                    M=1024, shape=Llama3-8B_TP1-AttnOut
     0.405ms     0.406ms     0.008ms     0.397ms     0.410ms     0.395ms     0.422ms       594.6  time_forward                    M=1024, shape=Llama3-8B_TP1-GateUp
```
with the direct method.

benchmarks/microbench/bench_attention.py
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env python3
```
matthiasdiener, line 1: Instead of creating a new attention microbenchmark, should we use the attention microbenchmark(s) already part of TE (in https://github.com/ROCm/TransformerEngine/tree/dev/benchmarks/attention)?

benchmarks/asv/driver.py
```diff
@@ -0,0 +1,386 @@
...
+        ).decode().strip()
+    except Exception:
+        return "unknown"
+
+
+def _compute_stats(samples):
+    """Return (median, mean, stdev, ci_lo, ci_hi, q25, q75) for *samples*."""
+    s = sorted(samples)
+    n = len(s)
+    mean = sum(s) / n
```
matthiasdiener: Why not use `statistics.median()` etc. here? The median calculated here   (`s[n//2]`) is probably not correct, for even numbers of elements it should be the average of the middle elements.

A similar issue exists for stdev, I think ( `/n` vs. `/(n-1)`).

benchmarks/asv/bench_attention.py
```diff
@@ -0,0 +1,97 @@
...
+
+Sources for model configs:
+  https://huggingface.co/meta-llama/Llama-3.1-8B/blob/main/config.json
+  https://huggingface.co/meta-llama/Llama-3.1-70B/blob/main/config.json
+  https://huggingface.co/meta-llama/Llama-3.1-405B/blob/main/config.json
+  https://huggingface.co/Qwen/Qwen2.5-7B-Instruct/blob/main/config.json
+  https://huggingface.co/Qwen/Qwen2.5-72B-Instruct/blob/main/config.json
+
+Forward FLOPs  = 4 * batch * num_q_heads * seq_len^2 * head_dim
+Backward FLOPs ~ 2x forward
```
matthiasdiener: Repeated from lines 17-19 in this file.

benchmarks/asv/driver.py
```diff
@@ -0,0 +1,386 @@
...
+            "python": sys.executable,
+            "requirements": {},
+            "env_vars": {},
+            "result_columns": [
+                "result", "params", "version",
+                "started_at", "duration",
+                "stats_ci_99_a", "stats_ci_99_b",
+                "stats_q_25", "stats_q_75",
+                "stats_number", "stats_repeat",
+                "samples",
```
matthiasdiener: What's the meaning of samples here? Is it ever written?

benchmarks/asv/driver.py
Micky774: Updated

benchmarks/asv/README.md
Micky774: I've actually now trimmed it so that we only run directly, simplifying the whole thing a bit. ASV is used strictly for publishing/viewing and regression tracking.

wangye805 (approved):

PR #492, opened 2026-03-17
Add fsdp2 fp8 unit tests TE 2.10

sudhu2k:
# Description

This PR adds unit test covering different configurations such as:
1) delayed scaling + fp8 autocast only
2) delayed scaling + fp8 init only (new)
3) delayed scaling + fp8 init + fp8 autocast (new)
4) current scaling + fp8 autocast only
5) current scaling + fp8 init only (new)
6) current scaling + fp8 init + fp8 autocast (new)
7) MXFP8 scaling + fp8 autocast only
8) MXFP8 scaling + fp8 init only (new)
9) MXFP8 scaling + fp8 init + fp8 autocast (new)
10) fp32 (new)

All the unit tests compare FSDP2 vs DDP grads/output.

This PR also cleans up fsdp2_all_gather_tensor to match upstream's methods.
1. Removes keep_fp8_weight_transpose_cache dependency
2. Removes storing module reference to the tensor.

This PR also fixes issue with fused_adam when using it with FSDP2.

Fixes # (https://github.com/ROCm/frameworks-internal/issues/15291)

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:


transformer_engine/pytorch/tensor/fsdp2_allgather_tensor.py
```diff
@@ -123,31 +114,48 @@ def unwrap(x):
...
+        if not module.fp8:
+            return (self._data,), (self._data.requires_grad, module)
         # Use the actual data
         base = self._data
         # Access the quantizer using fp8_meta_index
-        quantizer = self._module.quantizers["scaling_fwd"][self._fp8_meta_index]
-        if not isinstance(quantizer, MXFP8Quantizer) and not self._keep_fp8_weight_transpose_cache:
+        quantizer = module.quantizers["scaling_fwd"][self._fp8_meta_index]
+        if not isinstance(quantizer, MXFP8Quantizer):
             quantizer.set_usage(columnwise=False)
```
sudhu2k, line 160: For FSDP2 with FP8, keep_fp8_weight_transpose_cache should be False. Caching the transposed weight would imply an all-gather of the transposed tensor as well, increasing memory and communication and negating the advantages of FSDP2’s sharded parameter layout.

transformer_engine/pytorch/optimizers/fused_adam.py
```diff
@@ -376,9 +376,9 @@ def _initialize_state(
         """
         dtype = self.name_to_dtype_map[state_name]
         if store_param_remainders:
-            data = torch.zeros(param.shape, dtype=torch.int16, device=param.device)
+            data = torch.zeros_like(param, dtype=torch.int16)
         else:
-            data = torch.empty(param.shape, dtype=dtype, device=param.device)
+            data = torch.empty_like(param, dtype=dtype)
```
sudhu2k: When using FSDP2, parameters are DTensors, and when we do torch.zeros() or torch.empty() we create regular pytorch Tensors.
This was causing 
`[rank1]: RuntimeError: aten.copy_.default: got mixed torch.Tensor and DTensor, need to convert all torch.Tensor to DTensor before calling distributed operators!`

```
[rank7]:   File "/workspace/TransformerEngine/transformer_engine/pytorch/optimizers/fused_adam.py", line 422, in initialize_state
[rank7]:     self.set_scaled_state(param, "master_param", param.clone().detach().float())
[rank7]:   File "/workspace/TransformerEngine/transformer_engine/pytorch/optimizers/fused_adam.py", line 363, in set_scaled_state
[rank7]:     state[state_name].copy_(unscaled_state)
```

Fix:
Keep optimizer state consistent with the parameter type: when parameters are DTensors, state should be DTensors as well. Using torch.empty_like(param, ...) (and the same idea for other state buffers) creates state as a DTensor with the same placement as param, so both sides of copy_ are DTensors and the error is avoided.

tests/pytorch/distributed/run_fsdp2_fp8_model.py
```diff
@@ -253,7 +261,10 @@ def _train(args):
 
         # Zero the parameter gradients
         optimizer.zero_grad()
-        with te.fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):
```
ipanfilo, line 259: Does with te.fp8_autocast(enabled=args.fp8_autocast,.. ) do the same?

tests/pytorch/distributed/test_torch_fsdp2_fp8.py
```diff
@@ -17,15 +17,20 @@
...
-    l1: List[torch.Tensor], l2: List[torch.Tensor]) -> bool:
-    """Ensures two lists are exactly equal."""
+def assert_allclose(
+    l1: List[torch.Tensor], l2: List[torch.Tensor], atol: float, rtol: float = None
+) -> bool:
+    """Ensures two lists are equal."""
     assert len(l1) == len(l2), "Unequal number of outputs."
     for i, (t1, t2) in enumerate(zip(l1, l2)):
-        result = torch.allclose(t1, t2, atol=0, rtol=0)
+        tols = dict(atol=atol)
```
ipanfilo: Move tolls calculation out of the loop

tests/pytorch/distributed/run_fsdp2_fp8_model.py
sudhu2k: It does do the same but since with TEv2.10, te.fp8_autocast is replaced with te.autocast, I've made the change to be consistent.

ipanfilo: So will 'with te.autocast(enabled=args.fp8_autocast, recipe=...)' do the same as if/else?

transformer_engine/pytorch/optimizers/fused_adam.py
ipanfilo: Is it upstream fix cherry-picking?

sudhu2k: Upstream fixes this in TEv2.12, along with few other fixes.
https://github.com/NVIDIA/TransformerEngine/commit/fe8fad59c9add7f8aa841fed2a0b4087b931856f#diff-0801a8d92a56d458946da1439b62e0add1613b7da83d31bc218a852b6b9e42b1
This wasn't cherry picked.

tests/pytorch/distributed/run_fsdp2_fp8_model.py
sudhu2k: Yes, it should. I'll make the changes.

sudhu2k: Done.

ipanfilo (approved):

tests/pytorch/distributed/run_fsdp2_fp8_model.py
```diff
@@ -89,6 +89,9 @@ def _parse_args(argv=None, namespace=None):
     parser.add_argument(
         "--fp8-init", action="store_true", default=False, help="Initialize primary weights in FP8."
     )
+    parser.add_argument(
+        "--fp8-autocast", action="store_true", default=False, help="Enable FP8 autocast."
```
alextmagro: fp8 autocast has been deprecated, so maybe we should rename this to autocast?

tests/pytorch/distributed/test_torch_fsdp2_fp8.py
```diff
@@ -17,49 +17,82 @@
...
+    # scales and produce bit-identical FP8 GEMMs — strict tolerance (0) is used.
+    if fp_init or (not fp_init and not fp8_autocast):
+        atol = 1e-6
+        rtol = 5e-5
     
     for idx, (te_output_no_cache, te_output_cache) in enumerate(zip(output_fsdp, output_dp)):
     
         print(f"Comparing FSDP {te_output_no_cache[0]}, DDP {te_output_cache[0]} at index {idx}...")
-        assertEqual(te_output_no_cache[1], te_output_cache[1]) # expects exact match
+        assert_allclose(te_output_no_cache[1], te_output_cache[1], atol=atol, rtol=rtol) # expects exact match
```
alextmagro: Let's remove the comment regarding exact match here

tests/pytorch/distributed/test_torch_fsdp2_fp8.py
```diff
@@ -70,13 +103,24 @@ def cleanup_artifacts():
...
+# FP8 disabled case (only once)
+test_cases.append((False, False, "delayed"))
+
+
 @pytest.mark.skipif(NUM_PROCS < 4, reason="Requires 4+ GPUs")
 @pytest.mark.skipif(NUM_PROCS % 2 != 0, reason="Requires even number of GPUs")
 @pytest.mark.skipif(not torch_version() >= (2, 4, 0), reason="Requires PyTorch 2.4.0+")
-@pytest.mark.parametrize("fp8_init", ([False]))
-@pytest.mark.parametrize("recipe", (["delayed", "current", "mxfp8"]))
+@pytest.mark.parametrize("fp8_init,fp8_autocast,recipe", test_cases)
```
alextmagro: nit: spacing

transformer_engine/pytorch/tensor/fsdp2_allgather_tensor.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/python3
 # Copyright (c) 2025-2026, Advanced Micro Devices, Inc. All rights reserved.
 # See LICENSE for license information.
-
```
alextmagro, line 4: nit: missing empty line

transformer_engine/pytorch/tensor/fsdp2_allgather_tensor.py
```diff
@@ -123,31 +114,48 @@ def unwrap(x):
...
+                return cls(x, fp8_meta_index=meta)
             return x
 
         out = pytree.tree_map_only(torch.Tensor, rewrap, out)
         return out
 
     # Must return (list_of_tensors_to_all_gather, user_metadata)
-    def fsdp_pre_all_gather(self, mesh):
+    def fsdp_pre_all_gather(self, mesh, orig_size, contiguous_orig_stride, module, mp_policy):
+        """Functions FSDP2 calls before all-gather of the
```
alextmagro: nit: new line after """

transformer_engine/pytorch/tensor/fsdp2_allgather_tensor.py
```diff
@@ -168,14 +176,26 @@ def fsdp_post_all_gather(
         *,
         out: Optional[torch.Tensor] = None,
     ):
-        (requires_grad, ) = metadata
-        if not self._module.fp8:
+        """Functions FSDP2 calls after all-gather of the
```
alextmagro: New line after """

transformer_engine/pytorch/tensor/fsdp2_allgather_tensor.py
```diff
@@ -197,3 +217,13 @@ def fsdp_post_all_gather(
             out._scale_inv = 1 / quantizer.scale
             out._data = data
         return out, all_gather_outputs
+
+    def __reduce_ex__(self, protocol: int) -> tuple:
+        """Custom pickling - unwrap to inner tensor
```
alextmagro: NL

transformer_engine/pytorch/tensor/fsdp2_allgather_tensor.py
```diff
@@ -197,3 +217,13 @@ def fsdp_post_all_gather(
...
+        
+        During checkpointing, save just the underlying high-precision tensor.
+        FSDPAGTensor is a transient wrapper for FSDP2 communication - when the
+        model is loaded and FSDP2 is re-initialized, parameters get wrapped again.
+        """
+        # Delegate to the inner tensor's serialization
+        return self._data.__reduce_ex__(protocol)
```
alextmagro, line 231: possibly missing EOF

alextmagro (approved):

tests/pytorch/distributed/test_torch_fsdp2_fp8.py
```diff
@@ -17,49 +17,82 @@
...
+    # - No FP8 (quantized_init=False, autocast=False): gradient reduction order differs
+    #   (all-reduce vs reduce-scatter), so float non-associativity produces last-bit
+    #   differences in the reduced gradients and updated weights. Hence we use a relaxed tolerance.
+    #
+    # When autocast=True and quantized_init=False, FP8 quantization happens after the
+    # FSDP2 AllGather reconstructs the full weight, so both paths compute identical
+    # scales and produce bit-identical FP8 GEMMs — strict tolerance (0) is used.
+    if quantized_init or (not quantized_init and not autocast):
+        atol = 1e-6
+        rtol = 5e-5
```
wangye805, line 102: If our reference is ddp with the same fp8 primary weight, then the same cast from fp32 master weight to fp8 happens in both target and reference flow. Then we will have exact match?

wangye805 (changes requested):

tests/pytorch/distributed/test_torch_fsdp2_fp8.py
sudhu2k: When quantized_init=True, weights are stored as FP8 parameters. 
After each optimizer step, the FP8-aware Adam kernel (AdamFunctorMaster<fp8_e4m3> in adam.cu) re-quantizes the updated FP32 master weights back to FP8. This re-quantization requires a scale, which is derived from the amax written during the previous optimizer step for delayed scaling.

With FSDP2, each rank only holds a shard of the weight, so the optimizer computes amax = max_abs(master_weight_shard) —> a shard-local value. 

With DDP, each rank holds the full weight, so it computes amax = max_abs(master_weight_full).

Although the shard-local amaxes are all-reduced before computing the next iteration's scale, the scale used for the current re-quantization was already derived from the previous iteration's shard-local amax. Since max_abs(shard) might not be equal to max_abs(full_tensor), the FSDP2 and DDP scales can differ, leading to different FP8 rounding. This compounds over iterations.

With autocast-only (no quantized_init), this doesn't happen because weights remain in BF16/FP32 — the optimizer uses regular Adam with no FP8 re-quantization.

https://github.com/ROCm/TransformerEngine/blob/f6efbbf3b377b9f5338c67ec49f60ef26ab8d494/transformer_engine/common/multi_tensor/adam.cu#L161

https://github.com/ROCm/TransformerEngine/blob/f6efbbf3b377b9f5338c67ec49f60ef26ab8d494/transformer_engine/common/multi_tensor/adam.cu#L170-L181

wangye805: I see. Good catch. 

Then I would consider this a bug in the flow that adam optimizer didn't update the scale inv with the global scale inv (from all-reduced amax). I presume this is also a bug in NV upstream? 

If it's hard to fix, we can keep the atol/rtol for now and leave a TODO for the future

sudhu2k: So I checked this again and I see that there's no issue with the amax update here. It is correct that amax calculation is done only on sharded tensors, but before the scale gets updated all reduce on amax is anyways done in quantization.py.

https://github.com/ROCm/TransformerEngine/blob/f6efbbf3b377b9f5338c67ec49f60ef26ab8d494/transformer_engine/pytorch/quantization.py#L526

Since the amaxes would be same in this case before the scales are computed, this should result in equivalent scales.

The rootcause of the mismatch between DDP and FSDP when using fp8 parameters is because of dequantization and quantization of fp8 parameters.
image

image


In DDP, I noticed 4 dequantization calls plus 4 quantization calls for 4 parameters. If you check the image of the trace here, it seems to happen during broadcast stage. 

When using PyTorch's built-in DistributedDataParallel (DDP) with TransformerEngine's quantized_model_init (FP8 primary weights), DDP's parameter synchronization during initialization causes FP8 weight corruption. 

Specifically, DDP calls _sync_params_and_buffers → _broadcast_coalesced → flatten_dense_tensors → aten::cat to concatenate all model parameters into a single flat buffer for broadcasting from rank 0. However, modules like LayerNormLinear and LayerNormMLP contain a mix of FP8 weight parameters (stored as Float8Tensor with underlying uint8 data) and FP32 LayerNorm parameters (layer_norm_weight, layer_norm_bias). Since aten::cat of 2 different types of paramters is not handled in Float8Tensor.__torch_dispatch__, it falls through to the default path in QuantizedTensor.__torch_dispatch__, which dequantizes all Float8Tensors to float so they can be concatenated with the FP32 parameters.

After broadcast, unflatten_dense_tensors followed by copy_ writes the dequantized float values back into the original Float8Tensor parameters, triggering re-quantization with a freshly computed amax that differs from the original amax set during quantized_model_init. This amax mismatch produces different FP8 scales between DDP and FSDP2 (which just scatters the parameters without dequantization), seeding numerical divergence that propagates through subsequent training iterations. Megatron-LM avoids this entirely by using a custom DDP implementation that groups parameters by dtype — placing FP8 params in a separate uint8 buffer from FP32 params — so mixed-type concatenation never occurs.

wangye805: I got a little bit confused. Based on your description:
1). each amax and scale was set in quantized_model_init? I thought they are set only in reduce_and_update_fp8_tensors during autocast_exit?
2). Say a fp8 primary weight got dequantized to fp32 during this broadcasting, will it change its scale_inv and amax history? 
during the next re-quantization, I understand it will create a fresh amax, but will it use a new scale_inv? If not, then the re-quanitized fp8 will be the same as the old fp8 primary weight, right?

sudhu2k: 1) Correct. amax and scale gets updated in reduce_and_update_fp8_tensors during autocast_exit
2) Initially the scale is 1, so when we quantize fp32 (1) to fp8 (2) and back to fp32 (3) and fp8 (4), I would expect fp8(2) to be same as fp8(4) but the amax calculation when quantizing for fp8(2) and fp8(4) would change, which would throw the upcoming quantizations. With current scaling, fp8(2) won't be the same as fp8(4) since the amax would differ.

wangye805: Synced with Sudharshan about this issue. 

Notice that fp8(2) and fp(8) could also be different in the first iteration since 
fp32(3) = fp8(2)*scale_inv, then saved in fp32 format,
fp8(4) = fp32*scale, saved into fp8 format. Though the scale = 1/scale_inv mathematically, but accumulated truncation error will make fp8(4) different from fp8(2)

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -822,6 +822,31 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+                if src_q is not None and dst_q is not None:
+                    if hasattr(src_q, "amax") and hasattr(dst_q, "amax"):
+                        dst_q.amax.copy_(src_q.amax)
+                    if hasattr(src_q, "scale") and hasattr(dst_q, "scale"):
+                        dst_q.scale.copy_(src_q.scale)
+                return dst
+            # DDP broadcast path: _broadcast_coalesced dequantizes Float8Tensors
+            # (via aten::cat fallback) then copies the plain tensor back.
+            # Re-quantize while preserving the original quantizer state.
+            if isinstance(dst, Float8Tensor) and not isinstance(src, Float8Tensor):
```
sudhu2k: When PyTorch's native DDP broadcasts parameters, _broadcast_coalesced concatenates all module parameters (FP8 weights and FP32 biases) via aten::cat, which triggers Float8Tensor's fallback dispatch path and dequantizes the FP8 weights to FP32. After the broadcast, aten::copy_ writes the FP32 data back into the Float8Tensor parameter, calling quantize_() which recomputes amax and scale from the dequantized values — but these differ slightly from the originals due to FP8 round-trip error, causing numerical divergence from FSDP2. The fix intercepts this plain-tensor-to-Float8Tensor copy_ path, saves the quantizer's amax and scale before re-quantization, and restores them afterward, so the DDP broadcast becomes a no-op with respect to quantizer state. This makes the test pass for a model with layernormMLP, LayernormLinear and Linear module with 0 tolerance.

PR #497, opened 2026-03-19
MXFP4 cast tranpose hadamard shuffle fused hip [WIP]

sarthak-amd:


sarthak-amd:
Will create a separate PR with full integration

PR #505, opened 2026-03-25
NVFP4 dequantization

aris134:
# Description

Fixes https://github.com/ROCm/frameworks-internal/issues/15998

Enable NVFP4 dequantization on AMD GPU (gfx950) and add unit test.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Enable compilation of NVFP4 dequantization kernel for AMD GPU
- Add unit test that verifies NVFP4 dequantization works on gfx950

# Checklist:


tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,237 @@
...
+                    fp8_dis);
+
+    auto err = hipMemcpy(input.rowwise_dptr(), host_input.get(), data_bytes, hipMemcpyHostToDevice);
+    ASSERT_EQ(err, hipSuccess) << hipGetErrorString(err);
+
+    err = hipMemcpy(input.rowwise_scale_inv_dptr(), host_scales.get(), scale_bytes, hipMemcpyHostToDevice);
+    ASSERT_EQ(err, hipSuccess) << hipGetErrorString(err);
+
+    const float amax = 1.0f;
+    input.set_tensor_amax(amax);
```
matthiasdiener: set_scale() instead?

ipanfilo (commented):
It is based on PR#472. Not to review the same changes twice let's wait for that PR to merge

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,235 @@
...
+                rows,
+                cols,
+                scale_stride);
+
+    auto [atol, rtol] = getTolerances(otype);
+    compareResults("output", output, ref_output.get(), true, atol, rtol);
+}
+
+std::vector<std::pair<size_t, size_t>> tensor_dims = {
+    {32, 32},
```
wangye805, line 215: Like mxfp8, NV fp4 has its own scale_inv layout agreement for rowwise/colwise data: https://github.com/ROCm/TransformerEngine/blob/98ccd2ebcf16fc8d24a3d3ebe5cc0ed7f42509bd/tests/cpp/test_common.h#L348 https://github.com/ROCm/TransformerEngine/blob/98ccd2ebcf16fc8d24a3d3ebe5cc0ed7f42509bd/transformer_engine/common/common.h#L783

Take tensor dim {32,32} as an example, the rowwise scale inv will not be a continuous array for the first and the second row because nvfp4_scale_tensor_alignment_Y_rowwise=128, so padding is needed from 32/16=2 to 128 per row

tests/cpp/operator/test_dequantize_nvfp4.cu
wangye805: Yeah, I think for dequantization, the scale is needed

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,235 @@
...
+        std::make_unique<fp8e4m3[]>(scale_numel);
+    std::unique_ptr<OutputType[]> ref_output =
+        std::make_unique<OutputType[]>(rows * cols);
+
+    static std::mt19937 gen(42);
+    std::uniform_int_distribution<int> fp4_dis(0, 15);
+    std::uniform_int_distribution<int> fp8_dis(0, 255);
+
+    generate_data(host_input.get(), rows, cols, gen, fp4_dis);
+    generate_scales(host_scales.get(),
```
wangye805: According to the layout alignment requirement, the data and scale for nvfp4 are not continuous in memory. Probably we can reuse the nvfp4 quantization here to generate a valid nvfp4 tensor

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,235 @@
...
+void performTest(const size_t rows, const size_t cols, DType otype) {
+    constexpr size_t block_size_1d = 16;
+    ASSERT_EQ(cols % block_size_1d, 0u);
+    ASSERT_EQ(cols % 2, 0u);
+
+    const DType itype = DType::kFloat4E2M1;
+    const size_t blocks_per_row = cols / block_size_1d;
+
+    Tensor input("input", std::vector<size_t>{rows, cols}, itype,
+                 true, false, NVTE_NVFP4_1D_SCALING);
```
wangye805: Try to also test with 2D scaling, and with columnwise data

wangye805 (changes requested):

tests/cpp/operator/test_dequantize_nvfp4.cu
aris134: This leads to memory fault run-time error, whereas my current method (set_tensor_amax) works fine. Leaving as is for now.

tests/cpp/operator/test_dequantize_nvfp4.cu
aris134: scale tensor alignment fixed in 4475cf8

tests/cpp/operator/test_dequantize_nvfp4.cu
aris134: Tests added in 4475cf8

tests/cpp/operator/test_dequantize_nvfp4.cu
aris134: Adapted in 4475cf8

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
```diff
@@ -157,12 +157,25 @@ static_assert(kNumThreadsLoad <= kThreadsPerWarp, "kNumThreadsLoad must be <= kT
...
 
 // max for every group_size elements in warp
 template <int group_size, int shfl_down_stride>
-__device__ __forceinline__ float groupMax(float val, unsigned int groupMask) {
+__device__ __forceinline__ float groupMax(float val,
+#ifdef __HIP_PLATFORM_AMD__
+                                          uint64_t groupMask) {
+#else
+                                          unsigned int groupMask) {
+#endif
```
matthiasdiener: I think the changes in this file are due to a merge error and should not be necessary.

tests/cpp/operator/test_dequantize_nvfp4.cu
matthiasdiener: Please double-check. Quantization does not need amax, dequant should not have it either.

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,410 @@
...
+
+namespace {
+
+constexpr size_t kFP4BlockSize1D = 16;
+constexpr size_t kFP4BlockSize2DY = 16;
+constexpr size_t kFP4BlockSize2DX = 16;
+
+size_t divide_round_up(size_t x, size_t y) {
+    return (x + y - 1) / y;
+}
```
matthiasdiener: Isn't this part of test_common.h?

matthiasdiener (changes requested):

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,410 @@
...
+                const uint8_t bits = static_cast<uint8_t>(dis(gen));
+
+                fp8e4m3 candidate;
+                std::memcpy(&candidate, &bits, sizeof(bits));
+
+                const float decoded = static_cast<float>(candidate);
+                if (std::isfinite(decoded)) {
+                    scale_buffer[idx] = candidate;
+                    break;
+                }
```
wangye805: This section of codes generating a valid fp8e4m3 are reused in the 2d scale as well, let's consolidate them to avoid maintaining duplicated copies

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,410 @@
...
+                     std::mt19937& gen,
+                     std::uniform_int_distribution<int>& dis) {
+    const size_t total_elems = mathematical_rows * physical_row_stride;
+    std::memset(scale_buffer, 0, total_elems * sizeof(fp8e4m3));
+
+    for (size_t row = 0; row < mathematical_rows; ++row) {
+        for (size_t block = 0; block < mathematical_blocks_per_row; ++block) {
+            const size_t idx = row * physical_row_stride + block;
+
+            while (true) {
```
wangye805: By the way, is there a way to generate fp8e4m3 without using a while loop try? I understand with multiple-tryout, we will finally find a non-infinite fp8e4m3, but it's a little bit waste for the random seed and execution time. Fp8e4m3 is well documented, probably we can study which bit-patterns give non-infinite values?

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,410 @@
...
+                const uint8_t bits = static_cast<uint8_t>(dis(gen));
+
+                fp8e4m3 candidate;
+                std::memcpy(&candidate, &bits, sizeof(bits));
+
+                const float decoded = static_cast<float>(candidate);
+                if (std::isfinite(decoded)) {
```
wangye805: Scales also need to be non-negative, right?

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,410 @@
...
+                                gen,
+                                fp4_dis);
+
+    // Row-wise 1D scales on [rows, cols]
+    generate_1d_scales(host_scales_rowwise_1d.get(),
+                    unpadded_blocks_Y,
+                    unpadded_blocks_X,
+                    scales_stride,
+                    gen,
+                    fp8_dis);
```
wangye805: alignment?

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,410 @@
...
+
+    // Col-wise 1D scales on [cols, rows]
+    generate_1d_scales(host_scales_colwise_1d.get(),
+                    unpadded_blocks_Y_t,
+                    unpadded_blocks_X_t,
+                    scales_stride_t,
+                    gen,
+                    fp8_dis);
```
wangye805: alignment?

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,410 @@
...
+// Generates random FP8 (E4M3) scale values by sampling raw 8-bit patterns
+// and rejects non-finite values (i.e., NaN) before storing.
+// Values are written using memcpy to preserve exact
+// bit patterns rather than relying on numeric conversion.
+void generate_1d_scales(fp8e4m3* scale_buffer,
+                     const size_t mathematical_rows,
+                     const size_t mathematical_blocks_per_row,
+                     const size_t physical_row_stride,
+                     std::mt19937& gen,
+                     std::uniform_int_distribution<int>& dis) {
```
wangye805: alignment?

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,410 @@
...
+}
+
+// Generates random FP8 (E4M3) scale values by sampling raw 8-bit patterns
+// and rejects non-finite values (i.e., NaN) before storing.
+// Values are written using memcpy to preserve exact
+// bit patterns rather than relying on numeric conversion.
+void generate_1d_scales(fp8e4m3* scale_buffer,
+                     const size_t mathematical_rows,
+                     const size_t mathematical_blocks_per_row,
+                     const size_t physical_row_stride,
```
wangye805: You can reuse the existing names (unpadded_blocks_Y, unpadded_blocks_X, and scales_stride)

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,410 @@
...
+            set_fp4_nibble(data, idx, nibble);
+
+            const size_t idx_t = j * rows + i;
+            set_fp4_nibble(data_t, idx_t, nibble);
+        }
+    }
+}
+
+// Decode a single FP4 (E2M1) value from packed storage.
+float get_fp4_value(const fp4e2m1* data, const size_t mathematical_idx) {
```
wangye805: Only the scale has the padding/alignment distinction, mathemtical (unpadded) idx vs padded index. I recall for rowwise /columnwise data, we don't have this padding issue? If so, we can mathematical_idx -> idx

tests/cpp/test_common.h
```diff
@@ -301,6 +301,21 @@ class Tensor {
...
+    } else {
+      *amax_cpu_data_ = amax;
+    }
+
+    float *amax_gpu = nullptr;
+    NVTE_CHECK_CUDA(cudaMalloc(&amax_gpu, sizeof(float)));
+    NVTE_CHECK_CUDA(cudaMemcpy(amax_gpu, amax_cpu_data_.get(),
+                              sizeof(float), cudaMemcpyHostToDevice));
+
+    tensor_.set_amax(amax_gpu, DType::kFloat32, tensor_.defaultShape);
```
wangye805: use from_cpu() https://github.com/ROCm/TransformerEngine/blob/1d0a70eb69ff84bdfd4f32f308e6b4873a1639bd/tests/cpp/test_common.cu#L481

tests/cpp/test_common.h
```diff
@@ -301,6 +301,21 @@ class Tensor {
     tensor_.set_amax(nullptr, DType::kFloat32, tensor_.defaultShape);
   }
 
+  void set_tensor_amax(float amax) {
```
wangye805: Guard our rocm specific changes by macro

tests/cpp/test_common.h
```diff
@@ -356,6 +371,11 @@ constexpr size_t scale_tensor_alignment_Y_colwise = 4;
 constexpr size_t scale_tensor_alignment_X_colwise = 128;
 #endif
 
+static constexpr float E2M1_LUT[16] = {
```
wangye805, line 364: rocm specific contents

wangye805 (changes requested):

tests/cpp/test_common.h
aris134: Replaced with set_scale so no longer needed. See d6837f6

tests/cpp/test_common.h
aris134: Done in d6837f6

tests/cpp/test_common.h
aris134: Replaced set_tensor_amax with pre-existing set_scale so no longer needed. See d6837f6

tests/cpp/operator/test_dequantize_nvfp4.cu
aris134: Agreed, done in d6837f6

tests/cpp/operator/test_dequantize_nvfp4.cu
aris134: Done in d6837f6

tests/cpp/operator/test_dequantize_nvfp4.cu
aris134: Done. d6837f6

tests/cpp/operator/test_dequantize_nvfp4.cu
aris134: Done. d6837f6

tests/cpp/operator/test_dequantize_nvfp4.cu
aris134: Yes. Done in d6837f6

tests/cpp/operator/test_dequantize_nvfp4.cu
aris134: Yes, I found a more efficient way to do it without the while loop. See d6837f6

tests/cpp/operator/test_dequantize_nvfp4.cu
aris134: Removing the while loop removed most of the duplication. See d6837f6

tests/cpp/operator/test_dequantize_nvfp4.cu
aris134: Yes, removed in d6837f6

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
aris134: Yes, patched in d3f8dd1

tests/cpp/operator/test_dequantize_nvfp4.cu
aris134: Handled in d6837f6

transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh
```diff
@@ -57,7 +57,7 @@ __global__ void __launch_bounds__(512)
   fp4vec value;
   value.vec = input_vectorized[my_index];
   fp8e4m3 scale = scales[my_scale_index];
-  float amax = *tensor_amax;
+  float amax = (tensor_amax == nullptr) ? 1.0f : *tensor_amax;
```
matthiasdiener: I think this change needs an #ifdef. 

Also, please add a comment here on why this change is needed (e.g., does upstream have the same issue or solve it in another way?)

aris134: Done in 25d72ff

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,386 @@
...
+                                cols,
+                                blocks_Y,
+                                blocks_X,
+                                scales_stride,
+                                amax,
+                                otype);
+}
+
+std::vector<std::pair<size_t, size_t>> tensor_dims = {
+    {32, 32},
```
matthiasdiener, line 215: Can you also test shapes that are not multiples of 16, like {33, 47},
{129, 65}?

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,386 @@
...
+    auto [atol, rtol] = getTolerances(otype);
+    compareResults(case_name, output, ref_output.get(), true, atol, rtol);
+}
+
+// End-to-end test: generate random FP4 input and FP8 scales, then exercise
+// 1) row-wise 1D dequant
+// 2) col-wise 1D dequant (by running the same dequant kernel on transposed data)
+// 3) 2D dequant semantics using row-wise replicated scales
+template <typename OutputType>
+void performTest(const size_t rows, const size_t cols, DType otype) {
```
matthiasdiener, line 166: Can we also test with:
- amax != 1.0
- shapes that are not multiples of 16
?

aris134: I believe it is a requirement for the flattened 2D first and last dimensions to be a multiple of 32. 

See in get_scales: https://github.com/ROCm/TransformerEngine/blob/f6efbbf3b377b9f5338c67ec49f60ef26ab8d494/tests/cpp/test_common.cu#L177-L186

the above function is called in 
https://github.com/ROCm/TransformerEngine/blob/f6efbbf3b377b9f5338c67ec49f60ef26ab8d494/tests/cpp/test_common.cu#L434

The 2D flattened dims for all test cases in test_cast_nvfp4_transpose.cu are in fact multiples of 32:

```
{32, 32}           -> first_dim = 32,    last_dim = 32
{32, 64}           -> first_dim = 32,    last_dim = 64
{64, 32}           -> first_dim = 64,    last_dim = 32
{64, 96}           -> first_dim = 64,    last_dim = 96
{128, 128}         -> first_dim = 128,   last_dim = 128
{256, 256}         -> first_dim = 256,   last_dim = 256
{512, 512}         -> first_dim = 512,   last_dim = 512
{1024, 1024}       -> first_dim = 1024,  last_dim = 1024
{2048, 2048}       -> first_dim = 2048,  last_dim = 2048
{128, 256}         -> first_dim = 128,   last_dim = 256
{8192, 128}        -> first_dim = 8192,  last_dim = 128
{2048, 160}        -> first_dim = 2048,  last_dim = 160
{8, 32, 1024}      -> first_dim = 256,   last_dim = 1024
{16, 8, 4, 512}    -> first_dim = 512,   last_dim = 512
{1024, 16384}      -> first_dim = 1024,  last_dim = 16384
{4096, 13312}      -> first_dim = 4096,  last_dim = 13312
```

Thus, running a non-multiple of 32 for a tensor shape trips this error:
`C++ exception with description "/nvfp4_dequant_work/TransformerEngine/tests/cpp/test_common.hip:188 in function get_scales: Assertion failed: first_dim % 32 == 0. " thrown in the test body.`

aris134: In theory, it would be useful to test amax != 1.0, but with the current test_common.h helper there is no way as far as I am aware to explicitly populate a separate global amax buffer. The set_scale() call used in the quant test does not appear to change that, since it is wired for delayed scaling in the helper, so that path also ends up exercising the nullptr fallback with amax = 1.0.

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,387 @@
...
+// even mathematical index -> low nibble, odd mathematical index -> high nibble.
+void set_fp4_nibble(fp4e2m1* data, const size_t mathematical_idx, const uint8_t nibble) {
+    ASSERT_TRUE(nibble < 16);
+    auto* raw = reinterpret_cast<uint8_t*>(data);
+    const size_t byte_idx = mathematical_idx / 2;
+    const uint8_t val = nibble;
+
+    if ((mathematical_idx % 2) == 0) {
+        // set low nibble
+        raw[byte_idx] = static_cast<uint8_t>((raw[byte_idx] & 0xF0) | val);
```
ipanfilo: Why does it need cast here?

aris134: It was not needed for correctness, so I went ahead and removed it in 92de4bd

transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh
matthiasdiener, line 1: Needs AMD copyright

transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh
```diff
@@ -57,7 +57,16 @@ __global__ void __launch_bounds__(512)
...
+  // NVFP4 may reach this path with scale present but no separate amax buffer.
+  // Use 1.0f as the neutral fallback when tensor_amax is not provided on HIP.
+  float amax = 1.0f;
+#ifndef __HIP_PLATFORM_AMD__
+  amax = *tensor_amax;
+#else
+  if (tensor_amax != nullptr) {
+    amax = *tensor_amax;
+  }
+#endif
```
matthiasdiener: (just to maintain the exact line in the #ifdef)

```suggestion
#ifndef __HIP_PLATFORM_AMD__
  float amax = *tensor_amax;
#else
  float amax = 1.0f;
  if (tensor_amax != nullptr) {
    amax = *tensor_amax;
  }
#endif
```

aris134: Done in 458803f

transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh
aris134: Done in 458803f

tests/cpp/operator/test_dequantize_nvfp4.cu
matthiasdiener: I think that's fine. Thanks for checking.

matthiasdiener (approved):

transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh
```diff
@@ -57,7 +59,16 @@ __global__ void __launch_bounds__(512)
   fp4vec value;
   value.vec = input_vectorized[my_index];
   fp8e4m3 scale = scales[my_scale_index];
+  // NVFP4 may reach this path with scale present but no separate amax buffer.
+  // Use 1.0f as the neutral fallback when tensor_amax is not provided on HIP.
+#ifndef __HIP_PLATFORM_AMD__
   float amax = *tensor_amax;
+#else
+  float amax = 1.0f;
```
ipanfilo: nit: it can be initialized with ternary operator (tensor_amax != nullptr) ? *tensor_amax : 1.0f

ipanfilo (approved):

tests/cpp/operator/test_dequantize_nvfp4.cu
wangye805: For scales or input tensors? I didn't recall nvfp4 supports input tensors not 16 aligned in the blockwise dimension. For example, (33, 48) is okay for rowwise, but not (33, 47)? Better double check. Thanks

aris134: I believe I addressed this comment in another response pictured here:

Screenshot 2026-04-10 154012

transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh
aris134: Done in dfc0028

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,387 @@
...
+                output[idx] = static_cast<OutputType>(x * scale);
+            }
+        }
+    }
+}
+
+template <typename OutputType>
+void run_single_case(const std::string& case_name,
+                     const fp4e2m1* host_input,
+                     const fp8e4m3* host_scales,
```
wangye805: Just realized that you are copying the input data/scale generated outside into the test tensors. In fact, test Tensor contains both cpu and gpu data ptrs: https://github.com/ROCm/TransformerEngine/blob/098115728f7eb1838b8e750eb54e8ae2dcc7bdcd/tests/cpp/test_common.h#L327-L337. (Probably the std::mt19937 can also be reused in your nvfp4 data/scale random generation). 

Your current way is okay since you reuse them across three checks to save random number generation time. But according to my analysis below regarding NV upstream dequantization kernel only takes the rowwise input/scale, probably this reuse is not needed.

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,387 @@
...
+                          host_input,
+                          data_bytes,
+                          cudaMemcpyHostToDevice);
+    ASSERT_EQ(err, cudaSuccess) << case_name << ": " << cudaGetErrorString(err);
+
+    err = cudaMemcpy(input.rowwise_scale_inv_dptr(),
+                     host_scales,
+                     scale_bytes,
+                     cudaMemcpyHostToDevice);
+    ASSERT_EQ(err, cudaSuccess) << case_name << ": " << cudaGetErrorString(err);
```
wangye805: you can reuse the from_cpu() https://github.com/ROCm/TransformerEngine/blob/098115728f7eb1838b8e750eb54e8ae2dcc7bdcd/tests/cpp/test_common.h#L309 util

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,387 @@
...
+    run_single_case<OutputType>("colwise_1d_dequant",
+                                host_input_t.get(),
+                                host_scales_colwise_1d.get(),
+                                cols,
+                                rows,
+                                blocks_Y_t,
+                                blocks_X_t,
+                                scales_stride_t,
+                                amax,
+                                otype);
```
wangye805: Technically this isn't testing the dequantization from the columnwise data/scale pointers. You feed the transposed input/scale still to the rowwise input/scale in the input Tensor inside run_single_case. Probably not needed here

This is not our problem, since their dequantization kernel (from nv upstream) seems only taking input and scale from rowwise https://github.com/ROCm/TransformerEngine/blob/516f565fa3cbcca09b307a2408beb6779815dd61/transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh#L110-L114

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,387 @@
...
+    run_single_case<OutputType>("replicated_2d_dequant",
+                                host_input.get(),
+                                host_scales_2d_replicated.get(),
+                                rows,
+                                cols,
+                                blocks_Y,
+                                blocks_X,
+                                scales_stride,
+                                amax,
+                                otype);
```
wangye805: Duplication for the same/similar reasons. This single case is different from the "rowwise_1d_dequant" only in terms of the scales. Essentially it's still testing the dequantizing from rowwise data and scale in a special format.

wangye805 (changes requested):

tests/cpp/operator/test_dequantize_nvfp4.cu
wangye805: Input data/scale generation can happen inside [Tensor::fillUniform](https://github.com/ROCm/TransformerEngine/blob/098115728f7eb1838b8e750eb54e8ae2dcc7bdcd/tests/cpp/test_common.cu#L1100) or a new dedicated method for fp4 tensors

aris134: I kept this as dedicated NVFP4-specific generation in a15cd1e rather than using fillUniform(), since the test is specifically exercising the NVFP4 dequant kernel. That also matches the style used in test_dequantize_mxfp8.cu, which uses dedicated helpers for input and scale generation instead of the generic fill utility.

aris134: removed external host vectors in a15cd1e. Now uses the build-in cpu and gpu data ptrs from the test tensor class

tests/cpp/operator/test_dequantize_nvfp4.cu
aris134: Done in a15cd1e

tests/cpp/operator/test_dequantize_nvfp4.cu
aris134: Done in a15cd1e

tests/cpp/operator/test_dequantize_nvfp4.cu
aris134: Done in a15cd1e

wangye805 (approved):

PR #506, opened 2026-03-26
Full MXFP4 Training Recipe

sarthak-amd:
1. Introduces `cast_transpose_mxfp4_fused_shuffle`  . The existing Triton based MXFP4 Cast Transpose does not fused shuffle and hadamard.  Replaced `te_quantize_triton` path in `MXFP4Quantizer.update_quantized()` with fully fused hip path. Using the Triton path for training would required an unfused hadamard call for stable training 

2. `fp4_gemm_handler` that dispatches a4w4 gemm calls to the AITER backend with the right layouts expeceted by AITER. 

3. Add MXFP4 weight caching in Linear.forward() and LayerNormLinear.forward() that persists quantized MXFP4TensorStorage weights across forward passes, 

Training recipe can be enabled using 
```
export FP4_RECIPE=mxfp4
export NVTE_MXFP4_USE_HADAMARD=1
```

- [ ] remove GEMM Shape logger from fp4_gemm_handler
- [ ] Add Unit Tests


build_tools/pytorch.py
```diff
@@ -38,8 +38,11 @@ def setup_pytorch_extension(
 ) -> setuptools.Extension:
     """Setup CUDA extension for PyTorch support"""
 
-    # Source files
+    # Source files - include both .cpp and .cu files
+    # .cu files will be hipified to .hip for ROCm builds
     sources = all_files_in_dir(Path(csrc_source_files), name_extension="cpp")
+    cu_sources = all_files_in_dir(Path(csrc_source_files), name_extension="cu")
```
ipanfilo: GPU code should live in the main TE not in extensions.

ipanfilo (changes requested):

wangye805 (changes requested):
Below is the list of required unitests (cpp level gtest or pytorch pytests):
1). cpp level gtest for cast_transpose_mxfp4_shuffled. Can refer to our nvfp4 gtest (https://github.com/ROCm/TransformerEngine/blob/dev/tests/cpp/operator/test_cast_nvfp4_transpose.cu) as an example
2). pytorch pytest for aiter a4w4 gemm. You can refer to this fp8gemm in test_numerics as an example: https://github.com/ROCm/TransformerEngine/blob/82617fea8380d4fd4ec7145ae952bcc28af9f56a/tests/pytorch/test_numerics.py#L3025
3). extend the test_numerics with test_layernorm_linear_* (for example, https://github.com/ROCm/TransformerEngine/blob/82617fea8380d4fd4ec7145ae952bcc28af9f56a/tests/pytorch/test_numerics.py#L1618) and test_linear_*(https://github.com/ROCm/TransformerEngine/blob/82617fea8380d4fd4ec7145ae952bcc28af9f56a/tests/pytorch/test_numerics.py#L1302) with mxfp4 format

build_tools/pytorch.py
sarthak-amd: Done.

PR #507, opened 2026-03-26
Mxfp8 cast optimization

alextmagro:
# Description

Improves performance of HIP MXFP8 Cast Kernels -- see [confluence page](https://amd.atlassian.net/wiki/spaces/~712020f71796901d2a44b39d0a65dcc0c69ce7/pages/1563878722/MXFP8+Cast+Kernel+Optimization) for details
Adds benchmarking scripts for reproducability 

A potential bug in scale updating will be fixed in https://github.com/ROCm/TransformerEngine/pull/519
image


benchmarks/cpp/cast/bench_dequantize_mxfp8.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+#include <benchmark/benchmark.h>
+#include <hip/hip_runtime.h>
+#include <hip/hip_fp16.h>
+#include <hip/hip_bfloat16.h>
+#include "amd_detail/hip_float8.h"
+
+#include "benchmark_utils.h"
+
+#include "amd_detail/hip_float8.h"
```
ipanfilo: Dup of line 11

benchmarks/cpp/cast/bench_dequantize_mxfp8.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+#include <transformer_engine/cast_hip.h>
```
ipanfilo: Is this file hipified?

benchmarks/cpp/utils/test_common.cpp
```diff
@@ -0,0 +1,1241 @@
+// !!! This is a file automatically generated by hipify!!!
```
ipanfilo: Please add original file, not hipified

transformer_engine/common/common.h
```diff
@@ -736,6 +736,21 @@ struct TypeInfo {
     }                                                               \
   }
 
+#define TRANSFORMER_ENGINE_CHUNK_DIM_SWITCH(use_large, CHUNK_Y, CHUNK_X, THREADS, ...) \
+  [&] {                                                                                \
```
ipanfilo: Those values are selected for specific kernel, aren't they? The define can bereplaced with TRANSFORMER_ENGINE_SWITCH_CONDITION

ipanfilo (changes requested):

benchmarks/cpp/cast/bench_dequantize_mxfp8.cpp
alextmagro: It is written in hip, but references the hipified headers in te. Would you prefer we write this in CUDA then hipify it?

benchmarks/cpp/utils/test_common.cpp
alextmagro: Whoops. This file was copied over from the test/cpp directory to use the tensor capabilities, but forgot to clean it up to just keep what we need for the benchmarks. I will work on this, and push in a later commit.

transformer_engine/common/common.h
alextmagro: Good idea, will make that change.

benchmarks/cpp/utils/test_common.cpp
alextmagro: Regarding this, made a design change that's hopefully cleaner. I use the test_common files directly, but added a macro that ensures we only compile what we need within the executable.

benchmarks/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
...
+endif()
+
+set(COMMON_COMPILE_OPTIONS
+  -Wall
+  -Wextra
+  -O3
+  -DNDEBUG
+  -DUSE_ROCM
+  --offload-arch=${GPU_TARGETS}
+  -w
```
matthiasdiener: Why the -w here?

benchmarks/cpp/run_benchmarks.sh
```diff
@@ -0,0 +1,122 @@
+#!/bin/bash
+# Builds benchmarks, runs them, and consolidates results into a single CSV
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+BUILD_DIR="${SCRIPT_DIR}/build"
+RESULTS_DIR="${SCRIPT_DIR}/results"
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
```
matthiasdiener: I'd recommend not using colors here, these might not show up well in e.g. CI logs.

benchmarks/cpp/cast/bench_gated_mxfp8.cpp
```diff
@@ -0,0 +1,212 @@
...
+    state.SetIterationTime(ms / 1000.0);
+  }
+
+  HIP_CHECK(hipEventDestroy(start));
+  HIP_CHECK(hipEventDestroy(stop));
+
+  const size_t bytes_write_data   = rows * output_cols * sizeof(OType) *
+                                   ((USE_ROWWISE ?: 0) + (USE_COLWISE ?: 0));
+  const size_t bytes_write_scales = (USE_ROWWISE ? rows * scale_cols_row : 0) +
+                                     (USE_COLWISE ? scale_rows_col * scale_cols_col : 0);
```
matthiasdiener, line 168: Does this need a dtype?

alextmagro: Scales are E8M0, so 1 byte. We don't have that type defined in these benchmarks, so skipped the dtype. I added a comment to that extent.

benchmarks/cpp/CMakeLists.txt
alextmagro: during debugging, I found the warnings to be clutter, especially since they came from TE rather than these benchmarks. I have removed the -w now that things are working.

benchmarks/cpp/run_benchmarks.sh
alextmagro: Good point, will remove.

benchmarks/cpp/run_benchmarks.sh
```diff
@@ -5,18 +5,13 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
...
-NC='\033[0m'
-
 setup_test_common_symlinks() {
     local utils_dir="${SCRIPT_DIR}/utils"
     local test_common_hip="../../tests/cpp/test_common.hip"
     local test_common_h="../../tests/cpp/test_common_hip.h"
 
     if [ ! -f "${SCRIPT_DIR}/${test_common_hip}" ] || [ ! -f "${SCRIPT_DIR}/${test_common_h}" ]; then
-        echo -e "${RED}Error: hipified test_common files not found. Build tests before running benchmarks."
+        echo -e "Error: hipified test_common files not found. Build tests before running benchmarks."
```
matthiasdiener: Many of these echos don't need -e anymore. I'd also suggest removing the non-ASCII symbols like "✓" and "✗" - these don't always render well in terminals.

benchmarks/cpp/cast/bench_dequantize_mxfp8.cpp
ipanfilo: It is OK to run it directly with HIP. The question is if hipify is called on it - from the headers it seems it is not. BWT, why <> notation is used for TE headers

benchmarks/cpp/run_benchmarks.sh
```diff
@@ -5,27 +5,47 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
...
+    local utils_dir="${SCRIPT_DIR}/utils"
+    local test_common_hip="../../tests/cpp/test_common.hip"
+    local test_common_h="../../tests/cpp/test_common_hip.h"
+
+    if [ ! -f "${SCRIPT_DIR}/${test_common_hip}" ] || [ ! -f "${SCRIPT_DIR}/${test_common_h}" ]; then
+        echo -e "Error: hipified test_common files not found. Build tests before running benchmarks."
+        return 1
+    fi
+
+    if [ ! -L "${utils_dir}/test_common.hip" ] || [ ! -e "${utils_dir}/test_common.hip" ]; then
```
ipanfilo: Why cannot include test_common.cu and test_common.h directly? And how big piece of functionality is used from test_common in benchmarks?

benchmarks/cpp/cast/bench_dequantize_mxfp8.cpp
alextmagro: Since the benchmarks are built separate from TE library, best practice is to use <> to ensure we are using the installed library's headers.

benchmarks/cpp/run_benchmarks.sh
alextmagro: As we don't do any hipification in the benchmarks/ folder, we need files that have been hipified already. My thought was that if these ever made it into a CI run (although I plan to use variants on them for other kernel benchmarking during development), it would be after tests were run, so we would have the hipified files already.

Regarding the functionality, we use most of the Tensor functions. We explicitly guard against anything gtest related to avoid the dependency, and compare functions.

ipanfilo: It does not answer why use symlinks instead adding those files as a source in CMake. Also, to make benchmarks build-able independently, it may make sense to add hipification to it.

benchmarks/cpp/cast/bench_dequantize_mxfp8.cpp
ipanfilo: Well, it uses TE headers paths from local TE tree, not system paths

benchmarks/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,89 @@
...
+find_package(HIP REQUIRED)
+
+include(FetchContent)
+FetchContent_Declare(
+  benchmark
+  GIT_REPOSITORY https://github.com/google/benchmark.git
+  GIT_TAG v1.8.3
+)
+set(BENCHMARK_ENABLE_TESTING OFF CACHE BOOL "Disable benchmark tests" FORCE)
+set(BENCHMARK_ENABLE_GTEST_TESTS OFF CACHE BOOL "Disable gtest in benchmark" FORCE)
```
ipanfilo, line 23: What are those two parameters? I didn't find they are used

benchmarks/cpp/cast/bench_dequantize_mxfp8.cpp
alextmagro: That's true, I have changed everything to use "" to make that clear.

benchmarks/cpp/CMakeLists.txt
alextmagro: Those are Google Benchmark variables that are used to avoid building the Google Benchmark test suite which we don't need.

benchmarks/cpp/run_benchmarks.sh
alextmagro: I have cleaned this up.

benchmarks/cpp/run_benchmarks.sh
alextmagro: I had used symlink so that I can see the file in the folder, but hipification makes sense for stand alone builds, so have replaced things with that.

benchmarks/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,95 @@
...
+
+include_directories(
+  ${CMAKE_CURRENT_SOURCE_DIR}/../../transformer_engine/common/include
+  ${CMAKE_CURRENT_SOURCE_DIR}/../../transformer_engine/common
+  ${CMAKE_CURRENT_SOURCE_DIR}/../../transformer_engine
+  ${CMAKE_CURRENT_SOURCE_DIR}/utils
+  ${TESTS_CPP_DIR}
+)
+
+if(DEFINED ENV{NVTE_ROCM_ARCH})
```
ipanfilo: Consider using build_tools/rocm_utils.cmake instead

benchmarks/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,95 @@
...
+  set(GPU_TARGETS "gfx942;gfx950")
+endif()
+
+set(COMMON_COMPILE_OPTIONS
+  -Wall
+  -Wextra
+  -O3
+  -DNDEBUG
+  -DUSE_ROCM
+  --offload-arch=${GPU_TARGETS}
```
ipanfilo: offload-arch does not support lists.

benchmarks/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,95 @@
+cmake_minimum_required(VERSION 3.18)
+
+if(NOT DEFINED CMAKE_CXX_COMPILER)
+  set(CMAKE_CXX_COMPILER hipcc)
+endif()
+
+project(transformer_engine_benchmarks LANGUAGES CXX)
```
ipanfilo: HIP? Looks like only CXX file is test_common

benchmarks/cpp/CMakeLists.txt
alextmagro: I have refactored the Cmake file to use rocm_utils.cmake. This also added HIP to the project, and removed the offload-arch list.

ipanfilo (approved):

PR #508, opened 2026-03-27
[TE] Enable deterministic mode for fused attention

AllenFarcas:
# Description

Fixes https://github.com/ROCm/frameworks-internal/issues/15875

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Added deterministic functionality to fused attention
- Added test for the introduced deterministic functionality

# Checklist:


Micky774:
Unless we want to support non-deterministic CK **only for the JAX integration**, we should probably also add some tests to the pytorch integration side since it'll be enabled there too.

Also I think you still need to adjust https://github.com/ROCm/TransformerEngine/blob/82617fea8380d4fd4ec7145ae952bcc28af9f56a/transformer_engine/pytorch/attention/dot_product_attention/utils.py#L1070-L1078

tests/jax/test_fused_attn.py
```diff
@@ -1252,3 +1253,165 @@ def test_jax_new_rng():
...
+    # Verify the CK backend is selected, otherwise test is meaningless
+    backend = FusedAttnHelper(
+        True, dtype, dtype, qkv_layout, AttnBiasType.NO_BIAS, attn_mask_type,
+        0.0, h_q, h_kv, s, s, d, d, (-1, -1),
+    ).get_fused_attn_backend()
+    if backend == NVTE_Fused_Attn_Backend.NVTE_No_Backend:
+        pytest.skip("No fused attention backend available for this config")
+    assert backend == NVTE_Fused_Attn_Backend.NVTE_CK, (
+        f"Expected CK backend but got {backend}."
+    )
```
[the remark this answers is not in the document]
Micky774: This will indeed report it as a fail, so let's skip instead.

tests/jax/test_fused_attn.py
```diff
@@ -1252,3 +1253,165 @@ def test_jax_new_rng():
...
+    produces bitwise-reproducible gradients.  Optionally checks numerical
+    correctness against an unfused JAX reference (O(s^2)); this is skipped for
+    large seq_len to keep CI fast.
+
+    All seq_len values should be >= 256 so that nsplits = ceil(s/kN0) > 1
+    (kN0=128 for d<=128), ensuring the kernel actually exercises the
+    deterministic split-accumulator path rather than the trivial single-split.
+    """
+    if check_numerical is None:
+        check_numerical = seq_len <= 256
```
wangye805: Why do we skip checking the numerical for cases with seqlen<=256

tests/jax/test_fused_attn.py
```diff
@@ -26,7 +27,7 @@
 
 from transformer_engine.jax.cpp_extensions.misc import is_hip_extension
 from transformer_engine.jax import autocast
-from transformer_engine.jax.sharding import MeshResource
+from transformer_engine.jax.sharding import MeshResource, global_shard_guard
```
wangye805: Why do we need this?

tests/jax/test_fused_attn.py
```diff
@@ -1252,3 +1253,165 @@ def test_jax_new_rng():
...
+    large seq_len to keep CI fast.
+
+    All seq_len values should be >= 256 so that nsplits = ceil(s/kN0) > 1
+    (kN0=128 for d<=128), ensuring the kernel actually exercises the
+    deterministic split-accumulator path rather than the trivial single-split.
+    """
+    if check_numerical is None:
+        check_numerical = seq_len <= 256
+    s = seq_len
+    dtype = jnp.bfloat16
```
wangye805: Let's check for both bf16 and fp16

tests/jax/test_fused_attn.py
```diff
@@ -1252,3 +1253,165 @@ def test_jax_new_rng():
...
+    scaling_factor = 1.0 / sqrt(d)
+
+    # Set deterministic mode before any TE calls so the flag is visible
+    # throughout backend selection and kernel dispatch.
+    _orig_nondeterministic = os.environ.get("NVTE_ALLOW_NONDETERMINISTIC_ALGO")
+    os.environ["NVTE_ALLOW_NONDETERMINISTIC_ALGO"] = "0"
```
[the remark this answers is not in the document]
wangye805: Yeah, copilot gave a good comment. You will need to cache the outside env values and reset it after we finish this pytest

tests/jax/test_fused_attn.py
wangye805: Technically, if we specify NVTE_ALLOW_NONDETERMINISTIC_ALGO=0, the backend selection should take this env and choose deterministic backend for us, not restricting to CK. As I recall, aotriton by its nature is deterministic @xinyazhang

tests/jax/test_fused_attn.py
```diff
@@ -1252,3 +1253,165 @@ def test_jax_new_rng():
...
+
+
+@pytest.mark.skipif(
+    not is_hip_extension(), reason="CK deterministic backward only applies to AMD hardware"
+)
+@pytest.mark.parametrize(
+    "attn_mask_type",
+    [
+        pytest.param(AttnMaskType.NO_MASK, id="NO_MASK"),
+        pytest.param(AttnMaskType.CAUSAL_MASK, id="CAUSAL"),
```
wangye805: Let's not restrict to NO_MASK or CAUSAL, let's add causal, padding, padding causal, padding causal bottom right as well

tests/jax/test_fused_attn.py
```diff
@@ -1252,3 +1253,165 @@ def test_jax_new_rng():
...
+    ],
+)
+def test_deterministic_bwd_gqa(attn_mask_type):
+    """GQA variant: BSHD_BSHD_BSHD with h_q != h_kv."""
```
wangye805: Also, extend to nonGQA cases as well

tests/jax/test_fused_attn.py
```diff
@@ -1252,3 +1253,165 @@ def test_jax_new_rng():
...
+    _run_deterministic_bwd_case(
+        qkv_layout=QKVLayout.BSHD_BSHD_BSHD,
+        attn_mask_type=attn_mask_type,
+        b=2, seq_len=2048, h_q=12, h_kv=4, d=128,
```
wangye805: Also, check some sequence packing cases

wangye805 (changes requested):
BTW, add some deterministic testcases in pytorch side as well

tests/jax/test_fused_attn.py
ipanfilo: Well, backend selection API does not support deterministic flag. And yes, TE considers AOTriton as deterministic. The question is if we want to test it for AOTrioton

tests/jax/test_fused_attn.py
```diff
@@ -1252,3 +1253,165 @@ def test_jax_new_rng():
...
+            ref_dv = ref_grads[2].block_until_ready()
+
+            assert_allclose(jnp.array(fused_dq1), ref_dq, dtype=dtype)
+            assert_allclose(jnp.array(fused_dk1), ref_dk, dtype=dtype)
+            assert_allclose(jnp.array(fused_dv1), ref_dv, dtype=dtype)
+    finally:
+        if _orig_nondeterministic is None:
+            os.environ.pop("NVTE_ALLOW_NONDETERMINISTIC_ALGO", None)
+        else:
+            os.environ["NVTE_ALLOW_NONDETERMINISTIC_ALGO"] = _orig_nondeterministic
```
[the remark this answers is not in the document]
ipanfilo: Since this is the end of the test no explicit monkeypatch action is needed. Idiomatic way to use it is add monkeypatch as the test argument and use it to set env varables so it cleans everything up automatically

tests/pytorch/attention/test_attention.py
```diff
@@ -2963,3 +2963,317 @@ def forward(
...
+        config,
+        qkv_dtype=dtype,
+        qkv_layout=qkv_layout,
+        deterministic=True,
+        is_training=True,
+    )
+    if not available_backends[1]:
+        pytest.skip("FusedAttention not available for this config")
+    if not fused_attn_backends:
+        pytest.skip("No fused attention sub-backend available")
```
Micky774: Note that `available_backends[1] = len(fused_attn_backends) > 0` by definition so this is redundant.

tests/pytorch/attention/test_attention.py
```diff
@@ -2963,3 +2963,317 @@ def forward(
...
```
Micky774: Same here

tests/pytorch/attention/test_attention.py
```diff
@@ -2963,3 +2963,317 @@ def forward(
...
+        max_seqlen_q=seq_len,
+        num_heads=h_q,
+        head_dim_qk=d,
+        num_gqa_groups=h_kv,
+        attn_mask_type=attn_mask_type,
+    )
+    _run_deterministic_bwd_pytorch(monkeypatch, dtype, config, qkv_layout)
+
+
+def _run_deterministic_bwd_thd_pytorch(
```
Micky774: There's a lot of code overlap between this and `_run_deterministic_bwd_pytorch`, can we refactor a bit?

Micky774 (changes requested):

tests/pytorch/attention/test_attention.py
```diff
@@ -2990,28 +3023,23 @@ def _run_deterministic_bwd_pytorch(
...
+    qkv_format = "thd" if is_thd else "bshd"
 
-    _attention_backends["backend_selection_requires_update"] = True
+    # THD requires a padding mask type.
+    if is_thd and "padding" not in config.attn_mask_type:
+        config.attn_mask_type = (
+            "padding_" + config.attn_mask_type
+            if config.attn_mask_type != "no_mask"
+            else "padding"
+        )
```
Micky774: `test_deterministic_bwd_thd` already is parameterized only by padding type masks, why is this section needed?

tests/jax/test_fused_attn.py
```diff
@@ -1507,3 +1507,285 @@ def test_jax_new_rng():
...
+    ).get_fused_attn_backend()
+    if backend == NVTE_Fused_Attn_Backend.NVTE_No_Backend:
+        pytest.skip("No fused attention backend available for this config")
+    if backend not in (
+        NVTE_Fused_Attn_Backend.NVTE_CK,
+        NVTE_Fused_Attn_Backend.NVTE_AOTriton,
+    ):
+        pytest.skip(
+            f"Deterministic test requires CK or AOTriton backend, got {backend}."
+        )
```
wangye805: We only had three backends (no_backend, ck, and aotrition). This check is not needed

tests/jax/test_fused_attn.py
```diff
@@ -1507,3 +1507,285 @@ def test_jax_new_rng():
...
+        mask = make_mask(segment_ids, segment_ids, None, None, attn_mask_type)
+    else:
+        seqlens = jnp.full((b,), s, dtype=jnp.int32)
+        seq_desc = SequenceDescriptor.from_seqlens((seqlens, seqlens))
+
+        if attn_mask_type == AttnMaskType.NO_MASK:
+            mask = None
+        else:
+            segment_ids = jnp.ones((b, s), dtype=jnp.int32)
+            mask = make_mask(segment_ids, segment_ids, None, None, attn_mask_type)
```
wangye805: For segment id, segment pos, mask, and seqlens generation, can you consolidate with existing util functions like those applied in https://github.com/ROCm/TransformerEngine/blob/098115728f7eb1838b8e750eb54e8ae2dcc7bdcd/tests/jax/test_fused_attn.py#L627-L660?

tests/pytorch/attention/test_attention.py
```diff
@@ -2963,3 +2963,242 @@ def forward(
...
+    """Run backward twice with the same seed and assert bitwise reproducibility."""
+    dq1, dk1, dv1 = run_once_fn(42)
+    dq2, dk2, dv2 = run_once_fn(42)
+
+    torch.testing.assert_close(dq1, dq2, atol=0, rtol=0, msg="dQ not bitwise reproducible")
+    torch.testing.assert_close(dk1, dk2, atol=0, rtol=0, msg="dK not bitwise reproducible")
+    torch.testing.assert_close(dv1, dv2, atol=0, rtol=0, msg="dV not bitwise reproducible")
+
+
+def _run_deterministic_bwd_pytorch(
```
wangye805: nit: _pytorch is not needed in the function name as you are already under tests/pytorch dir

tests/pytorch/attention/test_attention.py
```diff
@@ -2963,3 +2963,242 @@ def forward(
...
+        Data type (torch.bfloat16 or torch.float16).
+    config : ModelConfig
+        Model configuration.
+    qkv_layout : str
+        QKV memory layout string (e.g. "bshd_bshd_bshd" or "thd_thd_thd").
+    """
+    import copy
+    config = copy.copy(config)
+
+    is_thd = "thd" in qkv_layout
```
wangye805: What if the qkv_layout is th3d? or bs3hd? There should be util functions to get qkv_layout to qkv_format

tests/pytorch/attention/test_attention.py
```diff
@@ -2963,3 +2963,242 @@ def forward(
...
+    qkv_format = "thd" if is_thd else "bshd"
+
+    # THD requires a padding mask type.
+    if is_thd and "padding" not in config.attn_mask_type:
```
wangye805: thd format always comes with padding in mask type. Otherwise the config is invalid

tests/pytorch/attention/test_attention.py
```diff
@@ -2963,3 +2963,242 @@ def forward(
...
+
+def _setup_deterministic_env(monkeypatch, config, dtype, qkv_layout):
+    """Set environment variables and check backend availability for deterministic bwd tests."""
+    monkeypatch.setenv("NVTE_ALLOW_NONDETERMINISTIC_ALGO", "0")
+    monkeypatch.setenv("NVTE_FUSED_ATTN", "1")
+    monkeypatch.setenv("NVTE_FLASH_ATTN", "0")
+    monkeypatch.setenv("NVTE_UNFUSED_ATTN", "0")
+    _attention_backends["backend_selection_requires_update"] = True
+
+    available_backends, _, _ = get_available_attention_backends(
```
wangye805: Unfortunately, get_available_attention_backends always reset NVTE_*_ATTN envs inside: https://github.com/ROCm/TransformerEngine/blob/098115728f7eb1838b8e750eb54e8ae2dcc7bdcd/tests/pytorch/utils.py#L280-L298

You don't need the monkeypatch above

tests/pytorch/attention/test_attention.py
```diff
@@ -2963,3 +2963,242 @@ def forward(
...
+        else:
+            fwd_kwargs["attention_mask"] = attention_mask
+
+        out = block(q, k, v, **fwd_kwargs)
+        out.backward(d_out)
+        return (
+            q.grad.clone().detach(),
+            k.grad.clone().detach(),
+            v.grad.clone().detach(),
+        )
```
wangye805: Is there a way to consolidate the test input generation, fwd, and bwd with existing functions like https://github.com/ROCm/TransformerEngine/blob/098115728f7eb1838b8e750eb54e8ae2dcc7bdcd/tests/pytorch/attention/test_attention.py#L1006?

tests/pytorch/attention/test_attention.py
```diff
@@ -2963,3 +2963,242 @@ def forward(
...
+    ],
+)
+@pytest.mark.parametrize(
+    "b, seq_len, h_q, h_kv, d",
+    [
+        pytest.param(2, 2048, 8, 8, 128, id="b2_s2048_MHA"),
+        pytest.param(2, 2048, 12, 4, 128, id="b2_s2048_GQA"),
+    ],
+)
+def test_deterministic_bwd_thd(
```
wangye805: If I recall correctly, only flash-attn and our fused-attn support thd, but flash-attn is not deterministic. So for thd format, you can only check the determinism. If you wanted to check whether the result is correct, you will need to enable flash-attn with non-determinism

tests/pytorch/attention/test_attention.py
```diff
@@ -2963,3 +2963,242 @@ def forward(
...
+    monkeypatch, dtype, attn_mask_type, b, seq_len, h_q, h_kv, d
+):
+    """Test deterministic backward with THD (sequence packing): bitwise reproducibility."""
```
wangye805: Sequence packing is not thd, it's another jax term which packs sub-sequences into a sequence

wangye805 (changes requested):

tests/jax/test_fused_attn.py
```diff
@@ -1507,3 +1507,285 @@ def test_jax_new_rng():
     )
     runner = FusedAttnRunner(**kwargs)
     runner.test_forward()
+
+
+def _run_deterministic_bwd_case(
+    monkeypatch, qkv_layout, attn_mask_type, b, seq_len, h_q, h_kv, d,
+    dtype=jnp.bfloat16, check_numerical=None,
```
Micky774: When is `check_numerical` actually ever used as `None`? If we only pass it as `False` we can just hard-code that path.

tests/jax/test_fused_attn.py
Micky774, line 1: Like Ye mentioned, there's a lot of opportunities to re-use existing code -- for example, could you use `FusedAttnRunner._setup_inputs()` and extract the generated inputs?

tests/jax/test_fused_attn.py
```diff
@@ -1507,3 +1507,285 @@ def test_jax_new_rng():
...
+
+        ref_val_grad = jit(jax.value_and_grad(ref_fn, argnums=(0, 1, 2)))
+        _, ref_grads = ref_val_grad(q, k, v)
+        ref_dq = ref_grads[0].block_until_ready()
+        ref_dk = ref_grads[1].block_until_ready()
+        ref_dv = ref_grads[2].block_until_ready()
+
+        assert_allclose(jnp.array(fused_dq1), ref_dq, dtype=dtype)
+        assert_allclose(jnp.array(fused_dk1), ref_dk, dtype=dtype)
+        assert_allclose(jnp.array(fused_dv1), ref_dv, dtype=dtype)
```
Micky774: This already only triggers for a narrow subset of runs, do we really need it at all? If we validate purely determinism, we can leave the rest of the suite to validate correctness.

tests/jax/test_fused_attn.py
```diff
@@ -1507,3 +1507,285 @@ def test_jax_new_rng():
...
+        pytest.param(QKVLayout.THD_THD_THD, id="THD_THD_THD"),
+    ],
+)
+def test_deterministic_bwd_seqpack(monkeypatch, dtype, attn_mask_type, qkv_layout):
+    """Sequence packing (THD) variants for deterministic backward."""
+    _run_deterministic_bwd_case(
+        monkeypatch,
+        qkv_layout=qkv_layout,
+        attn_mask_type=attn_mask_type,
+        b=2, seq_len=2048, h_q=8, h_kv=8, d=128,
```
Micky774: btw why hardcode these specific values? Is it an arbitrary choice?

tests/jax/test_fused_attn.py
```diff
@@ -1507,3 +1507,285 @@ def test_jax_new_rng():
...
+)
+@pytest.mark.parametrize(
+    "b, seq_len, h_q, h_kv, d",
+    [
+        pytest.param(2, 256, 8, 8, 128, id="b2_s256_MHA"),
+        pytest.param(2, 512, 8, 8, 128, id="b2_s512_MHA"),
+        pytest.param(2, 2048, 8, 8, 128, id="b2_s2048_MHA"),
+    ],
+)
+def test_deterministic_bwd(monkeypatch, dtype, qkv_layout, attn_mask_type, b, seq_len, h_q, h_kv, d):
```
Micky774: Can we consolidate the various tests into a singular test with a covering parameterization?

tests/pytorch/attention/test_attention.py
Micky774: Ditto this point, especially since it returns its output gradients so you can harvest them to check for determinism

tests/pytorch/attention/test_attention.py
Micky774: Does this need a separate test or can it be consolidated as part of the parameterization of `test_deterministic_bwd` w/ skips as needed for invalid padding configs?

tests/pytorch/attention/test_attention.py
```diff
@@ -2963,3 +2963,242 @@ def forward(
                 self.quantizers,
             )
         return out
+
+
+# ---------------------- Deterministic Backward Tests ----------------------
+
+
+def _setup_deterministic_env(monkeypatch, config, dtype, qkv_layout):
```
Micky774: Can you use something like
```python
    _, _, fused_attn_backends = get_available_attention_backends(
        config,
        qkv_dtype=dtype,
        qkv_layout=qkv_layout,
        pad_between_seqs=pad_between_seqs,
    )
    if FusedAttnBackend["CK"] not in fused_attn_backends:
        pytest.skip("This test requires the CK fused attention backend.")
```
instead?

tests/pytorch/attention/test_attention.py
```diff
@@ -2963,3 +2963,242 @@ def forward(
...
+            if config.attn_mask_type != "no_mask"
+            else "padding"
+        )
+
+    _setup_deterministic_env(monkeypatch, config, dtype, qkv_layout)
+
+    def _run_once(seed_val):
+        reset_rng_states()
+        torch.manual_seed(seed_val)
+        torch.cuda.manual_seed(seed_val)
```
Micky774: redundant with `reset_rng_states()`

tests/pytorch/attention/test_attention.py
```diff
@@ -2963,3 +2963,242 @@ def forward(
...
+
+
+def _assert_deterministic_bwd(run_once_fn):
+    """Run backward twice with the same seed and assert bitwise reproducibility."""
+    dq1, dk1, dv1 = run_once_fn(42)
+    dq2, dk2, dv2 = run_once_fn(42)
+
+    torch.testing.assert_close(dq1, dq2, atol=0, rtol=0, msg="dQ not bitwise reproducible")
+    torch.testing.assert_close(dk1, dk2, atol=0, rtol=0, msg="dK not bitwise reproducible")
+    torch.testing.assert_close(dv1, dv2, atol=0, rtol=0, msg="dV not bitwise reproducible")
```
Micky774: Something more pythonic would be along the lines of:
```suggestion
    for name, x, y in zip(("dQ", "dK", "dV"), run_once_fn(42), run_once_fn(42)):
        torch.testing.assert_close(x, y, atol=0, rtol=0, msg=f"{name} not bitwise reproducible")
```

tests/pytorch/attention/test_attention.py
```diff
@@ -2963,3 +2963,242 @@ def forward(
...
+    monkeypatch : pytest.MonkeyPatch
+        Pytest fixture for safe env var management.
+    dtype : torch.dtype
+        Data type (torch.bfloat16 or torch.float16).
+    config : ModelConfig
+        Model configuration.
+    qkv_layout : str
+        QKV memory layout string (e.g. "bshd_bshd_bshd" or "thd_thd_thd").
+    """
+    import copy
```
Micky774: Let's move to the top of the file, though copying likely isn't needed (see comments below)

Micky774 (changes requested):

tests/pytorch/attention/test_attention.py
```diff
@@ -3161,44 +3019,39 @@ def test_deterministic_bwd(monkeypatch, dtype, attn_mask_type, b, seq_len, h_q,
...
+    _, _, grads2 = _run_dot_product_attention(
+        dtype=dtype,
+        config=config,
+        backend="FusedAttention",
+        ckpt_attn=False,
+        qkv_layout=qkv_layout,
+        workspace_opt=False,
+        pad_between_seqs=False,
+        is_training=True,
+    )
```
Micky774: Can we absorb these keyword args into a shared `kwargs ={...}` dict to re-use? Also, is there a reason we shouldn't include the output in the final comparison? I've observed that even if the gradients are equal, sometimes numerical failures can occur in the output, so it might be worth including it for caution.

Micky774 (commented):
I wanted to point out that upstream has existing infrastructure for deterministic correctness tests (see [this commit](https://github.com/NVIDIA/TransformerEngine/commit/27fc168e2166e2338a540e3fcd59f9c8491f3b10)). Maybe we should adopt their approach. Furthermore, do we need determinism tests at all levels, or should we isolate to level 3? cc: @wangye805

tests/jax/test_fused_attn.py
```diff
@@ -1507,3 +1507,123 @@ def test_jax_new_rng():
...
+
+    _, grads2 = fused_val_grad(q, k, v)
+    fused_dq2 = np.array(grads2[0].block_until_ready())
+    fused_dk2 = np.array(grads2[1].block_until_ready())
+    fused_dv2 = np.array(grads2[2].block_until_ready())
+
+    # Bitwise reproducibility across consecutive runs
+    np.testing.assert_array_equal(fused_dq1, fused_dq2, err_msg="dQ not bitwise reproducible")
+    np.testing.assert_array_equal(fused_dk1, fused_dk2, err_msg="dK not bitwise reproducible")
+    np.testing.assert_array_equal(fused_dv1, fused_dv2, err_msg="dV not bitwise reproducible")
```
Micky774: Couple notes. First, for consistency can we use `assert_allclose` (which uses `np.testing.assert_allclose`)? Second, similarly to the other test, a more pythonic solution might look like:
```suggestion
    _, grads1 = fused_val_grad(q, k, v)
    _, grads2 = fused_val_grad(q, k, v)
    for name, x, y in zip(("dQ", "dK", "dV"), grads1, grads2):
        # Bitwise reproducibility across consecutive runs
        assert_allclose(x, y, atol=0, rtol=0, err_msg=f"{name} not bitwise reproducible")
    ```

File: tests/pytorch/attention/test_attention.py [hunk 33]
```diff
@@ -80,6 +80,14 @@
 seed = 1234
 reset_rng_states()
 
+# Determinism mode is selected by the NVTE_ALLOW_NONDETERMINISTIC_ALGO env var
+# (or by torch's global deterministic flag). When enabled, deterministic-only
+# tests run and non-determinism-only tests are skipped.
+_deterministic = (
+    not bool(int(os.getenv("NVTE_ALLOW_NONDETERMINISTIC_ALGO", "1")))
+    or torch.are_deterministic_algorithms_enabled()
+)
```
Micky774, line 89: Does this mean that if somehow `torch.are_deterministic_algorithms_enabled()` then `NVTE_ALLOW_NONDETERMINISTIC_ALGO=0` is overruled? That seems like incorrect behavior to me. Is this from upstream?

tests/pytorch/attention/test_attention.py
```diff
@@ -2963,3 +2971,87 @@ def forward(
...
+        "padding",
+        "padding_causal",
+        "padding_causal_bottom_right",
+    )
+    if not (layout.startswith("thd") and "padding" not in mask)
+]
+
+
+@pytest.mark.skipif(
+    not IS_HIP_EXTENSION, reason="Deterministic backward only applies to AMD hardware"
```
Micky774: This currently reads as "NV does not support deterministic backwards" -- is that true, or does the message need to be updated?

AllenFarcas: Changed to "CK split-accumulator deterministic backward is ROCm-only"

tests/pytorch/attention/test_attention.py
AllenFarcas: Upstream has already this implemented to just making sure it is compatible and adjusted according to upstream.

docs/envvars.rst
```diff
@@ -156,7 +156,7 @@ Attention Backend Selection
 
    :Type: ``int`` (0 or 1)
    :Default: ``1``
-   :Description: Allow non-deterministic algorithms for Transformer Engine execution. When set to ``0``, only deterministic algorithms are allowed. This is relevant for both PyTorch and JAX attention implementations.
+   :Description: Allow non-deterministic algorithms for Transformer Engine execution. When set to ``0``, only deterministic algorithms are allowed. This is relevant for both PyTorch and JAX attention implementations. On AMD/HIP builds, setting this to ``0`` enables the deterministic backward pass of the CK FusedAttention backend (which uses a split-accumulator workspace for deterministic ``dQ``); on NVIDIA builds it disables FusedAttention paths that are known to be non-deterministic.
```
wangye805: As far as I know, our aotriton backend is also deterministic. Why setting this only enables the CK backend?

wangye805 (approved):
My only concern right now is, aotriton backend is also determinstic. But your current PR should be okay since CK backend has higher priority than aotriton backend

tests/pytorch/attention/test_attention.py
Micky774: nvm I misread the actual logic

Micky774:
Currently it remains untested -- we still need to add it to the CI tests

docs/envvars.rst
AllenFarcas: Since AOTriton is always deterministic, there is no action required when NVTE_ALLOW_NONDETERMINISTIC_ALGO=0, so this envvar description is technically accuracy since only CK needs that flag to change its behavior.

Micky774: I think the wording is accurate, but confusing. The behavior really isn't different from NV. I don't think the description needs to be changed from the original.

Micky774:
Two questions regarding tests:
1. Do we want to have them be level 1 or 3?
2. Do we want to have determinism tests be mutually exclusive w/ regular tests, i.e. add a decorator to skip normal tests for AMD if the flag is set?

If we don't make them mutually exclusive, I think they should be level 3. If we do, then level 1 is fine.

cc: @wangye805

wangye805:
> Two questions regarding tests:
> 
> 1. Do we want to have them be level 1 or 3?
> 2. Do we want to have determinism tests be mutually exclusive w/ regular tests, i.e. add a decorator to skip normal tests for AMD if the flag is set?
> 
> If we don't make them mutually exclusive, I think they should be level 3. If we do, then level 1 is fine.
> 
> cc: @wangye805

Previously Ilya and I agreed that level 1 should be running things with default config, like hip cast transpose, instead of triton cast transpose. So if this deterministic flow is not set to true by default, we can put it into level 3 ci

docs/envvars.rst
```diff
@@ -156,7 +156,7 @@ Attention Backend Selection
 
    :Type: ``int`` (0 or 1)
    :Default: ``1``
-   :Description: Allow non-deterministic algorithms for Transformer Engine execution. When set to ``0``, only deterministic algorithms are allowed. This is relevant for both PyTorch and JAX attention implementations.
+   :Description: Allow non-deterministic algorithms for Transformer Engine execution. When set to ``0``, only deterministic algorithms are allowed.
```
Micky774: Let's minimize the diff here

Micky774 (approved):

Micky774:
Let's make sure to wait for a green level 3 CI first

ipanfilo (approved):

Micky774:
CI failure is unrelated

PR #509, opened 2026-03-27
NVFP4 Random Hadamard Transform (butterfly permutation-based)

matthiasdiener:
# Description

Implements RHT via a butterfly permutation-based algorithm for NVFP4.

Has similar restrictions as upstream:
- BF16 only
- Transpose path only (no identity path)

Fixes https://github.com/ROCm/frameworks-internal/issues/15732

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/common/hadamard_transform/hadamard_transform.cu
```diff
@@ -658,8 +662,332 @@ __global__ void HadamardTransformKernel(const T* __restrict__ input, T* __restri
...
+    return r;
+}
+
+// BF16 helpers
+__device__ __forceinline__ float          to_f32 (__hip_bfloat16 v) { return static_cast<float>(v); }
+__device__ __forceinline__ __hip_bfloat16 to_bf16(float v)          { return static_cast<__hip_bfloat16>(v); }
+
+// Bit-cast __hip_bfloat16->uint16_t without address-of-temporary.
+__device__ __forceinline__ uint16_t bf16_to_bits(__hip_bfloat16 v) {
+    uint16_t bits; __builtin_memcpy(&bits, &v, sizeof(uint16_t)); return bits;
```
ipanfilo: __builtin_bit_cast ?

transformer_engine/common/include/transformer_engine/hadamard_transform.h
```diff
@@ -13,8 +13,6 @@
 #ifndef TRANSFORMER_ENGINE_HADAMARD_TRANSFORM_H_
 #define TRANSFORMER_ENGINE_HADAMARD_TRANSFORM_H_
 
-#ifndef __HIP_PLATFORM_AMD__
```
ipanfilo, line 16: Is it now identical to upstream variant?

matthiasdiener: As far as I can tell, yes. Would you like me to remove the AMD copyright header?

transformer_engine/common/hadamard_transform/hadamard_transform.cu
matthiasdiener: Thanks, changed in ca1aacf7

transformer_engine/common/include/transformer_engine/hadamard_transform.h
ipanfilo: Yes, if there is no difference let's use pure upstream file

matthiasdiener: Removed in [bc9f0a3](https://github.com/ROCm/TransformerEngine/pull/509/commits/bc9f0a392b1af6f11e5e8414f582a1e8228599f2). From what I can see, the file is now identical to upstream.

transformer_engine/common/hadamard_transform/hadamard_transform.cu
```diff
@@ -16,13 +18,16 @@
...
 #include "common/util/ptx.cuh"
 #include "common/utils.cuh"
+#ifndef __HIP_PLATFORM_AMD__
 #include "hadamard_transform_utils.cuh"
+#endif
 
 namespace transformer_engine {
 namespace {
 
 constexpr int kThreadsPerWarp = 32;
```
ipanfilo, line 29: It also seems unused on ROCm now so whole namespace could be guarded

transformer_engine/pytorch/csrc/pybind.h
```diff
@@ -110,12 +110,8 @@ constexpr std::array custom_types_converters = {
                     CreateQuantizer<MXFP8Quantizer>),
     std::make_tuple(IsFloat8BlockwiseQTensor, IsFloat8BlockwiseQuantizers,
                     NVTETensorFromFloat8BlockwiseQTensor, CreateQuantizer<Float8BlockQuantizer>),
-#ifdef USE_ROCM
```
ipanfilo, line 113: The file seems identical to upstream now, so clear copyright

tests/pytorch/nvfp4/test_nvfp4_rht_quantize_exact.py
```diff
@@ -246,3 +248,124 @@ def test_nvfp4_quantization_noncontiguous_inputs(
...
+        h = 1
+        while h < 16:
+            for i in range(0, 16, h * 2):
+                a = tile[:, i:i+h].clone()
+                b = tile[:, i+h:i+2*h].clone()
+                tile[:, i:i+h]     = a + b
+                tile[:, i+h:i+2*h] = a - b
+            h *= 2
+        out[:, c:c+16] = tile * 0.25
+    return out
```
wangye805: Can we reuse the _apply_rht in upstream's experimental quantizer?

tests/pytorch/nvfp4/test_nvfp4_rht_quantize_exact.py
```diff
@@ -246,3 +248,124 @@ def test_nvfp4_quantization_noncontiguous_inputs(
...
+
+
+def _ref_quantize_wht16_tiled(
```
wangye805: This is like NVFP4QuantizerRef + padding + unpadding. Why do we need padding and unpadding?

tests/pytorch/nvfp4/test_nvfp4_rht_quantize_exact.py
```diff
@@ -246,3 +248,124 @@ def test_nvfp4_quantization_noncontiguous_inputs(
...
+        eps=ref_quantizer.eps,
+    )
+
+    qx_t_ref = ref_quantizer._rm_pad_tensor(qx_t_ref, (x_t_rht.shape[0], x_t_rht.shape[1] // 2))
+
+    return qx_t_ref, sx_t_ref
+
+
+@pytest.mark.parametrize("rows,cols", [(64, 64), (128, 128)])
+def test_hadamard_transform_amax(rows, cols):
```
wangye805: In addition to the quantization result check, I saw your new test also checks the intermediate results of the rht. Why not add just a hadmard transform check without the quantization result checking? I think the quantization result checking is already in test_nvfp4_quantization_noncontiguous_inputs and test_rht_with_quantization_block_tiling_versus_reference

tests/pytorch/nvfp4/test_nvfp4_rht_quantize_exact.py
```diff
@@ -246,3 +248,124 @@ def test_nvfp4_quantization_noncontiguous_inputs(
...
+    qx_t_ref, sx_t_ref = _ref_quantize_wht16_tiled(x, sign_mask_t, out._amax_columnwise)
+
+    qx_t = unpack_fp4(out._columnwise_data.view(torch.uint8))
+    qx_t_ref = unpack_fp4(qx_t_ref.view(torch.uint8))
+    torch.testing.assert_close(qx_t, qx_t_ref, atol=0.0, rtol=0.0)
+
+    sx_t = out._columnwise_scale_inv
+    sx_t_ref = sx_t_ref.view(dtype=torch.uint8)
+    sx_t_valid = sx_t[: sx_t_ref.shape[0], : sx_t_ref.shape[1]]
+    torch.testing.assert_close(sx_t_valid, sx_t_ref, atol=0.0, rtol=0.0)
```
wangye805: Also, don't forget guarding our rocm specific codes

wangye805 (changes requested):

transformer_engine/common/hadamard_transform/hadamard_transform.cu
```diff
@@ -485,8 +490,332 @@ __global__ void HadamardTransformKernel(const T* __restrict__ input, T* __restri
...
+        float r0=v0, r1=v1, r2=v2, r3=v3;
+        float lam = 0.f;
+        if (global_row < num_rows) {
+            wht16(r0, r1, r2, r3, thread_in_grp, random_sign_mask, apply_pre);
+            if constexpr (kUpdateAmax) {
+                // Match the stored/output precision when reporting amax.
+                const float r0_bf16 = to_f32(to_bf16(r0));
+                const float r1_bf16 = to_f32(to_bf16(r1));
+                const float r2_bf16 = to_f32(to_bf16(r2));
+                const float r3_bf16 = to_f32(to_bf16(r3));
```
wangye805, line 719: Emm, I presume your downcast to bf16 then upcast to fp32 to "match the stored/output precision when reporting amax". But is this required? Or are you doing this because NV upstream is doing this?

transformer_engine/common/hadamard_transform/hadamard_transform.cu
```diff
@@ -485,8 +490,332 @@ __global__ void HadamardTransformKernel(const T* __restrict__ input, T* __restri
...
+    __shared__ float block_amax_t[kWarpsPerBlock];
+
+    float v0=0.f, v1=0.f, v2=0.f, v3=0.f;
+    if (in_bounds) {
+        unpack_bf16x4(*reinterpret_cast<const uint64_t*>(
+            &input[global_row * row_length + col_base]), v0, v1, v2, v3);
+    }
+
+    // Identity path: WHT along row dimension
+    if constexpr (kComputeIdentity || kUpdateAmax) {
```
wangye805, line 708: In this if branch, kcomputeIdentity and kUpdateAmax seem to be independent but the rht is always done. Two questions:
1). Do we need the row-wise rht at all? From their nvfp4 paper, based on my understanding, we only need rht in colwise data? Unless this is for future extension?

Image
2). If the kComputeIdentity is False but kUpdateAmax is true, do we still need rht to get the amax?

transformer_engine/common/hadamard_transform/hadamard_transform.cu
```diff
@@ -485,8 +490,332 @@ __global__ void HadamardTransformKernel(const T* __restrict__ input, T* __restri
...
+        __syncthreads();
+
+        // Re-read: row_in_warp -> column index, thread_in_grp -> 4 rows
+        const int t_col      = row_in_warp;
+        const int smem_rbase = warp_id*kRowsPerWarp + thread_in_grp*kElemsPerThread;
+
+        float c0=to_f32(smem[smem_rbase+0][t_col]), c1=to_f32(smem[smem_rbase+1][t_col]);
+        float c2=to_f32(smem[smem_rbase+2][t_col]), c3=to_f32(smem[smem_rbase+3][t_col]);
+
+        wht16(c0, c1, c2, c3, thread_in_grp, random_sign_mask_t, apply_pre);
```
wangye805, line 754: Same question here about whether wht16 need to be done if kUpdateAmaxT is true but kComputeTranspose is false.

But this seems to be a smaller problem since they should be true or false at the same time in NV paper

transformer_engine/common/hadamard_transform/hadamard_transform.cu
```diff
@@ -485,8 +490,332 @@ __global__ void HadamardTransformKernel(const T* __restrict__ input, T* __restri
...
+
+        if constexpr (kUpdateAmaxT) {
+            // Match the stored/output precision when reporting amax.
+            const float c0_bf16 = to_f32(to_bf16(c0));
```
wangye805, line 759: Same question about the downcasting then upcasting

transformer_engine/common/hadamard_transform/hadamard_transform.cu
```diff
@@ -485,8 +490,332 @@ __global__ void HadamardTransformKernel(const T* __restrict__ input, T* __restri
...
+    float v0=0.f, v1=0.f, v2=0.f, v3=0.f;
+    if (in_bounds) {
+        unpack_bf16x4(*reinterpret_cast<const uint64_t*>(
+            &input[global_row * row_length + col_base]), v0, v1, v2, v3);
+    }
+
+    // Identity path: WHT along row dimension
+    if constexpr (kComputeIdentity || kUpdateAmax) {
+        float r0=v0, r1=v1, r2=v2, r3=v3;
+        float lam = 0.f;
```
wangye805: just curious, what does this lam stands for? local amax?

transformer_engine/common/hadamard_transform/hadamard_transform.cu
```diff
@@ -485,8 +490,332 @@ __global__ void HadamardTransformKernel(const T* __restrict__ input, T* __restri
...
+
+                if (lane_id == 0)
+                  atomicMaxFloat(amax_t, block_lam_t);
+            }
+        }
+    }
+}
+
+// Pre-RHT amax: max|input| before any transform.
+__global__ void PreRhtAmaxKernel(const __hip_bfloat16* __restrict__ input,
```
wangye805: Just curious, should we have existing same or similar functions for this global amax for a tensor? I recall you previously worked on a current scaling performance improvement PR which also need an per-tensor amax?

wangye805 (changes requested):

transformer_engine/common/hadamard_transform/hadamard_transform.cu
```diff
@@ -485,8 +490,332 @@ __global__ void HadamardTransformKernel(const T* __restrict__ input, T* __restri
...
+
+static constexpr int kHadamardDim     = 16;
+static constexpr int kWarpSize        = 64;
+static constexpr int kThreadsPerWHT   = 4;
+static constexpr int kElemsPerThread  = 4;
+static constexpr int kRowsPerWarp     = kWarpSize / kThreadsPerWHT;   // 16
+static constexpr int kWarpsPerBlock   = 4;
+static constexpr int kRowsPerBlock    = kRowsPerWarp * kWarpsPerBlock; // 64
+static constexpr int kThreadsPerBlock = kWarpSize   * kWarpsPerBlock;  // 256
+static constexpr float kHadamardScale = 0.25f;
```
aris134: These do not seem like arbitrary tuning knobs. Some comment describing the layout scheme here could be helpful

transformer_engine/common/hadamard_transform/hadamard_transform.cu
```diff
@@ -485,8 +490,332 @@ __global__ void HadamardTransformKernel(const T* __restrict__ input, T* __restri
...
+__device__ __forceinline__ float          to_f32 (__hip_bfloat16 v) { return static_cast<float>(v); }
+__device__ __forceinline__ __hip_bfloat16 to_bf16(float v)          { return static_cast<__hip_bfloat16>(v); }
+
+// Bit-cast __hip_bfloat16->uint16_t without address-of-temporary.
+__device__ __forceinline__ uint16_t bf16_to_bits(__hip_bfloat16 v) {
+    return __builtin_bit_cast(uint16_t, v);
+}
+
+// Unpack/pack 4 BF16 values as uint64_t (vectorised global load/store).
+// Same trick as cast_transpose_mxfp4_kernel_shuffled.cu::bf16x4_to_float4.
```
aris134, line 552: Since this function seems redundant with the referenced function from #506, I wonder if it would make sense to have it in a separate header

transformer_engine/common/hadamard_transform/hadamard_transform.cu
```diff
@@ -612,6 +965,33 @@ void hadamard_transform_amax(const Tensor& input_, Tensor& output_, uint16_t ran
...
+
+  if (pre_amax_ptr) {
+    NVTE_CHECK_CUDA(cudaMemsetAsync(pre_amax_ptr, 0, sizeof(float), stream));
+  }
+  if (id_amax_ptr) {
+    NVTE_CHECK_CUDA(cudaMemsetAsync(id_amax_ptr, 0, sizeof(float), stream));
+  }
+  if (tr_amax_ptr) {
+    NVTE_CHECK_CUDA(cudaMemsetAsync(tr_amax_ptr, 0, sizeof(float), stream));
+  }
```
aris134, line 989: I noticed upstream zeroes these amax outputs via ZeroAmaxKernel, whereas this ROCm change uses cudaMemsetAsync on each pointer. What was the motivation for switching from the kernel-based approach to memset here?

transformer_engine/common/hadamard_transform/hadamard_transform.cu
```diff
@@ -485,8 +490,332 @@ __global__ void HadamardTransformKernel(const T* __restrict__ input, T* __restri
...
+
+    __syncthreads();
+
+    if (warp_id == 0) {
+      float block_lam = (lane_id < kWarpsPerBlock) ? block_amax[lane_id] : 0.f;
+      for (int off=kWarpSize/2; off>=1; off>>=1)
+        block_lam=fmaxf(block_lam,__shfl_xor(block_lam,off));
+
+      if (lane_id == 0)
+        atomicMaxFloat(amax_out, block_lam);
```
aris134: This seems correct, but from a performance perspective, did you consider a hierarchical/two-pass reduction instead of atomically combining block-local amax values into global memory? Since it is only one atomic per block, I can see the simplicity argument, but I was curious about the tradeoff.

transformer_engine/common/hadamard_transform/hadamard_transform.cu
```diff
@@ -485,8 +490,332 @@ __global__ void HadamardTransformKernel(const T* __restrict__ input, T* __restri
...
+
+                for (int off=kWarpSize/2; off>=1; off>>=1)
+                  block_lam_t = fmaxf(block_lam_t, __shfl_xor(block_lam_t, off));
+
+                if (lane_id == 0)
+                  atomicMaxFloat(amax_t, block_lam_t);
+            }
+        }
+    }
+}
```
aris134, line 800: It looks like the `kUpdateAmax` and `kUpdateAmaxT` branches are doing the same warp-0 reduction pattern, just with different shared-memory sources / output pointers. Would it make sense to factor that into a small local helper?

transformer_engine/common/hadamard_transform/hadamard_transform.cu
wangye805: Right, Matthias hadamard transform was adapting from PR506. But looks like this PR509 will be ready first? So probably Sudharshan will need to consolidate PR506 against this PR :-)

transformer_engine/common/hadamard_transform/hadamard_transform.cu
wangye805: I think I can help answer partial of your questions. 
kHadamardDim is the dimension of the hadamard transform matrix. In this specific case, the hadamard transform matrix is of size 16x16. 
And kHadamardScale is 1/sqrt(hadamard matrix dim) 

For the tiling constants:
kWarpSize is the number of threads per warp (or wavefront in our amd platform)
kThreadsPerWHT is how many threads are needed for one 16-point hadamard transform. Here it's set to be 4, which means that each thread will manage 4 inputs
kRowsPerWarp is defined as kWarpSize/kThreadsPerWHT probably because Matthias assign one warp (64 threads) to deal with a 2D data block of size 16x16 at the same time. So one row of 16 input data can be handle by 4 threads

But regarding why those tiling parameters are chosen like this, I'm not quite sure either

tests/pytorch/nvfp4/test_nvfp4_rht_quantize_exact.py
matthiasdiener: All of the pytorch nvfp4 tests from upstream, including the ones you mentioned, require the nvfp4 recipe, which will only be enabled as part of https://github.com/ROCm/TransformerEngine/pull/518. For this reason, I added this new test here.

tests/pytorch/nvfp4/test_nvfp4_rht_quantize_exact.py
matthiasdiener: Yes, done in cf2c8f63

tests/pytorch/nvfp4/test_nvfp4_rht_quantize_exact.py
matthiasdiener: Removed the unnecessary padding/unpadding in cf2c8f63, the test dimensions are always multiples of 16 (as required by the kernel), so it was dead code.

tests/pytorch/nvfp4/test_nvfp4_rht_quantize_exact.py
matthiasdiener: Done in cf2c8f63

transformer_engine/common/hadamard_transform/hadamard_transform.cu
matthiasdiener: Yes, this is to match upstream. The WHT is computed in FP32 but the result is stored as BF16. The downcast rounds to BF16 precision so the reported amax matches the actual stored values rather than the FP32 intermediates

transformer_engine/common/hadamard_transform/hadamard_transform.cu
matthiasdiener: Right, from the NVFP4 paper, RHT is only needed for the colwise data. The identity path (rowwise RHT + its amax) exists to match upstream's kernel. So for both questions, we don't need them at the moment, but perhaps in the future and to match upstream.

transformer_engine/common/hadamard_transform/hadamard_transform.cu
matthiasdiener: Right, both are always set to the same value in the current recipe. The template is inherited from upstream's interface

transformer_engine/common/hadamard_transform/hadamard_transform.cu
matthiasdiener: Same answer as above, to reflect actual bf16 values.

transformer_engine/common/hadamard_transform/hadamard_transform.cu
matthiasdiener: Yes, I renamed it to local_amax in cf2c8f63

transformer_engine/common/hadamard_transform/hadamard_transform.cu
matthiasdiener: Yes, current_scaling.cu has a general-purpose amax_kernel multi-dtype support. I kept PreRhtAmaxKernel separate here since it's a small BF16-only kernel specific to this file.

transformer_engine/common/hadamard_transform/hadamard_transform.cu
matthiasdiener: Thanks @wangye805 for answering. 

I've added a comment regarding these comments in cf2c8f63. These values aren't tuning knobs, they're determined by the problem structure. kThreadsPerWHT=4 follows from the Kronecker decomposition H16 = H4 x H4, where we reshape the 16-element vector into a 4×4 matrix with one column per thread. This means each thread hold 4 values and the cross-thread butterfly stages use ds_swizzle for the H4. 

Given that and a 64-wide wavefront, the rest follows: kRowsPerWarp = 64/4 = 16 rows per wavefront, and kWarpsPerBlock = 4 gives 64 rows per block

transformer_engine/common/hadamard_transform/hadamard_transform.cu
matthiasdiener: Yes, I think we can consolidate later.

transformer_engine/common/hadamard_transform/hadamard_transform.cu
matthiasdiener: The operations in ZeroAmaxKernel are tiny, so I felt it wasn't necessary to launch that as a kernel.

transformer_engine/common/hadamard_transform/hadamard_transform.cu
matthiasdiener: Like you said, for this kernel, the atomic contention should be relatively small. The two-stage reduction requires a workspace allocation plus another kernel launch. We can revisit if profiling shows this as a bottleneck.

transformer_engine/common/hadamard_transform/hadamard_transform.cu
matthiasdiener: Thanks for the suggestion. I factored this out into reduce_block_amax in cf2c8f63

transformer_engine/pytorch/csrc/pybind.h
matthiasdiener: Done in cf2c8f63

transformer_engine/common/hadamard_transform/hadamard_transform.cu
matthiasdiener: Done in cf2c8f63

wangye805 (approved):

aris134 (approved):

ipanfilo (approved):

PR #511, opened 2026-03-30
TE building over TheRock

ipanfilo:
# Description

Support TE wheels building over ROCm TheRock

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

## Changes

- Update wheels Docker file to use TheRock
- Update different components build and initialization to use TheRock default location.
- Update CI scripts
- Make RTC code changes to be aware of TheRock default layout
- Support comma separated list of GPU arch

# Checklist:


Micky774:
May we also swap these two lines, or modify the second into an append: https://github.com/ROCm/TransformerEngine/blob/2cb27a3f5f256d0c1c70be226ddddcfe1f350435/setup.py#L102-L103

.github/scripts/aiter_prebuild_upload.sh
```diff
@@ -12,7 +12,14 @@ set -euo pipefail
...
 ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
-ROCM_PATH="${ROCM_PATH:-/opt/rocm}"
+if [ -n "$ROCM_PATH" ]; then
+    true # Use provided ROCM_PATH
+elif [ -d "/opt/rocm/core" ]; then
+    ROCM_PATH="/opt/rocm/core"
+else
+    ROCM_PATH="/opt/rocm"
+fi
+export ROCM_PATH
```
AllenFarcas: This fallback logic from `$ROCM_PATH` -> `/opt/rocm/core` -> `/opt/rocm` appears multiple times in multiple files. Not a blocker, but if the fallback order changes there are multiple places to update. Maybe some refactoring to share this logic in one place could help.

build_tools/wheel_utils/Dockerfile.rocm.manylinux.x86
```diff
@@ -16,22 +16,23 @@
...
 
 # Setup packages
-RUN dnf install -y --disablerepo=epel rocm-dev hipblaslt hipblaslt-devel hipcub hipcub-devel hiprand-devel rocrand-devel
-RUN dnf group install -y "Development Tools" && dnf install -y git cmake llvm-toolset gcc-toolset-12
+RUN dnf install -y --disablerepo=epel amdrocm-core-devel-gfx950
 
-#Uncomment the next line for ROCm 6.4 cmake workaround: remove newer incomnpatible cmake preinstalled on base image
-#RUN rm /usr/local/bin/cmake || true
+# xz-devel installs lzma needed by AOTriton
+RUN dnf install -y gcc-toolset-12 xz-devel
```
AllenFarcas: Previously, `git`, `cmake` `llvm-toolset` and `gcc-toolset-12` were installed explicitly and now the only explicity install is for `gcc-toolset-12`. Are they provided by the `amdrocm-core-devel-gfx950` or the base image now?

build_tools/wheel_utils/Dockerfile.rocm.manylinux.x86
```diff
@@ -16,22 +16,23 @@
...
 RUN dnf clean all
 RUN rm -rf /var/cache/dnf/*
 
-ENV CMAKE_PREFIX_PATH=/opt/rocm/lib/cmake
+ENV HIP_PLATFORM=amd
+ENV ROCM_PATH=/opt/rocm/core
+
 ENV NVTE_RELEASE_BUILD=1
 
 ARG GPU_TARGETS="gfx942;gfx950"
```
AllenFarcas, line 40: Since these are the supported GPU targets, does the previously installed `amdrocm-core-devel-gfx950` still support `gfx942`?

build_tools/utils.py
```diff
@@ -196,30 +196,16 @@ def rocm_build() -> bool:
         ValueError: If NVTE_USE_ROCM is set to invalid value.
         FileNotFoundError: If required tools (hipcc or nvcc) are not found.
     """
```
AllenFarcas, line 197: The docstring should be updated to reflect the changes below. Something like:

```
"""
Determines which build platform to use:

- If `NVTE_USE_ROCM` is set:
    - "0": Use CUDA, if nvcc is detected.
    - Any other value: Require ROCm and use it if hipcc is detected.
- If `NVTE_USE_ROCM` is not set:
    - If `HIP_PLATFORM=amd`, require ROCm and use it if hipcc is detected.
    - Otherwise, attempt to auto-detect: Check for ROCm first, then CUDA.

Returns:
    bool: `True` for ROCm, `False` for CUDA.

Raises:
    FileNotFoundError: If ROCm is required but hipcc is not found, or if
        neither ROCm nor CUDA can be detected.
"""
```

.github/scripts/aiter_prebuild_upload.sh
AllenFarcas, line 5: Bug: The `-u` options tells bash to treat unset variables as an error and exit immediately. The line 15 below `if [ -n "$ROCM_PATH" ]; then` expands `ROCM_PATH` before it is set, so the script exits immediately when ROCM_PATH is unset. 

Using `if [ -n "${ROCM_PATH:-}" ]; then` is safer since it expands to an empty string instead of erroring, so the fallback logic can run.

build_tools/wheel_utils/Dockerfile.rocm.manylinux.x86
ipanfilo: This is quite counterintuitive but for building it does not actually make difference which GPU ROCm is installed if all builds libraries provide the same API - ROCm .so are just needed for TE linking. Thus, we use the latest released HW for building and runtime will use amdrocm-core corresponding to actual GPU.

.github/scripts/aiter_prebuild_upload.sh
ipanfilo: I thought about it. Python and shell scripts need it for different purposes and they are also totally independent entities that do not use common shared code.
Good thing is that such changes like switching to different distribution model are extremely rare.
For CMake files it might however be possible to have some rocm_utils.cmake with ROCM_PATH configuring.

build_tools/wheel_utils/Dockerfile.rocm.manylinux.x86
ipanfilo: They are part of base image now

build_tools/utils.py
```diff
@@ -184,16 +184,16 @@ def rocm_build() -> bool:
     Determines which build platform to use:
 
     - If `NVTE_USE_ROCM` is set:
-        - Non-zero value: Use ROCm, if hipcc is detected.
+        - Any value excet "0": Use ROCm, if hipcc is detected.
```
Micky774: Typo: "except"

build_tools/utils.py
```diff
@@ -196,30 +196,16 @@ def rocm_build() -> bool:
         ValueError: If NVTE_USE_ROCM is set to invalid value.
         FileNotFoundError: If required tools (hipcc or nvcc) are not found.
     """
-    nvte_use_rocm = os.getenv("NVTE_USE_ROCM")
-    if nvte_use_rocm:
+    nvte_use_rocm = os.getenv("NVTE_USE_ROCM",
+                              "1" if os.getenv("HIP_PLATFORM", "") == "amd" else "")
+    if nvte_use_rocm != "0":
```
Micky774: When `NVTE_USE_ROCM` is unset and `HIP_PLATFORM` is not "amd", `nvte_use_rocm` defaults to `"" != "0"` so we'll need to either change the default value or the check value.

transformer_engine/common/__init__.py
```diff
@@ -421,7 +421,10 @@ def _load_core_library():
     if te_rocm_build:
         try:
             # Get installed ROCm version
-            with open(os.getenv("ROCM_PATH", "/opt/rocm") + "/.info/version", "r") as f:
+            for rocm_path in (os.getenv("ROCM_PATH"), "/opt/rocm/core", "/opt/rocm"):
+                if rocm_path and os.path.exists(os.path.join(rocm_path, ".info/version")):
+                    break
```
Micky774, line 426: Can we reuse `build_tools.utils.rocm_path`?

build_tools/utils.py
```diff
@@ -231,19 +217,23 @@ def rocm_build() -> bool:
 @functools.lru_cache(maxsize=None)
 def rocm_path() -> Tuple[str, str]:
     """ROCm root path and HIPCC binary path as a tuple"""
-    """If ROCm installation is not specified, use default /opt/rocm path"""
+    """If ROCm installation is not specified, use default ROCm path"""
```
Micky774: style nit: why do we use two docstrings instead of combining into one?

Micky774 (commented):
Overall looks good, just a comment on env variable usage and some small nits.

.github/scripts/aiter_prebuild_upload.sh
Micky774: I ditto Allen's comment. I think it would be good to do so at least for the cmake files. For the python files, we could maybe move `rocm_path()` into the actual runtime library and have the build tools import from it?

build_tools/utils.py
ipanfilo: Well, this is tricky part. HIP_PLATFORM="nvidia" is technically not the same as NVTE_USE_ROCM=0. The former supposes using of HIP compiler to build for NV HW while the 2nd is to use native NV build. But for TE I can indeed for NV native path when HIP_PLATFORM is set to nvidia

transformer_engine/common/__init__.py
ipanfilo: No, build_tools are not part of run-time package

Micky774 (approved):
Looks good, thanks!

PR #519, opened 2026-04-04
"castonly/casttranspose HIP kernel optimization in fp8

alextmagro:
Improvements to cast_transpose and cast for FP8 delayed scaling

Introduced rocm specific cast and cast+transpose functions tuned for MI350s and MI300s

Results are in https://github.com/ROCm/frameworks-internal/issues/16001

### Cast+Transpose Kernel (`rocm_cast_transpose.cuh`)
1. **OVecT packed FP8 shared memory** — smem stores CVec<fp8,8> instead of float, 4x smaller footprint, avoids bank conflicts with +1 padding
2. **Register transpose during load** — accumulate transposed FP8 into `local_t[j2][iter].val[i2]` during the load phase, avoiding a separate transpose pass
3. **Non-temporal stores** for output_c (rowwise) and output_t (transposed) — `__builtin_nontemporal_store` confirmed as `global_store_* ... nt` in assembly
4. **gfx950 FP8 packed intrinsics** — `__builtin_amdgcn_cvt_scalef32_pk_fp8_f32` with scale=1.0 and pre-multiply. Values clamped to FP8 range via `fmed3f` before the intrinsic to prevent Inf-to-NaN (the intrinsic's E8M0 scale param is a post-quantization scale, not a pre-multiply)
5. **word_select packing** — two intrinsic calls with `word_select=false/true` pack 4 FP8 values into one uint32
6. **Two-launch row strategy** — STORE=8 for bulk, then best-fit single launch for remainder (STORE=4 if rem%128==0, STORE=2 if rem%64==0, general kernel otherwise). Max 2 launches for any M value
7. **Single-launch for small tensors** — shapes with M<512 attempt a single kernel launch with the best-fit STORE size, avoiding multi-launch overhead
8. **Column cascade** — LOAD_SZ checks for single-launch alignment, then cascades to smaller LOAD sizes for remainder columns
9. **CVec standalone vector type** — aligned vector struct with `load()`, `store()`, `nt_store()` methods. Shared across cast, cast+transpose, and MXFP8 kernels via `rocm_device_utils.cuh`
10. **BF16/FP16 LOAD capped at 8** — LOAD=16 for BF16 uses 211 VGPRs (2 waves/SIMD). Capping at LOAD=8 uses 125 VGPRs (4 waves/SIMD), doubling occupancy

### Cast-Only Kernel (`rocm_cast.cuh`)
1. **Dedicated 1D grid-stride kernel** — flat 1D indexing over `M*N` elements. No tiling, no cascade, single kernel launch for any shape
2. **256 threads/block, 16 elements/thread** — 42 VGPRs (from assembly), 10 waves/SIMD max, 0 scratch, 32 bytes LDS (amax only)
3. **Direct FP8 packing into OVec** — intrinsic results written directly into the output CVec via `reinterpret_cast<uint32_t*>`. No intermediate `converted[]` array, which preserves the NT store hint through the compiler
4. **Non-temporal stores** — `CVec::nt_store()` confirmed as `global_store_dwordx4 ... nt` in assembly (required eliminating the intermediate array to prevent the compiler from dropping the NT hint)
5. **gfx950 FP8 packed intrinsics** — same as cast+transpose, with `fmed3f` clamping
6. **Dynamic grid sizing** — `cu_count` blocks for FP32 and small BF16 tensors; `cu_count*2` for BF16 tensors >128M elements (crossover point determined empirically)
7. **Scalar tail** for non-aligned element counts (rarely exercised — model dimensions are multiples of 16+)


transformer_engine/common/transpose/rocm_cast_transpose.cuh
```diff
@@ -0,0 +1,390 @@
+/*************************************************************************
+ * Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * License for AMD contributions = MIT. See LICENSE for more information
+ ************************************************************************/
+#pragma once
+//#include "hip/hip_runtime.h" //dummy include to prevent hipification adding this header
```
ipanfilo: Contrary to ROCm specific files in common/cast/mxfp8, this one uses CUDA API and/or datatypes so it will be hipified. 
Let's switch to HIP API then

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -17,6 +17,10 @@
 #include "../utils.cuh"
 #include "cast_transpose.h"
 
+#ifdef __HIP_PLATFORM_AMD__
+#include "rocm_cast_transpose.cuh"
```
ipanfilo, line 23: Should it be in detail namespace? And let's guard cast_transpose_general_kernel ans so on since they are not used anymore

ipanfilo (changes requested):

transformer_engine/common/transpose/rocm_cast_transpose.cuh
alextmagro: I have removed CUDA API code. One note -- I needed to replace cuda::sm_count() to avoid hipification, so have added a static const lambda to grab that value once.

transformer_engine/common/transpose/cast_transpose.cu
alextmagro: It should be like upstream, so have added the dispatch function to detail. unused NV code is now guarded.

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -259,102 +266,41 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *output_, cu
...
-#ifdef __HIP_PLATFORM_AMD__
-            // do_general_config means using the cost model like NVTE to generate kernel configs
-            bool do_general_config = false;
-#endif
             if (aligned && rtc::is_enabled()) {  // Runtime-compiled tuned kernel
-#ifdef __HIP_PLATFORM_AMD__
-              do_general_config = true;
-              // even if we enforce to use OPTIMIZED_HIPIFIED_CAST_TRANSPOSE, may fall back to general kernel configs from NVTE cost model
-              bool nvte_use_optimized_hipified_cast_transpose = false;
-              if (const char* env_p = std::getenv("NVTE_USE_OPTIMIZED_HIPIFIED_CAST_TRANSPOSE") ) {
```
wangye805, line 271: Are you proposing to remove the previous optimized hipified cast transpose PR completely? If so, is it because your new flow will always be faster than the previous one?

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -259,102 +266,41 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *output_, cu
             constexpr size_t itype_size = sizeof(InputType);
             constexpr size_t otype_size = sizeof(OutputType);
 
+#ifdef __HIP_PLATFORM_AMD__
```
wangye805, line 266: Based on the PR description, there are still a small portion of scenarios that your new code is slower than the baseline. For those scenarios, do you think we can keep the old flow and add an env? So that our customers can choose which flow to use?

transformer_engine/common/cast/fp8/rocm_cast.cuh
```diff
@@ -0,0 +1,171 @@
+/*************************************************************************
```
wangye805, line 1: It's a little bit weird to put cast only kernel under the transpose dir. How about put it under cast/fp8/?

transformer_engine/common/transpose/rocm_cast.cuh
```diff
@@ -0,0 +1,171 @@
...
+            rocm_atomicMaxFloat(amax_ptr, amax);
+        }
+    }
+
+    if (blockIdx.x == 0 && threadIdx.x == 0 && scale_inv_ptr != nullptr) {
+        *scale_inv_ptr = __frcp_rn(scale);
+    }
+}
+
+namespace transformer_engine::detail {
```
wangye805: transformer_engine::detail is not appropriate if this is only for fp8 delay scaling casting

transformer_engine/common/cast/fp8/rocm_cast.cuh
```diff
@@ -0,0 +1,171 @@
...
+                      OType       *__restrict__ output,
+                      const float *__restrict__ scale_ptr,
+                      float       *__restrict__ amax_ptr,
+                      float       *__restrict__ scale_inv_ptr,
+                      const size_t              total_elems) {
+    if (noop != nullptr && noop[0] == 1.0f) {
+        return;
+    }
+
+    constexpr int NVEC_IN  = 16 / sizeof(IType);
```
wangye805: Is 16 the same as ROCM_CAST_ELEMS?

transformer_engine/common/transpose/rocm_cast.cuh
```diff
@@ -0,0 +1,171 @@
...
+                float s3 = static_cast<float>(in[l3].val[j3]) * scale;
+
+                v2i16_t r = {0, 0};
+                if constexpr (is_e4m3) {
+                    r = __builtin_amdgcn_cvt_scalef32_pk_fp8_f32(r, s0, s1, 1.0f, false);
+                    r = __builtin_amdgcn_cvt_scalef32_pk_fp8_f32(r, s2, s3, 1.0f, true);
+                } else {
+                    r = __builtin_amdgcn_cvt_scalef32_pk_bf8_f32(r, s0, s1, 1.0f, false);
+                    r = __builtin_amdgcn_cvt_scalef32_pk_bf8_f32(r, s2, s3, 1.0f, true);
+                }
```
wangye805: Why not multiply the scale in __builtin_amdgcn_cvt_scalef32_pk_fp8_fp32? I mean 1.0f-->scale and s0 is just static_cast<float>(in[l0].val[j0])

wangye805 (changes requested):

transformer_engine/common/transpose/rocm_cast_transpose.cuh
```diff
@@ -0,0 +1,393 @@
...
+
+                    if constexpr (NVEC_IN > 2) {
+                        float s2 = (j2 + 2 < NVEC_IN) ? static_cast<float>(in.val[j2 + 2]) * scale : 0.0f;
+                        float s3 = (j2 + 3 < NVEC_IN) ? static_cast<float>(in.val[j2 + 3]) * scale : 0.0f;
+                        if constexpr (is_e4m3) {
+                            r = __builtin_amdgcn_cvt_scalef32_pk_fp8_f32(r, s2, s3, 1.0f, true);
+                        } else {
+                            r = __builtin_amdgcn_cvt_scalef32_pk_bf8_f32(r, s2, s3, 1.0f, true);
+                        }
+                    }
```
wangye805: Looks like this trick is widely used by your recent PRs. Perhaps wrap it up as an inline function to reduce the code duplication?

transformer_engine/common/transpose/rocm_cast_transpose.cuh
```diff
@@ -0,0 +1,393 @@
...
+        }
+        val = v;
+    }
+    return val;
+}
+
+template <int LOAD_SIZE, int STORE_SIZE, int WARPS_PER_TILE,
+          typename IType, typename OType>
+__global__ void __launch_bounds__(ROCM_CT_WARP_SIZE * WARPS_PER_TILE)
+rocm_cast_transpose_kernel(const IType *__restrict__ input,
```
wangye805, line 25: Besides the gfx950 specific optimization, how is this different from the current hiprtc kernel: https://github.com/ROCm/TransformerEngine/blob/f6efbbf3b377b9f5338c67ec49f60ef26ab8d494/transformer_engine/common/transpose/rtc/cast_transpose.cu#L24-L128

transformer_engine/common/transpose/rocm_cast_transpose.cuh
```diff
@@ -0,0 +1,393 @@
...
+template <typename IType, typename OType>
+size_t rocm_cast_transpose_dispatch(const IType *in, const float *noop,
+                                    OType *out_c, OType *out_t,
+                                    const float *scale, float *amax, float *scale_inv,
+                                    size_t row_length, size_t num_rows, hipStream_t stream) {
+    constexpr int WPT = 16;
+    constexpr int OSZ = sizeof(OType);
+
+    size_t rows_done = 0;
+    constexpr int TM8 = ROCM_CT_WARP_SIZE * (8 / OSZ);
```
wangye805: What does the name TM8 mean?

transformer_engine/common/transpose/rocm_cast_transpose.cuh
```diff
@@ -0,0 +1,393 @@
...
+    size_t chunk = (num_rows / TM8) * TM8;
```
wangye805: For this chunk, is it chunk_size of num of chunk?

transformer_engine/common/transpose/rocm_cast_transpose.cuh
```diff
@@ -0,0 +1,393 @@
...
+    if (chunk > 0) {
+        rocm_ct_launch_cols<8, WPT>(in, noop, out_c, out_t, scale, amax, scale_inv,
```
wangye805: Is this 8 the same as the "8" in line 360? If so, can we use a macro/const so that it's easier for future maintenance

transformer_engine/common/transpose/rocm_cast_transpose.cuh
```diff
@@ -0,0 +1,393 @@
...
+                in + row_off * row_length + done, noop,
+                out_c + row_off * row_length + done,
+                out_t + done * num_rows + row_off,
+                scale, amax, scale_inv,
+                sub_rows, rem, row_length, num_rows);
+    }
+}
+
+template <typename IType, typename OType>
+size_t rocm_cast_transpose_dispatch(const IType *in, const float *noop,
```
wangye805, line 286: Can you describe more on your partition logic in this dispatch function? It's hard for me to understand 1). why it works correctly and 2). why it can be better than no partition

transformer_engine/common/transpose/rocm_cast_transpose.cuh
```diff
@@ -0,0 +1,393 @@
...
+ *
+ * License for AMD contributions = MIT. See LICENSE for more information
+ ************************************************************************/
+#pragma once
+//#include "hip/hip_runtime.h" //dummy include to prevent hipification adding this header
+
+#define ROCM_CT_WARP_SIZE 32
+
+template <typename T, int N>
+struct alignas(sizeof(T) * N) CVec {
```
wangye805: Why this is named CVec? Is this a direct equivalent to NV's Vec https://github.com/ROCm/TransformerEngine/blob/f6efbbf3b377b9f5338c67ec49f60ef26ab8d494/transformer_engine/common/transpose/rtc/cast_transpose.cu#L34?

wangye805 (changes requested):

transformer_engine/common/cast/fp8/rocm_cast.cuh
```diff
@@ -0,0 +1,171 @@
...
+    CheckOutputTensor(output, "rocm_cast_only_output");
+
+    NVTE_CHECK(input.has_data(), "Input is not allocated");
+    NVTE_CHECK(output.has_data(), "Output data is not allocated");
+
+    const size_t row_length = input.flat_last_dim();
+    const size_t num_rows   = input.flat_first_dim();
+    const size_t total      = num_rows * row_length;
+
+    static const size_t cu_count = []() {
```
aris134, line 137: Could we use sm_count(0) here? https://github.com/ROCm/TransformerEngine/blob/f6efbbf3b377b9f5338c67ec49f60ef26ab8d494/transformer_engine/common/util/cuda_runtime.cpp#L90

transformer_engine/common/cast/fp8/rocm_cast.cuh
```diff
@@ -0,0 +1,171 @@
...
+        int dev;
+        hipGetDevice(&dev);
+        hipDeviceProp_t prop;
+        hipGetDeviceProperties(&prop, dev);
+        return static_cast<size_t>(prop.multiProcessorCount);
+    }();
+
+    TRANSFORMER_ENGINE_TYPE_SWITCH_INPUT(
```
aris134, line 145: The PR mentions paths for FP32 and BF16, but I think this potentially allows FP16 input?

transformer_engine/common/transpose/rocm_cast.cuh
```diff
@@ -0,0 +1,171 @@
+/*************************************************************************
+ * Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * License for AMD contributions = MIT. See LICENSE for more information
+ ************************************************************************/
+#pragma once
+
+#include "../common.h"
+#include "../utils.cuh"
+#include "rocm_cast_transpose.cuh"
```
aris134: This dependence feels a bit awkward. rocm cast should not depend on transpose. Should the helpers defined in rocm_cast_transpose.cuh be moved into a common header?

transformer_engine/common/transpose/cast_transpose.cu
alextmagro: I think most of those scenarios are latency issues. We can add a env variable, but I want to do that at the end after finishing up the rest of the review, so we know where the performance stands at the end.

transformer_engine/common/transpose/cast_transpose.cu
alextmagro: Not only is my new flow always faster than the previous one, it appears that the upstream flow is always faster than the optimized hipified cast for both MI300 and MI350s anyway.

For real model shapes:
MI350 results: 1.73 TiB/s BW with upstream path, vs 1.17 TiB/s for hip only path

MI300 results: 1.48 TiB/s BW with upstream path, vs 0.90 TiB/s for hip only path


For reference, my path obtains 2.2 TiB/s on MI300 on average.

transformer_engine/common/transpose/rocm_cast_transpose.cuh
aris134: I second this. The logic is a bit subtle and could benefit from a brief high-level comment explaining the dispatch flow.

aris134 (dismissed):

transformer_engine/common/cast/fp8/rocm_cast.cuh
alextmagro: Good point -- I had worked on cast_tranpose first and was in that folder, but have moved it now.

transformer_engine/common/transpose/rocm_cast.cuh
alextmagro: I agree, but couldn't find a pre-existing rocm utils file before. I have created one in utils/rocm_device_utils.cuh, that we can use moving forward.

transformer_engine/common/cast/fp8/rocm_cast.cuh
alextmagro: No, 16 corresponds to the width of our vector load in bytes.

transformer_engine/common/transpose/rocm_cast.cuh
alextmagro: __builtin_amdgcn_cvt_scalef32_pk_fp8_fp32 expects a e8m0 scale, so discards the mantissa that is associated with fp8 scales. I have made changes to this as we discussed briefly to avoid NaN issues, so we now pass the scale's exponent directly in.

transformer_engine/common/transpose/rocm_cast.cuh
alextmagro: Understood, I have moved it to the fp8 namespace which matches upstream's kernels for fp8 casts.

transformer_engine/common/cast/fp8/rocm_cast.cuh
alextmagro: This is a good idea in theory, unfortunately though this is the unhipified version of that file, so referencing cuda_runtime.h would cause the file to be hipified, which we want to avoid. We could, in theory, reference the hipified header but I think we want to avoid that.

transformer_engine/common/cast/fp8/rocm_cast.cuh
alextmagro: That is correct, this would be used for FP16 too. As we have native FP16 support, it should actually be faster than BF16 on average, but I have not benchmarked this yet -- MLPerf is asking specifically for BF16 so that was my main focus. Also of note is that we only benchmark e4m3 out, but this code is used for e5m2. Experiments showed there was no significant difference in the performance between the two.

transformer_engine/common/transpose/rocm_cast_transpose.cuh
alextmagro: This is not a direct equivalent, but similar functionality. The main differences are we have a nontemporal stores which are not possible in the current format of NV Vec. We could refactor some things to make it possible, but I preferred a self contained struct that only uses what we need.

As far as naming goes, CVec just referred to the fact that this is a Clike struct (although I later templated it so it no long really applies). I could rename it to HIPVec or something if you prefer.

transformer_engine/common/transpose/rocm_cast_transpose.cuh
alextmagro: The optimizations made that don't exist in the rtc kernel are listed at the bottom of the Claude comment discussing results -- let me know if those make sense or not.

transformer_engine/common/transpose/rocm_cast_transpose.cuh
alextmagro: Done -- now mxfp8 and fp8 reference this same packed store.

transformer_engine/common/transpose/rocm_cast_transpose.cuh
alextmagro: Tile M 8 -- The tile height that we use with STORE_SIZE=8.

transformer_engine/common/transpose/rocm_cast_transpose.cuh
alextmagro: This is actually the number of rows to compute that are STORE_SIZE=8 aligned. I have renamed it aligned_rows_s8 as such.

transformer_engine/common/transpose/cast_transpose.cu
alextmagro: I decided to add a check for small kernels -- as explained below, we often do more than one kernel launch. For these tiny tensors, it's better to limit it to 1 launch -- I have done that, and will be updating bench results shortly.

transformer_engine/common/transpose/rocm_cast_transpose.cuh
alextmagro: This is the same 8, regarding STORE_SZ. I have added variables to be used.

transformer_engine/common/transpose/rocm_cast_transpose.cuh
alextmagro: Sure, I have added a basic comment going over the flow. Basically, we launch multiple kernels that handle different parts of the tensor based on how it is aligned -- so, if it is not 256 aligned, but it is 64 aligned, we do the portion that is 256 aligned in one launch, then handle the 64 aligned portion in a second launch.

This is particularly relevant for GPT-OSS, which include sizes like 2880 and 5760: If we only launched one kernel, they would fall into Store=2 and Store=4 in a single launch. Doing Store=8 + Store=2 and Store=8 + Store=4 is significantly more efficient.

transformer_engine/common/transpose/cast_transpose.cu
alextmagro: Per Ye's request, we default back to general_kernel for launch dominated tensors (tiny, under 512 rows) so now we only guard the rtc kernel.

transformer_engine/common/cast/fp8/rocm_cast.cuh
aris134: Got it, thanks for the clarification

transformer_engine/common/cast/fp8/rocm_cast.cuh
aris134: Perhaps this is obvious, but why do we want to avoid hipification in this particular case? My understanding was that substantial parts of the codebase already go through hipify, so I want to better understand what makes this file different

transformer_engine/common/transpose/rocm_cast.cuh
aris134: Seems reasonable to me, thanks

transformer_engine/common/transpose/rocm_cast_transpose.cuh
aris134: Thanks for the clarification

transformer_engine/common/cast/fp8/rocm_cast.cuh
alextmagro: My understanding is that any substantial rocm specific code is supposed to go into its own file for licensing and maintenance purposes. Then, for C it sort of makes sense to just write code in HIP instead of CUDA and skip hipification since there's no reason to align with upstream.

transformer_engine/common/util/rocm_device_utils.cuh
```diff
@@ -0,0 +1,99 @@
...
+    v2i16_t r = {0, 0};
+    if constexpr (std::is_same_v<OType, transformer_engine::fp8e4m3>) {
+        r = __builtin_amdgcn_cvt_scalef32_pk_fp8_f32(r, s0, s1, scale, false);
+        r = __builtin_amdgcn_cvt_scalef32_pk_fp8_f32(r, s2, s3, scale, true);
+    } else {
+        r = __builtin_amdgcn_cvt_scalef32_pk_bf8_f32(r, s0, s1, scale, false);
+        r = __builtin_amdgcn_cvt_scalef32_pk_bf8_f32(r, s2, s3, scale, true);
+    }
+    uint32_t result;
+    memcpy(&result, &r, 4);
```
ipanfilo: return __builtin_bit_cast?

transformer_engine/common/util/rocm_device_utils.cuh
ipanfilo, line 1: It does not look like generic device util so it probbaly should got to cast

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -431,6 +393,7 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *output_, cu
                       static_cast<CType *>(output.scale_inv.dptr), row_length, num_rows);
               NVTE_CHECK_CUDA(cudaGetLastError());
             }
+#endif
```
ipanfilo: nit: please add comment to this endif - is seems far from its #ifdef

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -17,18 +17,26 @@
...
+using CType = float;
+#endif
+
 namespace transformer_engine::detail {
 
 namespace {
 
+#ifndef __HIP_PLATFORM_AMD__
 // String with RTC kernel implementation
 #include "string_code_transpose_rtc_cast_transpose_cu.h"
```
ipanfilo, line 29: Need to remove it's generation in CMakeLists.txt then

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -259,102 +268,55 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *output_, cu
...
+              size_t rows_done = rocm_cast_transpose_dispatch<InputType, OutputType>(
+                  static_cast<const InputType *>(input.data.dptr),
+                  reinterpret_cast<const CType *>(noop.data.dptr),
+                  static_cast<OutputType *>(output.data.dptr),
+                  static_cast<OutputType *>(output.columnwise_data.dptr),
+                  static_cast<const CType *>(output.scale.dptr),
+                  static_cast<CType *>(output.amax.dptr),
+                  static_cast<CType *>(output.scale_inv.dptr),
+                  row_length, num_rows, stream);
+              if (rows_done == 0 && rows_done < num_rows) {
```
ipanfilo: Isn't the second condition redundant - num_rows should not be 0 ?

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -259,102 +268,55 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *output_, cu
             constexpr size_t itype_size = sizeof(InputType);
             constexpr size_t otype_size = sizeof(OutputType);
 
+#ifdef __HIP_PLATFORM_AMD__
+            {
+              size_t rows_done = rocm_cast_transpose_dispatch<InputType, OutputType>(
```
ipanfilo, line 269: consider moving using CType =  here

transformer_engine/common/transpose/cast_transpose.cu
ipanfilo: Why aren't move everything including kernels to detail

ipanfilo (changes requested):

transformer_engine/common/util/rocm_device_utils.cuh
alextmagro: Good idea, I have made that change.

transformer_engine/common/util/rocm_device_utils.cuh
alextmagro: Utils here are used by cast/fp8, cast/mxfp8, and transpose/. I also am hoping to use these for normalization in the future, so am hoping we can keep this in the utils/ folder and expand as needed

transformer_engine/common/transpose/cast_transpose.cu
alextmagro: There were a couple uncommented #endifs in this PR, I went ahead and added comments to all for consistency

transformer_engine/common/transpose/cast_transpose.cu
alextmagro: Yes, it's redundant. I have removed it.

transformer_engine/common/transpose/cast_transpose.cu
alextmagro: using CType on line 34 is where upstream has it. I had an duplicate definition, and removed that.

transformer_engine/common/transpose/cast_transpose.cu
alextmagro: I had thought the upstream kernels were in an anonymous namespace, but the general kernel is actually anonymous within the detail namespace -- rtc is included in detail as it is included inside the namespace as well. I have moved our include inside the namespace as well, so it should be included in detail now too.

transformer_engine/common/transpose/cast_transpose.cu
wangye805: Emm, did you try the configs listed in our previous PR https://github.com/ROCm/TransformerEngine/pull/89. For example, I didn't see [2048, 12288] in your benchmark table. I didn't recall every details of that PR but I remember heyi's optimized flow is better than the baseline on MI300X and MI308X.

wangye805: The configs in PR89 was probably from MLPerf's previous submission. And https://amd.atlassian.net/wiki/spaces/~yewang12/pages/435389025/Cast+Transpose+Kernel+Optimization was from semianalysis. Better to try those as well

transformer_engine/common/cast/fp8/rocm_cast.cuh
wangye805: I see. Can you re-encode it using a macro, or constexpr int for future readers?

transformer_engine/common/util/rocm_device_utils.cuh
```diff
@@ -0,0 +1,97 @@
...
+//#include "hip/hip_runtime.h" // prevent hipification of this rocm_ file
+
+#include <type_traits>
+
+#define ROCM_CT_WARP_SIZE 32
+
+#if defined(__gfx950__) && __HIP_DEVICE_COMPILE__
+template <typename OType>
+__device__ __forceinline__
+uint32_t rocm_cvt_4xfp8(float s0, float s1, float s2, float s3, float scale) {
```
wangye805: An ambiguous name: fp8 in TE usually mean fp8e4m3, but here can mean both fp8 and bf8

transformer_engine/common/util/rocm_device_utils.cuh
```diff
@@ -0,0 +1,97 @@
...
+    // Clamp to FP8 max to prevent NaNs from polluting
+    constexpr float FP8_MAX = std::is_same_v<OType, transformer_engine::fp8e4m3>
+                              ? 448.0f : 57344.0f;
```
wangye805: Avoid hardcoding the maximum values. I believe amd_details/fp8.h should have some util functions for those max definition

transformer_engine/common/transpose/rocm_cast_transpose.cuh
wangye805: I didn't know a good tool to show you the logically duplicated parts so I got the following screenshots. 
The left side is our current dev hiprtc cast_transpose kernel and the right side is your new rocm_cast_transpose.cuh. 

Probably I missed your new optimization techniques? I believe those logically duplicated parts can be consolidated to save our future maintenance burden. 

image
image
image

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -17,6 +17,9 @@
 
 #include "../../common.h"
 #include "../../transpose/cast_transpose.h"
+#ifdef __HIP_PLATFORM_AMD__
+#include "../fp8/rocm_cast.cuh"
+#endif
```
ipanfilo: Can it be lower with other ../fp8 headers?

transformer_engine/common/transpose/rocm_cast_transpose.cuh
wangye805: Hi @alextmagro , thanks for the clarification. But I still got some questions. When you say a tensor is 256 aligned or 64 aligned, are you saying their # or rows are 256 aligned, 64 aligned, or their # of columns are 256/64 aligned? Or # of bytes are 256/64 aligned? Thanks

transformer_engine/common/transpose/rocm_cast_transpose.cuh
wangye805: Oh, I'm just trying to understand what the "C" means in your "CVec". I thought it would be a compute type vector. 
Does it live in the same namespace as the upstream Vec? If not we can override with the same name. Or based on your name, something like VecWithNTStore (or something else aligned with existing naming convention)

wangye805 (changes requested):

transformer_engine/common/cast/mxfp8/rocm_quantize_mxfp8.cuh
```diff
@@ -168,15 +170,21 @@ __global__ void __launch_bounds__(THREADS_PER_CHUNK)
...
         uint32_t my_scale = static_cast<uint32_t>(biased_exponent);
         if constexpr (SCALES_PER_GROUP >= 4) {
-          uint32_t s1 = __shfl_down(my_scale, 1 * THREADS_PER_SCALE_X_ROWWISE, THREADS_PER_CHUNK_X_ROWWISE);
-          uint32_t s2 = __shfl_down(my_scale, 2 * THREADS_PER_SCALE_X_ROWWISE, THREADS_PER_CHUNK_X_ROWWISE);
-          uint32_t s3 = __shfl_down(my_scale, 3 * THREADS_PER_SCALE_X_ROWWISE, THREADS_PER_CHUNK_X_ROWWISE);
-          uint32_t packed = (my_scale & 0xFF) | ((s1 & 0xFF) << 8) | ((s2 & 0xFF) << 16) | ((s3 & 0xFF) << 24);
-          if (tid_rowwise_X == 0 && row_valid && col_valid) {
-            const int scale_idx = row * scale_stride_rowwise + scales_rowwise_block_offset_X;
-            reinterpret_cast<uint32_t*>(&scales_rowwise[scale_idx])[0] = packed;
+#pragma unroll
```
wangye805, line 176: I didn't see section of codes used in fp8 delay-scaling cast or cast transpose. Let's put it into a separate PR for better tracking

wangye805 (changes requested):

wangye805:
By the way, can you make the PR title accurate and concise, for example, "cast transpose hipified kernel optimization in fp8 delay-scaling". Therefore it can be easily found by future readers and TE maintainers

transformer_engine/common/util/rocm_device_utils.cuh
```diff
@@ -0,0 +1,97 @@
...
+#include <type_traits>
+
+#define ROCM_CT_WARP_SIZE 32
+
+#if defined(__gfx950__) && __HIP_DEVICE_COMPILE__
+template <typename OType>
+__device__ __forceinline__
+uint32_t rocm_cvt_4xfp8(float s0, float s1, float s2, float s3, float scale) {
+    // Clamp to FP8 max to prevent NaNs from polluting
+    constexpr float FP8_MAX = std::is_same_v<OType, transformer_engine::fp8e4m3>
```
ipanfilo: Add static assert for valid OTypes?

transformer_engine/common/util/rocm_device_utils.cuh
```diff
@@ -0,0 +1,97 @@
...
+    if constexpr (std::is_same_v<OType, transformer_engine::fp8e4m3>) {
+        r = __builtin_amdgcn_cvt_scalef32_pk_fp8_f32(r, s0, s1, scale, false);
+        r = __builtin_amdgcn_cvt_scalef32_pk_fp8_f32(r, s2, s3, scale, true);
+    } else {
+        r = __builtin_amdgcn_cvt_scalef32_pk_bf8_f32(r, s0, s1, scale, false);
+        r = __builtin_amdgcn_cvt_scalef32_pk_bf8_f32(r, s2, s3, scale, true);
+    }
+    return __builtin_bit_cast(uint32_t, r);
+}
+#endif  // #if defined(__gfx950__)
```
ipanfilo: Would it be easier for remaining code to have non gfx950 version of rocm_cvt_4xfp8 that may just do 4 static_cast<OType>?

transformer_engine/common/cast/fp8/rocm_cast.cuh
```diff
@@ -0,0 +1,166 @@
...
+        if constexpr (sizeof(OType) == 1) {
+#pragma unroll
+            for (int e = 0; e < ROCM_CAST_ELEMS; e += 4) {
+                const int l0 = e / NVEC_IN,     j0 = e % NVEC_IN;
+                const int l1 = (e+1) / NVEC_IN, j1 = (e+1) % NVEC_IN;
+                const int l2 = (e+2) / NVEC_IN, j2 = (e+2) % NVEC_IN;
+                const int l3 = (e+3) / NVEC_IN, j3 = (e+3) % NVEC_IN;
+
+                uint32_t packed = rocm_cvt_4xfp8<OType>(
+                    static_cast<float>(in[l0].val[j0]) * scale,
```
ipanfilo, line 73: Why don't you pass actual scale as the last parameter?

transformer_engine/common/cast/fp8/rocm_cast.cuh
```diff
@@ -0,0 +1,166 @@
...
+                    static_cast<float>(in[l1].val[j1]) * scale,
+                    static_cast<float>(in[l2].val[j2]) * scale,
+                    static_cast<float>(in[l3].val[j3]) * scale,
+                    1.0f);
+                memcpy(&out[0].val[e], &packed, 4);
```
ipanfilo: Is memcpy converted to something builtin? And strictly speaking it should be &out[e/NVEC_OUT]

transformer_engine/common/cast/fp8/rocm_cast.cuh
alextmagro: Sure. I have added the macro ROCM_VEC_BTYES to rocm_device_utils and replaced explicit 16s throughout.

transformer_engine/common/cast/fp8/rocm_cast.cuh
alextmagro: The intrinsic's scale parameter is a E8M0 scale. If we were to pass the scale to the intrinsic, we would need to separate the exp from the mantissa. I tried this, but got some issues with NaNs occurring as we couldn't do the clamping necessary.

As far as I can tell, there is only the scaled intrinsic for OCP, so we need to pass in 1.0f.

transformer_engine/common/cast/fp8/rocm_cast.cuh
alextmagro: memcpy coverts into a simple store instruction for constant sizes -- 4 should fit in a single instruction. 

Good catch on the indexing, I have fixed that.

transformer_engine/common/cast/mxfp8/rocm_quantize_mxfp8.cuh
alextmagro: Hi Ye, we discussed this offline but this change is related to your request to use a helper function for the intrinsic trick, rocm_cvt_4xfp8.

transformer_engine/common/transpose/cast_transpose.cu
alextmagro: Results for those configs have all been added to the issue.

transformer_engine/common/transpose/rocm_cast_transpose.cuh
alextmagro: We discussed offline, but that is # of rows. How many rows for each STORE_SZ is determined by the IType.

transformer_engine/common/util/rocm_device_utils.cuh
alextmagro: I have renamed it to float8 instead of fp8 to be more clear.

transformer_engine/common/util/rocm_device_utils.cuh
alextmagro: I don't think so, because of how scales are handled differently between mxfp8 and fp8. I also added a default value for scale in the function, hopefully to make it more clear what paths use it (MXFP) vs those that don't.

transformer_engine/common/transpose/rocm_cast_transpose.cuh
alextmagro: While we share a some code with the rtc kernel, we have a few static optimizations that would make unifying things hard, like gfx950 intrinsics and NT stores that are type based.

I think we discussed this, but I am not sure there is much benefit to doing things rtc instead of templating other than keeping build times slightly shorter.

alextmagro: Regarding maintenance, I agree that it is a bit extra work sometimes to keep things separate. However, for performance sensitive kernels, I think a pattern is coming out where upstream kernels perform poorly as is for us in most cases, and I think the tradeoff is worth it in cases like these.

transformer_engine/common/transpose/rocm_cast_transpose.cuh
alextmagro: I think Vec lives in the transformer_engine namespace, so would be a bit confusing if we had the same name. I have renamed it to NTVec to specific that it is to be used for situations where a non-temporal load/store is beneficial.

transformer_engine/common/transpose/rocm_cast_transpose.cuh
wangye805: Oh, I'm not asking to choose just one implementation over the other, nor making your impl into hiprtc. I'm asking you whether you can consolidate those two similar/duplicated pieces of codes, in the pure coding style point of view. For example, put the common parts into some header, use macro of constexpr to add your specific things like gfx950 intrinsics and NT stores

transformer_engine/common/transpose/rocm_cast_transpose.cuh
wangye805: NTVec is a better name. Thanks

transformer_engine/common/cast/mxfp8/rocm_quantize_mxfp8.cuh
wangye805: Okay, we can put this in this PR. I'll edit the previous PR description to mention about this fix. 

But just wanted to clarify, in my mind, this fix is independent of the following rocm_cvt_4xfp8 below. plus, rocm_cvt_4xfp8 is gfx950 only but this __shfl_down optimization can be in gfx942 as well?

wangye805 (changes requested):

alextmagro:
> By the way, can you make the PR title accurate and concise, for example, "cast transpose hipified kernel optimization in fp8 delay-scaling". Therefore it can be easily found by future readers and TE maintainers

I have changed the title -- one difference, we do touch FP8 Current scaling with this change as well -- hoping to run some e2e benchmarks w/ and w/o this PR to see the overall effect on both options. Will sync with Sudharshan for that.

transformer_engine/common/cast/fp8/rocm_cast.cuh
```diff
@@ -0,0 +1,165 @@
...
+        hipGetDevice(&dev);
+        hipDeviceProp_t prop;
+        hipGetDeviceProperties(&prop, dev);
+        return static_cast<size_t>(prop.multiProcessorCount);
+    }();
+
+    TRANSFORMER_ENGINE_TYPE_SWITCH_INPUT(
+        input.dtype(), InputType,
+        TRANSFORMER_ENGINE_TYPE_SWITCH_OUTPUT(
+            output.dtype(), OutputType,
```
ipanfilo, line 148: It is in FP8 space but neither Input nor Output datatype switches are FP8

alextmagro: This cast is for FP8 delayed scaling only, we can change the output switch to FP8. This also aligns us with VectorizedUnaryKernel which this function replaces

wangye805 (approved):

benchmarks/cpp/cast/bench_casttranspose.cpp
```diff
@@ -138,6 +143,9 @@ static void BM_CastOnly(benchmark::State &state) {
   HIP_CHECK(hipEventCreate(&stop));
 
   warmup_gpu();
+  // Untimed call to trigger any RTC compilation before measurement
```
ipanfilo, line 133: nit: It can be instead of before warmup_gpu, in the latter case no extra hipDeviceSynchronize is needed

transformer_engine/common/util/rocm_device_utils.cuh
```diff
@@ -24,6 +25,26 @@ uint32_t rocm_cvt_4xfloat8(float s0, float s1, float s2, float s3, float scale =
...
+    } else {
+        r = __builtin_amdgcn_cvt_pk_bf8_f32(s0, s1, r, false);
+        r = __builtin_amdgcn_cvt_pk_bf8_f32(s2, s3, r, true);
+    }
+    return static_cast<uint32_t>(r);
+}
+#endif  // __has_builtin(__builtin_amdgcn_cvt_pk_fp8_f32)
+
+// Scaled FP8 packed conversion for MXFP8 (E8M0 block scale)
+#if defined(__gfx950__) && __HIP_DEVICE_COMPILE__
```
ipanfilo: Should it also be `__HIP_DEVICE_COMPILE__ && __has_builtin(__builtin_amdgcn_cvt_scalef32_pk_fp8_f32)`?

alextmagro: Yes, missed this.

ipanfilo: So, maybe the code where rocm_cvt_4xfloat8 is used should also be guarded similar way or add synthetic define like _HAS_CVT_4xFLOAT8_

alextmagro: I've added synthetic defines for both the scaled and unscaled pack

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -269,53 +265,144 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *output_, cu
...
+                          ot + rows_done,
+                          static_cast<const CType *>(output.scale.dptr),
+                          static_cast<CType *>(output.amax.dptr),
+                          static_cast<CType *>(output.scale_inv.dptr),
+                          rem, row_length, row_length, num_rows);
+                }
+                NVTE_CHECK_CUDA(cudaGetLastError());
+              } else
+#endif  // #ifdef __HIP_PLATFORM_AMD__
+              {
```
ipanfilo, line 312: lint: keep { in-line with else; it will require to guard corresponding  too }
nit: de-intent the above block one step. it will look better to have the block above aligned with opening { than having inner code below having less indent than outer indent. and btw, is opening { at line 267 is needed at all?

transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -259,6 +263,54 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *output_, cu
             constexpr size_t itype_size = sizeof(InputType);
             constexpr size_t otype_size = sizeof(OutputType);
 
+#ifdef __HIP_PLATFORM_AMD__
+            {
+              const char *env_p = std::getenv("NVTE_USE_OPTIMIZED_HIPIFIED_CAST_TRANSPOSE");
+              if (!(env_p && std::string(env_p) == "1")) {
```
ipanfilo: nit: env_p == nullptr || std::string(env_p) != "1" will be more readable

transformer_engine/common/transpose/cast_transpose.cu
ipanfilo: it is always true because false condition is handled at line 269

transformer_engine/common/transpose/cast_transpose.cu
alextmagro: Done -- I had needed it before when dealing with some local vars, now it is no longer needed.

ipanfilo (approved):
TODO: lint at cast_transpose.cu line 312

PR #527, opened 2026-04-07
Gfx1250 changes

ipanfilo:
# Description

Port gfx1250 support changes from NPI branch.
It only includes changes in TE itself, no attention backends update.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Modify wavefront size to 32 where 64 was used
- Add gfx1250 to TE GPU specific logic
- Add gfx1250 to GEMM tests
- Update AITER build to build for both V3 and V2 GPU architectures simultaneously
- Add mode to use AITER Triton kernels for Flash Attention (controlled by NVTE_FLASH_ATTN_AITER env)

# Checklist:


examples/jax/encoder/common.py
```diff
@@ -25,8 +25,8 @@ def is_bf16_supported():
     """Return if BF16 has hardware supported"""
     gpu_arch = get_device_compute_capability(0)
     if is_hip_extension():
-        # only GFX9.4+ and MI200 machines support bf16
-        return 100 > gpu_arch >= 94 or is_mi200()
+        # only GFX9.4+, GFX12+ and MI200 machines support bf16, that excludes GFX10 and GFX11
```
wenchenvincent, line 28: gfx10* and gfx11* are Navi cards? Is is that the HW does not support bf16, or our TE ROCm does not support those cards?

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -266,7 +266,7 @@ __forceinline__ __device__ int binary_search(int32_t target, const int32_t *arra
   return left - 1;
 }
 
-constexpr int THREADS_PER_WAVEFRONT = 64;
+constexpr int THREADS_PER_WAVEFRONT = 32;
```
wenchenvincent, line 303: @wangye805 Did you check if there is any perf decrease on gfx942 or gfx950 if we use `THREADS_PER_WAVEFRONT = 32`?

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -87,8 +87,9 @@
 def get_cublas_workspace_size_bytes() -> None:
     """Return workspace size needed for current architecture"""
     if is_hip_extension():
-        """Return 64 MiB for gfx50x, 32 MiB for all other architectures."""
-        if get_device_compute_capability(0) == 95:
+        """Return 64 MiB for gfx50x+, 32 MiB for all other architectures."""
+        """TODO: gfx1250 WS size requirements"""
+        if get_device_compute_capability(0) in (95, 125):
```
wenchenvincent, line 91: Shall we check with hipblasLt? What's the observation on gfx1250?

transformer_engine/pytorch/utils.py
```diff
@@ -483,7 +483,7 @@ def assert_dim_for_all_gather(
 def is_bf16_compatible() -> None:
     if IS_HIP_EXTENSION:
         # only MI200 and newer machines support bf16
-        if get_device_compute_capability() in [(9, 4), (9, 5)] or is_mi200():
+        if get_device_compute_capability() in ((9, 4), (9, 5), (12, 5)) or is_mi200():
```
wenchenvincent, line 486: Shall we rename `is_mi200` to `is_cdna2`?

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -35,25 +35,6 @@ else()
   message(WARNING "Python interpreter not found; skipping AITER API validation.")
 endif()
 
-# so far, there are only gfx942 and gfx950 v3 kernels
```
wenchenvincent, line 56: What's the change in this file for?

transformer_engine/common/ck_fused_attn/CMakeLists.txt
wangye805, line 1: Please update this file as TE dev now switch to QoLA

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
wangye805: Well, it will need testing on real HW. But will you need this change in gfx942/gfx950?

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -91,7 +91,31 @@
...
+        )
+        from aiter.ops.triton.mha import (
+            flash_attn_onekernel_backward as _flash_attn_varlen_bwd,
+        )
+    except ImportError as e:
+        pass
+    else:
+        fa_utils.use_aiter_triton = True
+        # Setup Flash attention utils
+        fa_utils.version = PkgVersion("2.7.1")  #masqurade as FA 2.7.1
```
wangye805, line 116: Two questions: 
1). Do we plan to switch to aiter flash-attn for gfx942/gfx950/gfx1250 all together?
2). If we do, is this fa_utils.version still as 2.7.1?

wangye805 (changes requested):

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -132,8 +114,17 @@ set_target_properties(ck_fused_attn PROPERTIES INSTALL_RPATH "$ORIGIN")
 install(FILES ${__AITER_MHA_PATH}/libmha_fwd.so ${__AITER_MHA_PATH}/libmha_bwd.so DESTINATION ${CMAKE_INSTALL_PREFIX}/${AITER_MHA_INSTALL_PREFIX}/lib)
 install(TARGETS ck_fused_attn DESTINATION ${CMAKE_INSTALL_PREFIX}/${AITER_MHA_INSTALL_PREFIX}/lib)
 # copy v3 kernels to destination
-foreach(ARCH IN LISTS V3_ASM_ARCHS)
+if(NOT EXISTS "${__AITER_SOURCE_DIR}/hsa")
+  message(FATAL_ERROR "Cannot find AITER v3 kernels location at ${__AITER_SOURCE_DIR}/hsa.")
+  continue()
+endif()
```
AllenFarcas: The `message(FATAL_ERROR ...)` terminates the CMake configuration so the `continue()` is unreachable. Moreover, `continue()` is only valid inside a `foreach` / `while` loop and here it is at top level which would make it a CMake error if it would be reached.

transformer_engine/common/ck_fused_attn/CMakeLists.txt
ipanfilo: AITER/CK building bug that limits them to V3 archs only is fixed a while ago. Right now there is no need to perform this filtering on TE side. Since AITER might have no V3 kernels for GFX1250, this filtering is removed so it can be built despite V3 support in AITER

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -35,25 +35,6 @@ else()
   message(WARNING "Python interpreter not found; skipping AITER API validation.")
 endif()
 
-# so far, there are only gfx942 and gfx950 v3 kernels
-SET(V3_ASM_ARCHS_SUPPORTED "gfx942;gfx950")
```
AllenFarcas, line 57: Do we know AITER actually generates v3 kernels for gfx1250 without having the V3 support specified directly?

transformer_engine/pytorch/attention/dot_product_attention/backends.py
ipanfilo: This code is run on emu and real HW, so answer is yes. For extra safety it might make sense to disable this by default (make env default 0)

transformer_engine/pytorch/utils.py
ipanfilo: It is historically is_mi200 and it checks specifically for MI2.0, so current name better reflect what the method does

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
ipanfilo: For real Gfx1250 HW it should be 32. It also verified to work on Gfx9*. I changed it unconditionally because otherwise it will require runtime checking on host side

examples/jax/encoder/common.py
ipanfilo: No BF16 support in HW

transformer_engine/common/ck_fused_attn/CMakeLists.txt
ipanfilo: It is not to instruct AITER to make V3 kernels, it is to reflect what does AITER support. There used to be a bug in AITER that it could only compile for V3 supported archs,

transformer_engine/pytorch/quantization.py
```diff
@@ -75,9 +75,9 @@ def check_mxfp8_support() -> Tuple[bool, str]:
         if os.getenv("NVTE_ROCM_ENABLE_MXFP8", "0") == "0":
             return False, "MXFP8 support is not enabled."
         gpu_arch = get_device_compute_capability()
-        if gpu_arch == (9, 5):
+        if gpu_arch in ((9, 5), (12, 5)):
```
AllenFarcas, line 79: There is inconsistent gfx1250 detection across files:
- `gpu_arch in ((9, 5), (12, 5))` is for exact match
- examples/jax/encoder/common.py `gpu_arch >= 120` matches any gfx12x and tests/cpp/operator/test_cublaslt_gemm.cu `prop.major >= 12` has the same broad match

If only gfx1250 supports MXFP8/FP8 here, the JAX example and C++ test will incorrectly enable tests on gfx1200 for example. Maybe we could pick one canonical predicate and use it everywhere to be consistent in what the C++ and Python runtime accept.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -91,7 +91,31 @@
...
+        from aiter.ops.triton.mha import _flash_attn_forward as _flash_attn_fwd
+        from aiter.ops.triton.mha import flash_attn_onekernel_backward as _flash_attn_bwd
+        from aiter.ops.triton.mha import (
+            _flash_attn_forward as _flash_attn_varlen_fwd,
+        )
+        from aiter.ops.triton.mha import (
+            flash_attn_onekernel_backward as _flash_attn_varlen_bwd,
+        )
+    except ImportError as e:
+        pass
```
AllenFarcas: Maybe it would be better to log the error instead of silently fall through with no diagnostics?  Something like `attn_log.fa_logger.debug(...)`

transformer_engine/pytorch/quantization.py
```diff
@@ -53,10 +53,10 @@ def check_fp8_support() -> Tuple[bool, str]:
     """Return if fp8 support is available"""
     if IS_HIP_EXTENSION:
         gpu_arch = get_device_compute_capability()
-        if gpu_arch in ((9, 4), (9, 5)):
+        if gpu_arch in ((9, 4), (9, 5), (12, 5)):
             return True, ""
         else:
-            return False, "Device arch gfx94x or gfx95x required for FP8 execution."
+            return False, "Device arch gfx94x or newer required for FP8 execution."
```
AllenFarcas: Nit: Technically, gfx10xx and gfx11xx are newer but they are explicitly excluded in the if statement. Maybe we could just say `Device arch gfx94x, gfx95x or gfx125x required for FP8 execution.`

transformer_engine/pytorch/attention/dot_product_attention/backends.py
wangye805: Thanks for your confirmation. Emm, I'm okay to use aiter as default instead if aiter is available

transformer_engine/jax/cpp_extensions/gemm.py
ipanfilo: It is confirmed, I just need remove TODOs

transformer_engine/pytorch/quantization.py
ipanfilo: Actually, gfx10 and gfx11 are older than gfx9.4

transformer_engine/pytorch/quantization.py
ipanfilo: Feature support depends not on HW only but also on underlying libraries. For instance missing GEMMs do not allow to enable some features on Pytorch while JAX does not have problem because not using TE for it. So in general it is valid point bu we cannot blindly make the same conditions and we do not have enough test coverage on different platform.

build_tools/rocm_utils.cmake
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) 2022-2026, Advanced Micro Devices, Inc. All rights reserved.
 # License for AMD contributions = MIT. See LICENSE for more information
 
+include_guard(GLOBAL)
```
wangye805, line 4: Just curious, why do we need this guard here? To protect ROCM_PATH or CMAKE_HIP_ARCHITECTURE included multiple times?

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
wangye805: 32 should be okay with the logic correctness. But how about performance on gfx942/950?

build_tools/rocm_utils.cmake
ipanfilo: It is to make changes in ck_fiused_attn/CMakeLists.txt that disable gfx1250 working. Otherwise, rocm_utils.cmake is later included by aiter_prebuild.cmake and overwrites all modifications

wangye805: I see. Thanks

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
ipanfilo: That is open question. Because it is the only place where we actually use 64 I do not expect big difference. Did you see uplift from using 64 when implemented it originally?

wangye805 (approved):
LGTM since it's for 1250 branch only

PR #518, opened 2026-04-02
NVFP4 recipe with GEMM via BF16 dequant

matthiasdiener:
# Description

Part of https://github.com/ROCm/frameworks-internal/issues/15682

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Please list the changes introduced in this PR:

- Change A
- Change B

# Checklist:

- [ ] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [ ] The functionality is complete
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #537, opened 2026-04-13
Full MXFP4 Training Recipe

sarthak-amd:
## Summary

1. Introduces `nvte_cast_transpose_mxfp4_fused_shuffle` — fused HIP kernel for MXFP4 cast+transpose with optional Hadamard transform and memory-layout shuffling for GEMM. 

2. `fp4_gemm_handler` that dispatches A4W4 GEMM calls to the AITER backend with the right layouts.

3. MXFP4 weight caching in `Linear.forward()` and `LayerNormLinear.forward()` that persists quantized MXFP4TensorStorage weights across forward passes.

Training recipe can be enabled using:
```
export FP4_RECIPE=mxfp4
export NVTE_MXFP4_USE_HADAMARD=1
export FP4=True
```

## MXFP4 Flow:
image

## Loss graph:
image

## Test plan
- [ ] CI passes (linkage error resolved)
- [x] C++ gtest: `test_cast_mxfp4_transpose.cu` passes on gfx950
[mxfp4_cpp_gfx942.log](https://github.com/user-attachments/files/26913810/mxfp4_cpp_gfx942.log)
[mxfp4_cpp_gfx950.log](https://github.com/user-attachments/files/26913811/mxfp4_cpp_gfx950.log)
- [x] Python pytest: `test_mxfp4_*` tests pass on gfx950
[mxfp4_all_gfx942.log](https://github.com/user-attachments/files/26913816/mxfp4_all_gfx942.log)
[mxfp4_all_gfx950.log](https://github.com/user-attachments/files/26913817/mxfp4_all_gfx950.log)
- [x] End-to-end MXFP4 training with Megatron


tests/cpp/operator/test_cast_transpose_mxfp4_shuffled.cu
```diff
@@ -0,0 +1,139 @@
...
+namespace {
+
+constexpr int MXFP4_BLOCK_SIZE = 32;
+
+inline int cdiv(int a, int b) { return (a + b - 1) / b; }
+
+static const float E2M1_LUT[16] = {
+    0.0f, 0.5f, 1.0f, 1.5f, 2.0f, 3.0f, 4.0f, 6.0f,
+    -0.0f, -0.5f, -1.0f, -1.5f, -2.0f, -3.0f, -4.0f, -6.0f,
+};
```
wangye805: I recall I saw similar lut under nvfp4 cast_transpose or dequantization tests, maybe they are in test_common.h/cu

tests/cpp/operator/test_cast_transpose_mxfp4_shuffled.cu
```diff
@@ -0,0 +1,139 @@
...
+
+float decode_fp4(uint8_t nibble) { return E2M1_LUT[nibble & 0xF]; }
+
+bool is_gfx950() {
+    hipDeviceProp_t prop;
+    hipGetDeviceProperties(&prop, 0);
+    return prop.major == 9 && prop.minor == 5;
+}
```
wangye805: Try to see if existing functionalities can do this and consolidate if possible

tests/cpp/operator/test_cast_transpose_mxfp4_shuffled.cu
```diff
@@ -0,0 +1,139 @@
...
+
+struct MXFP4TestParams {
+    int M;
+    int N;
+    bool use_hadamard;
```
wangye805: I recall they also had a knob to choose whether or not shuffle the scale? Is it true? If true, let's add a bool scale_shuffle as well

tests/cpp/operator/test_cast_transpose_mxfp4_shuffled.cu
```diff
@@ -0,0 +1,139 @@
...
+    ASSERT_EQ(hipMalloc(&d_input, M * N * sizeof(hip_bfloat16)), hipSuccess);
+    ASSERT_EQ(hipMalloc(&d_row_fp4, M * N / 2), hipSuccess);
+    ASSERT_EQ(hipMalloc(&d_row_scale, rowwise_scale_M_pad * rowwise_scale_N_pad), hipSuccess);
+    ASSERT_EQ(hipMalloc(&d_col_fp4, N * M / 2), hipSuccess);
+    ASSERT_EQ(hipMalloc(&d_col_scale, colwise_scale_M_pad * colwise_scale_N_pad), hipSuccess);
+
+    ASSERT_EQ(hipMemset(d_row_fp4, 0, M * N / 2), hipSuccess);
+    ASSERT_EQ(hipMemset(d_col_fp4, 0, N * M / 2), hipSuccess);
+
+    ASSERT_EQ(hipMemcpy(d_input, h_input.data(), M * N * sizeof(hip_bfloat16),
```
wangye805: Avoid allocating host side memory via hipMalloc, test Tensor class already contain the host side memory management. See our previous nvfp4 quantization/dequantization PR as an example

tests/cpp/operator/test_cast_transpose_mxfp4_shuffled.cu
```diff
@@ -0,0 +1,139 @@
...
+
+    const int colwise_scale_M = N;
+    const int colwise_scale_N_val = cdiv(M, MXFP4_BLOCK_SIZE);
+    const int colwise_scale_M_pad = cdiv(N, 256) * 256;
+    const int colwise_scale_N_pad = cdiv(colwise_scale_N_val, 8) * 8;
+
+    std::mt19937 rng(42);
+    std::normal_distribution<float> dist(0.0f, 1.0f);
+    std::vector<hip_bfloat16> h_input(M * N);
+    for (auto& v : h_input) v = hip_bfloat16(dist(rng));
```
wangye805: bf16 random number generation is already in our test utils

tests/pytorch/mxfp4/test_mxfp4_gemm_exact.py
```diff
@@ -0,0 +1,179 @@
...
+import transformer_engine.pytorch as te
+import transformer_engine_torch as tex
+from transformer_engine.pytorch.tensor.mxfp4_tensor import MXFP4Quantizer
+from transformer_engine.pytorch.custom_recipes.quantization_mxfp4 import MXFP4QuantizerRef
+
+
+recipe_available, reason_for_no_recipe = te.is_mxfp4_available(return_reason=True)
+
+try:
+    import aiter  # noqa: F401
```
wangye805, line 21: What is noqa 401?

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -21,6 +21,7 @@
 #include "../core/common.cuh"
 #include "../fp8/quantize_fp8.cuh"
 #include "../mxfp8/quantize_mxfp8.cuh"
+#include "../mxfp4/quantize_mxfp4.cuh"
```
wangye805, line 28: Guard our rocm specific changes

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -93,6 +94,17 @@ void quantize_fwd_helper(const NVTETensor input, NVTETensor output,
           dummy_workspace_tensor, stream);
       break;
     }
+    #ifdef __HIP_PLATFORM_AMD__
```
wangye805: No space needed before macros like #ifdef or #endif. 

There seems to be several other places below

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -93,6 +94,17 @@ void quantize_fwd_helper(const NVTETensor input, NVTETensor output,
...
+    case NVTE_MXFP4_1D_SCALING: {
```
wangye805, line 110: Also, try to see if this mxfp4 flow is guarded on MI300X as I recall MI300X does not support mxfp4 format

transformer_engine/common/cast/mxfp4/cast_transpose_mxfp4_shuffled.cuh
```diff
@@ -0,0 +1,743 @@
...
+    uint32_t tmp = 0;
+    asm volatile("v_cvt_scalef32_pk_fp4_f32 %0, %1, %2, %3"
+                 : "+v"(tmp)
+                 : "v"(v2), "v"(v3), "v"(scale));
+    
+    // Combine into 16-bit result (4 FP4 values)
+    result |= (tmp << 8);
+    return (uint16_t)(result & 0xFFFF);
+#else
+    return 0;  // Fallback for non-gfx950 architectures
```
wangye805: Should we error out? Or fall back to naive quantization and pack?

transformer_engine/common/cast/mxfp4/cast_transpose_mxfp4_shuffled.cuh
```diff
@@ -0,0 +1,743 @@
...
+ * Permutation formula:
+ *   i0 = row // 32
+ *   i1 = (row % 32) // 16
+ *   i2 = row % 16
+ *   i3 = col // 8
+ *   i4 = (col % 8) // 4
+ *   i5 = col % 4
+ *   index = i0*(scale_n_pad//8)*256 + i3*256 + i5*64 + i2*4 + i4*2 + i1
+ */
+__device__ __forceinline__ int compute_shuffle_index(
```
wangye805: nit, rename this to compute_scale_shuffle_index? to avoid confusion with the fp4 data shuffle below

transformer_engine/common/cast/mxfp4/quantize_mxfp4.cuh
```diff
@@ -0,0 +1,91 @@
...
+      use_rowwise ? output->data.dptr : nullptr,
+      use_rowwise ? output->scale_inv.dptr : nullptr,
+      use_colwise ? output->columnwise_data.dptr : nullptr,
+      use_colwise ? output->columnwise_scale_inv.dptr : nullptr,
+      M, N,
+      use_rowwise, use_colwise,
+      scale_shuffle,
+      use_hadamard,
+      data_shuffle,
+      data_shuffle,
```
wangye805: So data shuffle controls both rowwise and columnwise shuffle at the same time?

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -99,6 +99,11 @@ enum NVTEScalingMode {
   /*! Single scale per block of 16 elements consecutive in either
    * rowwise or columnwise direction */
   NVTE_NVFP4_1D_SCALING = 4,
+  #ifdef USE_ROCM
+  /*! Single scale per block of 32 elements consecutive in either
+      rowwise or columnwise direction */
+  NVTE_MXFP4_1D_SCALING = 5,
```
wangye805: Not sure whether NV upstream wanted to have mxfp4 later, better to use a large number, like 50 to avoid future conflict since the NVTE_INVALID_SCALING is 100 here

transformer_engine/common/recipe/__init__.py
```diff
@@ -570,8 +574,10 @@ class MXFP4BlockScaling(Recipe):
 
     margin: int = 0
     fp4_format: Format = Format.E2M1
+    fp8_format: Format = Format.E4M3
     fp8_dpa: bool = False
     fp8_mha: bool = False
+    use_hadamard: bool = os.getenv("NVTE_MXFP4_USE_HADAMARD", "0") == "1"
```
wangye805, line 582: Is there a better way to pass this in than the env?

transformer_engine/common/common.h
```diff
@@ -421,6 +436,11 @@ struct QuantizationConfig {
   bool nvfp4_2d_quantization = false;
   bool stochastic_rounding = false;
   bool use_fast_math = false;
```
wangye805, line 451: Looks like nv upstream does not put the hadamard usage into the quantization config, can we align with them?

transformer_engine/common/transformer_engine.cpp
```diff
@@ -176,7 +180,34 @@ void CheckScaleTensorShape(const Tensor &t, const std::string &name) {
                    "\"  has invalid columnwise_scale_inv shape (expected ", expected, ", got ",
                    t.columnwise_scale_inv.shape, ")");
       }
+    } 
+    #ifdef __HIP_PLATFORM_AMD__
+    else if (t.scaling_mode == NVTE_MXFP4_1D_SCALING) {
+      const size_t row_alignment = 256;
+      const size_t col_alignment = 8;
+
+      const size_t block_size = 32;
```
wangye805, line 189: Let's rename them to some constants and share the usage somewhere else? To avoid hard coding

transformer_engine/common/transformer_engine.cpp
```diff
@@ -206,6 +237,27 @@ void CheckInputTensor(const Tensor &t, const std::string &name) {
...
+    }
+    if (t.has_columnwise_data()) {
+      NVTE_CHECK(t.columnwise_scale_inv.has_data(), "FP4 scaling factor input ", name,
+                 "_columnwise_scale_inverse must be allocated");
+      NVTE_CHECK(t.columnwise_scale_inv.dtype == expected_scale_dtype,
+                 "FP4 scaling factor input ", name,
+                 "_columnwise_scale_inverse has invalid dtype "
+                 "(expected ", to_string(expected_scale_dtype), ", got ",
+                 to_string(t.columnwise_scale_inv.dtype), ")");
+    }
```
wangye805, line 262: nit: avoid duplicating this scale check as it's already in the else branch, you can do something like
#if rocm
DType expected_scale_dtype=is_mxfp4? e8m0: e4m3
#else 
DType expected_scale_dtype=e4m3
#endif
Then the shared if(t.has_data), if (t.has_columnwise_data())

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -23,6 +23,10 @@
 from ..custom_recipes.gemm import custom_gemm
 from ...debug.pytorch.debug_quantization import DebugQuantizer
 
+from ..utils import cast_if_needed
```
wangye805: Don't forget to guard our rocm specific changes

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -99,6 +103,158 @@ def get_tensor_device(tensor: torch.Tensor) -> int:
...
+    accumulate: bool = False,
+):
+    """FP4 GEMM using layout notation (TN/NN/NT).
+
+    Layout mapping:
+        TN: A=weight, B=input       -> fprop: input @ weight^T
+        NN: A=weight, B=grad_output -> dgrad: grad_output @ weight
+        NT: A=input, B=grad_output  -> wgrad: grad_output^T @ input
+    """
+    with torch._C._DisableTorchDispatch():
```
wangye805: What's this torch._C._DisableTorchDispatch() used for?

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -99,6 +103,158 @@ def get_tensor_device(tensor: torch.Tensor) -> int:
...
+        result = aiter.gemm_a4w4_asm(
+            A_fp4, B_shuffled, A_scales_uint8, B_scales_uint8,
+            out_hp, kernel_name, None,
+            bpreshuffle=True, log2_k_split=log2_k_split,
+        )
+
+    return result[:M, :] if result.shape[0] > M else result
+
+
+def fp4_gemm_layout(
```
wangye805: This function name is misleading, it looks like checking some layout or return some layout

transformer_engine/pytorch/csrc/type_converters.cpp
```diff
@@ -126,6 +128,46 @@ TensorWrapper NVTETensorFromFloat8BlockwiseQTensor(py::handle tensor, Quantizer
...
+    const auto &data = tensor.attr("_columnwise_data").cast<at::Tensor>();
+    const auto &scale_inv = tensor.attr("_columnwise_scale_inv").cast<at::Tensor>();
+    ret.set_columnwise_data(data.data_ptr(), dtype,
+                            convert_shape_back_from_fp4(getTensorShape(data), true));
+    ret.set_columnwise_scale_inv(scale_inv.data_ptr(), DType::kFloat8E8M0,
+                                 getTensorShape(scale_inv));
+  }
+
+  // Scale layout
+  ret.set_with_gemm_swizzled_scales(with_gemm_swizzled_scales);
```
wangye805, line 163: Do we need to set the data shuffle flag to notify later gemm that data is also pre-shuffled?

transformer_engine/pytorch/custom_recipes/quantization_mxfp4.py
```diff
@@ -0,0 +1,586 @@
...
+    * E8M0 scale: ``floor(log2(amax_rounded)) - 2 + 127``
+    * FP4 encoding: threshold-based nearest-neighbour to E2M1 values
+    * Packing: ``(odd_nibble << 4) | even_nibble``
+    """
+
+    def __init__(
+        self,
+        rowwise: bool = True,
+        columnwise: bool = True,
+        shuffle_B_matrix_for_aiter: bool = False,
```
wangye805: Just curious, if this shuffle_B_matrix_for_aiter is set to true, this quantizer will apply data shuffle to the tensor it applied on, right? If so, this shuffle_B_matrix is kind of misleading, make it seems like it's trying to shuffle another tensor (B instead of A)

transformer_engine/pytorch/custom_recipes/quantization_mxfp4.py
```diff
@@ -0,0 +1,586 @@
...
+        # (M, num_blocks, 2, 16) — two independent 16-element halves per block
+        reshaped = data.reshape(M, N // MXFP4_BLOCK_SIZE, 2, dim)
+        transformed = np.einsum("...j,jk->...k", reshaped, transform)
+        return transformed.reshape(M, N)
+
+    # ------------------------------------------------------------------
+    # Core quantization (operates on a 2-D float tensor)
+    # ------------------------------------------------------------------
+
+    def _quantize_2d(self, tensor: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
```
wangye805: this _quantize_2d and _dequantize_2d are also misleading. NVFP4 has a dedicated 2d quantization scaling recipe. While for this mxfp4, we better name it _quantize or just _quanize_1d

transformer_engine/pytorch/tensor/storage/mxfp4_tensor_storage.py
```diff
@@ -77,7 +80,8 @@ def __new__(
         columnwise_data: Optional[torch.Tensor],
         columnwise_scale_inv: torch.Tensor,
         fp4_dtype: TE_DType,
-        quantizer: Optional[Quantizer] = None,
+        quantizer: Optional[Quantizer],
+        with_gemm_swizzled_scales: bool,
```
wangye805: Do we need to put data shuffle flag into mxfp4 tensor? Or put both data_shuffle and gemm shuffle both into quantizer?

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -372,6 +377,14 @@ enum NVTEQuantizationConfigAttribute {
...
    */
   kNVTEQuantizationConfigUseFastMath = 7,
+  #ifdef USE_ROCM
+  /*! Whether to apply Hadamard transform before MXFP4 quantization */
+  kNVTEQuantizationConfigMXFP4UseHadamard = 8,
+  /*! Whether to shuffle E8M0 scales for AITER GEMM (MXFP4) */
+  kNVTEQuantizationConfigMXFP4ScaleShuffle = 9,
+  /*! Whether to shuffle FP4 data for AITER GEMM (MXFP4) */
+  kNVTEQuantizationConfigMXFP4DataShuffle = 10,
+  #endif //#ifdef USE_ROCM
```
wangye805: Not sure whether hadamard transform, data shuffle, or scale shuffle should be added here. MXFP8 has scale shuffle and NVFP4 has hadamard, they are not here

wangye805 (changes requested):

tests/cpp/operator/test_cast_transpose_mxfp4_shuffled.cu
sudhu2k: Removed this file and developed test_cast_mxfp4_transpose.cu file.

tests/pytorch/mxfp4/test_mxfp4_gemm_exact.py
sudhu2k: noqa: F401 is a flake8/linter directive that suppresses the "imported but unused" warning (rule F401).
Here, aiter is imported only to check whether the package is installed (the try/except block sets _aiter_available). The module itself is never referenced directly in the test file — it's used later at runtime inside TE's GEMM dispatch. Without the # noqa: F401 comment, the linter would flag it as an unused import.

tests/cpp/operator/test_cast_transpose_mxfp4_shuffled.cu
sudhu2k: Removed this file and developed a new test_cast_mxfp4_transpose.cu file to match upstream's test files.

tests/cpp/operator/test_cast_transpose_mxfp4_shuffled.cu
sudhu2k: Removed this file and developed a new test_cast_mxfp4_transpose.cu file to match upstream's test files.

tests/cpp/operator/test_cast_transpose_mxfp4_shuffled.cu
sudhu2k: Removed this file and developed a new test_cast_mxfp4_transpose.cu file to match upstream's test files.

tests/cpp/operator/test_cast_transpose_mxfp4_shuffled.cu
sudhu2k: Removed this file and developed a new test_cast_mxfp4_transpose.cu file to match upstream's test files.

transformer_engine/common/cast/mxfp4/quantize_mxfp4.cuh
sudhu2k: Yes, 
https://github.com/ROCm/TransformerEngine/blob/58a3ff8d523ac4ad2ff5c18209ecf32f96e51818/transformer_engine/pytorch/cpp_extensions/gemm.py#L189-L209
https://github.com/ROCm/TransformerEngine/blob/58a3ff8d523ac4ad2ff5c18209ecf32f96e51818/transformer_engine/pytorch/quantization.py#L1382-L1392

AITER expects the B tensor (weight) to be shuffled for the TN (forward) and NN (backward dgrad) layouts, so we set data_shuffle=True for weights and pass b_pre_shuffled=True since both rowwise and columnwise data are shuffled.
However, for the wgrad case (layout == "NT"), the B tensor is the activation (input) rather than the weight. Here we pass b_pre_shuffled=False because both rowwise and columnwise data is not shuffled by quantizer. 

This case creates a divergence where rowwise and columnwise require different shuffle settings on the same tensor. It might be better to separate these two shuffle flags to handle this more cleanly.

transformer_engine/common/recipe/__init__.py
sudhu2k: Upstream also does similarly for nvfp4 recipe to enable or disable HADAMARD transform. So, I think this is okay for now to stay consistent. Let me know what your thoughts are.
https://github.com/ROCm/TransformerEngine/blob/58a3ff8d523ac4ad2ff5c18209ecf32f96e51818/transformer_engine/common/recipe/__init__.py#L460-L466

transformer_engine/common/include/transformer_engine/transformer_engine.h
sudhu2k: NVFP4 restricts the Hadamard (RHT) transform to the wgrad path only, meaning it is applied exclusively during columnwise quantization. The rowwise path receives no RHT.

For rowwise data, the quantizer calls nvte_quantize_v2 directly on the raw input
https://github.com/ROCm/TransformerEngine/blob/58a3ff8d523ac4ad2ff5c18209ecf32f96e51818/transformer_engine/pytorch/csrc/quantizer.cpp#L1572-L1589

For columnwise data, there are two paths controlled by an eligible_for_rht_cast_fusion check 
https://github.com/ROCm/TransformerEngine/blob/58a3ff8d523ac4ad2ff5c18209ecf32f96e51818/transformer_engine/pytorch/csrc/quantizer.cpp#L1631-L1667

There's an if else case which controls which kernel is called.

1. Fused path: nvte_hadamard_transform_cast_fusion_columnwise — a single kernel that performs both RHT and FP4 quantization.
2. Non-fused path: nvte_hadamard_transform followed by nvte_quantize_v2 — two separate kernel launches.

In both cases, the caller (the C++ quantizer) decides which API to invoke. The nvte_quantize_v2 kernel itself has no awareness of Hadamard. This is why NVFP4 does not need a Hadamard flag in QuantizationConfig.

With MXFP8,

Note: MXFP8 scale swizzling is different from MXFP4 scale shuffling.
Swizzling is a memory layout transformation while shuffling is a permutation required by the AITER GEMM library. They serve different GEMM backends on different platforms.

https://github.com/ROCm/TransformerEngine/blob/58a3ff8d523ac4ad2ff5c18209ecf32f96e51818/transformer_engine/common/cast/mxfp8/quantize_mxfp8.cuh#L692

https://github.com/ROCm/TransformerEngine/blob/58a3ff8d523ac4ad2ff5c18209ecf32f96e51818/transformer_engine/common/cast/mxfp8/quantize_mxfp8.cuh#L734-L737

The option to swizzle is passed via the tensor itself.

With MXFP4,
All quantization goes exclusively through nvte_quantize_v2 → mxfp4::quantize → a single fused cast_transpose_mxfp4_shuffled kernel.
https://github.com/ROCm/TransformerEngine/blob/58a3ff8d523ac4ad2ff5c18209ecf32f96e51818/transformer_engine/pytorch/csrc/quantizer.cpp#L1960-L1975


This kernel handles Hadamard, scale shuffling, and data shuffling internally via template parameters. Since there is only one entry point and the kernel needs to know which features to enable, these settings must be communicated through QuantizationConfig flags (mxfp4_use_hadamard, mxfp4_scale_shuffle, mxfp4_data_shuffle).

This follows the same pattern NVFP4 uses for stochastic_rounding. It's a flag in QuantizationConfig because it controls behavior inside the quantization kernel, not the choice of which kernel to call.

We can probably add scale shuffle and data shuffle to the tensor, while I think use_hadamard should still be inside the quantization config because Hadamard is fundamentally an algorithmic choice and not a layout property of the output tensor. Putting it on the tensor would be semantically wrong. Let me know what you think.

transformer_engine/common/cast/dispatch/quantize.cuh
sudhu2k: https://github.com/ROCm/TransformerEngine/blob/58a3ff8d523ac4ad2ff5c18209ecf32f96e51818/transformer_engine/pytorch/quantization.py#L112-L120

Currently, MXFP4 recipe flow is guarded on MI300X, I'll also add a guard before the kernel launch.

transformer_engine/common/cast/mxfp4/cast_transpose_mxfp4_shuffled.cuh
sudhu2k: It's better to throw error before the kernel is launched instead of throwing error inside the kernel which might lead to crashes. I'll add that code.

transformer_engine/common/transformer_engine.cpp
sudhu2k: I think it's fine since it's consistent with NVFP4 code. Let me know if you think otherwise.

transformer_engine/pytorch/cpp_extensions/gemm.py
sudhu2k: torch._C._DisableTorchDispatch() is a context manager that temporarily disables PyTorch's dispatch mechanism. I think this is unnecessary, I can remove this.

transformer_engine/common/include/transformer_engine/transformer_engine.h
sarthak-amd: +1 on "Since there is only one entry point and the kernel needs to know which features to enable, these settings must be communicated through QuantizationConfig flags "

transformer_engine/pytorch/custom_recipes/quantization_mxfp4.py
sudhu2k: Right, I'll refactor this to move shuffle_B_matrix_for_aiter and shuffle_scale to MXFP4TensorStorage.

transformer_engine/common/cast/mxfp4/cast_transpose_mxfp4_shuffled.cuh
ipanfilo: Checking before calling is needed but it does not eliminate throwing error here. Imagine that MXPF4 is enabled on new arch and there is discrepancy between guarding here and at caller side.
The safest way though is to guard the whole method so it won't be declared on unsupported architectures

transformer_engine/common/cast/mxfp4/cast_transpose_mxfp4_shuffled.cuh
```diff
@@ -0,0 +1,743 @@
+/*************************************************************************
+ * Copyright (c) 2025-2026, Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * See LICENSE for license information.
+ ************************************************************************/
+ 
+/*
+ * MXFP4 Cast + Transpose Kernel (CUDA/HIP)
```
ipanfilo, line 8: It looks pure HIP. 
Also, the file seems not needing hipification. What are changes in cast_transpose_mxfp4_shuffled_hip.cuh after build comparing to this file?

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -23,6 +23,10 @@
 from ..custom_recipes.gemm import custom_gemm
 from ...debug.pytorch.debug_quantization import DebugQuantizer
 
+from ..utils import cast_if_needed
+
+_FP4_GEMM_TUNING = int(os.environ.get("NVTE_FP4_GEMM_TUNING", "1"))
```
ipanfilo: The name looks misleading - like if it will perform tuning. _FP4_USE_TUNED_GEMM_?

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -91,6 +96,22 @@ void init_float8blockwise_extension() {
...
+  MXFP4TensorPythonClass =
+  reinterpret_cast<PyTypeObject *>(PyObject_GetAttrString(fp4_module.ptr(), "MXFP4Tensor"));
+  auto fp4_base_module =
+      py::module_::import("transformer_engine.pytorch.tensor.storage.mxfp4_tensor_storage");
+  MXFP4TensorStoragePythonClass = reinterpret_cast<PyTypeObject *>(
+  PyObject_GetAttrString(fp4_base_module.ptr(), "MXFP4TensorStorage"));
+  NVTE_CHECK(MXFP4TensorPythonClass != nullptr,
+              "Internal error: could not initialize pyTorch MXFP4 extension.");
+}
+#endif //#ifdef USE_ROCM
```
ipanfilo, line 114: nit: there should be empty line(s) between methods

transformer_engine/common/cast/mxfp4/cast_transpose_mxfp4_shuffled.cuh
sudhu2k: The build script is different during the kernel launch
From
```
te_mxfp4::cast_transpose_mxfp4_shuffled<ROW, COL, SHUF_SCALES, HAD, SHUF_ROW, SHUF_COL> \
            <<<grid, block, 0, stream>>>
```
To
```
hipLaunchKernelGGL(( te_mxfp4::cast_transpose_mxfp4_shuffled<ROW, COL, SHUF_SCALES, HAD, SHUF_ROW, SHUF_COL>) \
            , dim3(grid), dim3(block), 0, stream,  \
```

But I think to be consistent with other files it's always better to have it as .cuh/.cu and let hipify take care of creating hip files.

transformer_engine/common/include/transformer_engine/transformer_engine.h
sudhu2k: New changes:

- Moved scale shuffle and data shuffle to MXFP4 Tensor class.
- Split data shuffle into rowwise data shuffle and columnwise data shuffle. I saw a minor boost in performance because of this since note that the input activation rowwise data isn't shuffled for fwd while it's columnwise data is shuffled for wgrad. This removes additional step of shuffling input activation columnwise data before calling GEMM and fuses it with the cast kernel. So it's better to split the shuffle_B_matrix_for aiter into shuffle_rowwise_data and shuffle_columnwise_data.
image

transformer_engine/pytorch/tensor/storage/mxfp4_tensor_storage.py
sudhu2k: I've made changes to put both data shuffle and scale shuffle into the MXFP4Tensor which is created via a quantizer.

transformer_engine/pytorch/custom_recipes/quantization_mxfp4.py
sudhu2k: I've split shuffle_B_matrix_for_aiter  into huffle_rowwise_data and shuffle_columnwise_data

transformer_engine/common/cast/mxfp4/quantize_mxfp4.cuh
sudhu2k: Separated rowwise and columnwise data shuffle.

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -74,6 +74,11 @@ enum NVTETensorParam {
   kNVTEColumnwiseScaleInv = 5,     /*!< Scale inverse tensor for decoding Columnwise Data */
   kNVTEColumnwiseAmax = 6,         /*!< Columnwise Amax tensor */
   kNVTEWithGEMMSwizzledScales = 7, /*!< Whether scaling factors are in format expected by GEMM */
+#ifdef USE_ROCM
+  kNVTEMXFP4ShuffleScales = 8,           /*!< Whether MXFP4 scales are shuffled for AITER GEMM */
+  kNVTEMXFP4ShuffleRowwiseData = 9,      /*!< Whether MXFP4 rowwise data is shuffled for AITER GEMM */
+  kNVTEMXFP4ShuffleColumnwiseData = 10,  /*!< Whether MXFP4 columnwise data is shuffled for AITER GEMM */
+#endif
```
matthiasdiener, line 80: If upstream adds their own params here, wouldn't these numbers conflict?

matthiasdiener: (same for NVTEQuantizationConfigAttribute )

transformer_engine/common/cast/mxfp4/cast_transpose_mxfp4_shuffled.cuh
```diff
@@ -0,0 +1,741 @@
...
+ * 
+ * Common patterns:
+ *   - 0x041F: XOR with lane 1 (exchange with adjacent thread)
+ *   - 0x081F: XOR with lane 2 (exchange 2 positions away)
+ *   - 0x101F: XOR with lane 4 (exchange 4 positions away)
+ * 
+ * Reference: AMD CDNA4 ISA, ds_swizzle_b32 (page 480)
+ */
+
+__device__ __forceinline__ float ds_swizzle_xor1(float val) {
```
matthiasdiener, line 121: Some of these have been merged as of #509, so perhaps they can be reused here?

transformer_engine/common/include/transformer_engine/transformer_engine.h
sudhu2k: It will yes, but contiguous numbers are required to make it work.
https://github.com/ROCm/TransformerEngine/blob/9b41fe2086bcfcb0cdd53c28a4544bec3b3ae977/transformer_engine/common/common.h#L169-L185
Here, attr_sizes is a plain C-style constexpr array indexed directly by the enum value. That's why the enum values must be contiguous (0, 1, 2, ... 10) — any gaps or jumps mean attr_sizes[enum_value] reads out of bounds.
I tried to make the numbers like 80, 90, 100 but that broke because attr_sizes[80] read garbage memory (out-of-bounds), causing the assertion:
`Buffer is too small for tensor parameter (parameter 80 needs 4294967297 bytes, but buffer has 1 bytes)
`

transformer_engine/common/cast/mxfp4/cast_transpose_mxfp4_shuffled.cuh
sudhu2k: Yes, I'm working on it now. Will let you know if I have any questions.

transformer_engine/common/include/transformer_engine/transformer_engine.h
sudhu2k: Same goes for NVTEQuantizationConfigAttribute

transformer_engine/common/cast/mxfp4/cast_transpose_mxfp4_shuffled.cuh
sudhu2k: I was looking into re-using the hadamard_transform files for MXFP4, I noticed that hadamard_transform kernel is written inside a .cu file and if I want to use it with cast_transpose_mxfp4_shuffled.cuh, I can't just #include a .cu file because that's not what .cu files are designed for and to share code .h or .cuh would be a better choice and also it's inside anonymous namespace which would mean I wouldn't be able to access it even if I #include it. I think the only way is to move ds_swizzle_xor1, ds_swizzle_xor2 and wht16 to a common file (hadamard_transform_utils.cuh?) and re-use it from there for both hadamard_transform.cu and cast_transpose_mxfp4_shuffled.cuh.
This refactor is being done by @matthiasdiener in another PR https://github.com/ROCm/TransformerEngine/pull/518/changes#diff-15b52a73d506d587d0ab0bc40968993ce688846c728fa9adb0889ae1a3f49404. We can wait and probably create a new PR in the future for this refactor.

transformer_engine/common/common.h
sudhu2k: Discussed offline on why use_hadamard should be part of QuantizationConfig compared to NVFP4.

transformer_engine/pytorch/csrc/type_converters.cpp
sudhu2k: with_gemm_swizzled_scales is just to be consistent with other Tensor storages and maybe a future addition if a GEMM config requires swizzled scales.

transformer_engine/common/cast/dispatch/quantize.cuh
sudhu2k: Added guard here
https://github.com/ROCm/TransformerEngine/blob/9a5579fb535f11a74648745e90ac70b07307c2d2/transformer_engine/common/cast/mxfp4/quantize_mxfp4.cuh#L31-L40

tests/cpp/operator/test_cast_transpose_mxfp4_shuffled.cu
wangye805: By the way, next time you can use git mv when you rename one file to another. This can help track the origin of changes

tests/cpp/operator/test_cast_mxfp4_transpose.cu
```diff
@@ -0,0 +1,527 @@
+/*************************************************************************
+ * Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * See LICENSE for license information.
+ ************************************************************************/
+
+ #include <cuda_bf16.h>
```
wangye805: nit: Looks like there is an empty space in front of all lines :-)

tests/cpp/operator/test_cast_mxfp4_transpose.cu
```diff
@@ -0,0 +1,527 @@
...
+     return {static_cast<double>(lo), static_cast<double>(hi)};
+#else
+     const __half2_raw raw_truncated_to_fp4e2m1_pair =
+         __nv_cvt_fp4x2_to_halfraw2(*reinterpret_cast<__nv_fp4x2_storage_t*>(&fp4_pair), __NV_E2M1);
+ 
+     const __half2 truncated_to_fp4e2m1_pair(raw_truncated_to_fp4e2m1_pair);
+     const double truncated_to_fp4e2m1_x = static_cast<double>(truncated_to_fp4e2m1_pair.x);
+     const double truncated_to_fp4e2m1_y = static_cast<double>(truncated_to_fp4e2m1_pair.y);
+     return {truncated_to_fp4e2m1_x, truncated_to_fp4e2m1_y};
+#endif //#ifndef __HIP_PLATFORM_AMD__
```
wangye805: NV upstream do not have this mxfp4 so the else branch will never be tested out. You can remove the else branch

tests/cpp/operator/test_cast_mxfp4_transpose.cu
```diff
@@ -0,0 +1,527 @@
...
+    }
+    uint32_t amax_bits = *reinterpret_cast<uint32_t*>(&amax);
+    amax_bits = (amax_bits + 0x200000u) & 0xFF800000u;
+    int exp = ((amax_bits >> 23) & 0xFF) - 127;
+    int scale_unbiased = exp - 2;
+    scale_unbiased = std::max(-127, std::min(127, scale_unbiased));
+    uint32_t rcp_bits = static_cast<uint32_t>(127 - scale_unbiased) << 23;
+    scale_reciprocal = *reinterpret_cast<float*>(&rcp_bits);
+    return static_cast<fp8e8m0>(scale_unbiased + 127);
+}
```
wangye805, line 69: Can we reuse the e8m0 conversion from mxfp8 gtests?

tests/cpp/operator/test_cast_mxfp4_transpose.cu
```diff
@@ -0,0 +1,527 @@
...
+void quantize_mxfp4_1d(float (*OP)(const float),
+                       const InputType* const input,
+                       fp4e2m1x2* const output,
+                       fp8e8m0* const scales,
+                        const size_t rows,
+                        const size_t cols,
+                        const size_t scales_stride,
+                        const float global_amax) {
+
+     constexpr size_t block_size_X = 32;
```
wangye805: Can we unify those 32 block size into some macro or constexpr in header files, instead of writing it in every mxfp4 source codes? How does mxfp8 handle this issue?

tests/cpp/operator/test_cast_mxfp4_transpose.cu
```diff
@@ -0,0 +1,527 @@
...
+         nvte_silu(input.data(), output.data(), 0);
+     } else if (OP == &relu) {
+         nvte_relu(input.data(), output.data(), 0);
+     } else if (OP == &qgelu) {
+         nvte_qgelu(input.data(), output.data(), 0);
+     } else if (OP == &srelu) {
+         nvte_srelu(input.data(), output.data(), 0);
+     } else {
+         nvte_quantize_v2(input.data(), output.data(), quant_config, 0);
+     }
```
wangye805: Do we need different ops? I thought mlperf only need the cast transpose

tests/cpp/operator/test_cast_mxfp4_transpose.cu
```diff
@@ -0,0 +1,527 @@
...
+         ::testing::Values(DType::kBFloat16)),
+     [](const testing::TestParamInfo<FusedCastTransposeMXFP4TestSuite::ParamType>& info) {
+         std::string name = to_string(std::get<0>(info.param));
+       const auto& shape = std::get<1>(info.param);
+       for ( const auto& s: shape) {
+         name += "X" + std::to_string(s);
+       }
+       name += "X" + test::typeName(std::get<2>(info.param));
+         return name;
+     });
```
wangye805: It looks like hadamard, data shuffle, and scale shuffle are not tested?

transformer_engine/common/cast/dispatch/quantize.cuh
wangye805: Yeah, guarding this in common dir makes the libtransformer_engine.so self-contained. Thanks

transformer_engine/common/cast/mxfp4/cast_transpose_mxfp4_shuffled.cuh
```diff
@@ -0,0 +1,743 @@
...
+    bool shuffle_scales, bool use_hadamard,
+    bool shuffle_rowwise_fp4, bool shuffle_colwise_fp4,
+    int rowwise_scale_stride, int colwise_scale_stride,
+    int rowwise_scale_N, int rowwise_scale_M_pad, int rowwise_scale_N_pad,
+    int colwise_scale_M, int colwise_scale_N,
+    int colwise_scale_M_pad, int colwise_scale_N_pad,
+    hipStream_t stream
+) {
+    dim3 grid((M + 127) / 128, (N + 63) / 64);
+    dim3 block(256);
```
wangye805: nit: express those 128, 256, 64 using macro or constexpr, instead of hard coding

transformer_engine/common/recipe/__init__.py
wangye805: Isee. Then let's use env as well

transformer_engine/common/recipe/__init__.py
```diff
@@ -570,8 +574,10 @@ class MXFP4BlockScaling(Recipe):
 
     margin: int = 0
     fp4_format: Format = Format.E2M1
+    fp8_format: Format = Format.E4M3
```
wangye805, line 579: WHy mxfp4 need fp8_format?

transformer_engine/common/common.h
```diff
@@ -149,6 +160,12 @@ struct Tensor {
    */
   bool with_gemm_swizzled_scales = false;
```
wangye805, line 161: Wait, I just realized that, can we re-use this for mxfp4?

transformer_engine/common/transformer_engine.cpp
wangye805: Okay then

transformer_engine/pytorch/csrc/common.h
```diff
@@ -290,6 +290,33 @@ class MXFP8Quantizer : public Quantizer {
...
 };
 
+#ifdef USE_ROCM
+class MXFP4Quantizer : public Quantizer {
+ public:
+  DType dtype;
+  bool use_hadamard;
+  bool shuffle_rowwise_data;
+  bool shuffle_columnwise_data;
+  bool shuffle_scales;
```
wangye805: Emm, do you still need those three fields since you already put those into mxfp4 tensor class?

transformer_engine/pytorch/csrc/type_converters.cpp
wangye805: Oh, I see, this NVTETensorFromMXFP4Tensor is not used anywhere like mxfp8 or nvfp4 tensors, right?

transformer_engine/pytorch/tensor/storage/mxfp4_tensor_storage.py
wangye805, line 68: If the quantizer already has the following 4 field info (with_gemm_swizzled_scales, shuffl_scales, shuffle_rowwise_data, shuffle_columnwise_data), can we retrieve those info from _quantizer directly?

transformer_engine/pytorch/tensor/storage/mxfp4_tensor_storage.py
wangye805: Well, I'm still puzzled about where to put those infos

wangye805 (changes requested):

transformer_engine/common/include/transformer_engine/transformer_engine.h
wangye805: Oh, actually the mxfp8 scale swizzling is the same thing as mxfp4 aiter scale shuffling (logically). They are both introduced to resolve the bank conflict issue for linear scale loading

transformer_engine/pytorch/csrc/type_converters.cpp
sudhu2k: Right, it's not used anywhere similar to NVTETensorFromNVFP4Tensor or NVTETensorFromMXFP8Tensor
It's just part of 
https://github.com/ROCm/TransformerEngine/blob/1d5fa8a0fa3585337c552b0d92be299600a2e28a/transformer_engine/pytorch/csrc/pybind.h#L119-L136

transformer_engine/common/include/transformer_engine/transformer_engine.h
sudhu2k: But operation wise, they are different right even though the purpose is for efficient access of scales. Swizzling does index remapping while shuffle does Block reshape + axis swap on 32-wide scale rows.
https://github.com/ROCm/TransformerEngine/blob/1d5fa8a0fa3585337c552b0d92be299600a2e28a/transformer_engine/pytorch/custom_recipes/quantization_mxfp4.py#L68-L73

tests/cpp/operator/test_cast_mxfp4_transpose.cu
sudhu2k: This method mimics the compute_e8m0_scale method from the kernel
https://github.com/ROCm/TransformerEngine/blob/1d5fa8a0fa3585337c552b0d92be299600a2e28a/transformer_engine/common/cast/mxfp4/cast_transpose_mxfp4_shuffled.cuh#L248-L286

MXFP8 path is not the same as the MXFP4 device rule: compute_e8m0_scale rounds the block max to the nearest power of two, applies exp − 2 fp4 headroom, clamps, and builds the scale for the FP4 conversion. The MXFP8 float_to_e8m0 pipeline is different, It pre-scales amax with max_reciprocal then applies float_to_e8m0 + exp2f_rcp for FP8 range and NaN/zero/exponent rules. So I think it's better to use a separate method to mimic the kernel.

tests/cpp/operator/test_cast_mxfp4_transpose.cu
sudhu2k: They only needed cast transpose yes, but I made it to be consistent with the NVFP4 test.

tests/cpp/operator/test_cast_mxfp4_transpose.cu
sudhu2k: Yes, those aren't tested here similar to NVFP4, they are only tested in  tests/pytorch/mxfp4/test_mxfp4_quantize_exact.py similar to  tests/pytorch/mxfp4/test_nvfp4_quantize_exact.py

transformer_engine/pytorch/csrc/common.h
sudhu2k: Yes, even though it's stored in MXFP4 tensor class, it is still set via the quantizer, hence it's needed here as well.
https://github.com/ROCm/TransformerEngine/blob/1d5fa8a0fa3585337c552b0d92be299600a2e28a/transformer_engine/pytorch/tensor/mxfp4_tensor.py#L63-L74

transformer_engine/common/recipe/__init__.py
sudhu2k: Just to be consistent with the fp8 scaling recipes where TE still uses and expects the fp8_format property like
https://github.com/ROCm/TransformerEngine/blob/1d5fa8a0fa3585337c552b0d92be299600a2e28a/transformer_engine/pytorch/module/base.py#L1084-L1087

transformer_engine/pytorch/tensor/storage/mxfp4_tensor_storage.py
sudhu2k: We can but I think the current design is much more consistent and safer because the storage describes the tensor's own data layout without depending on the external object.

transformer_engine/common/common.h
sudhu2k: We can remove this technically and just have shuffle_scales but I kept it for now accomodate any future GEMM configs which might need swizzling instead of shuffling. Let me know if you think we don't need this.

tests/cpp/operator/test_cast_mxfp4_transpose.cu
sudhu2k: MXFP8 also hardcodes all the literals

https://github.com/ROCm/TransformerEngine/blob/1d5fa8a0fa3585337c552b0d92be299600a2e28a/tests/cpp/operator/test_cast_mxfp8.cu#L61-L62

https://github.com/ROCm/TransformerEngine/blob/1d5fa8a0fa3585337c552b0d92be299600a2e28a/tests/cpp/operator/test_cast_mxfp8.cu#L200-L201

transformer_engine/common/cast/mxfp4/cast_transpose_mxfp4_shuffled.cuh
sudhu2k: HIP compiles every source file in two passes: one for the device and one for the host. Guarding the entire function with #if defined(__gfx950__) removes it from the host pass as well. However, the host compiler still needs to parse the kernel template that calls cvt_f32x4_to_fp4x4, causing an "undeclared identifier" error. 

Currently I've modified the code to keep the function signature visible to both passes and use #if defined(__gfx950__) && __HIP_DEVICE_COMPILE__ to gate only the inline asm body, with a software fallback in the #else branch for the host pass. This matches the convention used by the MXFP8 kernels in rocm_quantize_mxfp8.cuh.

https://github.com/ROCm/TransformerEngine/blob/790ee5db649c6dc3071b7c60bdc9af69023badb5/transformer_engine/common/cast/mxfp8/rocm_quantize_mxfp8.cuh#L338-L369

I've also ensured that the software fallback works correctly replacing with it and running all the tests.

transformer_engine/pytorch/csrc/type_converters.cpp
wangye805: Oh, if it's not used, probably not tested as well. Let's add a to-do to remind us it's not used/tested and should revisit in the future

transformer_engine/pytorch/tensor/storage/mxfp4_tensor_storage.py
wangye805: Chatted offline. Since upstream NVFP4 is doing the same thing that put scale shuffle flag in both quantizer and tensor, we will follow their pattern

transformer_engine/common/include/transformer_engine/transformer_engine.h
wangye805: chatted offline that hipblaslt will implement the same shuffling/swizzling as aiter

transformer_engine/pytorch/csrc/type_converters.cpp
sudhu2k: Nevermind, I re-checked the code flow and seems like the functions (NVTETensorFromNVFP4Tensor, NVTETensorFromMXFP8Tensor) are used in one of the makeTransformerEngineTensor  overloads.
https://github.com/ROCm/TransformerEngine/blob/3f7550999611ee3e94264c62b6bc0dcb5fce86a9/transformer_engine/pytorch/csrc/common.cpp#L81-L97

transformer_engine/common/include/transformer_engine/transformer_engine.h
sudhu2k: Refactored scale_shuffle to with_gemm_swizzled_scales

wangye805 (approved):

tests/cpp/operator/test_cast_mxfp4_transpose.cu
```diff
@@ -0,0 +1,505 @@
...
+    const fp4e2m1 *test_data_t = test.columnwise_cpu_dptr<fp4e2m1>();
+    const fp4e2m1 *ref_data = reinterpret_cast<const fp4e2m1*>(ref);
+    const fp4e2m1 *ref_data_t = reinterpret_cast<const fp4e2m1*>(ref_t);
+
+    // Print detailed element-by-element comparison
+    // print_detailed_tensor_comparison("output", test_data, ref_data, rows, cols);
+    // print_detailed_tensor_comparison("output_t", test_data_t, ref_data_t, cols, rows);
+
+    // Optionally dump tensor data to files for detailed analysis
+    if (dump_data) {
```
ipanfilo, line 315: How do we control dump_data?

tests/cpp/operator/test_cast_mxfp4_transpose.cu
```diff
@@ -0,0 +1,505 @@
...
+
+}  // namespace
+
+class FusedCastTransposeMXFP4TestSuite : public ::testing::TestWithParam
+    <std::tuple<ActivationType,
+                std::vector<size_t>,
+                transformer_engine::DType>> {};
+
+TEST_P(FusedCastTransposeMXFP4TestSuite, TestFusedCastTransposeMXFP4) {
+hipDeviceProp_t prop;
```
ipanfilo: indentation

tests/cpp/operator/test_cast_mxfp4_transpose.cu
```diff
@@ -0,0 +1,505 @@
...
+                        ref_scales_t.get(),
+                        output.scale(),
+                        rows,
+                        cols,
+                        scales_stride,
+                        scales_stride_t);
+
+    QuantizationConfigWrapper quant_config;
+
+    if (OP == &gelu) {
```
ipanfilo, line 374: nit: switch case?

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -99,6 +102,163 @@ def get_tensor_device(tensor: torch.Tensor) -> int:
...
+        B_scales = A._columnwise_scale_inv
+        b_pre_shuffled = A._shuffle_columnwise_data
+
+    elif layout == "NT":
+        A_fp4 = B._columnwise_data
+        A_scales = B._columnwise_scale_inv
+        B_fp4 = A._columnwise_data
+        B_scales = A._columnwise_scale_inv
+        b_pre_shuffled = A._shuffle_columnwise_data
+
```
ipanfilo, line 214: why not handle "TT"? and empty lines are actually not needed between elif branches

transformer_engine/common/recipe/__init__.py
ipanfilo: But in the above example it gracefully checks if the attribute is there, so it should not cause a problems if it is not set

tests/cpp/operator/test_cast_mxfp4_transpose.cu
sudhu2k: Have to control it via manual code change, similar to NVFP4 test
https://github.com/ROCm/TransformerEngine/blob/b68935bc8e8db267c2349c9581100dd56cbe3c6d/tests/cpp/operator/test_cast_nvfp4_transpose.cu#L506-L525

tests/cpp/operator/test_cast_mxfp4_transpose.cu
sudhu2k: True, yeah could be better but followed the same style as nvfp4 tests
https://github.com/ROCm/TransformerEngine/blob/b68935bc8e8db267c2349c9581100dd56cbe3c6d/tests/cpp/operator/test_cast_nvfp4_transpose.cu#L628-L644

transformer_engine/pytorch/cpp_extensions/gemm.py
sudhu2k: We never use TT layout for linear/layernorm linear modules and is not required for end to end training. I'm not sure if AITER even has a kernel to support TT.
 
image

transformer_engine/common/recipe/__init__.py
sudhu2k: Removed the attribute.

ci/pytorch.sh
```diff
@@ -85,8 +95,8 @@ run_test_config(){
     NVTE_USE_ATOMIC_AMAX=1 NVTE_USE_CAST_TRANSPOSE_TRITON=1 run_default_fa_lbl "amax+triton" 3 test_numerics.py
     NVTE_USE_ATOMIC_AMAX=1 NVTE_USE_CAST_TRANSPOSE_TRITON=1 run_default_fa_lbl "amax+triton" 3 test_fusible_ops.py
     NVTE_USE_ATOMIC_AMAX=1 run_default_fa_lbl "amax" 3 triton_kernels/test_cast.py
-
     run_default_fa 1 nvfp4/
+    with_aiter run_default_fa 1 mxfp4/
```
ipanfilo: Please add if [ $_fus_attn = "$_DEFAULT_FUSED_ATTN" ]; condition here, otherwise with_aiter will install/uninstall AITER on every pass

ci/pytorch.sh
```diff
@@ -38,6 +38,22 @@ run_default_fa_lbl() {
...
+    if [ $_fus_attn = "$_DEFAULT_FUSED_ATTN" ]; then
+        # flydsl is only on AMD nightlies; needed for aiter pyproject build-system requires.
+        (cd ${TE_PATH}3rdparty/aiter && \
+            PIP_FIND_LINKS="https://rocm.frameworks-nightlies.amd.com/whl/gfx942-gfx950/flydsl/" \
+            PIP_PRE=1 \
+            GPU_ARCHS=native pip install -e .)
+    fi
+    "$@"
+    if [ $_fus_attn = "$_DEFAULT_FUSED_ATTN" ]; then
+        pip uninstall -y amd-aiter
```
ipanfilo: WARNING: Skipping amd-aiter as it is not installed. 
The AITER is either not installed, or wrong name is used and it is not uninstalled

sudhu2k: Aiter isn't installed, that's why we are seeing 
WARNING: Skipping amd-aiter as it is not installed.

It's because of aiter's dependency on flydsl which is having some trouble getting installed from https://rocm.frameworks-nightlies.amd.com/whl/gfx942-gfx950/flydsl/ on the CI. Locally I'm able to install flydsl and aiter.

Right now, the GEMM/module tests for mxfp4 gets skipped if aiter isn't installed.

transformer_engine/common/recipe/__init__.py
sudhu2k: Actually, seems like it is hardcoded here
https://github.com/ROCm/TransformerEngine/blob/a62b7df1790989c37455660847bb65e16a1fdfcc/transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py#L725-L726

which is causing issues. I'll revert this back to have fp8_format.

ci/pytorch.sh
ipanfilo: ERROR: Could not find a version that satisfies the requirement flydsl==0.0.1.dev95158637 (from versions: none)
      ERROR: No matching distribution found for flydsl==0.0.1.dev95158637
So the problem of CI.
Also, add error handling - no need to run anything if AITER is not installed

ipanfilo (changes requested):

ci/pytorch.sh
sudhu2k: Done!

ipanfilo: No need to run anything - i.e. do not run  commands passed to this function: "$@"

ipanfilo: Literally, you need return from function if installation fails

sudhu2k: We would still want to run the tests even when the installation fails, the MXFP4 cast transpose tests doesn't need aiter to be installed. The tests that require aiter (GEMM/module level) tests gets skipped if aiter isn't installed anyways.

sudhu2k: Reverted this, since we would be using images with aiter pre-installed.

transformer_engine/common/recipe/__init__.py
ipanfilo: Please put comment because it is not obvious that MFXP4 reuses FP8 classes

sudhu2k: Done.

ipanfilo (approved):

sudhu2k:
All level 3 tests passed with the new aiter installed image:
https://github.com/ROCm/TransformerEngine/actions/runs/25014847020/job/73259852413

PR #461, opened 2026-02-24
[NO MERGE] Integrate CK varlen cross attention for small-seq (s_q=1, s_kv<=16)

VeeraRajasekhar:
Integrate the CK team's unfused variable-length attention HIP kernels from varlen_attn/ into Transformer Engine's ROCm fused-attn path as a specialized path for specialized cross-attention (Q length 1, KV length 2-16, large batch)..

- Add fused_attn_smallseq.hpp and fused_attn_smallseq.cpp under fused_attn_rocm/: declarations and implementation adapted from varlen_attn/attn_fwd.cpp and attn_bwd.cpp (scores, mask+softmax, output; grad_V, grad_attn, softmax bwd, grad_Q/grad_K). Runtime dispatch over max_seqlen_kv in {2,4,6,8,12,16}, head_dim 128, BF16.

- Add fused_attn_smallseq.cpp to the ROCm fused-attn build in transformer_engine/common/CMakeLists.txt.

- In fused_attn_ck_fwd: when THD and no bias, branch to small-seq path when max_seqlen_q==1 and 2<=max_seqlen_kv<=16. On shape query (Aux_CTX_Tensors->size == 0) skip get_runtime_max_seqlen (cu_seqlens pointers are null); use host max_seqlen_kv and set output_S to attention-weights shape {max_tokens_q, h_q, 1, runtime_max_seqlen_kv} and dtype QKV_type. On real run (size >= 2) call get_runtime_max_seqlen then fused_attn_smallseq_fwd. Use sequence count b_varlen = max_tokens_q (not segment count b) for get_runtime_max_seqlen, output_S shape, workspace size, and small-seq fwd so varlen kernel indexing matches Q and cu_seqlens_kv (THD may pass segment-level cu_seqlens; varlen kernel expects sequence-level batch).

- In fused_attn_ck_bwd: same THD/small-seq condition. On workspace query (workspace->data.dptr == nullptr) skip get_runtime_max_seqlen and use host max_seqlen_kv; on real run call get_runtime_max_seqlen then fused_attn_smallseq_bwd. Use b_varlen = max_tokens_q_bwd for get_runtime_max_seqlen, workspace size, and small-seq bwd.

- Reuse softmax LSE auxiliary buffer for attention weights in the small-seq path (forward write, backward read);

[...]

PR #542, opened 2026-04-15
[TE] Phase 2 of small-seq cross-attn integration: a separate cpp backend and a new jax api 

VeeraRajasekhar:
… backend

Refactor the ROCm small-sequence attention path so it is a first-class backend instead of branching from the generic CK fused-attention entry: add NVTE entry points and ROCm implementations under fused_attn_rocm, CMake wiring, and public declarations in fused_attn.h.

Rename the small-seq sources to fused_attn_small_seq.* so filenames match the new API. Extend kernel dispatch to head sizes 128, 256, and 512.

# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/common/fused_attn_rocm/fused_attn.cpp
```diff
@@ -875,6 +876,182 @@ void nvte_fused_attn_bwd(const NVTETensor Q, const NVTETensor K, const NVTETenso
...
+      "nvte_fused_attn_small_seq_fwd: configuration not supported for small-seq path.");
+
+  Tensor *softmax_aux_tensor = convertNVTETensorCheck(Aux_CTX_Tensors->tensors[0]);
+  void *attn_weights_buf = softmax_aux_tensor->data.dptr;
+
+  if (wkspace->data.dptr == nullptr) {
+    wkspace->data.shape = {1};
+    wkspace->data.dtype = DType::kByte;
+    return;
+  }
```
wangye805, line 1000: Compared to the bwd pass, it looks like you don't allocate the buffer for fwd pass? Do we need a workspace for runtime seqlen or segment check?

transformer_engine/common/fused_attn_rocm/fused_attn_small_seq.cpp
```diff
@@ -1037,5 +1002,124 @@ void fused_attn_smallseq_bwd(size_t b,
...
+  if (const char* env_p = std::getenv("NVTE_LOG_CK_CONFIG")) {
+    log = (env_p != nullptr && std::string(env_p) == "1");
+  }
+
+  if (num_gqa_groups == 0 || num_attn_heads % num_gqa_groups != 0) {
+    if (log) {
+      std::cout << "small-seq: heads must be divisible by GQA groups" << std::endl;
+    }
+    return false;
+  }
```
wangye805: I don't think new CK kernels can support gqa for those small cross attn for now. At least we didn't enable the test in their repo yet. Let's filter the gqa/mqa right now

transformer_engine/common/fused_attn_rocm/fused_attn_small_seq.cpp
```diff
@@ -1037,5 +1002,124 @@ void fused_attn_smallseq_bwd(size_t b,
...
+
+  NVTE_QKV_Layout_Group layout_group = nvte_get_qkv_layout_group(qkv_layout);
+  if (layout_group != NVTE_QKV_Layout_Group::NVTE_HD_HD_HD ||
+      nvte_get_qkv_format(qkv_layout) != NVTE_QKV_Format::NVTE_THD) {
+    if (log) {
+      std::cout << "small-seq: requires THD separate Q, K, V (THD_THD_THD)" << std::endl;
+    }
+    return false;
+  }
```
wangye805: This is redundant since the check from line 1063-1067 can fully cover it

transformer_engine/common/fused_attn_rocm/fused_attn_small_seq.cpp
```diff
@@ -1037,5 +1002,124 @@ void fused_attn_smallseq_bwd(size_t b,
...
+  if (attn_mask_type == NVTE_Mask_Type::NVTE_NO_MASK ||
+      attn_mask_type == NVTE_Mask_Type::NVTE_PADDING_MASK) {
+    if (!((window_size_left == -1 && window_size_right == -1) ||
+          (window_size_left >= 0 && window_size_right >= 0))) {
+      if (log) {
+        std::cout << "small-seq: invalid window size for mask type" << std::endl;
+      }
+      return false;
+    }
+  }
```
wangye805: Does CK small seq attn support sliding window for now?

transformer_engine/common/fused_attn_rocm/fused_attn.cpp
VeeraRajasekhar: Updated now, earlier I didn't add runtime seqlen check. Now its updated

transformer_engine/common/fused_attn_rocm/fused_attn_small_seq.cpp
VeeraRajasekhar: Updated the checks.

transformer_engine/common/fused_attn_rocm/fused_attn_small_seq.cpp
VeeraRajasekhar: Fixed

transformer_engine/common/fused_attn_rocm/fused_attn_small_seq.cpp
VeeraRajasekhar: No, fused_attn_small_seq_fwd/_bwd don't take window parameters — kernels use full seq_kv from cu_seqlens.

Updated these

tests/jax/test_fused_attn.py
```diff
@@ -497,8 +534,78 @@ def _check_configs(self):
                     "the F16_arbitrary_seqlen backend."
                 )
 
+    def _setup_thd_segments_small_seq(self, generate_random_segment_ids):
```
wangye805, line 537: Since our customer already mentioned that their sq is 1 for each segment, without padding at all, and their sq_kv <=16 including padding, we can create our own small_seq input generation to separate with the non-small-seq tests

transformer_engine/common/fused_attn_rocm/fused_attn.cpp
```diff
@@ -894,6 +896,218 @@ void nvte_fused_attn_bwd(const NVTETensor Q, const NVTETensor K, const NVTETenso
...
+    NVTE_QKV_Layout qkv_layout, NVTE_Bias_Type bias_type, NVTE_Mask_Type attn_mask_type,
+    int64_t window_size_left, int64_t window_size_right, NVTETensor workspace,
+    cudaStream_t stream) {
+  NVTE_API_CALL(nvte_fused_attn_small_seq_fwd);
+  using namespace transformer_engine;
+  const Tensor *input_cu_seqlens_q = convertNVTETensorCheck(cu_seqlens_q);
+  const Tensor *input_cu_seqlens_kv = convertNVTETensorCheck(cu_seqlens_kv);
+  const Tensor *input_cu_seqlens_q_padded = convertNVTETensorCheck(cu_seqlens_q_padded);
+  const Tensor *input_cu_seqlens_kv_padded = convertNVTETensorCheck(cu_seqlens_kv_padded);
+  const Tensor *input_rng_state = convertNVTETensorCheck(rng_state);
```
wangye805, line 961: Do we even support dropout with this rng_state?

transformer_engine/common/fused_attn_rocm/fused_attn.cpp
```diff
@@ -894,6 +896,218 @@ void nvte_fused_attn_bwd(const NVTETensor Q, const NVTETensor K, const NVTETenso
...
+    wkspace->data.dtype = DType::kByte;
+    return;
+  }
+
+  size_t workspace_bytes = 1;
+  for (size_t i = 0; i < wkspace->data.shape.size(); ++i) {
+    workspace_bytes *= wkspace->data.shape[i];
+  }
+  workspace_bytes *= fused_attn_rocm::nvte_dtype_size(wkspace->data.dtype);
+  NVTE_CHECK(workspace_bytes >= req_bytes, "nvte_fused_attn_small_seq_bwd: workspace too small.");
```
wangye805, line 1096: nit: now we don't have mixed old ck + new small_seq flow, will we still get more workspace than we need?

transformer_engine/common/fused_attn_rocm/fused_attn_small_seq.cpp
wangye805, line 1: nit: can we have their repo as a 3rdparty then reference their .h. No need to do this now

transformer_engine/common/fused_attn_rocm/fused_attn_small_seq.cpp
```diff
@@ -0,0 +1,1121 @@
...
+      workspace, Q, K, grad_Q, grad_K, scale, cu_seqlens_kv, cu_seqlens_kv_padded, b, head_num);
+}
+
+size_t fused_attn_small_seq_bwd_workspace_size(size_t b,
+                                              size_t h_q,
+                                              size_t max_seqlen_kv,
+                                              DType dtype) {
+  constexpr size_t elt_size = 2u;  // BF16 and FP16 are 2 bytes
+  return b * h_q * 1 * std::min(max_seqlen_kv, size_t(16)) * elt_size;
+}
```
wangye805, line 793: Let's have a comment on what's the actual bwd workspace for, namely, why it's b*h*q*1*std::min(max_seqlen_kv, size_t(16))*elt_size?
Probably it's because it's for dS or dP?

transformer_engine/common/fused_attn_rocm/fused_attn_small_seq.cpp
```diff
@@ -0,0 +1,1121 @@
...
+  dim3 grid_qk((b * seq_q * head_num + tasks_per_block_qk - 1) / tasks_per_block_qk);
+  compute_grad_qk_kernel<T, Config, tasks_per_block_qk><<<grid_qk, block, 0, stream>>>(
+      workspace, Q, K, grad_Q, grad_K, scale, cu_seqlens_kv, cu_seqlens_kv_padded, b, head_num);
+}
+
+size_t fused_attn_small_seq_bwd_workspace_size(size_t b,
+                                              size_t h_q,
+                                              size_t max_seqlen_kv,
+                                              DType dtype) {
+  constexpr size_t elt_size = 2u;  // BF16 and FP16 are 2 bytes
```
wangye805, line 791: Avoid hard coding?

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -373,11 +375,13 @@ def abstract(
             elif backend == NVTE_Fused_Attn_Backend.NVTE_CK:
                 if config.qkv_layout.is_thd():
                     softmax_shape = (*batch_shape, q_max_seqlen, attn_heads, 1)
+                    softmax_dtype = dtypes.canonicalize_dtype(jnp.float32)
                 else:
                     softmax_shape = (*batch_shape, attn_heads, q_max_seqlen, 1)
-                softmax_dtype = dtypes.canonicalize_dtype(jnp.float32)
+                    softmax_dtype = dtypes.canonicalize_dtype(jnp.float32)
```
wangye805, line 381: if both are fp32, we can revert to previous one

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -784,9 +788,800 @@ def abstract(
...
+            raise ImportError("JAX version 0.5.0 or later is required for shardy sharding.")
+        del mesh, result_types
+        input_spec = [(f"…{x}",) for x in range(len(value_types))]
+        rng_sharding = (f"…{len(value_types)}",)
+        input_spec[0] = ("…0", "seqlen", "head", "hidden")
+        out_sharding = ("…0", "seqlen", "head", "hidden")
+        softmax_aux_sharding = ("…0", "head", "seqlen", "i")
+        return SdyShardingRule(
+            tuple(input_spec), (out_sharding, softmax_aux_sharding, rng_sharding)
+        )
```
wangye805, line 1544: The new softmax_aux is [b, h, s_q, s_kv], probably the sharding rule need changes accordingly

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -784,9 +788,800 @@ def abstract(
...
+            SmallSeqAttnFwdPrimitive.outer_primitive.bind(*batched_args, config=config),
+            out_bdims,
+        )
+
+    @staticmethod
+    def infer_sharding_from_operands(config, mesh, arg_infos, result_infos):
+        del config, result_infos
+        q_spec = get_padded_spec(arg_infos[0])
+        out_sharding = NamedSharding(mesh, PartitionSpec(*q_spec))
+        softmax_aux_sharding = NamedSharding(
```
wangye805, line 1510: Not sure how sharding rule need to be changed

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -807,17 +1602,14 @@ def abstract(
...
+        bias_batch = bias_heads = 0
         deterministic = not FusedAttnHelper.is_non_deterministic_allowed()
-
-        input_batch = reduce(operator.mul, batch_shape)
-        wkspace_shape, wkspace_dtype = transformer_engine_jax.get_fused_attn_bwd_workspace_sizes(
-            input_batch,
+        # NVTE uses b = cu_seqlens_q.shape[0] - 1 (one packed segment per slot), not
+        # reduce(batch_shape). E.g. seqpack with max_seqlen_q>1 yields cu length
+        # batch*segments+1 while Q still has leading logical batch only.
+        small_seq_workspace_batch = q_seqlen_or_cu_seqlen_aval.shape[0] - 1
```
wangye805, line 1610: I think at jax side, we don't need to read the actual value of cu_seqlen's.

transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -351,6 +351,185 @@ static void FusedAttnForwardImpl(
...
+    NVTE_Mask_Type mask_type, NVTE_QKV_Layout qkv_layout, DType dtype, DType wkspace_dtype,
+    bool is_training, bool deterministic, int64_t window_size_left, int64_t window_size_right) {
+  (void)deterministic;
+  FUSED_ATTN_IMPL_COMMON_BLOCK;
+  NVTE_CHECK(layout_group == NVTE_QKV_Layout_Group::NVTE_HD_HD_HD && is_ragged,
+             "SmallSeqAttnForwardImpl: requires THD separate Q, K, V.");
+  NVTE_CHECK(bias_batch == 0 && bias_heads == 0,
+             "SmallSeqAttnForwardImpl: bias not supported for small-seq.");
+
+  auto bias_tensor = TensorWrapper(bias, bias_shape, dtype);
```
wangye805, line 430: Do we support bias?

transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -351,6 +351,185 @@ static void FusedAttnForwardImpl(
...
+
+  nvte_populate_rng_state_async(rng_state, seed, q_max_seqlen, kv_max_seqlen,
+                                NVTE_Fused_Attn_Backend::NVTE_CK, stream);
+
+  NVTETensorPack aux_output_tensors;
+  nvte_tensor_pack_create(&aux_output_tensors);
+  PrepareSmallSeqAttnForwardAuxTensors(&aux_output_tensors, input_batch, attn_heads, q_max_seqlen,
+                                       kv_max_seqlen, dtype, softmax_aux, rng_state);
+
+  auto dummy_page_table_tensor = TensorWrapper(nullptr, std::vector<size_t>{1}, DType::kInt32);
```
wangye805, line 453: If we know it's dummy and our nvte_fused_attn_small_seq_fwd do not use it actually, we can skip the passing

wangye805: Same as other tensors not used, for example rng_state

transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -351,6 +351,185 @@ static void FusedAttnForwardImpl(
...
+  (void)deterministic;
+  FUSED_ATTN_IMPL_COMMON_BLOCK;
+  NVTE_CHECK(layout_group == NVTE_QKV_Layout_Group::NVTE_HD_HD_HD && is_ragged,
+             "SmallSeqAttnForwardImpl: requires THD separate Q, K, V.");
+  NVTE_CHECK(bias_batch == 0 && bias_heads == 0,
+             "SmallSeqAttnForwardImpl: bias not supported for small-seq.");
+
+  auto bias_tensor = TensorWrapper(bias, bias_shape, dtype);
+
+  if (is_ragged) {
```
wangye805, line 432: Currently we only support is_ragged. Later we will support BSHD. Please add a TODO comment. Otherwise, we should include all following things into this if branch

wangye805 (changes requested):

Micky774:
This branch is currently missing commits `7faf099f` and `5f592b1a` from [the original PR](https://github.com/ROCm/TransformerEngine/pull/461).

transformer_engine/common/fused_attn_rocm/fused_attn.cpp
```diff
@@ -894,6 +896,218 @@ void nvte_fused_attn_bwd(const NVTETensor Q, const NVTETensor K, const NVTETenso
...
+                                                size_t workspace_bytes, const char *log_tag,
+                                                cudaStream_t stream) {
+  constexpr size_t runtime_seqlen_bytes = sizeof(uint64_t);
+  NVTE_CHECK(workspace_bytes >= runtime_seqlen_bytes, log_tag,
+             "workspace too small to compute runtime max seqlen (need at least ", runtime_seqlen_bytes,
+             " bytes).");
+  const size_t runtime_s_q = static_cast<size_t>(ck_fused_attn::get_runtime_max_seqlen(
+      b, dev_ptr_cu_seqlens_q, nullptr, workspace, stream));
+  const size_t runtime_s_kv = static_cast<size_t>(ck_fused_attn::get_runtime_max_seqlen(
+      b, dev_ptr_cu_seqlens_kv, nullptr, workspace, stream));
```
Micky774, line 931: This calls `ck_fused_attn::get_runtime_max_seqlen` unconditionally and breaks AOTriton-only builds.

transformer_engine/jax/attention.py
Micky774, line 1: Currently the code in this file unconditionally pulls in `tex.fused_attn_small_seq_{f,b}wd` -- we should guard these funcs.

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -784,9 +788,800 @@ def abstract(
...
+    """ROCm small-sequence cross-attention forward"""
+
+    name = "te_small_seq_attn_forward_ffi"
+    multiple_results = True
+    impl_static_args = (13,)
+    inner_primitive = None
+    outer_primitive = None
+
+    @staticmethod
+    def abstract(
```
Micky774, line 1229: Can we also proactively guard against GQA/MQA here?

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -784,9 +788,800 @@ def abstract(
...
+            qkv_layout=int(config.qkv_layout.value),
+            is_training=config.is_training,
+            deterministic=not FusedAttnHelper.is_non_deterministic_allowed(),
+            window_size_left=window_size_left,
+            window_size_right=window_size_right,
+            softmax_type=int(config.softmax_type.value),
+        )
+
+    @staticmethod
+    def impl(
```
Micky774, line 984: `_fix_len_take`, `convert_to_2d`, seqlen/offset processing are all copied between `FusedAttn*Primitive.impl` and `SmallSeqAttn*Primitive.impl` -- let's try to reuse them.

transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -351,6 +351,185 @@ static void FusedAttnForwardImpl(
...
+  nvte_set_tensor_param(&rng_state_aux, kNVTERowwiseData, &rng_state_aux_data);
+}
+
+void PrepareSmallSeqAttnBackwardAuxTensors(NVTETensorPack *tensor_pack, const size_t input_batch,
+                                         const size_t attn_heads, const size_t q_max_seqlen,
+                                         const size_t kv_max_seqlen, DType dtype, void *softmax_buf,
+                                         void *rng_state_buf) {
+  PrepareSmallSeqAttnForwardAuxTensors(tensor_pack, input_batch, attn_heads, q_max_seqlen,
+                                       kv_max_seqlen, dtype, softmax_buf, rng_state_buf);
+}
```
Micky774, line 386: This wrapper is trivial, can we forward directly?

transformer_engine/common/fused_attn_rocm/fused_attn.cpp
```diff
@@ -894,6 +896,218 @@ void nvte_fused_attn_bwd(const NVTETensor Q, const NVTETensor K, const NVTETenso
...
+  if (const char *env_ck = std::getenv("NVTE_LOG_CK_CONFIG");
```
Micky774, line 932: At this API level we use `NVTE_LOG_FUSED_ATTN_CONFIG`

Micky774 (changes requested):
Currently, if I'm understanding this correctly, the only way to utilize the new backend is via the Python-level `fused_attn_small_seq` entrypoint -- is this desired? We have no means of automatically dispatching to this backend through existing API.

VeeraRajasekhar:
> This branch is currently missing commits `7faf099f` and `5f592b1a` from [the original PR](https://github.com/ROCm/TransformerEngine/pull/461).

Yes @Micky774, this was intentional. Those missing commits are relevant to some of the corner cases which we decided to not cover.

VeeraRajasekhar:
> Currently, if I'm understanding this correctly, the only way to utilize the new backend is via the Python-level `fused_attn_small_seq` entrypoint -- is this desired? We have no means of automatically dispatching to this backend through existing API.

We cannot guarantee two conditions i.e., to determine max_seq_len_q = 1 and 2<=max_seq_len_kv<=16, which are known at runtime. i.e.,  these[ two tests ](https://github.com/ROCm/TransformerEngine/blob/b183024fab07d0db49254fdfa3ba554c6c8969be/tests/jax/test_fused_attn.py#L1818-L1819), even though these two donot satisfy the conditions but they will be satisfied on runtime, this is what is blocking us from automatically deploying to this backend. that being said we will throw assertion errors during runtime if these are not met.

Micky774:
> We cannot guarantee two conditions i.e., to determine max_seq_len_q = 1 and 2<=max_seq_len_kv<=16, which are known at runtime. i.e., these[ two tests ](https://github.com/ROCm/TransformerEngine/blob/b183024fab07d0db49254fdfa3ba554c6c8969be/tests/jax/test_fused_attn.py#L1818-L1819), even though these two donot satisfy the conditions but they will be satisfied on runtime, this is what is blocking us from automatically deploying to this backend. that being said we will throw assertion errors during runtime if these are not met.

If we're providing a static entry point anyways, I suspect this means that we expect users to know ahead of time whether their data complies with such a backend. In which case, even mediating with an environment variable should be feasible right? I'm just trying to understand the actual user experience we're trying to support here.

Micky774:
Alternatively, even just adding `max_segment_seqlen_q, max_segment_seqlen_kv` as helper descriptors to the `SequenceDescriptor` spec would suffice to enable an opt-in dispatch that requires user-ownership of data viability.

wangye805:
> Alternatively, even just adding `max_segment_seqlen_q, max_segment_seqlen_kv` as helper descriptors to the `SequenceDescriptor` spec would suffice to enable an opt-in dispatch that requires user-ownership of data viability.

Previously I discussed this with Veera and also our customers. There are two three things prevent us using a unified api as previous ck flow:
1). Due to jax's sequence packing, segment length could be quite different from the max_seqlen_q/kv
2). Due to the jit property, runtime sequence length cannot be obtained during the workspace buffer allocation phase
3). This set of new kernels are basically fused-attn which saves P = softmax(S) of shape [b, h, s_q, s_kv], which is different from ck flow which saves log(row sum of exp (S))

So any good ideas you have can work those around?

PR #544, opened 2026-04-16
[TE] Improve backward performance for CK Tile FP8 Group GEMM

aris134:
# Description

Fixes [#16263](https://github.com/ROCm/frameworks-internal/issues/16263)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Re-expresses the W_grad grouped GEMM in the alternate FP8 storage/layout encoding using columnwise_data, rather than feeding the original TN-style presentation directly into CK. This preserves the math but steers CK away from the less favorable transposed/shuffle handling path in the FP8 pipeline. A similar layout/storage rewrite had previously been applied to the D_grad grouped GEMM; this extends the same idea to W_grad. This change results in a 3.15x median speedup in the W_grad GEMM. See https://github.com/ROCm/frameworks-internal/issues/16263#issuecomment-4256774612
 for the performance breakdown.

# Checklist:


transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm.cpp
```diff
@@ -39,23 +39,53 @@ bool ck_tile_grouped_gemm(const NVTETensor* A,
...
+      if (has_a_col && has_b_col) {
+        use_a_columnwise_data = true;
+        use_b_columnwise_data = true;
+        transA_use = false;
+        transB_use = true;
+      } else if (has_a_col) {
+        // Fallback: preserves math, but does not become NT.
+        use_a_columnwise_data = true;
+        transA_use = false;
+      }
```
matthiasdiener: Is this `else if` part correct?

aris134: Yes mathematically it is correct because if we pass in A columnwise storage to NT GEMM we should not set transA_use for correctness. However, I acknowledge that the default behavior is already correct and this fallback doesn't really serve any particular benefit I know of, so I removed it in 8281de7

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm.cpp
```diff
@@ -61,30 +61,25 @@ bool ck_tile_grouped_gemm(const NVTETensor* A,
   // Cases handled:
   //   1) normalized NN -> NT by reading B from columnwise storage
   //   2) normalized TN -> NT by reading both A and B from columnwise storage
+  const bool has_a_col = A0_te->has_columnwise_data();
+  const bool has_b_col = B0_te->has_columnwise_data();
```
matthiasdiener: Can this go into the `if (is_8bit_float) {` below?

aris134: Yes I agree this makes sense, and I changed it here f83eaf9

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm.cpp
```diff
@@ -39,23 +39,48 @@ bool ck_tile_grouped_gemm(const NVTETensor* A,
...
+  // alternate physical storage view when available.
+  //
+  // Cases handled:
+  //   1) normalized NN -> NT by reading B from columnwise storage
+  //   2) normalized TN -> NT by reading both A and B from columnwise storage
+  if (is_8bit_float) {
+    const bool has_a_col = A0_te->has_columnwise_data();
+    const bool has_b_col = B0_te->has_columnwise_data();
+    // normalized NN: op(A_use)=A, op(B_use)=B
+    if (!transA_use && !transB_use) {
```
ipanfilo: Why do we need to check !transA_use here and !transB_use below? Why not check them independently - not for NN/TN but for A is N/T and B is N/T? The only one integrated check is needed is to make sure all alternative data layout exists, i.e. if ((!transA_use || has_a_col) and (transB_use || has_b_col))

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm.cpp
```diff
@@ -39,23 +39,48 @@ bool ck_tile_grouped_gemm(const NVTETensor* A,
...
-    } 
+  // FP8 special handling:
+  //
+  // Prefer reformulating problematic normalized layouts into NT using the
+  // alternate physical storage view when available.
+  //
+  // Cases handled:
+  //   1) normalized NN -> NT by reading B from columnwise storage
+  //   2) normalized TN -> NT by reading both A and B from columnwise storage
+  if (is_8bit_float) {
```
ipanfilo, line 80: Do we support Float8 A and non-Floaf8 B?

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp8.cpp
```diff
@@ -160,7 +160,17 @@ class QuantGroupedGemmRunner : public RunnerInterface {
       transformer_engine::Tensor* D_te =
           transformer_engine::convertNVTETensorCheck(ctx.D[i]);
 
-      const auto& a = data_view(*A_te);
+      const transformer_engine::SimpleTensor* a_src = nullptr;
+      if (ctx.use_a_columnwise_data) {
+        if (!A_te->has_columnwise_data()) {
```
ipanfilo: nit: NVTE_CHECK can be used for that assertion

ipanfilo (changes requested):

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm.cpp
```diff
@@ -39,23 +39,48 @@ bool ck_tile_grouped_gemm(const NVTETensor* A,
...
-      transB_use = true;
-    } 
+  // FP8 special handling:
+  //
+  // Prefer reformulating problematic normalized layouts into NT using the
+  // alternate physical storage view when available.
+  //
+  // Cases handled:
+  //   1) normalized NN -> NT by reading B from columnwise storage
+  //   2) normalized TN -> NT by reading both A and B from columnwise storage
```
alextmagro: I think there is a bit of confusion around this (looping in @matthiasdiener for reference), but I believe that this is a TN GEMM, not a NT GEMM. BLAS convention is that data should be in a colwise-major format. CK might do something different, but given that TE tends to follow BLAS convention I think we should update these comments to reflect that.

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_common.h
```diff
@@ -84,6 +84,7 @@ struct GroupedGemmRunContext {
     size_t workspace_bytes = 0;
     hipStream_t stream = nullptr;
 
+    bool use_a_columnwise_data = false;
```
alextmagro, line 87: This is declared and defined in both header and source.

alextmagro (changes requested):

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm.cpp
alextmagro: In hindsight, I think I might be wrong about this, since it is normalized. I'm wondering if we could make this comment a bit more clear about what exactly we're doing and why -- for one, I would expect CK performance for TN to be much better, so am still trying to figure out why this NT trick works...

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm.cpp
aris134: We keep the two explicit cases because the rewrite is layout-specific, not per-operand. In practice, we only want to transform two starting layouts: for normalized NN we rewrite B only, and for normalized TN we rewrite A and B together. Checking transA_use / transB_use as a pair is how we distinguish those two cases. A condition based only on whether alternate storage exists would not capture that distinction and would make the transformation look independent per operand, which is not the intent here.

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm.cpp
aris134: Not outside the FP8 family. This path does support mixed FP8-family combinations such as fp8/bf8 or bf8/fp8, but both operands still have to be FP8-family types. A non-FP8-family B operand is not supported here.

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm.cpp
aris134: Agreed — the comment was misleading. The underlying normalized GEMM here is TN, and the confusion came from me describing the post-rewrite encoding too loosely. What this code is actually doing is an FP8-specific CK pipeline rewrite: it re-expresses that same GEMM using columnwise_data to avoid the less favorable transposed/shuffle handling path in the CK FP8 pipeline. I have updated the comment in a5f679b

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_common.h
aris134: I think the confusion was the identical local/member names rather than a duplicate definition. I renamed the local variables to use_a_colwise_data / use_b_colwise_data to make the distinction clearer. See 291e8d9

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp8.cpp
aris134: Done in 1a72447

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm.cpp
```diff
@@ -39,23 +39,57 @@ bool ck_tile_grouped_gemm(const NVTETensor* A,
...
+//     -> rewrite B only:
+//        `data + N` -> `columnwise_data + T`
+//     The original NN form is not the preferred FP8 matmul encoding here.
+//
+//   normalized TN: `op(A_use)=A^T`, `op(B_use)=B`
+//     -> rewrite both operands:
+//        A: `data + T` -> `columnwise_data + N`
+//        B: `data + N` -> `columnwise_data + T`
+//     This avoids the extra transpose/shuffle handling path that the original
+//     TN presentation can trigger in the FP8 CK pipeline.
```
wangye805: nit: comment alignment not consistent with above or below

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm.cpp
```diff
@@ -39,23 +39,57 @@ bool ck_tile_grouped_gemm(const NVTETensor* A,
...
+    // normalized NN: op(A_use)=A, op(B_use)=B
+    if (!transA_use && !transB_use) {
+      if (has_b_col) {
+        use_b_colwise_data = true;
+        transB_use = true;
+      }
+    }
+
+    // normalized TN: op(A_use)=A^T, op(B_use)=B
+    else if (transA_use && !transB_use) {
```
wangye805: nit: can you put the else if closer with the if branch?

wangye805 (approved):
LGTM with minor coding style comments

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm.cpp
aris134: Done in 9cce253

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm.cpp
aris134: Done in 9cce253

alextmagro (approved):
LGTM!

matthiasdiener (approved):

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm.cpp
ipanfilo: Equivalent logic for what you have is:
if (!transB_use && has_b_col) {
if (!transA_use) { 
  //change B
} else if (has_a_col) {
  //change A&B
}
}
An the outer condition there is because you don't accept TT as a possible input layout.
My proposal is generalize it, not try to explicitly exclude layouts.
We need to pass NT to GEMM which means:
(!transA_use /*already N*/ or has_a_col /*can be normalized*/) and (transB_use /*alteady T*/ or has_b_col /*can be normalized*/) is the outermost condition when normalization is  possible. And inside we check separately if it is required for A and/or B

aris134: Ok generalized this in 8259446

ipanfilo (approved):

PR #545, opened 2026-04-16
[ROCm] fix the bug in hipfied optimized cast tranpose flow that two kernels got run

wangye805:
# Description

In ROCm TE when NVTE_USE_OPTIMIZED_HIPIFIED_CAST_TRANSPOSE is set, no matter the optimized kernel run or not, it will still run another kernel, either fallback to the cost model rtc one, or the general one. This PR fixes this issue

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes
1). fix the bug in hipfied optimized cast tranpose flow that two kernels got run
2). rename the do_general_config to make it easier to understand
3). added a new unit test

# Checklist:

### DIFF
--- a/ci/pytorch.sh
+++ b/ci/pytorch.sh
@@ -65,6 +65,7 @@ run_test_config(){
     run_default_fa 1 test_recipe.py
     run 1 test_sanity.py
     run_default_fa 1 test_sanity_import.py
+    run_default_fa 3 test_sanity_hipified_cast_transpose.py
     run_default_fa 1 attention/test_attention.py # Backend selection is controlled by the test
     run_default_fa 1 attention/test_cp_utils.py
     run_default_fa 1 attention/test_kv_cache.py
--- a/tests/pytorch/test_sanity_hipified_cast_transpose.py
+++ b/tests/pytorch/test_sanity_hipified_cast_transpose.py
@@ -1,0 +1,81 @@
+# Copyright (c) 2022-2026, Advanced Micro Devices, Inc. All rights reserved.
+# License for AMD contributions = MIT. See LICENSE for more information
+
+"""
+Non-regression tests for the hipified cast_transpose kernel dispatch.
+
+Verifies that tex.quantize (which routes through cast_transpose on AMD)
+dispatches exactly one GPU kernel per call.  A previous bug (commit 542b8c7b)
+caused the NVTE_USE_OPTIMIZED_HIPIFIED_CAST_TRANSPOSE path to launch both
+cast_transpose_optimized_kernel AND cast_transpose_general_kernel per call,
+doubling the GPU work.
+"""
+
+import os
+import pytest
+import torch
+from torch.profiler import profile, ProfilerActivity
+
+from transformer_engine.pytorch import cpp_extensions as tex
+from transformer_engine.pytorch.tensor.float8_tensor import Float8Quantizer
+
+
+def _fill_uniform(shape, dtype):
+    """Create a deterministic random tensor on GPU."""
+    gen = torch.Generator(device="cuda")
+    gen.manual_seed(12345)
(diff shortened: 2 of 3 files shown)


transformer_engine/common/transpose/cast_transpose.cu
```diff
@@ -412,7 +412,12 @@ void cast_transpose(const Tensor &input, const Tensor &noop, Tensor *output_, cu
                                  static_cast<const CType *>(output.scale.dptr),
                                  static_cast<CType *>(output.amax.dptr),
                                  static_cast<CType *>(output.scale_inv.dptr), row_length, num_rows);
+#ifdef __HIP_PLATFORM_AMD__
+            }
+            if (!aligned || !rtc::is_enabled()) {  // Statically-compiled general kernel
```
alextmagro: Instead of adding another if statement, can't we align the else on L419 with the if on L269? We would then move the close bracket on L355 to the end of the function.

Alternatively, could we make this an else if to make sure readers understand it is connected.

alextmagro (approved):
LGTM -- one comment on the fix itself but is a suggestion

transformer_engine/common/transpose/cast_transpose.cu
wangye805: Done. Thanks!

tests/pytorch/test_sanity_hipified_cast_transpose.py
```diff
@@ -0,0 +1,81 @@
...
+    (NVTE_USE_OPTIMIZED_HIPIFIED_CAST_TRANSPOSE=1).
+    """
+    input_tensor = _fill_uniform(shape, dtype=in_dtype)
+    scale = torch.rand(1, dtype=torch.float32, device="cuda") * 3.0 - 2.0
+    amax = torch.zeros(1, dtype=torch.float32, device="cuda")
+
+    env_key = "NVTE_USE_OPTIMIZED_HIPIFIED_CAST_TRANSPOSE"
+    old_val = os.environ.get(env_key)
+
+    try:
```
ipanfilo, line 51: If you want to restore the env, better use monkeypatch instead of try/finally. Also will it be more clear to set NVTE_USE_OPTIMIZED_HIPIFIED_CAST_TRANSPOSE in both cases so the test behaviour does not depend on default value?

transformer_engine/common/transpose/cast_transpose.cu
ipanfilo, line 1: For me it looks like refactoring - rename do_general_config -> fallback_to_cost_model_rtc and move fallback path under 'if (aligned && rtc::is_enabled())' but corresponding condition is set under that if anyway.

ci/pytorch.sh
```diff
@@ -65,6 +65,7 @@ run_test_config(){
     run_default_fa 1 test_recipe.py
     run 1 test_sanity.py
     run_default_fa 1 test_sanity_import.py
+    run_default_fa 3 test_sanity_hipified_cast_transpose.py
```
ipanfilo, line 68: nit: put before sanity_import

wangye805: Done in https://github.com/ROCm/TransformerEngine/pull/549

tests/pytorch/test_sanity_hipified_cast_transpose.py
wangye805: Done in https://github.com/ROCm/TransformerEngine/pull/549

transformer_engine/common/transpose/cast_transpose.cu
wangye805: Yeah, it's a refactoring and a bug fixing. 

Previously do_general_config also means do the cost model rtc run, which falls into the aligned && rtc_enabled flow. So this refactoring will keep the same logic (but fix the bug)

ipanfilo: Thank you, I finally figured out the problem with original logic

PR #547, opened 2026-04-17
Enable CI lint gh action on ROCm

VeeraRajasekhar:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
```diff
@@ -76,7 +76,13 @@ void set_aiter_asm_dir() {
...
   std::call_once(aiter_asm_dir_once, []() {
     Dl_info info;
-    dladdr((void*)set_aiter_asm_dir, &info);
+    // dladdr expects void*; avoid reinterpret_cast<void*>(fn) (not ISO C++).
+    union {
+      void (*fn)();
+      void *addr;
+    } sym{};
+    sym.fn = set_aiter_asm_dir;
+    dladdr(sym.addr, &info);
```
Micky774, line 85: IMO this is unnecessary. Yes, it quiets the warning, but the warning is irrelevant for us granted our support is POSIX focused to begin with. From the [`dlopen` man page](https://man7.org/linux/man-pages/man3/dlopen.3.html):
```
           /* According to the ISO C standard, casting between function
              pointers and 'void *', as done above, produces undefined results.
              POSIX.1-2001 and POSIX.1-2008 accepted this state of affairs and
              proposed the following workaround:

                  *(void **) &cosine = dlsym(handle, "cos");

              This (clumsy) cast conforms with the ISO C standard and will
              avoid any compiler warnings.

              The 2013 Technical Corrigendum 1 to POSIX.1-2008 improved matters
              by requiring that conforming implementations support casting
              'void *' to a function pointer.  Nevertheless, some compilers
              (e.g., gcc with the '-pedantic' option) may complain about the
              cast used in this program.  */
```
the union trick here provides no additional safety -- it's still undefined behavior technically speaking -- and will break in the same circumstances (non-POSIX risk).

All things considered, I'd rather we keep things as-is, and if we really want to deal with the warning, we can make a small utility to use pragmas to suppress the warnings locally around the cast.

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -117,7 +117,7 @@ public:
     data[key][stream] = item; 
   }
 
-  ObjCache(void (*a_offload)(const Data&)): offload(a_offload) {}
+  explicit ObjCache(void (*a_offload)(const Data&)): offload(a_offload) {}
```
Micky774, line 120: Do we really want these constructors to be explicit? Are they even used implicitly anywhere in our codebase?

commit.txt
Micky774, line 1: I'm assuming this won't be part of the final PR?

Micky774 (commented):
Does this cause divergence from upstream's source code in inherited files? Also, why do we need linting as a CI action instead of e.g. a pre-commit?

commit.txt
VeeraRajasekhar: Yes, This is for me to keep track of the issues fixed, will remove this.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
```diff
@@ -130,9 +136,10 @@ std::ostream* get_ck_log_stream() {
       if (!log_dir_str.empty() && log_dir_str != "0") {
         if (log_dir_str == "1") {
           log_stream = &std::cout;
-        }
-        else if (open_ck_fused_attn_log_file(log_file, "ck_fused_attn", log_dir_str)) {
```
ipanfilo, line 134: What is a warning for if-else if? I think it is used a lot in our code

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -769,14 +769,17 @@ protected:
       }
 
 #if HIPBLASLT_VERSION_MAJOR > 0 || HIPBLASLT_VERSION_MINOR >= 15
-      if (cfg.scaling_mode < 0 || cfg.scaling_mode >= (int)HIPBLASLT_MATMUL_MATRIX_SCALE_END)
```
ipanfilo, line 772: Line length and { at the end of the line are understood but lint does not require duplicate body of If.

transformer_engine/common/normalization/common.cpp
```diff
@@ -544,24 +544,26 @@ NormalizationPlanBase* NormalizationPlanRegistry::getNormalizationPlan(
     plan = std::make_unique<CudnnNormalizationPlan>(NormType, NormStage, wtype, itype, otype, ctype,
                                                     batch_size, hidden_size, sm_count,
                                                     zero_centered_gamma, mode, training);
-  } else
+  }
 #endif
```
ipanfilo: Let's get rid of nested ifs. If splitting else-if does not work, better add dummy 'if (false) {' for ROCm instead of 'if (NOrmBAckend...'

transformer_engine/common/__init__.py
```diff
@@ -51,8 +51,8 @@ def _is_package_installed_from_wheel(package) -> bool:
     if not te_wheel_file:
         return False
 
-    with te_wheel_file.open("r") as f:
-        for line in f:
```
ipanfilo, line 55: It is upstream code. Why change it?

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -9,9 +9,10 @@
 import warnings
 from dataclasses import dataclass, replace
 from functools import partial, reduce
-from packaging import version
 from typing import Optional, Tuple
 
+from packaging import version
```
ipanfilo: Why is it needed?

transformer_engine/jax/cpp_extensions/normalization.py
```diff
@@ -8,16 +8,18 @@
...
 import jax
 import jax.numpy as jnp
 from jax import dtypes, ffi
-from jax.experimental.custom_partitioning import SdyShardingRule, BATCHING
+if version.parse(jax.__version__) >= version.parse("0.5.0"):
+    from jax.experimental.custom_partitioning import SdyShardingRule
+from jax.experimental.custom_partitioning import BATCHING
 from jax.interpreters.mlir import ir
 from jax.sharding import PartitionSpec
-from .misc import is_hip_extension
```
ipanfilo, line 20: Does the file have ROCm code?

transformer_engine/jax/cpp_extensions/normalization.py
```diff
@@ -8,16 +8,18 @@
...
-from packaging import version
 from typing import Optional, Union
 
+from packaging import version
+
 import jax
 import jax.numpy as jnp
 from jax import dtypes, ffi
-from jax.experimental.custom_partitioning import SdyShardingRule, BATCHING
+if version.parse(jax.__version__) >= version.parse("0.5.0"):
```
ipanfilo: IT is not needed, we don;t actually support JAX < 0.5

transformer_engine/jax/quantize/helper.py
```diff
@@ -26,14 +26,8 @@
 import jax.numpy as jnp
 from flax.core.frozen_dict import FrozenDict
 
-from ..util import is_hip_extension, get_jnp_float8_e4m3_type, get_jnp_float8_e5m2_type
+import transformer_engine_jax as tejax
```
ipanfilo: All those changes will result in IFU conflicts

transformer_engine/jax/setup.py
```diff
@@ -51,11 +51,11 @@
 from build_tools.te_version import te_version
 from build_tools.jax import setup_jax_extension, install_requirements, test_requirements
 
-from pybind11.setup_helpers import build_ext as BuildExtension
```
ipanfilo, line 54: Why is it needed?

ipanfilo (changes requested):

Micky774 (changes requested):
@VeeraRajasekhar I still have questions from my initial review unanswered.

The main one which @ipanfilo has also asked is: why such divergence? It will be a huge maintenance burden to reconcile against _every_ subsequent IFU, unless NV upstream implements these _exact same_ decisions (afaik they don't?).

Enforcing linting on ROCm-only files is fine, but we have constraints regarding minimizing diff with upstream even if it means letting some things go "un-linted".

transformer_engine/common/gemm/rocm_gemm.cu
VeeraRajasekhar: they aren’t used implicitly anywhere; the only constructions are direct ObjCache<T,K>(nullptr) from ObjPool and direct init of service_stream_cache with a lambda. explicit was added only to satisfy cpplint runtime/explicit. If we prefer not to mark these callbacks as explicit, we can drop explicit and suppress that line with NOLINT for cpplint instead.

There are no APIs that take an ObjCache by value. So explicit is not required for correctness, only for style / tooling.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
VeeraRajasekhar: The category was readability/braces. The message was along the lines of: “If an else has a brace on one side, it should have it on both.”

So the warning showed up because cpplint’s readability/braces heuristic fired on this if / else if layout, not because else if is forbidden. Nesting as else { if (...) { ... } } makes the structure obvious to the linter and cleared the warning.

transformer_engine/common/gemm/rocm_gemm.cu
VeeraRajasekhar: Yes, the duplicate was to fix the readability/braces warning, I have restructured it to compute a bool in preprocessor-only branches, then use one if body.

transformer_engine/common/normalization/common.cpp
VeeraRajasekhar: Dropped the nested if (!plan) / inner if (Forward) structure. and added the following more clear structure:
> CUDA: if (Cudnn) … else if (Forward TE) … else (backward TE) in one #ifndef __HIP_PLATFORM_AMD__ block.
> ROCm: if (Forward TE) … else (backward TE) in #else, with mode/training on TE constructors.

transformer_engine/jax/cpp_extensions/attention.py
VeeraRajasekhar: This is used to check the following

```
if version.parse(jax.__version__) >= version.parse("0.5.0"):
    from jax.experimental.custom_partitioning import SdyShardingRule
```

I will update the codebase to remove all these checks as we no longer support Jax<0.5.0 as mentioned in https://github.com/ROCm/TransformerEngine/pull/547#discussion_r3121416919

transformer_engine/jax/cpp_extensions/normalization.py
VeeraRajasekhar: Yes, jax version checks were added.

transformer_engine/jax/cpp_extensions/normalization.py
VeeraRajasekhar: Updated

transformer_engine/jax/setup.py
VeeraRajasekhar: Reverted and added # pylint: disable=ungrouped-imports

transformer_engine/pytorch/triton_kernels/gmm/gmm_kernels.py
```diff
@@ -255,7 +256,7 @@ def gmm_kernel(
         tl.device_assert(num_tiles >= 0, "num_tiles < 0")
 
         # Loop through tiles of current MM problem.
-        while tile >= last_mm_tile and tile < last_mm_tile + num_tiles:
+        while last_mm_tile <= tile < last_mm_tile + num_tiles:
```
Micky774, line 259: Triton doesn't support this multi-comparison -- it looks like the linter is trying to enforce python semantics on it

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_utils.cpp
ipanfilo: Move them to single line then but do not create nested ifs

transformer_engine/common/gemm/rocm_gemm.cu
ipanfilo: It looks like there is discrepancy here. .clang-format sets line width 100 and this is what IDE uses. So cpplint should be configured accorfingly

ipanfilo (changes requested):

PR #569, opened 2026-05-01
Fix canonicalize MXFP8 GEMM to TN layout for other layouts.

sudhu2k:
# Description
This PR fixes a bug and the below warning when using MXFP8 GEMM for wgrad computation (layout = NN).
```
shared/rocroller/lib/source/KernelGraph/Transformations/SwizzleScale.cpp:793: FatalError(m_params->transposeMemoryAccess[LayoutType::MATRIX_A] && !m_params->transposeMemoryAccess[LayoutType::MATRIX_B])
Non-TN is not supported by SwizzleScale
```

Issue:
ret.A = is_A_transposed ? A.data.dptr : A.columnwise_data.dptr;
ret.transA = transA;

ret.B = is_B_transposed ? B.columnwise_data.dptr : B.data.dptr;
ret.transB = transB;

For NN (transA=N, transB=N), the MXFP8 branch selects columnwise_data for A and data (rowwise) for B, but historically left transA and transB as N in the descriptor, which is incorrect.

If MXFP8 GEMM supports any layout, we should completely use rowwise data only or columnwise data only. But looking at the warning message above it only supports TN layout, which means the data selection is correct, but the flags aren't getting updated. This PR fixes it.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/transformer_engine/common/gemm/rocm_gemm.cu
+++ b/transformer_engine/common/gemm/rocm_gemm.cu
@@ -314,7 +314,7 @@ GemmParam CanonicalizeGemmInput(const transformer_engine::Tensor &A, const cubla
       NVTE_CHECK(A.has_columnwise_data(), "Input A is missing column-wise usage");
     }
     ret.A = is_A_transposed ? A.data.dptr : A.columnwise_data.dptr;
-    ret.transA = transA;
+    ret.transA = is_A_transposed ? transA : CUBLAS_OP_T;
     ret.Atype = is_A_transposed ? A.data.dtype : A.columnwise_data.dtype;
     ret.A_scale_inv = is_A_transposed ? A.scale_inv.dptr : A.columnwise_scale_inv.dptr;
     ret.lda = is_A_transposed ? k : m;
@@ -360,7 +360,7 @@ GemmParam CanonicalizeGemmInput(const transformer_engine::Tensor &A, const cubla
       NVTE_CHECK(B.has_data(), "Input B is missing row-wise usage");
     }
     ret.B = is_B_transposed ? B.columnwise_data.dptr : B.data.dptr;
-    ret.transB = transB;
+    ret.transB = is_B_transposed ? CUBLAS_OP_N : transB;
     ret.Btype = is_B_transposed ? B.columnwise_data.dtype : B.data.dtype;
     ret.B_scale_inv = is_B_transposed ? B.columnwise_scale_inv.dptr : B.scale_inv.dptr;
     ret.ldb = is_B_transposed ? n : k;


transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -314,7 +314,7 @@ GemmParam CanonicalizeGemmInput(const transformer_engine::Tensor &A, const cubla
       NVTE_CHECK(A.has_columnwise_data(), "Input A is missing column-wise usage");
     }
     ret.A = is_A_transposed ? A.data.dptr : A.columnwise_data.dptr;
-    ret.transA = transA;
+    ret.transA = is_A_transposed ? transA : CUBLAS_OP_T;
```
wangye805, line 317: This is confusing, 

is_A_transposed = (transA == CUBLAS_OP_T), then you suggested to change ret.transA to always CUBLAS_OP_T? Even if we are in NN layout?

NV upstream follows the same behavior: https://github.com/NVIDIA/TransformerEngine/blob/5671fd3675906cda1ade26c24a65d3dedd88eb89/transformer_engine/common/gemm/cublaslt_gemm.cu#L186. 

So is there a behavior difference btw our hipblaslt and cublaslt?

wangye805 (changes requested):

transformer_engine/common/gemm/rocm_gemm.cu
sudhu2k: Yes, because if you see the data
`ret.A = is_A_transposed ? A.data.dptr : A.columnwise_data.dptr;`  
`ret.B = is_B_transposed ? B.columnwise_data.dptr : B.data.dptr;
`
For A, if it's T -> read rowwise data, if it's N -> read columnwise data.
For B, if it's T -> read columnwise data, if it's N -> read rowwise data.

For NN layout, the code then reads, A-> columnwise and B-> rowwise.
To make it a proper NN, the layout of A has to be changed to T, since reading a columnwise data as T would be equivalent to reading a rowwise data as N. No change is needed for B in this case, since it's already N. Let me know if this makes sense.

Either there's a different behavior or a bug on the upstream as well.

sudhu2k:
Currently working on checking why unit tests didn't catch this issue.

transformer_engine/common/gemm/rocm_gemm.cu
sudhu2k: Also since we've transformed the data and not updated the layouts, we are seeing
```
shared/rocroller/lib/source/KernelGraph/Transformations/SwizzleScale.cpp:793: FatalError(m_params->transposeMemoryAccess[LayoutType::MATRIX_A] && !m_params->transposeMemoryAccess[LayoutType::MATRIX_B])
Non-TN is not supported by SwizzleScale
```
From the rocroller.
It is still passed as NN, but the data is now transformed to be read as a TN...

alextmagro: Rowwise data is not the transpose of colwise data in MXFP8, so they are not interchangeable. As for the rocroller warning, this is a non issue according to hipBLASlt. Basically, rocroller looks for a solution, and prints out these "fatal errors" that are warnings about a solution not being supported.

sudhu2k:
Closing this PR, since the columnwise data isn't stored as a transpose, this would make this PR wrong and changes the Transpose flags when the columnwise data itself is stored as rowwise.

PR #574, opened 2026-05-05
ck_tile grouped gemm: more padding options

matthiasdiener:
# Description

Enabling padding always causes a significant (~15%) reduction in speed, so only enable it when necessary.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16.cpp
```diff
@@ -216,6 +219,37 @@ bool ck_tile_grouped_gemm_fp16_dispatch(DType a_dtype,
...
+      if (need_m_pad && need_k_pad)
+        break;
+    }
+  }
+
+  // CK tile kernel produces incorrect results with kPadK + ColMajor B.
+  // Fall back to cuBLAS for this combination.
+  if (need_k_pad && ctx.transB) {
+    return false;
+  }
```
aris134, line 70: Assuming we have access to B column-wise storage, could we avoid this fallback by selecting B’s column-wise buffer and calling CK with transB=false, while preserving the same logical GEMM? In other words, is the incorrect-result issue specific to CK’s kPadK + transB=true / ColMajor-B path, or would the columnwise-buffer normalization still hit the same underlying issue?

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_common.h
```diff
@@ -140,7 +140,17 @@ static inline bool launch_grouped_gemm_kernel(const DescContainer& descs,
 
   if (!Kernel::IsSupportedArgument(kargs)) {
     NVTE_WARN("ck_tile_grouped_gemm: CK_Tile kernel arguments not supported for this config. "
-              "Falling back.");
+              "transA=", ctx.transA, " transB=", ctx.transB,
+              " accumulate=", ctx.accumulate, " groups=", ctx.group_num,
+              ". Falling back. "
+              "CK_Tile constraints for bf16/fp16: "
+              "contiguous dim of A and B must be dword-aligned (even), "
+              "N must be multiple of 16 (GetVectorSizeC).");
```
aris134: I'm not sure this function (GetVectorSizeC) implies that N must be a multiple of 16. For the M16N16K32 warp-gemm path, the relevant attributes appear to be kN = 16 and kCNLane = 16, so in the row-major non-TransposeC case this returns kCNLane / kN = 1. That seems to describe the per-thread contiguous C vector size rather than an N divisibility requirement. Is there another place where the N % 16 == 0 constraint is enforced or assumed?

tests/pytorch/test_numerics.py
```diff
@@ -3078,6 +3078,128 @@ def test_grouped_gemm(shape, dtype, layout, accumulate, use_cutlass):
         os.environ.pop("NVTE_USE_CUTLASS_GROUPED_GEMM", None)
 
 
+@pytest.mark.skipif(
+    torch.cuda.get_device_capability() != (9, 0) and not IS_HIP_EXTENSION,
+    reason="Only enable CUTLASS/CK grouped gemm on Hopper or ROCm",
+)
+@pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float16], ids=str)
+@pytest.mark.parametrize("layout", ["TN", "NN"])
```
aris134: NT?

aris134: Oh perhaps the reason NT is left out has to do with your comment about kPadK + column major B

tests/pytorch/test_numerics.py
```diff
@@ -3078,6 +3078,128 @@ def test_grouped_gemm(shape, dtype, layout, accumulate, use_cutlass):
...
+        total_m = sum(m_vals)
+        out = [torch.randn(total_m, n_val, dtype=dtype, device="cuda")]
+        out_ref = [o.clone() for o in torch.split(out[0], m_vals)]
+        m_splits = m_vals
+        grad = False
+        single_output = True
+    else:  # NN
+        # NN GEMM: M=m_splits[i], N=A.cols, K=A.rows
+        if pad_dim == "K":
+            gemm_k = unaligned_k
```
aris134: This test seems to cover one unaligned dimension at a time, but not the combined kPadM && kPadK case. Since the dispatch logic can instantiate a runner with padding set for both dimensions, should we add a case where both M and K are unaligned?

aris134 (changes requested):

tests/pytorch/test_numerics.py
matthiasdiener: Corrected in https://github.com/ROCm/TransformerEngine/pull/574/commits/225c3dca9e86143214bd9ffdd640c03416e41e5a

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16.cpp
matthiasdiener: I implemented the fallback in 29390173 and it seems to work.

tests/pytorch/test_numerics.py
matthiasdiener: I added NT in https://github.com/ROCm/TransformerEngine/pull/574/commits/29390173c5b0cca9e5a040276f166ad5180b3aa7

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_common.h
matthiasdiener: Removed this part of the comment in https://github.com/ROCm/TransformerEngine/pull/574/commits/225c3dca9e86143214bd9ffdd640c03416e41e5a

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16_nn.cpp
aris134, line 1: At the very least, if we have to introduce multiple files to avoid template explosion, can we remove the code redundancy? I'm wondering if you can template ck_tile_grouped_gemm_fp16_dispatch on the A/B layouts, so that you end up with something like this maybe:

```
namespace transformer_engine::grouped_gemm {

bool ck_tile_grouped_gemm_fp16_dispatch_nn(
    DType a_dtype,
    DType d_dtype,
    bool need_m_pad,
    bool need_k_pad,
    const GroupedGemmRunContext& ctx) {
  return ck_tile_grouped_gemm_fp16_dispatch_layout<RowMajor, RowMajor>(
      a_dtype, d_dtype, need_m_pad, need_k_pad, ctx);
}

```

This way each ck_grouped_gemm_fp16_**.cpp would at least become thinner, and you're not having to repeat the nested TE SWITCH statements in each file.

aris134 (changes requested):

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16_nn.cpp
matthiasdiener: Good idea, thanks. What do you think of https://github.com/ROCm/TransformerEngine/pull/574/commits/01f62d08e3575009d2e4581e9a4e0d58e2deda55 ?

aris134: Yeah this looks a lot better! Thanks

tests/pytorch/test_numerics.py
```diff
@@ -3075,6 +3075,166 @@ def test_grouped_gemm(shape, dtype, layout, accumulate, use_cutlass):
         os.environ.pop("NVTE_USE_CUTLASS_GROUPED_GEMM", None)
 
 
+@pytest.mark.skipif(
+    torch.cuda.get_device_capability() != (9, 0) and not IS_HIP_EXTENSION,
+    reason="Only enable CUTLASS/CK grouped gemm on Hopper or ROCm",
+)
+@pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float16], ids=str)
+@pytest.mark.parametrize("layout", ["TN", "NN", "NT"])
```
aris134: TT?

tests/pytorch/test_numerics.py
```diff
@@ -3075,6 +3075,166 @@ def test_grouped_gemm(shape, dtype, layout, accumulate, use_cutlass):
...
+            n_val = n_aligned
+        elif pad_dim == "MK":
+            k_val = unaligned_k
+            m_vals = unaligned_m
+            n_val = n_aligned
+        else:  # N
+            k_val = k_aligned
+            m_vals = [m_aligned] * z
+            n_val = unaligned_n
+
```
aris134: Would we want an MKN unaligned test? Would that cover something that isn't included in the current test sweep?

tests/pytorch/test_numerics.py
```diff
@@ -3075,6 +3075,166 @@ def test_grouped_gemm(shape, dtype, layout, accumulate, use_cutlass):
...
+            m_vals = unaligned_m
+            n_val = n_aligned
+        elif pad_dim == "MK":
+            k_val = unaligned_k
+            m_vals = unaligned_m
+            n_val = n_aligned
+        else:  # N
+            k_val = k_aligned
+            m_vals = [m_aligned] * z
+            n_val = unaligned_n
```
aris134: Can we factor out this if-elif-elif-else block that seems repeated for each layout?

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16.cpp
```diff
@@ -6,250 +6,80 @@
...
-  bool run(const ck_tile::stream_config& stream_cfg,
-           const GroupedGemmRunContext& ctx) override {
-    auto descs = build_descs(ctx);
-    if (descs.empty()) {
+    if (!all_have_columnwise) {
       return false;
     }
-    return launch_grouped_gemm_kernel<Kernel>(descs, ctx, stream_cfg);
+    // Dispatch with B's columnwise buffer as RowMajor (transB=false).
+    GroupedGemmRunContext ctx_nn = ctx;
```
aris134: nit: ctx_nn seems a bit misleading since this only rewrites B as non-transposed via columnwise_data; A can still be T or N. Maybe rename to something like ctx_b_colwise?

tests/pytorch/test_numerics.py
```diff
@@ -3075,6 +3075,166 @@ def test_grouped_gemm(shape, dtype, layout, accumulate, use_cutlass):
...
+        A = [torch.randn(gemm_k, n_out, dtype=dtype, device="cuda") for _ in range(z)]
+        B = [torch.randn(m, gemm_k, dtype=dtype, device="cuda") for m in m_vals]
+        total_m = sum(m_vals)
+        out = [torch.randn(total_m, n_out, dtype=dtype, device="cuda")]
+        out_ref = [o.clone() for o in torch.split(out[0], m_vals)]
+        m_splits = m_vals
+        grad = True
+        single_output = True
+    else:  # NT
+        # NT GEMM: out[i] = A[i]^T @ B[i], A[i]: (m_i, k), B[i]: (m_i, n), out[i]: (n, k)
```
aris134: nit: this comment is a little confusing. For the grouped path, the user-facing NT inputs are A=(m_i,k), B=(m_i,n), out=(n,k), but normalization swaps operands/layouts before dispatch, so the actual dispatched gemm is B^T @ A = (n,m_i) @ (m_i,k).

aris134 (changes requested):

tests/pytorch/test_numerics.py
matthiasdiener: Added TT in aee2c4c2.

tests/pytorch/test_numerics.py
matthiasdiener: I added an MKN test in aee2c4c2.

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16.cpp
matthiasdiener: Renamed in aee2c4c2.

tests/pytorch/test_numerics.py
matthiasdiener: Right, I removed this comment in aee2c4c2

tests/pytorch/test_numerics.py
matthiasdiener: Restructured this in aee2c4c2

aris134 (approved):
LGTM!

aris134:
Quick follow-up question: are there certain padding cases/shapes where we should prefer fallback due to the performance penalty of the padded path?

matthiasdiener:
> Quick follow-up question: are there certain padding cases/shapes where we should prefer fallback due to the performance penalty of the padded path?

I looked at this briefly but could not find a config where this would be profitable, at least for bf16.

tests/pytorch/test_numerics.py
```diff
@@ -3075,6 +3075,144 @@ def test_grouped_gemm(shape, dtype, layout, accumulate, use_cutlass):
...
+            dtype,
+            m_splits=m_splits,
+            grad=grad,
+            accumulate=accumulate,
+            layout=layout,
+            single_output=single_output,
+        )
+
+        for o, o_ref in zip(out, out_ref):
+            if IS_HIP_EXTENSION and accumulate and dtype == torch.bfloat16 and get_device_compute_capability() == (9, 4):
```
ipanfilo: The test itself is IS_HIP_EXTENSION only

tests/pytorch/test_numerics.py
```diff
@@ -3075,6 +3075,144 @@ def test_grouped_gemm(shape, dtype, layout, accumulate, use_cutlass):
...
+        m_aligned = 256
+        n_aligned = 2048
+
+        # Select (un)aligned values based on pad_dim.
+        k_val = unaligned_k if "K" in pad_dim else k_aligned
+        m_vals = unaligned_m if "M" in pad_dim else [m_aligned] * z
+        n_val = unaligned_n if "N" in pad_dim else n_aligned
+
+        total_m = sum(m_vals)
+        os.environ["NVTE_USE_CUTLASS_GROUPED_GEMM"] = "1"
```
ipanfilo: nit: better use monkeypath to make sure the envs are cleared if tests fails

tests/pytorch/test_numerics.py
```diff
@@ -3075,6 +3075,144 @@ def test_grouped_gemm(shape, dtype, layout, accumulate, use_cutlass):
...
+        """
+        torch.manual_seed(0)
+        z = 8
+
+        # Unaligned values per dimension (all satisfy CK vector-load constraints).
+        # K: even but not multiple of tile (64). Same for all groups.
+        # M: not multiples of tile (256), varies per group.
+        # N: multiple of 16 but not multiple of tile (128).
+        unaligned_k = 2016
+        unaligned_m = [100, 300, 150, 200, 50, 350, 250, 180]
```
ipanfilo, line 3110: I think z should be derived as len of unaligned_m, or it should be asserted that they are equal

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16_nn.cpp
ipanfilo, line 1: But why separate translation units are needed for every ck_tile_grouped_gemm_fp16_dispatch_* and why the methods themselves are needed instead of directly calling ck_tile_grouped_gemm_fp16_dispatch_layout<>() ?

ipanfilo (changes requested):

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16_nn.cpp
matthiasdiener: These are split due to long compile times (~20min) when put into a single translation unit.

tests/pytorch/test_numerics.py
matthiasdiener: Done in https://github.com/ROCm/TransformerEngine/pull/574/commits/dff56356eed952ef8fcb5cc801ec1727ffcdeccc

tests/pytorch/test_numerics.py
matthiasdiener: Done in https://github.com/ROCm/TransformerEngine/pull/574/commits/dff56356eed952ef8fcb5cc801ec1727ffcdeccc

tests/pytorch/test_numerics.py
matthiasdiener: Thanks, removed the extra if in https://github.com/ROCm/TransformerEngine/pull/574/commits/dff56356eed952ef8fcb5cc801ec1727ffcdeccc.

ipanfilo (approved):

PR #576, opened 2026-05-06
CK Tile Group GEMM gfx1250

aris134:
# Description

Extend the present CK tile grouped GEMM (F16/F8) implementation for compatibility with gfx1250. Replaces 3rdparty/aiter with 3rdparty/rocm-libraries for the gfx1250 changes from CK.

Fixes #16490

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/common/CMakeLists.txt
+++ b/transformer_engine/common/CMakeLists.txt
@@ -354,7 +354,7 @@ set_property(
   PROPERTY
   COMPILE_OPTIONS "-g0;-dopt=on")
 else()
-  set(CK_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../../3rdparty/aiter/3rdparty/composable_kernel)
+  set(CK_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../../3rdparty/rocm_libraries/projects/composablekernel)
   target_include_directories(transformer_engine PRIVATE ${CK_ROOT}/include)
 endif() #USE_CUDA
 
--- a/transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_common.h
+++ b/transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_common.h
@@ -7,6 +7,7 @@
 #pragma once
 
 #include <hip/hip_runtime.h>
+#include "common/util/cuda_runtime.h"
 
 #include <array>
 #include <type_traits>
@@ -68,6 +69,28 @@ static inline const transformer_engine::SimpleTensor& data_view(const transforme
 
 static inline const transformer_engine::SimpleTensor& scale_inv_view(const transformer_engine::Tensor& t) {
   return t.scale_inv;
+}
+
+enum class GPUArch {
+  GFX942,
+  GFX950,
+  GFX1250,
+  UNKNOWN
+};
+
+static inline GPUArch detect_gpu_arch() {
+  int arch = cuda::sm_arch(0);
+
+  if (arch == 94) {
+    return GPUArch::GFX942;
+  }
+  if (arch == 95) {
+    return GPUArch::GFX950;
+  }
+  if (arch == 125 || arch == 1250) {
+    return GPUArch::GFX1250;
+  }
+  return GPUArch::UNKNOWN;
 }
 
 struct GroupedGemmRunContext {
--- a/transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_common.h
+++ b/transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_common.h
@@ -87,7 +87,7 @@ static inline GPUArch detect_gpu_arch() {
   if (arch == 95) {
     return GPUArch::GFX950;
   }
-  if (arch == 125 || arch == 1250) {
+  if (arch == 1250) {
     return GPUArch::GFX1250;
   }
   return GPUArch::UNKNOWN;
--- a/transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16.cpp
+++ b/transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16.cpp
@@ -233,28 +233,10 @@ class GroupedGemmRunner : public RunnerInterface {
   })
 
 template <GPUArch Arch>
-struct FP16TileCfg;
-
-template <>
-struct FP16TileCfg<GPUArch::GFX942> {
-  using type = TileCfg_256x256x64_MFMA;
-};
-
-template <>
-struct FP16TileCfg<GPUArch::GFX950> {
-  using type = TileCfg_256x256x64_MFMA;
-};
-
-template <>
-struct FP16TileCfg<GPUArch::GFX1250> {
-  using type = TileCfg_256x256x64_WMMA;
-};
-
-template <GPUArch Arch>
 bool ck_tile_grouped_gemm_fp16_dispatch_arch(DType a_dtype,
-                                        DType b_dtype,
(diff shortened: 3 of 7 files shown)


transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -1122,7 +1122,8 @@ void nvte_multi_tensor_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor
   if (!use_cutlass || num_gemms == 1) {
 #else
   // Currently only support cutlass group gemm on Hopper Arch
-  if (!(is_hopper && use_cutlass)) {
+  //if (!(is_hopper && use_cutlass)) {
+  if (!use_cutlass) {
```
ipanfilo, line 1126: It is CUDA path

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16.cpp
```diff
@@ -209,7 +232,26 @@ class GroupedGemmRunner : public RunnerInterface {
...
+struct FP16TileCfg<GPUArch::GFX950> {
+  using type = TileCfg_256x256x64_MFMA;
+};
+
+template <>
+struct FP16TileCfg<GPUArch::GFX1250> {
+  using type = TileCfg_256x256x64_WMMA;
+};
+
+template <GPUArch Arch>
```
ipanfilo, line 235: Why does it need template over reguler if-else or switch-case?

ipanfilo (changes requested):

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16.cpp
aris134: The template is needed because the arch selection affects CK kernel template instantiation, not just runtime control flow. GPUArch must be a compile-time value so if constexpr can prune unsupported tile/kernel combinations for a given architecture. In this case, it prevents the MFMA configs from being instantiated for gfx1250.

transformer_engine/common/gemm/cublaslt_gemm.cu
aris134: Reverted in 752e0d3

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_common.h
```diff
@@ -70,6 +71,28 @@ static inline const transformer_engine::SimpleTensor& scale_inv_view(const trans
...
+
+  if (arch == 94) {
+    return GPUArch::GFX942;
+  }
+  if (arch == 95) {
+    return GPUArch::GFX950;
+  }
+  if (arch == 125 || arch == 1250) {
+    return GPUArch::GFX1250;
+  }
```
wangye805: Why do we want to host two possible arch ids for gfx1250? Is it because in some docker image, it shows 125 but in other docker images it shows 1250?

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16.cpp
```diff
@@ -37,14 +37,37 @@ struct TileCfg_256x256x64 {
...
+  static constexpr ck_tile::index_t N_Warp = 2;
+  static constexpr ck_tile::index_t K_Warp = 1;
+
+  static constexpr ck_tile::index_t M_Warp_Tile = 16;
+  static constexpr ck_tile::index_t N_Warp_Tile = 16;
+  static constexpr ck_tile::index_t K_Warp_Tile = 32;
+
+  static constexpr bool kPadM = true;
+  static constexpr bool kPadN = true;
+  static constexpr bool kPadK = true;
```
wangye805, line 63: so the difference btw TileCfg_256x256x64_MFMA and TileCfg_256x256x64_WMMA is inside M, N, K warp tile and kPads?

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16.cpp
wangye805: I didn't compile it with gfx1250 arch only but I was still puzzled about this templated dispatch. In line 298, you still rely on runtime detect_gpu_arch() to branch to specific ck_tile_grouped_gemm_fp16_dispatch_arch<arch_id>'s. So I presume all three arches verions will still be instantiated? And I didn't see any compile time guarding?

transformer_engine/common/CMakeLists.txt
```diff
@@ -354,7 +354,7 @@ set_property(
   PROPERTY
   COMPILE_OPTIONS "-g0;-dopt=on")
 else()
-  set(CK_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../../3rdparty/aiter/3rdparty/composable_kernel)
+  set(CK_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../../3rdparty/rocm_libraries/projects/composablekernel)
```
wangye805, line 357: nit: Will the whole rocm_libraries too big? Do we have a way to have sparse check out for this ck subdir?

wangye805 (changes requested):

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_common.h
aris134: Kept it to 1250 in b55fe29

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16.cpp
aris134: The difference is not just the warp tile shape or kPads. MFMA and WMMA are different warp-level MMA instruction paths, so they lower through different warp dispatch/pipeline configurations with different tile and padding requirements.

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16.cpp
aris134: Good point. Before the latest change the runtime switch still referenced all `ck_tile_grouped_gemm_fp16_dispatch_arch<...>` specializations, so the compiler could still instantiate all arch variants. b55fe29 adds compile-time `#if defined(__gfxXXX__)` guards around each runtime dispatch case so unsupported arch paths are no longer instantiated.

transformer_engine/common/CMakeLists.txt
aris134: Good point. The full rocm_libraries checkout is fairly large (~8G locally), while projects/composablekernel alone is much smaller (~167M). Yeah, sparse checkout probably makes sense here, but I am wondering if it would be better handled in a separate PR.

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16.cpp
wangye805: Oh, I was comparing those two struct classes TileCfg_256x256x64_MFMA and TileCfg_256x256x64_WMMA. Inside those two defined structs, contents just differ by warp tile and kPads?

transformer_engine/common/CMakeLists.txt
wangye805: Yeah, we can do it in separate PR

wangye805 (approved):

PR #583, opened 2026-05-12
RMS Norm Optimization

aris134:
# Description

Fixes # (16527)

RMSNorm falls back to general kernel implementation on several DeepSeek and Qwen shapes, causing poor performance. These shapes have been registered with the tuned kernel cache, and a performance benchmark for RMSNorm has been added.

Additionally, a fallback warning is printed the first time at which a tuned config is not found for a requested kernel. For example:

```
in function getKernel: Falling back to general normalization kernel because no tuned kernel is available for this config. hidden_size=128, wtype=bf16, itype=bf16, otype=bf16, ctype=fp32
```

E2E TFLOPS/s/GPU for proxy models (Previous -> Current with RMSNorm tuning) :
 
Qwen:
bf16: 369.4 -> 374.7
fp8: 352.1 ->358.2
 
Deepseek:
bf16: 501.4 -> 529.4
fp8: 463.9 -> 511.4

Also added matching tuned configs for LayerNorm.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


benchmarks/cpp/normalization/bench_normalization.cpp
```diff
@@ -0,0 +1,299 @@
...
+    nvte_layernorm_fwd(input.data(), gamma.data(), beta.data(), epsilon,
+                       output.data(), mu.data(), rsigma.data(), workspace.data(),
+                       prop.multiProcessorCount, zero_centered_gamma, stream);
+  } else {
+    nvte_rmsnorm_fwd(input.data(), gamma.data(), epsilon,
+                     output.data(), rsigma.data(), workspace.data(),
+                     prop.multiProcessorCount, zero_centered_gamma, stream);
+  }
+
+  HIP_CHECK(hipStreamSynchronize(stream));
```
ipanfilo: Is synchronization needed before warmup?

transformer_engine/common/normalization/common.h
```diff
@@ -24,6 +24,7 @@
 #include <typeindex>
 #include <unordered_map>
 #include <vector>
+#include <unordered_set>
```
ipanfilo, line 26: nit: move it after unordered_map

transformer_engine/common/normalization/common.h
```diff
@@ -202,6 +203,58 @@ TupleKeyType get_key(NVTE_Norm_Backend NormBackend, NVTE_Norm_Type NormType,
                      bool is_tuned, NVTEScalingMode mode = NVTE_DELAYED_TENSOR_SCALING,
                      bool training = true, bool gamma_in_weight_dtype = false);
 
+inline DType decode_itype(uint64_t general_key) {
```
ipanfilo, line 208: This code is fragile because encoding could change. At least put comments here and at encoding block that they should match

benchmarks/cpp/normalization/bench_normalization.cpp
aris134: Good point. These are in fact redundant since the warmup already calls a device-wide sync anyway. Removed in 4256e3c

transformer_engine/common/normalization/common.h
aris134: Done in 2f9ff47

transformer_engine/common/normalization/common.h
aris134: Good point. I updated this in d548d54 to make the coupling between encoding/decoding explicit by introducing shared norm_key bit-layout constants and using them in both get_key() and the decode helpers. I also added comments documenting that the layouts must remain in sync, so future changes to the packed key format are less likely to silently diverge.

benchmarks/cpp/normalization/bench_normalization.cpp
```diff
@@ -0,0 +1,297 @@
...
+    float ms = 0.0f;
+    HIP_CHECK(hipEventElapsedTime(&ms, start, stop));
+    state.SetIterationTime(ms / 1000.0);
+  }
+
+  HIP_CHECK(hipEventDestroy(start));
+  HIP_CHECK(hipEventDestroy(stop));
+  HIP_CHECK(hipStreamDestroy(stream));
+
+  size_t bytes_read =
```
alextmagro: nit: line splits aren't needed here

benchmarks/cpp/normalization/bench_normalization.cpp
```diff
@@ -0,0 +1,297 @@
...
+  }
+
+  set_bytes_processed(state, bytes_read + bytes_write);
+}
+
+template <BenchNormType Norm, typename WType, typename IType, typename OType, typename CType>
+static void BM_NormBackward(benchmark::State& state) {
+  const size_t N = state.range(0);
+  const size_t H = state.range(1);
+  const float epsilon = 1e-5f;
```
alextmagro: epsilon is the same between forward and backward, can probably make a const global

transformer_engine/common/normalization/layernorm/ln_bwd_semi_cuda_kernel.cu
```diff
@@ -214,6 +220,12 @@ REGISTER_NORM_LAUNCHER(LayerNorm, Backward, tuned, 6144, fp16, fp32, fp16, fp32,
 REGISTER_NORM_LAUNCHER(LayerNorm, Backward, tuned, 6144, bf16, bf16, bf16, fp32, 1, 1, 8, 16, 4);
 REGISTER_NORM_LAUNCHER(LayerNorm, Backward, tuned, 6144, bf16, fp32, bf16, fp32, 1, 1, 8, 16, 4);
 
+REGISTER_NORM_LAUNCHER(LayerNorm, Backward, tuned, 7168, fp32, fp32, fp32, fp32, 1, 1, 7, 8, 4);
```
alextmagro: Does BYTES_PER_LDG=8 outperform 16 for this config? If so, I wonder if the configs around it would perform better that way too.

transformer_engine/common/normalization/layernorm/ln_fwd_cuda_kernel.cu
```diff
@@ -280,6 +292,12 @@ REGISTER_NORM_LAUNCHER(LayerNorm, Forward, tuned, 6144, fp32, fp32, fp16, fp32,
 REGISTER_NORM_LAUNCHER(LayerNorm, Forward, tuned, 6144, bf16, bf16, bf16, fp32, 1, 1, 4, 16);
 REGISTER_NORM_LAUNCHER(LayerNorm, Forward, tuned, 6144, fp32, fp32, bf16, fp32, 1, 1, 4, 16);
 
+REGISTER_NORM_LAUNCHER(LayerNorm, Forward, tuned, 7168, fp32, fp32, fp32, fp32, 1, 1, 4, 16);
```
alextmagro: BWD you have 7 warps set, but here you have 4. Is this optimal?

transformer_engine/common/normalization/rmsnorm/rmsnorm_bwd_semi_cuda_kernel.cu
```diff
@@ -175,6 +183,10 @@ REGISTER_NORM_LAUNCHER(RMSNorm, Backward, tuned, 4096, fp32, fp32, fp32, fp32, 1
 REGISTER_NORM_LAUNCHER(RMSNorm, Backward, tuned, 4096, fp16, fp16, fp16, fp32, 1, 1, 4, 16, 4);
 REGISTER_NORM_LAUNCHER(RMSNorm, Backward, tuned, 4096, bf16, bf16, bf16, fp32, 1, 1, 4, 16, 4);
 
+REGISTER_NORM_LAUNCHER(RMSNorm, Backward, tuned, 7168, fp32, fp32, fp32, fp32, 1, 1, 4, 16, 4);
```
alextmagro: Same here, is 4 better than 7?

transformer_engine/common/normalization/common.cpp
```diff
@@ -58,12 +58,18 @@ TupleKeyType get_key(NVTE_Norm_Backend NormBackend, NVTE_Norm_Type NormType,
   static_assert(NVTE_INVALID_SCALING < 1024,
                 "This function assumes at most 10 bits used in the scaling mode.");
   static_assert(kNVTENumTypes < 32, "This function assumes at most 5 bits used in the NVTEDType");
-  uint64_t general_key = static_cast<uint64_t>(itype) | (static_cast<uint64_t>(otype) << 5) |
-                         (static_cast<uint64_t>(ctype) << 10) |
-                         (static_cast<uint64_t>(wtype) << 15) | (uint64_t(NormType) << 20) |
-                         (uint64_t(NormStage)) << 22 | (uint64_t(NormBackend) << 24) |
-                         (uint64_t(zero_centered_gamma) << 26) | (uint64_t(mode) << 27) |
-                         (uint64_t(training) << 37) | (uint64_t(gamma_in_weight_dtype) << 38);
+  uint64_t general_key =
```
alextmagro: I get the motivation behind this change, but this affects upstream code. I feel like we're more likely to miss a key change from upstream if we have diverged here.

alextmagro (changes requested):

benchmarks/cpp/normalization/bench_normalization.cpp
aris134: Removed in b73062c

benchmarks/cpp/normalization/bench_normalization.cpp
aris134: Good point. Re-factored it into a global constant in b73062c

transformer_engine/common/normalization/common.cpp
aris134: Good point — I think I overcorrected this by refactoring the key layout into shared named constants, which does increase divergence from upstream and could make future key-layout changes easier to miss during syncs.

I've reverted the get_key() refactor back to the original upstream-style encoding layout and instead added explicit comments at both the encoding and decode sites documenting that the bit layouts must stay in sync. See 0949b9a

alextmagro: Why not add a static assert at the decode so the code doesn't compile for us if upstream changes the encoding?

transformer_engine/common/normalization/layernorm/ln_bwd_semi_cuda_kernel.cu
aris134: Actually, BYTES_PER_LDG=16 is definitely the better configuration here. 

## LayerNorm Backward H=7168: BYTES_PER_LDG 8 → 16

| DType | Old BW (LDG=8) | New BW (LDG=16) | BW Delta |
|---|---:|---:|---:|
| BF16 | 2.61985 TiB/s | 2.89315 TiB/s | +10.43% |
| FP16 | 3.85141 TiB/s | 3.86147 TiB/s | +0.26% |
| FP32 | 4.32618 TiB/s | 4.29753 TiB/s | -0.66% |

| DType | Old Time | New Time | Time Delta |
|---|---:|---:|---:|
| BF16 | 122 us | 111 us | -9.0% |
| FP16 | 83.2 us | 83.0 us | -0.24% |
| FP32 | 148 us | 149 us | +0.68% |

I've changed it in e8afcc8. Note this config is now almost identical to the H=6144 config except with WARPS_N=7 instead of WARPS_N=8. It turns out setting WARPS_N=8 for H=7168 results in an illegal configuration, with the same issue described here: https://github.com/ROCm/frameworks-internal/issues/16527#issuecomment-4435402706

transformer_engine/common/normalization/layernorm/ln_fwd_cuda_kernel.cu
aris134: For BWD, 7 warps is indeed more performant than 4 warps across all DTypes tested. For FWD, 7 warps gives a performance boost to fp16/fp32, but regresses on bf16, so I kept 4 warps for bf16 for the h=7168 config. See 3d22d82

transformer_engine/common/normalization/common.cpp
aris134: I'm not sure how to accomplish this without modifying the get_key definition itself (e.g. introducing a shared layout definition/helper used by both get_key and the decode helpers). Otherwise the decode functions only have their own hardcoded assumptions about the bit layout and cannot independently detect drift in get_key.

transformer_engine/common/normalization/rmsnorm/rmsnorm_bwd_semi_cuda_kernel.cu
aris134: Right, 7 is marginally better here, so I updated that in 63d0e26

transformer_engine/common/normalization/layernorm/ln_fwd_cuda_kernel.cu
aris134: Did not see the same performance boost for RMSNorm forward however, so I left warps=4 there

transformer_engine/common/normalization/common.cpp
alextmagro: Right... We'd have to make encoder a constexpr and move to header. That would be ideal, but ends up with more upstream changes anyway.

alextmagro (approved):
LGTM! After merge, please work with Sudharshan to run the E2E configs again and get updated performance numbers

transformer_engine/common/normalization/common.cpp
alextmagro: Actually, could we add a runtime check instead of a static assert that verifies the round trip is valid? Maybe something like this:

```C
  namespace {                                                                                                                                                                                                               
  [[maybe_unused]] const bool kNormKeyLayoutCheck = [] {
      auto [key, b, h, t] = get_key(                                                                                                                                                                                        
          NVTE_Norm_Backend::Te, NVTE_Norm_Type::RMSNorm, NVTE_Norm_Stage::Forward,                                                                                                                                         
          DType::kFloat16,DType::kBFloat16, DType::kFloat8E4M3, DType::kFloat32, 1, 1, false, false);                                                                                                                                                                                              
      NVTE_CHECK(decode_itype(key)     == DType::kBFloat16);                                                                                                                                                                
      NVTE_CHECK(decode_otype(key)     == DType::kFloat8E4M3);                                                                                                                                                              
      NVTE_CHECK(decode_ctype(key)     == DType::kFloat32);                                                                                                                                                                 
      NVTE_CHECK(decode_wtype(key)     == DType::kFloat16);                                                                                                                                                                 
      NVTE_CHECK(decode_norm_type(key) == NVTE_Norm_Type::RMSNorm);
      return true;                                                                                                                                                                                                          
  }();                                                                                                                                                                                                                      
  }  
```

aris134: This is helpful, thanks for the suggestion. I've gone ahead and implemented this in db7b017

PR #584, opened 2026-05-12
Harden claude-pr-action.yml

Micky774:
# Description

Triaged from a third-party security review of the workflow. This commit addresses the must-fix issues presented in the review.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Block fork PRs from non-collaborators before any PR code is staged. pull_request_target gives this job write-scoped GITHUB_TOKEN and the CLAUDE_CODE_OAUTH_TOKEN secret. Without this gate, a fork PR can hijack both via e.g. a .gitattributes textconv driver invoked by Claude's `git diff` calls. The new step runs first and exits 1 unless the head repo is the same as the base repo or the PR author has write/maintain/admin permission on the base.

- Run the Claude warmup in mktemp -d instead of $GITHUB_WORKSPACE. Claude Code auto-discovers CLAUDE.md and .claude/ from CWD; with the PR tree checked out and CLAUDE_CODE_OAUTH_TOKEN in env, an attacker workspace CLAUDE.md could prompt-inject the warmup into echoing the token into workflow logs.

- Pin third-party actions to commit SHAs:
  * actions/checkout@v4              -> @34e114876b0b... (v4.3.1)
  * actions/upload-artifact@v4       -> @ea165f8d65b6... (v4.6.2)
  * anthropics/claude-code-action@v1 -> @dde2242db6af... (v1.0.120)

- Replace blocklist allowedTools with a strict allowlist for both the review and summary steps. The previous `Bash(gh *)` + denylist allowed e.g. `gh api --method DELETE repos/...` because the denylist only covered specific subcommands. New allowlist only matches read-shaped invocations (`gh api repos/*`, `git diff *`, etc.); summary additionally permits `gh api --method PATCH repos/*` so it can update prior walkthrough comments.

# Checklist:

### DIFF
--- a/.github/workflows/claude-pr-action.yml
+++ b/.github/workflows/claude-pr-action.yml
@@ -332,14 +332,18 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           allowed_bots: "github-actions[bot]"
           show_full_output: true
-          # Strict allowlist (no disallowedTools). Patterns match against the
-          # literal command string, so `Bash(gh api repos/*)` permits read
-          # paths like `gh api repos/foo/bar/pulls/1/comments` but rejects
-          # `gh api --method DELETE repos/...` because the next token after
-          # `gh api` isn't `repos/*`. Add new entries here when a skill needs
-          # a new subcommand — never widen to `Bash(gh *)` or `Bash(git *)`.
+          # Strict allowlist for Bash(...) and MCP tools (no disallowedTools).
+          # The matcher splits chained commands on `;`/`&&`/`|` and checks
+          # each segment, so `Bash(git diff *)` covers `BASE=…; git diff …`
+          # but `Bash(echo *)` is still needed for the `echo "---"` segment.
+          # Built-in tools (Read, Grep, Glob, Skill, TodoWrite, ToolSearch,
+          # Write, Edit) bypass --allowedTools enforcement, but Read/Grep/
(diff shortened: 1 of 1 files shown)


.github/workflows/claude-pr-action.yml
```diff
@@ -200,13 +200,45 @@ jobs:
...
+          esac
+
       # refs/pull/<n>/merge is GitHub's synthetic merge commit (base tip
       # merged with PR head). Checking it out gives us both parents in one
       # shot: HEAD^1 = base tip, HEAD^2 = PR head — no separate base fetch
       # needed. Caveat: the merge ref only exists when the PR is mergeable;
       # if it's missing/stale, the workflow will fail fast on checkout, which
       # is the right behavior for an unreviewable PR.
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4.3.1
```
matthiasdiener: Perhaps mention here why the pinning is done.

Micky774: Pinning is pretty standard practice for CI security just to avoid accidentally using a new and compromised version -- this isn't pinned to that _exact_ version for a particular version, other than it is "known to work".

I think we can probably avoid an extra comment on it, but I'm open to it if you still think it would be useful to folks. What do you think?

wenchenvincent (approved):

PR #592, opened 2026-05-20
speed up nvte_multi_padding / nvte_multi_unpadding

matthiasdiener:
# Description

Fixes https://github.com/ROCm/frameworks-internal/issues/16530

See https://github.com/ROCm/frameworks-internal/issues/16530#issuecomment-4502138388 for performance.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/common/util/padding.cu
```diff
@@ -83,6 +95,35 @@ __global__ void __launch_bounds__(threads_per_block) multi_padding_kernel(MultiP
...
+  // Process subtiles with vectorized loads/stores
+#pragma unroll
+  for (int iter = 0; iter < n_iterations; ++iter) {
+    const int i1 = tidy + iter * bdimy;
+    const int j1 = tidx;
+#pragma unroll
+    for (int i2 = 0; i2 < nvec; ++i2) {
+      const int row = tile_row + i1 * nvec + i2;
+      const int col = tile_col + j1 * nvec;
+      const int remaining = row_length - col;
```
aris134: Can we factor out these lines since they don't depend on `i2`? This also lets us compute `valid_cols` once per `iter` instead of repeating the ternary in both load/store calls:

```
const int col = tile_col + j1 * nvec;
const int remaining = row_length - col;
const int valid_cols = remaining > 0 ? min(remaining, nvec) : 0;
#pragma unroll
for (int i2 = 0; i2 < nvec; ++i2)
```

transformer_engine/common/util/padding.cu
aris134, line 65: Did you try tuning any of these parameters by chance? The current block size seems small. I am also wondering if you can squeeze out any more performance by tuning the load/store size

transformer_engine/common/util/padding.cu
```diff
@@ -167,6 +216,26 @@ __global__ void __launch_bounds__(threads_per_block) multi_unpadding_kernel(Mult
...
+#ifdef __HIP_PLATFORM_AMD__
+  // Process subtiles with vectorized loads/stores
+#pragma unroll
+  for (int iter = 0; iter < n_iterations; ++iter) {
+    const int i1 = tidy + iter * bdimy;
+    const int j1 = tidx;
+#pragma unroll
+    for (int i2 = 0; i2 < nvec; ++i2) {
+      const int row = tile_row + i1 * nvec + i2;
+      const int col = tile_col + j1 * nvec;
```
aris134: nit: similar to above for the multi_padding_kernel, I think `col` and the ternary op can be hoisted up

aris134 (changes requested):

transformer_engine/common/util/padding.cu
matthiasdiener: Good point, thanks. Bumping n_warps_per_tile (in https://github.com/ROCm/TransformerEngine/pull/592/commits/dc708c6b0ea990ffff4de2939483822a62b7e724) does lead to a significant performance increase. Increasing desired_load_store_size reduced performance.

transformer_engine/common/util/padding.cu
matthiasdiener: Done in https://github.com/ROCm/TransformerEngine/pull/592/commits/cb9221d44f36f8254292a9a8cf85bb7c5c45e12d. Does not really affect performance; I suspect the compiler is able to factor these out by itself.

transformer_engine/common/util/padding.cu
matthiasdiener: Done in https://github.com/ROCm/TransformerEngine/commit/cb9221d44f36f8254292a9a8cf85bb7c5c45e12d.

transformer_engine/common/util/padding.cu
alextmagro: I wonder if templating out load and store separately would help -- previously I found that 16 outperformed for some kernels for loads, but store was well optimized at 8. This was for FP8 though so may be different here. Also non-temporal stores may be a potential improvement, check out rocm device utils for reference

matthiasdiener: > I wonder if templating out load and store separately would help -- previously I found that 16 outperformed for some kernels for loads, but store was well optimized at 8. This was for FP8 though so may be different here. 

No, that also seems to reduce performance in this case.

> Also non-temporal stores may be a potential improvement, check out rocm device utils for reference

This did help with performance, done in https://github.com/ROCm/TransformerEngine/pull/592/commits/84b7d09fd8a3ed1f37eee1061ddc1e118f42e0ed

aris134 (approved):
LGTM!

transformer_engine/common/util/padding.cu
```diff
@@ -13,13 +15,42 @@
...
 namespace transformer_engine {
 
 namespace {
 
+#ifdef __HIP_PLATFORM_AMD__
+// Non-temporal store helper: uses NT store for full aligned vectors,
+// falls back to element-wise for partial/unaligned cases.
+// Note: NT loads were also benchmarked but hurt performance.
+template <uint32_t nvec, typename Type>
+__device__ __forceinline__ void nt_store_to_elts(const Vec<Type, nvec>& v,
```
alextmagro: What is the case where we hit non-aligned vectors? Isn't FP8/MXFP8 always padded to a multiple of 16 by default? Ideally we would template out the NT vs elementwise stores

transformer_engine/common/util/padding.cu
```diff
@@ -83,6 +121,36 @@ __global__ void __launch_bounds__(threads_per_block) multi_padding_kernel(MultiP
   const int tile_row = tile_id_m * tile_dim_m;
   const int tile_col = tile_id_n * tile_dim_n;
 
+#ifdef __HIP_PLATFORM_AMD__
+  // Process subtiles with vectorized loads/stores
+#pragma unroll
```
alextmagro, line 135: Did you try #pragma unroll 2 here? If we have the registers available that might help performance.

transformer_engine/common/util/padding.cu
matthiasdiener: Templated out in https://github.com/ROCm/TransformerEngine/pull/592/commits/c98de38d2a393e2ec3e984bf81ecfc37e8414a1d. It does improve performance by ~1%.

transformer_engine/common/util/padding.cu
matthiasdiener: Yes, but no performance improvement was observed.

alextmagro (approved):

PR #593, opened 2026-05-20
Triton RMSNorm Optimizations

Micky774:
# Description

Optimizes the Triton RMSNorm forward and backward kernels and adds an LDS-tiled FP8 transpose path. Measured 10%-50% improvements across a representative shape sweep for bf16 w/ no quantization or FP8 quant, and improvements of 3x-8x on FP8 Transpose outputs.

[Benchmarks](https://amd.atlassian.net/wiki/x/jSVvZQ) generated by [this script](https://gist.github.com/Micky774/70367aa54c88ae39b77200e8a0284fc8).

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- **Loop-invariant hoisting.**
  - Fwd non-blocked path: `gamma` load + `ZERO_CENTERED_GAMMA` adjustment + `1/n_cols` hoisted outside the persistent row loop.
  - Bwd non-blocked path: same `gamma` hoist; `inv_n_cols` hoisted.
  - Bwd both paths: per-row `c_scalar = nf*nf*grad_sum*inv_n_cols` computed once before the dx/dg loop; `dx` expression refactored to `nf * (dz*g - c*x)` (saves one multiply per element).
- **Autotune wiring for bwd kernels.** `_rmsnorm_bwd_triton` and `_rmsnorm_bwd_dg_reduce_triton` now follow the impl + autotune-wrapper dispatch pattern already used by the fwd kernel. `te_rmsnorm_bwd_triton` takes an `autotune: bool = True` kwarg; when off it uses the previously-hardcoded `num_warps=8` + fixed `BLOCK_SIZE_M/N=128/64` reduce tile.
- **External LDS-tiled FP8 transpose kernel.** New `_fp8_transpose_2d_impl` (+ autotune wrapper) replaces the in-kernel `out_transpose_ptr + cols * stride + row_idx` strided byte stores that were uncoalesced (one byte per thread to a different cache line). The new kernel does a coalesced `(BLOCK_M, BLOCK_N)` read, `tl.trans()` for LDS-staged transpose, then coalesced strided write. 

# Checklist:


transformer_engine/pytorch/triton_kernels/norms_common.py
```diff
@@ -216,6 +250,29 @@ def _te_norm_fwd_triton(
...
+                triton.cdiv(N, meta['BLOCK_M']),
+                triton.cdiv(H, meta['BLOCK_N']),
+            )
+            transpose_kernel[grid_t](
+                out._data, out._transpose,
+                N, H,
+                out._data.stride(0), out._transpose.stride(0),
+            )
+        else:
+            BLOCK_M, BLOCK_N = 64, 64
```
aris134, line 321: Is there a reason we default to BLOCK_M=64, BLOCK_N=64 here? Just wondering whether this was benchmarked for the expected N/H ranges or chosen as a conservative default.

transformer_engine/pytorch/triton_kernels/rmsnorm.py
```diff
@@ -349,3 +395,87 @@ def _rmsnorm_bwd_dg_reduce_triton(dg_in_ptr, dg_out_ptr, dg_in_stride, n_rows, n
...
+#
+# This kernel reads a (BLOCK_M, BLOCK_N) tile coalesced from the row-major
+# fp8 output, transposes it through LDS via `tl.trans`, and writes the
+# (BLOCK_N, BLOCK_M) tile coalesced to the column-major transpose buffer.
+#
+# Operates on the raw uint8 storage so the fp8 dtype is irrelevant to
+# correctness.
+# --------------------------------------------------------------------------- #
+@triton.jit
+def _fp8_transpose_2d_impl(
```
aris134: Since this doesn't seem specific to RMSNorm, would it be better to put this in a shared Triton utility?

README.rst
```diff
@@ -281,6 +281,8 @@ At runtime, you can enable specific triton kernels using the specific environmen
 * NVTE_USE_CAST_TRANSPOSE_TRITON=1 can be used to enable cast transpose (bgrad) triton kernels;
 * NVTE_USE_LAYERNORM_TRITON=1 can be used to enable layernorm triton kernels.
 * NVTE_USE_RMSNORM_TRITON=1 can be used to enable rmsnorm triton kernels.
+* NVTE_RMS_EXTERNAL_TRANSPOSE=0 can be used to enable external transpose in rmsnorm triton kernels,
```
aris134: The README wording reads backward to me. In norms_common.py, the code defaults NVTE_RMS_EXTERNAL_TRANSPOSE to "1", and the comment says "Set NVTE_RMS_EXTERNAL_TRANSPOSE=0 to fall back."

So I think the README should say that NVTE_RMS_EXTERNAL_TRANSPOSE=0 disables external transpose in RMSNorm Triton kernels and recovers the legacy in-kernel transpose behavior.

aris134 (changes requested):

tests/pytorch/triton_kernels/test_norms.py
```diff
@@ -298,7 +298,13 @@ def test_norm_triton(
             zero_centered_gamma=zero_centered_gamma,
 
         )
-        triton_bwd_outs = triton_bwd_func(*args["triton"])
+        # te_rmsnorm_bwd_triton accepts an `autotune` kwarg; te_layernorm_bwd_triton does not.
+        # Honor the same NVTE_TEST_TRITON_AUTOTUNE env toggle as the fwd path so
+        # default test runs avoid the autotune compile/sweep cost.
+        if norm == "rms":
```
alextmagro: This logic could be combined with the following if/elses

transformer_engine/pytorch/triton_kernels/norms_common.py
```diff
@@ -20,8 +21,22 @@
     _rmsnorm_fwd_triton,
     _rmsnorm_fwd_triton_impl,
     _rmsnorm_bwd_triton,
+    _rmsnorm_bwd_triton_impl,
     _rmsnorm_bwd_dg_reduce_triton,
+    _rmsnorm_bwd_dg_reduce_triton_impl,
+    _fp8_transpose_2d_triton,
```
alextmagro: nit: _fp8 should come before _rmsnorm

transformer_engine/pytorch/triton_kernels/norms_common.py
```diff
@@ -20,8 +21,22 @@
...
+    _fp8_transpose_2d_impl,
 )
+
+# Use the external LDS-tiled byte transpose instead of the in-kernel strided
+# stores. Default on -- the in-kernel path is uncoalesced and bottlenecks
+# every fp8_t shape. Set NVTE_RMS_EXTERNAL_TRANSPOSE=0 to fall back.
+_USE_EXTERNAL_TRANSPOSE = os.environ.get("NVTE_RMS_EXTERNAL_TRANSPOSE", "1") == "1"
```
alextmagro: Have you benchmarked shapes to see if there are any that outperform with the old logic?

transformer_engine/pytorch/triton_kernels/norms_common.py
```diff
@@ -152,6 +177,10 @@ def _te_norm_fwd_triton(
     out_transpose_ptr = None
     out_transpose_stride = None
     FP8_MAX = None
+    # When True, skip in-kernel strided transpose stores and dispatch a
```
alextmagro: This comment more or less a repeat of above so can be removed I think

transformer_engine/pytorch/triton_kernels/norms_common.py
```diff
@@ -248,25 +305,72 @@ def te_rmsnorm_bwd_triton(dz, x, rsigma, gamma, sm_margin, zero_centered_gamma):
     blk_size = block_size(x_)
     USE_BLOCKED = use_blocked(x_)
     NUM_PRGMS = num_programs(x_, sm_margin)
-    need_reduction = N > 1
-    dg_tmp_rows =  x_.shape[0] if use_blocked(x_) else num_programs(x_, sm_margin)
-    dg_tmp = torch.empty(dg_tmp_rows, N, device=x.device, dtype=torch.float32, requires_grad=False) if need_reduction else None
+    # dg accumulation strategy:
+    #   * Large M (rows_per_program > 1): per-program partial buffer of shape
```
alextmagro: How do we know these claims are true in triton? Since we don't know what the compiler is doing under the hood, it seems like these are assumptions or behavior we want rather than confirmed

transformer_engine/pytorch/triton_kernels/norms_common.py
```diff
@@ -248,25 +305,72 @@ def te_rmsnorm_bwd_triton(dz, x, rsigma, gamma, sm_margin, zero_centered_gamma):
...
     dg_target = dg_tmp if need_reduction else dgamma
     dg_aligned_16 = (dg_target.data_ptr() % 16 == 0) and (dg_target.stride(0) * dg_target.dtype.itemsize % 16 == 0)
+
     grid_bwd = lambda meta: (NUM_PRGMS, )
-    _rmsnorm_bwd_triton[grid_bwd](dz_, x_, gamma_, rsigma_, dx, dg_tmp if need_reduction else dgamma,
-                                  x_.stride(0), dz_.stride(0), M, N, zero_centered_gamma, blk_size,
-                                  USE_BLOCKED, NUM_PRGMS, input_aligned_16, grad_output_aligned_16,
-                                  dx_aligned_16, dg_aligned_16, num_warps=8)
+    bwd_kernel = _rmsnorm_bwd_kernels[autotune]
+    bwd_kwargs = dict(
```
alextmagro, line 374: nit: I'm not sure this change is needed, and makes things a bit harder to read. We should probably just define a var for num_warps based on autotune and use the old code path for clarity.

transformer_engine/pytorch/triton_kernels/norms_common.py
```diff
@@ -248,25 +305,72 @@ def te_rmsnorm_bwd_triton(dz, x, rsigma, gamma, sm_margin, zero_centered_gamma):
...
-        _rmsnorm_bwd_dg_reduce_triton[grid_reduce](dg_tmp, dgamma, dg_tmp.stride(0), dg_tmp.shape[0], dg_tmp.shape[1],
-                                                   BLOCK_SIZE_M=128, BLOCK_SIZE_N=64)
+        reduce_kernel = _rmsnorm_bwd_dg_reduce_kernels[autotune]
+        if autotune:
+            grid_reduce = lambda meta: [triton.cdiv(N, meta['BLOCK_SIZE_N'])]
+            reduce_kernel[grid_reduce](
+                dg_tmp, dgamma, dg_tmp.stride(0), dg_tmp.shape[0], dg_tmp.shape[1],
+            )
+        else:
+            # Match the previously-hardcoded tile when autotune is disabled.
```
alextmagro: nit: Comment can be removed, since we know where the values came from

transformer_engine/pytorch/triton_kernels/rmsnorm.py
```diff
@@ -140,22 +140,24 @@ def _rmsnorm_fwd_triton_impl(
 
     else:
         mask = col_offsets < n_cols
+        # gamma is invariant across rows -- load + ZERO_CENTERED adjustment once per program.
+        g = tl.load(g_ptr + col_offsets, mask=mask, other=0.0).to(tl.float32)
```
alextmagro, line 132: We still do one load per row, right? Might be better to take this one step further and load all values for all rows that the block handles outside the loop. That might give the compiler some more hints on where it can overlap loads

transformer_engine/pytorch/triton_kernels/rmsnorm.py
```diff
@@ -349,3 +395,87 @@ def _rmsnorm_bwd_dg_reduce_triton(dg_in_ptr, dg_out_ptr, dg_in_stride, n_rows, n
...
+    tile_t = tl.trans(tile)
+
+    # Coalesced write of (BLOCK_N, BLOCK_M) tile (innermost dim = rows).
+    dst_offs = cn[:, None] * dst_stride + rm[None, :]
+    dst_mask = (cn[:, None] < n_cols) & (rm[None, :] < n_rows)
+    tl.store(dst_ptr + dst_offs, tile_t, mask=dst_mask)
+
+
+def _get_fp8_transpose_configs():
+    # 1 B/elem on AMD CDNA3. Keep tile <= ~16 KB so the LDS staging buffer
```
alextmagro: These numbers don't seem to be right. MI300s have 64 KiB, MI350 has 160 KiB. We can probably get away with a higher value. Also, double buffering probably isn't a benefit for this kernel, so we might want to consider using the full LDS for the transpose

alextmagro (changes requested):

transformer_engine/pytorch/triton_kernels/norms_common.py
Micky774: Yes I had run a sweep across various shapes we test w/ and generally 64x64 was an optimal/close-to-optimal choice. Still arbitrary, but at least an _informed_ heuristic.

transformer_engine/pytorch/triton_kernels/rmsnorm.py
Micky774: Updated!

README.rst
Micky774: Updated!

transformer_engine/pytorch/triton_kernels/norms_common.py
Micky774: Yes, this was uniformly better. You can check out the benchmark output linked in the body of the PR.

alextmagro: I don't seem to have access, I requested it. If we are uniformly better for MI300 and MI350, I think we can just remove the old code as this is just a solid optimization then.

tests/pytorch/triton_kernels/test_norms.py
Micky774: Updated

transformer_engine/pytorch/triton_kernels/norms_common.py
Micky774: Sure, updated.

transformer_engine/pytorch/triton_kernels/norms_common.py
Micky774: Good point -- profiling w/ hw counters seems to falsify this claim as is anyways. Although in practice it helped for some shapes, more exhaustive testing shows that the heuristic boundary is complex, so I'm removing this from this PR for now.

transformer_engine/pytorch/triton_kernels/norms_common.py
Micky774: I think moving to keyword args is helpful here for clarity, and this mechanism follows the same pattern as the existing fwd dispatch function. I'd prefer to keep it as-is

transformer_engine/pytorch/triton_kernels/norms_common.py
Micky774: Updated

transformer_engine/pytorch/triton_kernels/rmsnorm.py
Micky774: Not sure what you mean, this is already a single load outside of any loop. For the blocked paths, that would essentially require loading all of g anyways, which is infeasible unless we restructure the loops and change the algorithm itself a bit.

transformer_engine/pytorch/triton_kernels/rmsnorm.py
Micky774: I've removed the comment. I'm leaving the autotuning configs as-is only because I didn't see significant changes on other sizes.

transformer_engine/pytorch/triton_kernels/norms_common.py
alextmagro: Gotcha. If it helps on average I think it is ok to keep or add in a followup PR, just wanted to check if we had evidence or if these were empirical improvements.

transformer_engine/pytorch/triton_kernels/rmsnorm.py
alextmagro: Sorry -- python indentation confused me. Thought we were inside the for loop on L51.

alextmagro (approved):
LGTM!

transformer_engine/pytorch/triton_kernels/rmsnorm.py
```diff
@@ -238,6 +223,9 @@ def _rmsnorm_bwd_triton(grad_output_ptr, input_ptr, g_ptr, rsigma_ptr, dx_ptr, d
 
             # Load r_sigma
             norm_factor = tl.load(rsigma_ptr + row_idx).to(tl.float32)
+            # Precomputed per-row invariant: dx = nf * (dz*g - c*x) where
```
aris134: nit: This comment makes it sound like dx itself is the row-invariant value, but only c is row-invariant. Maybe clarify that c is the precomputed per-row invariant used in the dx expression.

aris134 (approved):
LGTM!

README.rst
```diff
@@ -281,6 +281,8 @@ At runtime, you can enable specific triton kernels using the specific environmen
 * NVTE_USE_CAST_TRANSPOSE_TRITON=1 can be used to enable cast transpose (bgrad) triton kernels;
 * NVTE_USE_LAYERNORM_TRITON=1 can be used to enable layernorm triton kernels.
 * NVTE_USE_RMSNORM_TRITON=1 can be used to enable rmsnorm triton kernels.
+* NVTE_RMS_EXTERNAL_TRANSPOSE=0 disables external transpose in RMSNorm Triton kernels and
```
ipanfilo: It is not used in code

ipanfilo (changes requested):

transformer_engine/pytorch/triton_kernels/rmsnorm.py
Micky774: Updated!

README.rst
wenchenvincent: @ipanfilo Could you check if the comments has been addressed?

wenchenvincent:
@alextmagro @aris134 I saw you had approved the PR. For the inline comments, let's also resolve conversation if the comments has been addressed.

README.rst
Micky774: Updated!

ipanfilo (approved):

PR #594, opened 2026-05-21
Fix CK FP8 grouped GEMM dtype gating for columnwise operands

aris134:
# Description

During an E2E Qwen Megatron run, it was observed that certain FP8 grouped GEMM WGRAD paths were unexpectedly falling back from CK grouped GEMM to multi-stream hipBLASLt grouped GEMM. Further investigation showed that the CK backend was incorrectly rejecting certain FP8 WGRAD GEMMs due to dtype gating assuming row-wise tensor storage rather than the effective canonicalized operand.

This commit fixes the FP8 grouped GEMM dtype gating logic to use the effective operand dtype after canonicalization, eliminating erroneous fallback for columnwise FP8 operands. Additionally, this commit introduces support for FP8 padding when required to satisfy kernel requirements. Finally, this commit adds shape specialization for fully aligned operands.

Observed fallback warnings:

Number of fallback warnings originally: 7680
Number of fallback warnings after applying the fix from this PR: 960

Remaining fallbacks are due to lack of fp8 accumulate support.

These changes substantially reduce unintended fallback behavior in FP8 grouped GEMM execution paths while improving handling of padded FP8 grouped GEMM configurations.

Fixes # (16708)

## Type of change

- [] New feature (non-breaking change which adds functionality)
- [ ] Infra/Build change
- [ ] Code refactoring

Optimizations attempted but excluded:
CK Pipeline changes (i.e., compute V3 --> V4 with double-buffered ping-pong). This change consistently resulted in considerable performance degradation.
Tried a larger tile (256x256x128) with k-padding, which also reduced performance. 

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp8.cpp
```diff
@@ -261,7 +289,9 @@ class QuantGroupedGemmRunner : public RunnerInterface {
     if (descs.empty()) {
       return false;
     }
-    return launch_grouped_gemm_kernel<Kernel>(descs, ctx, stream_cfg);
+
+    const bool launched = launch_grouped_gemm_kernel<Kernel>(descs, ctx, stream_cfg);
```
ipanfilo: Why is it needed?

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_common.h
```diff
@@ -75,6 +75,7 @@ struct GroupedGemmRunContext {
     const NVTETensor* B = nullptr;
     NVTETensor* D = nullptr;
     int64_t N = 0;
+    int64_t K = 0;
```
ipanfilo: It does not seem used

aris134: You're right, I think I was using it at one point and then abandoned it, so I have removed it in a9202bf

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp8.cpp
aris134: Good point, didn't need to do this. Reverted in a9202bf

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp8.cpp
```diff
@@ -334,9 +457,12 @@ static bool ck_tile_grouped_gemm_fp8_dispatch_arch(DType a_dtype,
     return false;
   }
 
-  return runner->run(s, ctx);
+  const bool ok = runner->run(s, ctx);
```
ipanfilo: Similar to line 293 - it does not seem needed

aris134: Changed in deff9b4

wenchenvincent:
@aris134 Could you resolve the conflict now that the dev has moved?

ipanfilo (approved):

matthiasdiener (commented):
- What is the impact on compile time?
- I think bf16 also enables kPadM

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm.cpp
```diff
@@ -51,6 +51,7 @@ bool ck_tile_grouped_gemm(const NVTETensor* A,
 
   // Currently the accumulate path is only supported on fp16
   if (accumulate && is_8bit_float) {
+    NVTE_WARN("ck_tile_grouped_gemm: accumulate is currently unsupported on fp8");
```
matthiasdiener: Can you make this print only when warn_fallback is set? (perhaps put the check into nvte_multi_tensor_gemm if it is easier)

aris134:
> * What is the impact on compile time?
> * I think bf16 also enables kPadM

- Compile time does not seem much affected. I timed a full build at close to 6 and a half minutes.
- FP8 does not need m-padding. The FP8 GEMM is always canonicalized as NT. That, together with the fact that C is always RowMajor, means that `IsSupportedArgument` check within `QuantGemmKernel` only checks N and K for alignment requirements. This would seem to be further supported by the fact that there were no shape fallbacks in the Qwen run without explicit M-padding.

transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm.cpp
aris134: Done in 458acac

matthiasdiener (approved):

PR #596, opened 2026-05-26
Add timeout to distributed torch tests to fail on hang

ipanfilo:
# Description

Some distributed pytorch tests demonstrate random hang. Sometimes it triggers timeout in RCCL but in most cases it results in infinite wait.
The change in this PR adds test level timeout and torchrun wrapper to let it kill all child processes on timeout

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add subprocess.run wrapper that supports timeout and graceful shutdown

# Checklist:


alextmagro (approved):
LGTM -- One comment, would it make sense to use the utils.py already present in pytorch/ or do you want to keep distributed/utils.py separate?

ipanfilo:
> LGTM -- One comment, would it make sense to use the utils.py already present in pytorch/ or do you want to keep distributed/utils.py separate?

Because the tests are not loaded as a module import of ../utils is problematic and requires playing with sys.path. That is the main reason of creating separate utils.py

wangye805 (approved):

tests/pytorch/distributed/test_torch_fsdp2.py
```diff
@@ -32,7 +40,8 @@ def _run_test(fp_init, sharding_dims, recipe, layer_type):
     test_cmd += ["--recipe", recipe]
     test_cmd += ["--layer-type", layer_type]
 
-    result = subprocess.run(test_cmd, env=os.environ, check=True)
+    result = run_subprocess(test_cmd, 120 if IS_HIP_EXTENSION else None, env=os.environ,
+                            check=True)
```
matthiasdiener: Instead of using `run_proctree_with_timeout`, could we just use coreutils' `timeout` command?

This could perhaps even become a pytest fixture:

```python
# in conftest.py:
_DEFAULT_TIMEOUT = 120
_DEFAULT_KILL_AFTER = 30


def pytest_configure(config):
    config.addinivalue_line(
        "markers",
        "subprocess_timeout(seconds, kill_after=30): "
        "wrap subprocess.run calls with coreutils timeout to detect hangs",
    )


@pytest.fixture(autouse=True)
def subprocess_timeout(request, monkeypatch):
    marker = request.node.get_closest_marker("subprocess_timeout")
    if marker is None:
        return

    seconds = str(marker.args[0]) if marker.args else str(_DEFAULT_TIMEOUT)
    kill_after = str(marker.kwargs.get("kill_after", _DEFAULT_KILL_AFTER))

    original_run = subprocess.run

    def _run_with_timeout(cmd, *args, **kwargs):
        cmd = ["timeout", f"--kill-after={kill_after}", seconds] + list(cmd)
        result = original_run(cmd, *args, **kwargs)
        if result.returncode == 124:
            pytest.fail(f"Subprocess timed out after {seconds}s (hang detected)")
        return result

    monkeypatch.setattr(subprocess, "run", _run_with_timeout)

# for the test:
@pytest.mark.subprocess_timeout(20, kill_after=2)
def test_xyz():
```

This would have a few advantages:
- No sys.path modification, shorter code
- No code changes inside test functions
- signal/kill whole process group natively

ipanfilo: It is brilliant idea. It can be done even more simple by modifying command line right in tests. I've modified the PR. The first CI run is at https://github.com/ROCm/TransformerEngine/actions/runs/26611894136/job/78425262492

matthiasdiener: This looks good. 

There are no timeout-related messages in the distributed-test logs at https://github.com/ROCm/TransformerEngine/actions/runs/26611894136/job/78425262492 as far as I can tell, and the distributed tests passed, so the mechanism was not exercised in that run. Not sure if you want to test this further within this PR, but either way, good to go from my side.

matthiasdiener (approved):

PR #599, opened 2026-05-28
Update QoLA/AITER 

Micky774:
# Description

Updates QoLA as well as moves up the pinned AITER commit

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -22,34 +22,56 @@ if(NOT _gfx1250_idx EQUAL -1)
...
+  # Redirect GIT_CONFIG_GLOBAL to a tempfile carrying `safe.directory = *` so
+  # git operations inside the QoLA-managed AITER tree (and its recursive
+  # submodules) work in containerized builds where the bind-mounted .git is
+  # owned by a different UID than the build process. Mirrors the pattern in
+  # transformer_engine/common/CMakeLists.txt:get_git_commit().
+  execute_process(
+    COMMAND sh -c
+      "tmp=$(mktemp /tmp/gitconfig.XXXXXX) || exit 1; \
+GIT_CONFIG_GLOBAL=$tmp git config --global --add safe.directory '*' >/dev/null 2>&1; \
+GIT_CONFIG_GLOBAL=$tmp PYTHONPATH=\"${__QOLA_DIR}:$PYTHONPATH\" '${Python_EXECUTABLE}' -m qola.cli checkout \
```
ipanfilo: Wouldn't it make sense to integrate safe.directory overriding to qola? The pattern with dubious ownership is probably not TE specific

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -463,8 +458,11 @@ hipError_t ck_attn_bwd(const CkAttnBwdArgs& args, hipStream_t stream){
   }
 
   aiter::mha_bwd_args fmha_args{};
+  fmha_args.sink_ptr = nullptr;
+  fmha_args.d_sink_ptr = nullptr;
   fmha_args.mask_type = static_cast<int>(static_cast<mask_enum>(args.attn_mask_type));
-  fmha_args.use_asm_v3 = args.uses_bwd_v3;
+  // Mirrors AITER's small-seqlen guard at aiter/ops/mha.py:1689.
+  fmha_args.use_asm_v3 = (args.s_q < 16) ? false : args.uses_bwd_v3;
```
ipanfilo, line 536: Is it to avoid warning from AITER about unsupported config?

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
```diff
@@ -18,8 +18,22 @@ string(STRIP "${ROCM_VER_CONTENT}" ROCM_VER_CONTENT)
 string(REGEX MATCH "^[0-9]+\\.[0-9]+" ROCM_VER "${ROCM_VER_CONTENT}")
 string(REGEX MATCH "^[0-9]+" ROCM_VER_MAJOR "${ROCM_VER}")
 
-# AITER commit
-get_git_commit("${__AITER_SOURCE_DIR}" AITER_SHA)
+# AITER commit — read from the QoLA manifest so the cache key tracks the
+# commit QoLA will actually check out and build, not whatever happens to be
+# the submodule's current HEAD at configure time.
+set(__QOLA_MANIFEST "${CMAKE_CURRENT_LIST_DIR}/qola_manifest.toml")
+set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS "${__QOLA_MANIFEST}")
```
ipanfilo: AITER_SHA is not cached variable. Why is CMAKE_CONFIGURE_DEPENDS needed?

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -22,34 +22,56 @@ if(NOT _gfx1250_idx EQUAL -1)
...
+    RESULT_VARIABLE AITER_CHECKOUT_RESULT
+    OUTPUT_VARIABLE AITER_CHECKOUT_OUTPUT
+    ERROR_VARIABLE AITER_CHECKOUT_ERROR
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+    ERROR_STRIP_TRAILING_WHITESPACE
+  )
+  if(NOT AITER_CHECKOUT_RESULT EQUAL 0)
+    message(FATAL_ERROR
+            "Failed to sync AITER source tree at ${__AITER_SOURCE_DIR} to "
+            "manifest-pinned commit ${AITER_SHA}.\n"
```
ipanfilo, line 83: Should it also validate that actual commit matches one detected by prebuilt.cmake? If QoLA checkout AITER unconditionally, may be keep prebuilt.cmake as-is and where it is now? I.e. QoLA fetches AITER, prebuit.cmake checks for git commit as before. It will only loose AITER_SHA value in this error message

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: No, rather we get broken kernels that are currently unsupported. This is a gap in their C++ dispatch that's covered in their python layer (which we bypass).

transformer_engine/common/ck_fused_attn/CMakeLists.txt
Micky774: I worry that such behavior is a bit too authoritative for qola if that makes sense? My reasoning is that the permission scope here seems to be outside of qola and hence qola should not be the one in charge of it. I'm open to reconsidering that, but it's just my initial position.

ipanfilo: It is valid concern bearing in mind QoLA is intended to be reused by different components. May be make  this behavior controllable then. It is OK to keep the things in TE, it will just require doing things this overriding twice - here and when build is called - BTW, there are comments there but not actual code change

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -582,19 +579,63 @@ hipError_t ck_attn_bwd(const CkAttnBwdArgs& args, hipStream_t stream){
...
+  // and the dq_acc accumulator). aiter only contracts that the pointer remain
+  // valid for the duration of the kernels it enqueues; hipFreeAsync on the
+  // same stream defers the free until that work completes.
+  std::vector<void*> mha_bwd_workspaces;
+  fmha_args.workspace_alloc = [&mha_bwd_workspaces, stream](size_t bytes, bool zero_init) -> void* {
+    if(bytes == 0){
+      return nullptr;
+    }
+    void* ptr = nullptr;
+    if(hipMallocAsync(&ptr, bytes, stream) != hipSuccess){
```
wangye805: Emm, if we let hip runtime to allocate and manage our buffers, this will create a series of issues. For example, if Pytorch or JAX users pre-allocate 97% of HBM, then our hipMallocAsync will return out of memory. 

If what new aiter needs for fmha_args.workspace_alloc is a lambda, we can fake it to give jax/pytorch generated workspace buffer ptr?

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -582,19 +579,63 @@ hipError_t ck_attn_bwd(const CkAttnBwdArgs& args, hipStream_t stream){
...
+    mha_bwd_workspaces.push_back(ptr);
+    return ptr;
+  };
+  // Group mode requires a pinned host buffer for the async D2H seqstart
+  // pipeline; aiter keeps the shared_ptr alive past kernel completion via a
+  // stream-tail hipLaunchHostFunc keepalive. The deleter fires from that HIP
+  // callback thread, which holds runtime locks — calling any HIP API from it
+  // (including hipHostFree) deadlocks against concurrent main-thread HIP
+  // calls. Defer the free to ck_tile::pinned_host_releaser's worker thread.
+  fmha_args.pinned_host_alloc = [](size_t bytes) -> std::shared_ptr<void> {
```
wangye805: Emm, why do they need host memory allocated? Is it for inference?

Again, can we fake the lambda to use pytorch/jax generated workspace?

wangye805 (changes requested):

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
Micky774: I mainly have it so that direct `cmake` or `ninja` building will catch on manifest changes since I use it for incremental builds (and am thinking about including an incremental build option for TE's AITER FA backend). It's not strictly necessary, just a nice to have in such a workflow

transformer_engine/common/ck_fused_attn/CMakeLists.txt
Micky774: How about we add that explicit check for check-out commit = manifest commit but keep the current changes so we can explicitly track and use the manifest commit as a single source of truth for the build?

transformer_engine/common/ck_fused_attn/CMakeLists.txt
Micky774: Sounds good, but can we do that in a separate PR to QoLA first?

transformer_engine/common/ck_fused_attn/aiter_prebuilt.cmake
ipanfilo: Will it also trigger re-executing of QoLA cli in the parent (ck_used_attn/ ) CMAkeLists.txt?

VeeraRajasekhar (commented):
Verified that https://github.com/ROCm/rocm-libraries/pull/6764 exists in the current pinned aiter commit.

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -30,10 +57,11 @@ bool is_ck_backend_supported(
...
   size_t max_seqlen_q, size_t max_seqlen_kv,
-  size_t head_dim_qk, 
-  size_t head_dim_v, 
-  int64_t window_size_left, 
-  int64_t window_size_right) {
+  size_t head_dim_qk,
+  size_t head_dim_v,
+  int64_t window_size_left,
+  int64_t window_size_right,
+  bool is_training, bool cuda_graph) {
```
AllenFarcas: From my understanding, `is_ck_backend_supported` is the only consumer of the new graph-capture gate, but it's called without `deterministic` (`TransformerEngine\transformer_engine\common\fused_attn_rocm\fused_attn.cpp` line 320) even though `nvte_get_fused_attn_backend` has it in scope (`TransformerEngine\transformer_engine\common\fused_attn_rocm\fused_attn.cpp` line 282).

Since deterministic bwd forces the non-graph-safe CK v2 path, `is_ck_bwd_graph_capture_safe` cannot make the right call without it. See my other related comments above for what could be added to fix.

You should include here the `bool deterministic` as well

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -18,6 +18,33 @@
...
+  // v2 path off entirely.
+  bool uses_bwd_v3 = getenv<int>("NVTE_CK_USES_BWD_V3", 1);
+  const std::string& arch = cuda::sm_arch_name();
+  bool is_v3_arch = arch.find("gfx942") != std::string::npos ||
+                    arch.find("gfx950") != std::string::npos;
+  return uses_bwd_v3 &&
+         is_v3_arch &&
+         dropout == 0.f &&
+         bias_type == NVTE_Bias_Type::NVTE_NO_BIAS &&
+         max_seqlen_q >= 16;
```
AllenFarcas: Add ` && !deterministic`

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -145,6 +173,20 @@ bool is_ck_backend_supported(
...
   }
+
+  // Under HIP-graph capture, CK backward must take the graph-replay-safe v3 asm
+  // path; a config that would fall back to the CK v2 launcher is not graph-safe.
+  // Reject such graph-captured training configs so selection falls through to a
+  // graph-safe backend (the v2 host-pack hazard is backward-only, so inference is
+  // unaffected). Determinism also forces v2 but is invisible here, so it is handled
+  // on the framework side.
+  if(is_training && cuda_graph &&
+     !is_ck_bwd_graph_capture_safe(bias_type, dropout, max_seqlen_q)){
```
AllenFarcas: You should include here the `deterministic` as well

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -18,6 +18,33 @@
...
+// Returns false when a CK backward for this config would dispatch to the CK v2
+// launcher (fmha_bwd / prepare_workspace_async), which schedules self-deleting
+// hipLaunchHostFunc nodes that double-free on graph replay. Only the v3 asm bwd
+// path is HIP-graph-replay-safe. Mirrors AITER's fmha_v3_bwd gate (mha_bwd.cu)
+// for the conditions visible at backend-selection time; determinism is applied
+// separately on the framework side.
+static bool is_ck_bwd_graph_capture_safe(
+  NVTE_Bias_Type bias_type,
+  float dropout,
+  size_t max_seqlen_q) {
```
AllenFarcas: Add `bool deterministic`

transformer_engine/common/fused_attn_rocm/fused_attn.cpp
```diff
@@ -316,7 +316,8 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
         head_dim_qk,
         head_dim_v,
         window_size_left,
-        window_size_right)){
+        window_size_right,
+        is_training, cuda_graph)){
```
AllenFarcas: Pass here the `deterministic`

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -18,6 +18,33 @@
 namespace transformer_engine {
 namespace fused_attn_rocm {
 
+#ifdef USE_FUSED_ATTN_CK
+// Returns false when a CK backward for this config would dispatch to the CK v2
+// launcher (fmha_bwd / prepare_workspace_async), which schedules self-deleting
+// hipLaunchHostFunc nodes that double-free on graph replay. Only the v3 asm bwd
+// path is HIP-graph-replay-safe. Mirrors AITER's fmha_v3_bwd gate (mha_bwd.cu)
+// for the conditions visible at backend-selection time; determinism is applied
```
AllenFarcas: My understanding is that this comment (and two more below) justify the split design by claiming the C++ selector can't see determinism. But `deterministic` is literally a parameter of the `nvte_get_fused_attn_backend` and both frameworks pass it. 

The flag isn't invisible, the code just chooses not to forward it one level down, which should be fixed by my previous comment. Leaving these comments as-is will mislead the next person into thinking the Python guard is a hardware/architectural necessity rather than a fixable shortcut.

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -145,6 +173,20 @@ bool is_ck_backend_supported(
     }
     return false;
   }
+
+  // Under HIP-graph capture, CK backward must take the graph-replay-safe v3 asm
+  // path; a config that would fall back to the CK v2 launcher is not graph-safe.
+  // Reject such graph-captured training configs so selection falls through to a
+  // graph-safe backend (the v2 host-pack hazard is backward-only, so inference is
+  // unaffected). Determinism also forces v2 but is invisible here, so it is handled
```
AllenFarcas: Similar comment here

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -1103,6 +1103,24 @@ def _is_fa3_supported(num_heads, num_gqa_groups, head_dim_qk, head_dim_v, qkv_dt
             logger.debug("Disabling FusedAttention for determinism reasons with post_scale_bias")
             use_fused_attention = False
             fused_attention_backend = None
+    # ROCm: deterministic CK backward dispatches to the CK v2 launcher, which
+    # schedules self-deleting host nodes that are not HIP-graph-replay-safe. Under
+    # graph capture this corner is invisible to the C++ backend selector (determinism
+    # is not part of its signature), so fall back to the unfused backend here.
```
AllenFarcas: Similar comment here

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: Updated

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: Updated

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: Updated

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: Updated

transformer_engine/common/fused_attn_rocm/fused_attn.cpp
Micky774: Updated

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: Updated

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: Updated

transformer_engine/pytorch/attention/dot_product_attention/utils.py
Micky774: Updated

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -463,8 +521,11 @@ hipError_t ck_attn_bwd(const CkAttnBwdArgs& args, hipStream_t stream){
   }
 
   aiter::mha_bwd_args fmha_args{};
+  fmha_args.sink_ptr = nullptr;
+  fmha_args.d_sink_ptr = nullptr;
   fmha_args.mask_type = static_cast<int>(static_cast<mask_enum>(args.attn_mask_type));
-  fmha_args.use_asm_v3 = args.uses_bwd_v3;
+  // Mirrors AITER's small-seqlen guard at aiter/ops/mha.py:1689.
```
AllenFarcas, line 535: The next time AITER version is bumped this hardcoded line number will not be relevant anymore.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -677,20 +778,30 @@ hipError_t ck_attn_bwd(const CkAttnBwdArgs& args, hipStream_t stream){
...
+    }
+    // Fallback: no reservation exists (the size pass did not run before this dispatch).
+    // Under capture this hipHostMalloc returns hipErrorStreamCaptureUnsupported; surface a
+    // clear error rather than corrupting the capture. The deferred free mirrors aiter's
+    // stream-tail release contract: the deleter fires from a HIP callback thread holding
+    // runtime locks, so it must not call any HIP API directly.
     void* ptr = nullptr;
     if(hipHostMalloc(&ptr, bytes, hipHostMallocDefault) != hipSuccess){
-      throw std::runtime_error("ck_fused_attn bwd: hipHostMalloc failed for AITER pinned host buffer.");
+      throw std::runtime_error("ck_fused_attn bwd: pinned host staging was not reserved and "
```
ipanfilo: This message is specific for hipErrorStreamCaptureUnsupported only while there might be other allocation errors, so dispatch by hipGetLastError() is needed

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -723,13 +834,6 @@ hipError_t ck_attn_bwd(const CkAttnBwdArgs& args, hipStream_t stream){
     bool resolves_to_v3 = fmha_args.use_asm_v3 && !fmha_args.is_deterministic &&
                           !fmha_args.has_dbias && fmha_args.bias_type == 0 &&
                           !fmha_args.has_dropout && is_v3_arch;
-    if(!resolves_to_v3){
```
ipanfilo: W/o this throw the whole graph capture safety net is useless

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -491,6 +494,88 @@ size_t v3_dq_acc_bytes(const CkAttnBwdArgs& args){
...
+// hipStreamEndCapture). We therefore reserve the buffer once from
+// ck_attn_bwd_workspace_size (the host-side size pass, which runs outside capture) and
+// hand the captured dispatch a pre-reserved slice, so prepare_workspace_async issues only
+// capturable ops.
+//
+// A captured graph embeds its host-pack closure into this buffer and dereferences it on
+// every replay, so the buffer must outlive the graph: slabs are never freed (process
+// lifetime, bounded by the number of distinct backward configs). Distinct configs get
+// distinct slabs; configs that collide share one, which is correct for the serial,
+// single-stream replay used by JAX/XLA.
```
ipanfilo: But this API is generic, not JAX/XLA. And we know that torch tests pass. Would it make sense to make JAX extension CK aware and move this mechanics there? BTW JAX extensions always call nvte_fused_attn_bwd with graph=false which seems incorrect

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: I've updated the error string to be more descriptive, lmk if this is sufficient.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: The graph capture safety net actually needs to be removed entirely since now the CK kernels provided _should_ be graph safe. I've updated the PR to remove it.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: For now since this meant to be a short-term solution, are you okay with leaving things as-is? Also the graph arg is unused on the ROCm side entirely, so I've just updated it to be denoted explicitly as unused.

ipanfilo: Yes, like we discussed, it works as temporary w/a

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -66,13 +66,27 @@ if("${AITER_SHA}" STREQUAL "")
           "Failed to parse 'aiter_commit = \"...\"' line in ${__QOLA_MANIFEST}.")
 endif()
 
+# Resolve the manifest-pinned AITER commit (defines AITER_SHA) and bring the
+# QoLA-managed AITER source tree to that commit before any consumer reads it
+# (header validation below, header includes for the .cpp build later, and
+# QoLA's own kernel build if the prebuilt cache misses).
+include("${CMAKE_CURRENT_LIST_DIR}/aiter_prebuilt.cmake")
```
ipanfilo: It is not needed here - AITER_SHA is determined above, cache directory is only needed if neither AITER_MHA_PATH is set nor CK-JIT is enabled, so it is included at line 158

transformer_engine/common/ck_fused_attn/CMakeLists.txt
ipanfilo: Now, with qola expected to create aiter dir runtime (it is not a part of repository tree), is there a reason to modify git config safe.directory? It was added for specific case: repository tree is checked out by a user and then shared with a container run with another user credentials.

transformer_engine/common/ck_fused_attn/CMakeLists.txt
```diff
@@ -106,6 +120,8 @@ if(Python_EXECUTABLE)
             "${AITER_ARG_CHECK_OUTPUT}\n${AITER_ARG_CHECK_ERROR}")
   endif()
   message(STATUS "AITER API validation passed via check_aiter_mha_args.py")
+else()
+  message(WARNING "Python interpreter not found; skipping AITER source-tree sync and API validation.")
```
ipanfilo: There is e messages at line 43. If python is not found it won't just skip those steps - QoLA or CK-JIT builds will fail as well. So number of scenarios when Python is not needed is very limited

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -585,7 +585,16 @@ size_t ck_attn_bwd_workspace_size(const CkAttnBwdArgs& args){
   // local by QoLA's export script, so the v2 size is queried through QoLA.
   const size_t v2_bytes = QOLA_NS(mha_bwd_workspace_size)(make_bwd_traits(args));
   const size_t v3_bytes = v3_dq_acc_bytes(args);
-  return v2_bytes > v3_bytes ? v2_bytes : v3_bytes;
+  // Safety floor: when the CK-JIT cannot find workspace-query symbols for a
```
ipanfilo: For my curiosity - it happens with CK-JIT enabler only? How is it handled with CK-AOT compiling?

ipanfilo (commented):
It seems there are still some numeric failures in determenistic path

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: To clarify, the original comment is misleading/wrong. The issue is not intrinsic to CK-JIT but rather CK itself, and is only exposed through CK-JIT. AOT build would have the same issue. Fundamentally the host-side CK workspace sizing estimator is incorrect, and that is the crux of the issue.

ipanfilo:
The latest fixes for determinism seem not helping on gfx942

Micky774:
Alright looks like the failures are unrelated now.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -511,8 +682,13 @@ hipError_t ck_attn_bwd(const CkAttnBwdArgs& args, hipStream_t stream){
...
   fmha_args.seqlen_k_ptr = nullptr;
 
-  fmha_args.seqlen_q = args.s_q;
-  fmha_args.seqlen_k = args.s_kv;
+  // Group mode contract (matches aiter asm_mha_varlen_bwd.cu): seqlen_q/k
+  // carry the total token counts, max_seqlen_q/k the per-sequence maximum.
+  // aiter sizes dq_acc and related workspaces from seqlen_q; passing the
+  // per-sequence length in group mode under-sizes them and the kernel writes
+  // past the end.
+  fmha_args.seqlen_q = args.is_group_mode() ? args.max_tokens_q : args.s_q;
```
wangye805, line 588: I recall max_tokens are summarized over all batches

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -582,18 +753,96 @@ hipError_t ck_attn_bwd(const CkAttnBwdArgs& args, hipStream_t stream){
...
+  // call; only one allocation happens per dispatch, but the bump allocator stays
+  // correct if aiter splits the request.
+  void* ws_base = args.aiter_workspace_ptr;
+  const size_t ws_capacity = args.aiter_workspace_bytes;
+  size_t ws_offset = 0;
+  fmha_args.workspace_alloc = [ws_base, ws_capacity, &ws_offset, stream](size_t bytes, bool zero_init) -> void* {
+    if(bytes == 0){
+      return nullptr;
+    }
+    constexpr size_t kAlign = 256;
```
wangye805: Where is this kAlign=256 coming from?

transformer_engine/common/fused_attn_rocm/fused_attn.cpp
```diff
@@ -390,7 +390,7 @@ void nvte_fused_attn_fwd(const NVTETensor Q, const NVTETensor K, const NVTETenso
   NVTE_Fused_Attn_Backend fused_attention_backend = nvte_get_fused_attn_backend(
       is_training, Q_type, KV_type, qkv_layout, bias_type, attn_mask_type, softmax_type, dropout,
       h_q, h_kv, max_seqlen_q, max_seqlen_kv, d_qk, d_v, window_size_left, window_size_right,
-      return_max_logit, cuda_graph, false);
+      return_max_logit, false, false);
```
wangye805: Do we enable cuda_graph=true before?

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -892,8 +919,6 @@ void fused_attn_ck_bwd_impl(
   }
 
   // Initialize workspace buffers.
-  // dq_acc is of shape (nsplits, B, S, H, D_qk); CK requires zeroing
-  NVTE_CHECK_CUDA(cudaMemsetAsync(dq_acc_ptr, 0, sizeof(float)*nsplits*h*max_tokens_q*d_qk, stream));
```
wangye805, line 896: So now CK will handle the dq_acc zeroing out inside by our lambda fmha_args.workspace_alloc? 

Then how about v3 asm path? Do we need to provide the zero out?

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -441,20 +443,187 @@ void dump_bwd_timings(const char* dump_path, float average_runtime){
...
+  // under-report the dq_acc the actually-dispatched kernel uses. The deterministic
+  // dq_acc holds nsplits = ceil(s_kv / kN0) copies, so a smaller kN0 means MORE splits
+  // and a LARGER buffer. To stay an upper bound we must use the smallest kN0 the CK-tile
+  // bwd dispatcher can pick for any supported arch. Per the CK codegen bwd tile tables,
+  // the smallest non-tail kN0 is 64.
+  const size_t kN0 = 64u;
+  const size_t nsplits = args.deterministic ? ((args.s_kv + kN0 - 1) / kN0) : 1u;
+  const size_t tokens_q = args.is_group_mode() ? args.max_tokens_q : (args.b * args.s_q);
+  const size_t dq_acc_floor = nsplits * args.h * tokens_q * args.d_qk * sizeof(float);
+  return std::max({v2_bytes, v3_bytes, dq_acc_floor});
```
wangye805: So essentially v2_bytes can sometimes still be smaller than our dq_acc_floor which causes some issue?

wangye805 (changes requested):

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: Right, that floor is meant to be a safety mechanism since we don't know a-priori the exact amount of memory CK will use, and their upper-bound requires padding-aware quantities and assumes a uniform distribution of seqlen in the case of padding between sequences.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: Right, this is necessary with their new sizing API. In group-mode, that's the expected value of `fmha_args.seqlen_q`. Tests fail without this.

transformer_engine/common/fused_attn_rocm/fused_attn.cpp
Micky774: It was ignored before, so we can set it to false arbitrarily. I've added a comment clarifying this.

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: > So now CK will handle the dq_acc zeroing out inside by our lambda fmha_args.workspace_alloc?

Yes

> Then how about v3 asm path? Do we need to provide the zero out?

We provide the mechanism for that through the `fmha_args.workspace_alloc` callback w/ optional zero init that remains controlled by AITER. So in short, yes, we provide the zeroing for this.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -441,20 +440,117 @@ void dump_bwd_timings(const char* dump_path, float average_runtime){
...
+// past the granted dq_acc region -> NaN.
+//
+// The tight value would be the true max padded segment
+// max_i(cu_seqlen_kv_padded[i+1]-cu_seqlen_kv_padded[i]), but that is only knowable at
+// runtime from the device array: it cannot be computed in the lowering-time sizing pass
+// (seqstarts unpopulated) and a runtime reduction to fetch it syncs the stream, which is
+// illegal under HIP-graph capture. So reserve a capture-safe upper bound instead.
+//
+// The dq_acc split count is ceil(seqlen_k / kN0) where kN0 is the dispatched bwd KV tile
+// (bn0). The largest bn0 across supported archs is kCkBwdMaxKvTile (per the CK codegen bwd
```
wangye805: bn0 is the same as kN0?

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -441,20 +440,117 @@ void dump_bwd_timings(const char* dump_path, float average_runtime){
...
+// tile tables, codegen/ops/fmha_bwd.py). Adding one full max-tile of headroom guarantees at
+// least one extra split for any actually-dispatched kN0 (all <= kCkBwdMaxKvTile), which
+// covers inter-sequence padding of up to a whole tile per sequence -- far above the
+// few-token THD alignment padding used in practice.
+constexpr uint64_t kCkBwdMaxKvTile = 256;
```
wangye805: Emm, this const is copied from CK source code? If so, it may be changed later in CK which can cause our issue.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -441,20 +440,117 @@ void dump_bwd_timings(const char* dump_path, float average_runtime){
...
+  if(!use_asm_v3){
+    return 0;
+  }
+  const size_t seqlen_q = args.is_group_mode() ? args.max_tokens_q : args.s_q;
+  const size_t elem = args.is_v3_atomic_fp32 ? 4 : 2;
+  const size_t a16_seq = (args.s_q + 15) / 16 * 16;
+  const size_t a16_hdim = (args.d_qk == 192) ? 192 : 128;
+  const size_t dq_acc_seq = args.is_v3_atomic_fp32 ? seqlen_q : a16_seq;
+  const size_t dq_acc_hdim = args.is_v3_atomic_fp32 ? args.d_qk : a16_hdim;
+  const size_t eff_batch = (args.is_group_mode() && args.is_v3_atomic_fp32) ? 1 : args.b;
```
wangye805, line 491: In my mind, effective batch should be related to whether it's in group mode or not. bf16 or fp32 atomic mode should not affect it.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -441,20 +440,117 @@ void dump_bwd_timings(const char* dump_path, float average_runtime){
...
+    /* has_dropout      */ has_dropout,
+    /* is_store_randval */ false,
+    /* is_deterministic */ args.deterministic,
+  };
+}
+
+// dq_acc bytes the v3 asm path allocates via workspace_alloc. Mirrors aiter's
+// fmha_v3_bwd sizing (csrc/cpp_itfs/mha_bwd.cu); returns 0 when v3 can't run so
+// the CK launcher size dominates. Gating mirrors ck_attn_bwd's use_asm_v3.
+size_t v3_dq_acc_bytes(const CkAttnBwdArgs& args){
```
wangye805, line 479: Another thing I don't quite understand is: the whole dq_acc change is from CK side, not aiter asm. So v3 asm required dq_acc should not change too much?

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
wangye805: Emm, if that's a CK required change, not sure whether such large seqlen_q can affect the performance in AITER asm

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -441,20 +440,117 @@ void dump_bwd_timings(const char* dump_path, float average_runtime){
...
+}
+
+}  // namespace
+
+size_t ck_attn_bwd_workspace_size(const CkAttnBwdArgs& args){
+  // v2 (CK launcher) reports its full device workspace (host metadata + dq_acc)
+  // host-side; v3 (asm) allocates only dq_acc. v3 is tried first but may fall
+  // back to v2, so reserve the larger of the two. The launcher symbol is forced
+  // local by QoLA's export script, so the v2 size is queried through QoLA.
+  const size_t v2_bytes = QOLA_NS(mha_bwd_workspace_size)(make_bwd_traits(args));
```
wangye805, line 502: Can you remind me where the mha_bwd_workspace_size is defined? 

If the v2_bytes sometimes under estimate the real CK needs, can we change the mha_bwd_workspace_size there? Then we don't need to have std::max over three items. (We can just std::max over v3 asm needs and v2 ck needs)

wangye805 (changes requested):

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: Yes, it's a naming difference in different parts of the backend, but it's passed as the same var.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: This is based off of the `mha_bwd` section: https://github.com/ROCm/aiter/blob/028756633e4192785217838f4924dc16516f5780/csrc/cpp_itfs/mha_bwd.cu#L571-L601

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: We don't actually change anything -- this is just so that we can calculate the max between V2 (changed) / V3 (unchanged) needs.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
wangye805: Okay, let's put a comment of this code pointer so in the future we can understand where this piece of hardcoding code is from.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -441,6 +440,69 @@ void dump_bwd_timings(const char* dump_path, float average_runtime){
...
+// fmha_v3_bwd sizing (csrc/cpp_itfs/mha_bwd.cu); returns 0 when v3 can't run so
+// the CK launcher size dominates. Gating mirrors ck_attn_bwd's use_asm_v3.
+size_t v3_dq_acc_bytes(const CkAttnBwdArgs& args){
+  const bool use_asm_v3 = (args.s_q < 16) ? false : args.uses_bwd_v3;
+  if(!use_asm_v3){
+    return 0;
+  }
+  const size_t seqlen_q = args.is_group_mode() ? args.max_tokens_q : args.s_q;
+  const size_t elem = args.is_v3_atomic_fp32 ? 4 : 2;
+  const size_t a16_seq = (args.s_q + 15) / 16 * 16;
```
wangye805, line 486: In the future, let's add non 16 multiple sq's to our unit tests to see what will happen.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
wangye805: nit: if mha_bwd_workspace_size is defined in CK or qola, maybe should also put v3_dq_acc_bytes into qola

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -466,8 +528,11 @@ BwdFmhaArgs build_bwd_fmha_args(const CkAttnBwdArgs& args){
   }
 
   aiter::mha_bwd_args fmha_args{};
+  fmha_args.sink_ptr = nullptr;
+  fmha_args.d_sink_ptr = nullptr;
   fmha_args.mask_type = static_cast<int>(static_cast<mask_enum>(args.attn_mask_type));
-  fmha_args.use_asm_v3 = args.uses_bwd_v3;
+  // Mirrors AITER's small-seqlen guard at aiter/ops/mha.py:1689.
+  fmha_args.use_asm_v3 = (args.s_q < 16) ? false : args.uses_bwd_v3;
```
wangye805, line 536: nit: have such a testcase (s_q<16) to see if things work as our expectation

wangye805 (approved):

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: We already have several where `s_q=1` which trigger this code path. Specifically it was put in as a response to a failing test case originally.

3rdparty/ck_jit
ipanfilo, line 1: Is it expected to be in CK-JIT main?

transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -471,6 +471,18 @@ pybind11::tuple GetFusedAttnBackwardWorkspaceSizes(
...
+  // ROCm CK: the deterministic CK bwd workspace size differs between group and batch mode, and
+  // is_group_mode() keys off a non-null cu_seqlens pointer. The real cu_seqlens are unavailable at
+  // this lowering-time sizing pass, so with a null pointer JAX would size the workspace for batch
+  // mode while execution runs in group mode -> under-allocated dq_acc -> OOB write into the forward
+  // output. For group-mode configs (ragged/THD or a padding mask) pass a non-null sentinel so the
+  // sizing pass matches execution. The pointer is only null-checked (never dereferenced) here.
+  const bool mask_is_padding = (mask_type == NVTE_Mask_Type::NVTE_PADDING_MASK ||
+                                mask_type == NVTE_Mask_Type::NVTE_PADDING_CAUSAL_MASK ||
+                                mask_type == NVTE_Mask_Type::NVTE_PADDING_CAUSAL_BOTTOM_RIGHT_MASK);
+  void *cu_seqlens_sizing_sentinel =
```
ipanfilo: Should it be forced to nullptr for NV?

ci/ck_jit_prebuild.txt
ipanfilo, line 1: There are missing MI300 entries in the cache:
See 'Cache diff' in sGPU logs: https://github.com/ROCm/TransformerEngine/actions/runs/29612496923/job/88001450430?pr=599
fmha_bwd_convert_dq_d32_bf16_b64_group_o2_ps_ndeterministic_gfx9
fmha_bwd_d32_bf16_group_b32x128x32x32x32x32x64x32x32_r1x4x1_r4x1x1_r2x2x1_w16x16x32_w16x16x16_o1_maxq0_npad_nbias_ndbias_nmask_dropout_wg16_ndeterministic_ntrload_gfx9
fmha_bwd_dot_do_o_d32_bf16_b64_group_o2_ps_gfx9
fmha_fwd_d32_bf16_group_b128x64x16x32x32x32_r4x1x1_r4x1x1_w32x32x16_w32x32x16_qr_async_vr_psskddv_nlogits_nbias_nmask_lse_dropout_nskip_nqscale_ntrload_nsink_gfx9

For Mi350 sGPU no data because of no test logs

transformer_engine/jax/csrc/extensions/attention.cpp
Micky774: That would be safest, updated.

ci/ck_jit_prebuild.txt
Micky774: I've updated for these kernels -- since CI is having some issues, lets update the gfx950 ones in a follow-up PR?

ipanfilo (approved):

PR #609, opened 2026-06-03
enable blockwise FP8 quantization on rocm

asdfvg123:
# Description

Enable blockwise FP8 quantization on rocm

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

remove HIP guard in quantization.py
guard kernels using TMA in quantization. 
add branch to handle rocm for different threads per wave

# Checklist:


tests/pytorch/test_float8blockwisetensor.py
```diff
@@ -45,7 +45,10 @@ def _to_list(x: Union[Iterable, Any]) -> List:
 DimsType = Union[Iterable[int], int]
 
 # TODO replace with call to fp8.py when recipe added.
-recipe_available = not IS_HIP_EXTENSION and (get_device_compute_capability() >= (9, 0) and float(torch.version.cuda) >= 12.8)
+if IS_HIP_EXTENSION:
+    recipe_available = get_device_compute_capability() >= (9, 0)
```
matthiasdiener: Wouldn't this be always `True` on ROCm TE?

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
matthiasdiener, line 1: Needs AMD copyright

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
```diff
@@ -5,30 +5,50 @@
...
+#ifndef __HIP_PLATFORM_AMD__
 #include <cuda/barrier>
+#endif
 
 #include "common/common.h"
 #include "common/recipe/recipe_common.cuh"
 #include "common/util/cuda_runtime.h"
+#ifndef __HIP_PLATFORM_AMD__
 #include "common/util/ptx.cuh"
+#endif
```
matthiasdiener: These `#include`s should be already disabled via hipify, so probably no need for the `#ifndef`s here.

transformer_engine/common/common.h
```diff
@@ -636,8 +636,8 @@ struct TypeExtrema<fp8e4m3> {
   static constexpr float max = te_fp8_fnuz() ? 240.0f : 448.0f;
   static constexpr float max_inverse = 1.0 / max;
 #else
-  static float max;
-  static float max_inverse;
+  static constexpr float max = 448.0f;
+  static constexpr float max_inverse = 1.0 / max;
```
matthiasdiener: Is this change necessary? fp8e4m3 max depends on the device type on AMD.

alextmagro:
Could you give a description of what you want to achieve with this PR? My understanding is that block fp8 quantization relies on some upstream kernels that will need to be adapted for AMD.

If you're just trying to enable the interface, I would argue that we should do this last, after we have a working quantization and GEMM path (and enabled and passing C++/Python tests).

asdfvg123:
@alextmagro 
This PR is to enable only the quantization in the AMD gpus, not the GEMM. There are two kernels in the upstream which uses TMA for the quantization and does not uses TMA for the quantization. I guarded the kernels which uses TMA and used the non-TMA kernels to quantize for AMD. 

I tested with 
`tests/pytorch/test_float8blockwisetensor.py`
and it passes [175 passed / 32 xpassed / 5 warnings]

alextmagro:
> @alextmagro This PR is to enable only the quantization in the AMD gpus, not the GEMM. There are two kernels in the upstream which uses TMA for the quantization and does not uses TMA for the quantization. I guarded the kernels which uses TMA and used the non-TMA kernels to quantize for AMD.
> 
> I tested with `tests/pytorch/test_float8blockwisetensor.py` and it passes [175 passed / 32 xpassed / 5 warnings]

OK, in that case we need to add the cpp blockwise tests to the CMake file, and the pytorch test file to ci/pytorch.sh.

tests/pytorch/test_float8blockwisetensor.py
asdfvg123: This test targets MI300 and MI350 so I set to (9,0)

ci/pytorch.sh
```diff
@@ -51,6 +51,7 @@ run_test_config(){
     run_default_fa 1 test_deferred_init.py
     run_default_fa 1 test_quantized_tensor.py
     run_default_fa 1 test_float8_current_scaling_exact.py
+    run_default_fa 1 test_float8blockwisetensor.py
```
alextmagro, line 53: nit: This test and test_float8_current_scaling_exact should be ahead of test_quantized_tensor.

tests/cpp/operator/test_cast_float8blockwise.cu
```diff
@@ -67,7 +67,7 @@ void scales_from_amax(float amax, const QuantizationOptions& opts, float* qscale
     qscale = ldexpf(1.0f, static_cast<int32_t>(exp) - 127);
   }
 
-  float qscale_inv = 1.0 / qscale;
+  float qscale_inv = 1.0f / qscale;
```
alextmagro: We had discussed this as a potential solution offline, but if the tolerances are needed then we should remove this

tests/cpp/operator/test_cast_float8blockwise.cu
```diff
@@ -227,17 +227,22 @@ void ref_quantize_onedimensional_blocks(const ProcessingMethod processing_method
...
+  return inner_elements;
+#else
   return ((inner_elements + 4u - 1u) / 4u) * 4u;
+#endif
 };
 
 void compare_scaling_factors(const std::string& name, const float* test, const float* ref,
                              const size_t row_blocks, const size_t col_blocks,
-                             const size_t test_stride, const size_t ref_stride) {
+                             const size_t test_stride, const size_t ref_stride,
```
alextmagro: Both function signature and the function change should be guarded. We might also just want to put the atol into the function body itself for now, since it is only used for blockwise scaling.

tests/pytorch/test_float8blockwisetensor.py
```diff
@@ -45,7 +45,10 @@ def _to_list(x: Union[Iterable, Any]) -> List:
 DimsType = Union[Iterable[int], int]
 
 # TODO replace with call to fp8.py when recipe added.
-recipe_available = not IS_HIP_EXTENSION and (get_device_compute_capability() >= (9, 0) and float(torch.version.cuda) >= 12.8)
+if IS_HIP_EXTENSION:
+    recipe_available = get_device_compute_capability() >= (9, 5)
```
alextmagro: I think we should be able to support MI300s for this, right? If so, > (9, 0) should be correct.

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
```diff
@@ -130,10 +160,14 @@ __global__ void __launch_bounds__(THREADS_PER_BLOCK)
     }
   }
   // Reduce amax in the warp (32x32 tile)
+#ifdef __HIP_PLATFORM_AMD__
+  warp_tile_amax = blockwise_warp_reduce_max(amax);
+#else
   warp_tile_amax = warp_reduce_max<kThreadsPerWarp>(amax);
```
alextmagro, line 147: We should be able to just use warp_reduce_max here, and remove the kThreadsPerWarp=64 logic too. For the most part, the compiler will double up and we will be okay here.

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
```diff
@@ -456,6 +495,7 @@ __global__ void __launch_bounds__(THREADS_PER_BLOCK) block_scaled_cast_transpose
   }
 }
 
+#ifdef TMA_HW_SUPPORTED
```
alextmagro: I would prefer if we continued to use the AMD guard rather than TMA_HW_SUPPORTED here. Additionally, I am not sure we need to guard the def block around TMA_HW_SUPPORTED as all the cuda arch macros should be undefined.

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
```diff
@@ -561,7 +603,11 @@ void quantize_transpose_square_blockwise(const SimpleTensor& input, SimpleTensor
                         reinterpret_cast<float*>(scale_inv_t.dptr), row_length, num_rows,
                         scale_stride_x, scale_stride_y, scale_t_stride_x, scale_t_stride_y, epsilon,
                         tensor_map_output_trans, pow_2_scale, noop_ptr);
-              } else {
+              } else
```
alextmagro: Let's avoid splitting up the } else { line here. We can add another macro guard instead if needed.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
```diff
@@ -145,14 +155,23 @@ Step 3 (if columnwise transpose is False, COMPACT format): Skip Transpose, cast
...
+#endif
 
 // Hyperparameters for performance tuning
 constexpr int kTileDim = 128;  // Fixed to 128 beacause we are using 1x128 and 128x1 quantization
 constexpr int kNVecIn = 8;     // The number of elements each LDG touches
 constexpr int kNVecOut = 16;   // The number of elements each STG touches
 constexpr int kNVecSMem = 2;   // The number of elements each LDS/STS touches
+
+#ifdef __HIP_PLATFORM_AMD__
+constexpr int kThreadsPerBlock = 512;  // Thread block size, 8 warps (wave64) in total
```
alextmagro, line 163: Are there actual performance improvements for increasing the # of threads and the threads per warp? If not, we should use the already present values for now.

transformer_engine/pytorch/quantization.py
```diff
@@ -103,7 +103,7 @@ def check_nvfp4_support() -> Tuple[bool, str]:
 def check_fp8_block_scaling_support() -> Tuple[bool, str]:
     """Return if fp8 block scaling support is available"""
     if IS_HIP_EXTENSION:
-        return False, "FP8 block scaled gemm not yet supported for ROCm"
```
alextmagro, line 106: We need the same arch guard here as throughout the rest of the PR

alextmagro (changes requested):

alextmagro:
By the way, to run CI you need to add a CI level label. L3 is required before merging, L1 is for lighter testing, mostly sGPU tests, if you are midway through the ticket and expect to make more changes

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
asdfvg123: The kernel expects 8 waves / block , so I increased the number of threads

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: `warp_reduce_max` in `transformer_engine/common/utils.cuh` uses `THREADS_PER_WARP = 32` in the file which creates bug. Let me know if there is a better way

transformer_engine/common/common.h
asdfvg123: `quantize_transpose_square_blockwise.cu` and `quantize_transpose_vector_blockwise.cu` use
`compute_scale_from_types<IType, fp8e4m3>` for the first time, which exposed a latent bug in `common.h`

The `#else` branch of `TypeExtrema<fp8e4m3>` declared `max` as a `static float`, 
This caused the `constexpr static float max_finite_value` initializer in `TypeInfo` in the same file to fail when the template was instantiated on the host.

The fix uses `HIP_FP8_TYPE_FNUZ`, used in `hip_float8.h` for selecting FNUZ at compile time, to make the host-pass branch `constexpr` as well.

tests/cpp/operator/test_cast_float8blockwise.cu
ipanfilo, line 1: Copyright

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
```diff
@@ -18,17 +20,31 @@
...
 #define TMA_HW_SUPPORTED
 #endif
+#endif
 
 namespace transformer_engine {
 namespace {
 
+#ifdef __HIP_PLATFORM_AMD__
+using WarpSyncMask = uint64_t;
+constexpr WarpSyncMask kFullWarpMask = 0xFFFFFFFFFFFFFFFFULL;
```
ipanfilo: ROCm should not use it. See how *_sync calls are guarded in other places

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
```diff
@@ -130,10 +160,14 @@ __global__ void __launch_bounds__(THREADS_PER_BLOCK)
     }
   }
   // Reduce amax in the warp (32x32 tile)
+#ifdef __HIP_PLATFORM_AMD__
```
ipanfilo: The whole this code is under #ifndef __HIP_PLATFORM_AMD__

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
```diff
@@ -18,17 +20,31 @@
...
+constexpr WarpSyncMask kFullWarpMask = 0xFFFFFFFFFFFFFFFFULL;
+#else
+using WarpSyncMask = unsigned;
+constexpr WarpSyncMask kFullWarpMask = 0xFFFFFFFFu;
+#endif
+
 // const values configuration
 
+#ifdef __HIP_PLATFORM_AMD__
+constexpr size_t kThreadsPerWarp = 64;
```
ipanfilo, line 37: It is platform dependent.

transformer_engine/common/CMakeLists.txt
```diff
@@ -200,7 +200,8 @@ list(APPEND transformer_engine_cuda_sources
      transpose/cast_transpose_fusion.cu
      transpose/transpose_fusion.cu
      transpose/multi_cast_transpose.cu
-     transpose/quantize_transpose_vector_blockwise.cu #CUDA-only
+     transpose/quantize_transpose_vector_blockwise.cu
+     transpose/quantize_transpose_square_blockwise.cu
```
ipanfilo: It should stay in transformer_engine_cuda_arch_specific_sources

transformer_engine/common/common.h
ipanfilo: If the value really used host size, it should be runtime detected. If it is only for host translation of GPU code (i.e.. results are discarded), you can keep 448, no extra ifdefs is needed

ipanfilo (changes requested):

asdfvg123:
MI300 has 64KB of LDS which makes overflow when loading 128 * 128 FP32 data into LDS. I created a helper and branched the kernel. When loading FP32 data, the kernel loads 128 * 64 chunk of data and iterate to quantize. From the host's view, the kernel quantizes 128 * 128 elements.

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: added

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: fixed

transformer_engine/common/common.h
asdfvg123: reverted to the original upstream. I instead changed the recipe_common.cuh following the other convention in [quantize_transpose_vector_blockwise_fp4.cu L230](https://github.com/ROCm/TransformerEngine/blob/ff0ec480b96648809a7036388aa6e150a5dbaa34/transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu#L230)

ci/pytorch.sh
asdfvg123: fixed

tests/cpp/operator/test_cast_float8blockwise.cu
asdfvg123: fixed

tests/cpp/operator/test_cast_float8blockwise.cu
asdfvg123: fixed

tests/pytorch/test_float8blockwisetensor.py
asdfvg123: fixed, keeping (9,0)

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: Fixed and now using AMD guard. Keep the upstream's TMA_HW_SUPPORTED guard, to guard non-TMA kernel

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: I could not think of a better way than the current code. I guarded and called the kernel.

transformer_engine/pytorch/quantization.py
asdfvg123: fixed

tests/cpp/operator/test_cast_float8blockwise.cu
asdfvg123: added

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: removed the dead branch

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: fixed now guarded with gfx1250 for 32 threads

transformer_engine/common/CMakeLists.txt
asdfvg123: fixed

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: removed the mask and use ROCm __shfl instead of __shfl_sync

tests/pytorch/test_float8blockwisetensor.py
alextmagro: I believe MI250 is (9,0) so this should be a > rather than a >=, or (9,4)

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
```diff
@@ -357,10 +378,18 @@ __global__ void __launch_bounds__(THREADS_PER_BLOCK) block_scaled_cast_transpose
     }
   }
   // Reduce amax in the warp (32x32 tile)
+#ifdef __HIP_PLATFORM_AMD__
+  warp_tile_amax = warp_reduce_max_64(amax);
```
alextmagro: This is platform dependent. Can we just manually unroll the function inline here, and guard the last shuffle for 64 thread platforms?

Also, instead of a reduce-broadcast we can do an all reduce across threads using __shfl_xor?

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
```diff
@@ -543,9 +574,10 @@ void quantize_transpose_square_blockwise(const SimpleTensor& input, SimpleTensor
               return_transpose, kReturnTranspose,
 
               dim3 grid(num_blocks_x, num_blocks_y, 1);
+
+#ifndef __HIP_PLATFORM_AMD__
               const bool full_tile =
                   row_length % BLOCK_TILE_DIM == 0 && num_rows % BLOCK_TILE_DIM == 0;
-
```
alextmagro, line 548: unneeded whitespace change

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
alextmagro: This would be platform specific I think -- probably need the gfx1250 guard here too

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
```diff
@@ -166,6 +183,78 @@ static_assert(kNumThreadsLoad <= kThreadsPerWarp, "kNumThreadsLoad must be <= kT
 static_assert(kNumThreadsStore <= kThreadsPerWarp, "kNumThreadsStore must be <= kThreadsPerWarp");
 constexpr int kNumWarps = kThreadsPerBlock / kThreadsPerWarp;
 
+// gfx942 (MI300) has 64KB LDS; the full 128x128 fp32 staging tile overflows it. 
+#if defined(__HIP_PLATFORM_AMD__) && !defined(__gfx950__)
```
alextmagro: probably should specifically check for gfx942 here and elsewhere instead of !defined(gfx950)

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
alextmagro: Can we use warpSize from hipruntime here, since kThreadsPerWarp is only needed for device code?

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
```diff
@@ -419,6 +663,90 @@ __global__ void __launch_bounds__(kThreadsPerBlock) block_scaled_1d_cast_transpo
...
+    constexpr int num_smem_reads = kNVecOut / kNVecSMem;
+    constexpr int warps_per_chunk = kNumWarps / kNumChunks;
+    const int thr_idx_in_warp = threadIdx.x % kThreadsPerWarp;
+    const int warp_idx = threadIdx.x / kThreadsPerWarp;
+    const int warp_in_chunk = warp_idx % warps_per_chunk;
+    const int r_s = thr_idx_in_warp * kThreadTileRow;
+    const int c_s = warp_in_chunk * num_smem_reads;
+    size_t r_g = static_cast<size_t>(blockIdx.y) * kTileDim + r_s;
+    for (int chunk = 0; chunk < kNumChunks; ++chunk) {
+      __syncthreads();
```
alextmagro: We can probably skip the syncthreads for the first iteration, also a pragma unroll might help here.

transformer_engine/pytorch/quantization.py
alextmagro: note here we are still including gfx90a which is MI250

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
alextmagro: I think this kernel would work with 4 waves/block as well, right? That might give better performance given the waves are twice as large as upstream.

transformer_engine/common/recipe/recipe_common.cuh
ipanfilo, line 1: Copyright

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
ipanfilo: I think warpSize is not constexpr anymore. Or it is ?

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
```diff
@@ -83,7 +73,7 @@ constexpr size_t NUM_THREADS_Y_IN_WARP = kThreadsPerWarp / NUM_THREADS_X_IN_WARP
 #define MIN(a, b) (a < b ? a : b)
 
 #ifdef __HIP_PLATFORM_AMD__
-__device__ __forceinline__ float blockwise_warp_reduce_max(float val) {
+__device__ __forceinline__ float warp_reduce_max_64(float val) {
```
ipanfilo: why 64?

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
```diff
@@ -241,6 +333,89 @@ __global__ void __launch_bounds__(kThreadsPerBlock) block_scaled_1d_cast_transpo
...
+      CType amax = 0;
+#pragma unroll
+      for (int j = 0; j < kNVecOut; ++j) {
+        __builtin_assume(amax >= 0);
+        amax = fmaxf(amax, fabsf(in_vec.data.elt[j]));
+      }
+      // Step 2.3: Reduce amax
+#pragma unroll
+      for (int delta = kNumThreadsStore / 2; delta > 0; delta /= 2) {
+        const float other_amax = __shfl_down_sync(mask, amax, delta);
```
ipanfilo: Use __shfl_down on ROCm

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
```diff
@@ -26,6 +28,12 @@ namespace {
 using transformer_engine::detail::FP8BlockwiseColumnwiseOption;
 using transformer_engine::detail::FP8BlockwiseRowwiseOption;
 
+#ifdef __HIP_PLATFORM_AMD__
+using WarpSyncMask = uint64_t;
```
ipanfilo: Review where it is used. Wavefront level primitives on ROcm should not use mask

transformer_engine/pytorch/quantization.py
```diff
@@ -103,7 +103,10 @@ def check_nvfp4_support() -> Tuple[bool, str]:
...
+        gpu_arch = get_device_compute_capability()
+        if gpu_arch >= (9, 0):
```
ipanfilo: FP8 starts from 9.4

transformer_engine/common/recipe/recipe_common.cuh
asdfvg123: added

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: fixed. I changed to use _xor

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
asdfvg123: fixed and added

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: is now removed.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
asdfvg123: I removed all *__sync from both AMD only path and AMD & Nvidia common path. I added guard and use non-sync in the AMD path.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
asdfvg123: Removed the unnecessary mask definitions together with using only non-sync in the AMD path. Reverted to the upstream.

transformer_engine/pytorch/quantization.py
asdfvg123: fixed

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: fixed

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
asdfvg123: added gfx1250 guard. 
In line 779, there is a `static_assert(num_iterations == 1` which requires 8 waves / block for gfx942 and gfx950. This is necessary and 4 waves/block fails to compile.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
asdfvg123: fixed. 
`#if defined(__HIP_PLATFORM_AMD__) && (!defined(__HIP_DEVICE_COMPILE__) || defined(__gfx942__))`
the code block has to be compiled in both host pass and gfx942 device pass and also should NOT be compiled in the Nvidia pass. The macro guards and is only true if AMD host or gfx942.

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: I assume you are referring to 
```
inline __device__ const struct {
  __device__ __attribute__((always_inline, const)) operator int() const noexcept {
    return __builtin_amdgcn_wavefrontsize();
  }
} warpSize{};
```

in amd_warp_functions.h
and this is not constexpr (assigned in the runtime) so cannot used.

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
```diff
@@ -28,7 +30,11 @@ namespace {
 
 // const values configuration
 
+#if defined(__HIP_PLATFORM_AMD__) && !defined(__gfx1250__)
+constexpr size_t kThreadsPerWarp = 64;
```
ipanfilo, line 37: is kThreadsPerWarp only used by device code and not any dispatch functions?

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
```diff
@@ -357,10 +369,18 @@ __global__ void __launch_bounds__(THREADS_PER_BLOCK) block_scaled_cast_transpose
     }
   }
   // Reduce amax in the warp (32x32 tile)
+#ifdef __HIP_PLATFORM_AMD__
+#pragma unroll
+  for (int delta = kThreadsPerWarp / 2; delta > 0; delta /= 2) {
```
ipanfilo: Can you please clarify this logic with using xor?

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: It is used to compute NUM_THREADS_Y_IN_WARP in L71 for constexpr computation. And other than this, it is only used in the device code.

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: The purpose is to share the amax value across the lanes in a wave.
`shfl_xor(val, mask, width)` computes: lane i gives val to lane `i XOR mask` and returns lane `i XOR mask` val.
So in the first iteration (delta = 64 / 2 = 32) lane 0 exchanges amax with lane 32 and lane 1 exchanges amax with lane 33 so on.
in the next iteration (delta = 16) lane 0 exchanges the amax (accumulated amax from the previous step) with lane 16 and lane 1 exchanges the amax with lane 17 and so on.
At the end of the loop, all lanes compute the amax across the wave. So no separate broadcast is needed.

alextmagro: For reference: using xor for warp level all reduce is faster than shuffling down + broadcast since we skip the broadcast instructions, so is generally best practice.

Upstream has subwarp_reduce_max_broadcast and warp_reduce_max_broadcast using the slower reduce + broadcast implementation, so I will put in a PR to fix those up. We can then use those directly here?

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
alextmagro: Looking more closely, could we remove this syncthreads completely, and then do a double buffer for load_chunk_to_smem?

asdfvg123: Unfortunately, we cannot double the chunk size. It does not fit in the MI300 LDS. 
The chunk buffer size is

`kSMemSizeChunk = 128 * 33 * 2 = 8448 elements * 4B = 33KB (fp32)`

The 33 columns is to avoid the bank conflict. 

If we double buffer the chunk, it is larger than the LDS size (64KB). We need more experiments before the optimization. I would suggest doing it as a future PR.

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: I agree. For now I would like to keep the current XOR method (guarded for AMD), and once the upstream change is merged we can switch.

tests/pytorch/test_float8blockwisetensor.py
asdfvg123: fixed

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
ipanfilo: __gfz1250__ is device code define so it cannot be used host side.
You may refer transformer_engine/common/fused_attn_rocm/fused_attn.cpp to see how it solved there

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
```diff
@@ -166,6 +177,78 @@ static_assert(kNumThreadsLoad <= kThreadsPerWarp, "kNumThreadsLoad must be <= kT
...
+constexpr size_t kLdsLimitBytes = 64 * 1024;
+
+template <typename IType>
+constexpr bool use_chunked_lds() {
+  return sizeof(IType) * static_cast<size_t>(kSMemSize) > kLdsLimitBytes;
+}
+
+template <typename IType>
+size_t host_smem_bytes() {
+  if (transformer_engine::cuda::sm_arch() < 95 &&
```
ipanfilo: Should it be 94 explicitly?

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
```diff
@@ -196,6 +279,9 @@ __global__ void __launch_bounds__(kThreadsPerBlock) block_scaled_1d_cast_transpo
   SMemVec* smem = reinterpret_cast<SMemVec*>(&smem_base[0]);
 
   // Step 1: Load input to shared memory
+#if defined(__HIP_PLATFORM_AMD__) && (!defined(__HIP_DEVICE_COMPILE__) || defined(__gfx942__))
```
ipanfilo: nit: In device functions we don;t need to worry about !defined(__HIP_DEVICE_COMPILE__) condition so the ifdef here and in the rest of file can be simplified. Or if you prefer to keep the same ifdef condition across the file for easyier identification of changes, may be create some synthetic define like CHUNKED_LDS and use it

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
```diff
@@ -617,7 +967,11 @@ void quantize_transpose_vector_blockwise(const SimpleTensor& input, SimpleTensor
           TRANSFORMER_ENGINE_SWITCH_CONDITION(
               full_tile, kAligned,
 
+#if defined(__HIP_PLATFORM_AMD__) && (!defined(__HIP_DEVICE_COMPILE__) || defined(__gfx942__))
```
ipanfilo: if you go with conditions simplification this probably can be just __HIP_PLATFORM_AMD

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
alextmagro: Sure. For reference, sometimes padding to 34 is useful depending on dtype and instruction.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
asdfvg123: fixed

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: `kThreadsPerWarp` is used in the device code and is not used in any other host functions. 
However, I found the problem is that THREADS_PER_BLOCK is used in `__launch_bounds__` and also used in host kernel launch, which incurs bug. I fixed the `THREADS_PER_BLOCK` usage in the host dispatch. (L599)
```
              const size_t threads_per_block =
                  (transformer_engine::cuda::sm_arch(transformer_engine::cuda::current_device()) == 125) ? 256 : 512;
```
Now it checks the architecture in the runtime and launches the kernel

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
asdfvg123: Fixed to only use `#if defined(__gfx942__)` in the device. 
The define `CHUNKED_LDS` that you proposed is also ambiguous. gfx942 not always use the chunked-LDS helpers for the quantization but only uses when the input load to LDS is greater than the LDS size. This is determined in the runtime. Therefore, I decided to compile the chunked-LDS helpers for all AMD platforms but it is used when `sm_arch() == 94`.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
asdfvg123: fixed

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: Please also note that when I was fixing the threads_per_block problem for gfx1250, I also found there is the same problem in the `quantize_transpose_vector_blockwise.cu`, and fixed. Please refer to L972.

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
asdfvg123: @alextmagro now using your `rocm_subwarp_allreduce`.

ipanfilo (commented):
Review test failures

asdfvg123:
Add a temporal guard for the blockwise FP8 GEMM and skip the tests which expects the GEMM and only run the tests with the quantization. The temporal guard will be dropped in the following PR, integrating GEMM.

asdfvg123:
I tested the failures locally, and are passing. These failures are not related to my changes.

ipanfilo (approved):

alextmagro (approved):

matthiasdiener (approved):

PR #611, opened 2026-06-04
rocm-ci: scope test container to pod-allocated GPUs

okakarpa:
The sGPU/mGPU jobs launched the test container with '--device=/dev/dri --device=/dev/kfd', exposing ALL host GPUs to the nested (privileged-dind) container. Combined with the hard-coded absolute HIP_VISIBLE_DEVICES=0..3, two jobs co-scheduled on the same node both pinned physical GPUs 0-3 and collided (OOM/hangs/test failures) while GPUs 4-7 sat idle. Jobs only passed when the node was otherwise idle -- arch-independent (seen on mi300x and mi35x).

Forward only the GPUs the AMD device plugin allocated to this pod: it restricts /dev/dri/renderD*+card* in the runner container to the pod's GPUs, so enumerate those and pass only them. /dev/kfd (single system-wide compute node) is always passed. Fail fast if zero GPUs are allocated.

The container now sees only its allocated GPUs as 0..N-1, so the existing per-suite HIP_VISIBLE_DEVICES=0/1/2/3 split is correct and collision-free across co-scheduled pods.

### DIFF
--- a/.github/workflows/rocm-ci.yml
+++ b/.github/workflows/rocm-ci.yml
@@ -144,15 +144,28 @@ jobs:
 
       - name: Run Container
         run: |
+          # All GPUs are visible to the runner; per-suite visibility is set later
+          # via HIP_VISIBLE_DEVICES in the test step. Add render group for the
+          # container. Under kubernetes, GPU isolation comes from DEVICE_FLAG:
+          # the runner writes this pod's allocated render devices into
+          # /etc/podinfo/gha-render-devices; fall back to all GPUs on bare metal.
+          render_gid=$(cat /etc/group | grep render | cut -d: -f3)
+          if [ -f "/etc/podinfo/gha-render-devices" ]; then
+            DEVICE_FLAG=$(cat /etc/podinfo/gha-render-devices)
+          else
+            DEVICE_FLAG="--device /dev/dri"
+          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/rocm-ci.yml
```diff
@@ -144,15 +144,28 @@ jobs:
...
+          render_gid=$(cat /etc/group | grep render | cut -d: -f3)
+          if [ -f "/etc/podinfo/gha-render-devices" ]; then
+            DEVICE_FLAG=$(cat /etc/podinfo/gha-render-devices)
+          else
+            DEVICE_FLAG="--device /dev/dri"
+          fi
+          # --group-add daemon/bin cover the video-group GID -> subgid 1 mapping
+          # across Ubuntu 24.04 / AlmaLinux base images. /dev/kfd is the single
+          # system-wide compute node and is always required.
+          GPU_FLAG="--device=/dev/mem --device=/dev/kfd $DEVICE_FLAG --group-add video --group-add $render_gid --group-add daemon --group-add bin --cap-add=SYS_PTRACE --security-opt seccomp=unconfined"
```
ipanfilo: Because GID is not guaranteed to be consistent between host and container, please avoid using group names at least for video. And better used getent vs 'cat /etc/group'

sheralskumar (approved):

ipanfilo (approved):

PR #615, opened 2026-06-08
[Fix] TE RMSNorm Triton Kernel Optimization

AllenFarcas:
# Description

Optimize RMSNorm triton kernel for Qwen3 shapes.

Fixes # https://github.com/ROCm/frameworks-internal/issues/16614

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Pack multiple rows per program (ROWS_PER_PID, up to 16 for H≤128) instead of one program per row — slashes program-launch count for narrow-H shapes.
- Apply the same packing to the backward kernel (capped at 4 for H≤128), targeting the Qwen3 32768×128 QK-norm gradient pass.
- Eliminate redundant gamma reloads and launch overhead on narrow rows, the root cause of the 0.1× cliff on Qwen3 QK shapes.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/triton_kernels/norms_common.py
+++ b/transformer_engine/pytorch/triton_kernels/norms_common.py
@@ -157,6 +157,8 @@ def _rows_per_pid(rows, hidden, num_sms, cap_ladder):
             rpp = 4
         elif hidden <= 512 and rows % 2 == 0:
             rpp = 2
+    # The kernel packs rpp rows per program with no row-tail mask, so rpp must divide rows
+    assert rows % rpp == 0, f"rows_per_pid={rpp} does not divide rows={rows}"
     return rpp
 
 
--- a/transformer_engine/pytorch/triton_kernels/norms_common.py
+++ b/transformer_engine/pytorch/triton_kernels/norms_common.py
@@ -124,9 +124,9 @@ _rmsnorm_bwd_dg_reduce_kernels = {
 }
 
 
-_TARGET_PRGMS_PER_CU = 8 
-_MAX_PRGMS_PER_CU = 16 
-_I32_OFFSET_LIMIT = 1 << 31 
+_TARGET_PRGMS_PER_CU = 8
+_MAX_PRGMS_PER_CU = 16
+_I32_OFFSET_LIMIT = 1 << 31
 
 _FWD_FLAT_GRID_MAX_H = 2048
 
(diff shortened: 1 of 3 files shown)


Micky774:
Based on the [shared benchmarks](https://amd.atlassian.net/wiki/spaces/~712020d71119f77f2e48d98822f3c648e08ab9/pages/1701381099/2026-01-01+Issue+16614) it looks like there's actually a regression in the triton kernels across some shapes e.g. 

`Qwen3-235B/hidden, 8192, 4096, bfloat16, fwd: 4211.3 GB/s --> 3573.2 GB/s` similarly for the `bwd` of the same config. Can you run a general benchmark comparing the updated triton implementation to the current one on `dev` so we can see what the tradeoffs are? Maybe we can optimize heuristics a bit.

transformer_engine/pytorch/triton_kernels/utils.py
```diff
@@ -33,13 +33,21 @@ def get_inf_ln_sm_margin():
...
 
 
+# Per-device SM count cache. get_num_sms is called on every RMSNorm dispatch
+# (forward and backward), and get_device_properties is a driver round-trip;
+# the active device's SM count is constant, so cache it.
+_NUM_SMS_CACHE: "dict[int, int]" = {}
+
+
 def get_num_sms(sm_margin=None):
-    num_sms = torch.cuda.get_device_properties(
```
ipanfilo, line 37: Use get_sm_count from transformer_engine.pytorch.utils

ipanfilo (changes requested):

AllenFarcas:
@Micky774 I updated the benchmarks run and the results obtained [here](https://amd.atlassian.net/wiki/x/6wNpZQ) and solved all regressions.

ipanfilo (approved):
review n_rows to ROWS_PER_PID divisibility like suggested by bot

transformer_engine/pytorch/triton_kernels/norms_common.py
```diff
@@ -157,6 +157,8 @@ def _rows_per_pid(rows, hidden, num_sms, cap_ladder):
             rpp = 4
         elif hidden <= 512 and rows % 2 == 0:
             rpp = 2
+    # The kernel packs rpp rows per program with no row-tail mask, so rpp must divide rows
+    assert rows % rpp == 0, f"rows_per_pid={rpp} does not divide rows={rows}"
```
Micky774: Why not add a mask to allow for arbitrary shapes?

Micky774 (approved):
nit: `row_mask` should just be `mask` since otherwise it makes it sound like what `row_valid` actually is.

Otherwise, once CI is green LGTM

PR #621, opened 2026-06-09
Fix hipBLASLt GEMM algo cache misses caused by uninitialized struct padding

sudhu2k:
# Description

The ROCm hipBLASLt autotune cache (GemmAlgoCache) uses a std::multimap keyed by a Key struct. Lookup relies on a byte-wise comparator (Key::Comp) that compares the entire sizeof(Key) representation, including compiler-inserted tail padding bytes.

Key objects were previously constructed without zeroing those padding bytes. Two keys with identical logical fields (m, n, k, dtypes, etc.) could therefore compare as different if their padding contained different stack garbage. This caused:

1. Cache misses on repeated identical GEMM configurations
2. Redundant autotuning for the same problem

This PR zero-initializes the full Key object (including padding) in both constructors via memset before assigning members, making the byte-wise comparator deterministic and consistent with operator==.

Fixes https://github.com/ROCm/frameworks-internal/issues/16827
#16827

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Zero-initialize GemmAlgoCache::Key in both the parameterized and default constructors using std::memset(this, 0, sizeof(*this)) before member assignment# Checklist:

### DIFF
--- a/transformer_engine/common/gemm/rocm_gemm.cu
+++ b/transformer_engine/common/gemm/rocm_gemm.cu
@@ -24,6 +24,7 @@
 #include <cstdlib>
 #include <string>
 #include <cstdint>
+#include <cstring>
 
 #include "../common.h"
 #include "../util/vectorized_pointwise.h"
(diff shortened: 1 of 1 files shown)


ipanfilo (approved):

wenchenvincent:
@sudhu2k Just curious: why are there padding bytes in the object?

sudhu2k:
> @sudhu2k Just curious: why are there padding bytes in the object?

 The compiler inserts it automatically to satisfy alignment rules.
1. A struct's alignment equals the largest alignment of its members. (4 bytes in this case
2. A struct's total size must be a multiple of that alignment.

In our Key, almost every member is a 4-byte type (int, and the hipDataType/hipblasOperation_t/hipblasLtEpilogue_t enums), so the struct is 4-byte aligned. The members sum to 65 bytes which isn't a multiple of 4, so the compiler appends 3 bytes of tail padding to round the size up to 68.

Calculation:
https://github.com/ROCm/TransformerEngine/blob/6ac51951861731cbc87fa87f029c0e6a623113bf/transformer_engine/common/gemm/rocm_gemm.cu#L686-L695

 4  (deviceCap)
20  (5 × hipDataType)
12  (m, n, k)
12  (lda, ldb, ldd)
 8  (transa, transb)
 4  (scaling_mode)
 4  (epilogue)
 1  (fp4_alpha_device_vector)
_
65  bytes of real data

PR #628, opened 2026-06-16
Enable MultiCastTranspose for expert weights

sudhu2k:
# Description

This PR optimizes FP8 weight quantization for `GroupedLinear` by replacing the per-weight quantize loop with a single fused `multi_cast_transpose` kernel for delayed-scaling FP8 and MXFP8. Previously each weight in a group was cast and transposed with its own `quantize` call; now the whole group is processed in one fused call, reducing kernel-launch overhead and unnecessary tensor reallocations during FP8 operations.

To make this work efficiently with the existing weight-caching mechanism, `multi_tensor_quantize` now accepts an optional `outputs` argument so the fused kernel can write directly into cached workspace buffers in place rather than allocating fresh tensors on every update.

For performance numbers take a look at [#16929](https://github.com/ROCm/frameworks-internal/issues/16929)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Added `TransformerEngineBaseModule.quantize_multi_weight`, a group-aware analog of `quantize_weight` that casts and transposes a whole group of weights with a single fused `multi_tensor_quantize` kernel for delayed-scaling FP8 and MXFP8, with caching support. Falls back to the per-tensor `quantize_weight` path when fusion is not applicable (other recipes, rowwise-only usage, already-quantized weights, or CUDA-graph weight caching).
- Updated `_GroupedLinear` to use `quantize_multi_weight` instead of looping over `quantize_weight` per GEMM.
- Extended the `multi_tensor_quantize` C++ extension (`extensions.h`, `cast.cpp`, `pybind.cpp`) with an optional `outputs` parameter, enabling in-place quantization into pre-allocated/cached workspace buffers and avoiding redundant tensor allocations.
- Refactored the workspace allocation/caching logic to handle cache hits and misses through a unified fused path: a cache miss allocates fresh buffers, while a cache hit with `update_workspace` reuses cached buffers via the new `outputs` argument. Quantizers are temporarily forced to `internal=False` so cached workspaces survive `prepare_for_saving`.

# Checklist:


transformer_engine/pytorch/csrc/extensions.h
```diff
@@ -312,7 +312,8 @@ py::object group_quantize(const at::Tensor &tensor, py::handle quantizer, const
                           std::optional<at::Tensor> first_dims);
 
 std::vector<py::object> multi_tensor_quantize(const std::vector<at::Tensor> &tensor_list,
-                                              std::vector<py::handle> quantizer_list);
+                                              std::vector<py::handle> quantizer_list,
+                                              const py::object &outputs = py::none());
```
ipanfilo, line 326: Add # ROCm: comment so it is marked to keep on further IFUs

ipanfilo (approved):
Need CI results review before merging

PR #630, opened 2026-06-16
gfx1250 mxfp8 gemm: add NN/NT transpose workaround

matthiasdiener:
# Description

**NN -> TN conversion (DGRAD):**
The desired operation is $D = A \times B$ (both non-transposed). Since `CanonicalizeGemmInput` selects A's columnwise data for the non-transposed case, A is already stored in transposed physical layout. `make_mxfp8_rowwise_from_columnwise` transposes A's FP8 data bytes (`launch_transpose_u8`) and converts its columnwise MXFP8 scales to rowwise layout (`launch_mxfp8_colwise_scale_to_rowwise`), producing $A'$ where $A' = A^T$ in rowwise format. The GEMM then computes $D = A'^T \times B$ as a TN call, which equals $D = A \times B$, the original NN result.

**NT -> TN conversion (WGRAD):**
The desired operation is $D = A \times B^T$. Both A and B are transposed via `make_mxfp8_rowwise_from_columnwise` to produce rowwise $A'$ and $B'$. The identity $D = A \times B^T \implies D^T = B \times A^T = B'^T \times A'$ is exploited: the operands are swapped and hipBLASLt computes $D^T = B'^T \times A'$ as a TN call. A final `launch_transpose_output` kernel transposes the $m \times n$ result $D^T$ back to yield $D$.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1707,6 +1870,82 @@ void release_service_stream(hipStream_t stream, struct ServiceStreamCtl &ctl)
...
+
+  if (!((transa == CUBLAS_OP_N && transb == CUBLAS_OP_N) ||
+        (transa == CUBLAS_OP_N && transb == CUBLAS_OP_T))) {
+    return false;
+  }
+  if (beta != 0.0f || inputBias->data.dptr != nullptr || outputPreGelu->data.dptr != nullptr) {
+    return false;
+  }
+
+  const bool output_swizzled = cuda::sm_arch() == 125;
```
ipanfilo: all other arch are filtered out at line 1885

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1707,6 +1870,82 @@ void release_service_stream(hipStream_t stream, struct ServiceStreamCtl &ctl)
...
+    if (transb == CUBLAS_OP_T) {
+      launch_transpose_output(outputD->data.dtype, D_tn.data.dptr, outputD->data.dptr, m, n, stream);
+    }
+    launched = true;
+  } catch (...) {
+    free_async_temps(temp_buffers, stream);
+    throw;
+  }
+  free_async_temps(temp_buffers, stream);
+  return launched;
```
ipanfilo, line 1905: It looks like it can only be true here

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -199,6 +199,169 @@ struct GemmParam {
...
+  for (void* ptr : buffers) {
+    if (ptr != nullptr) {
+      NVTE_CHECK_CUDA(hipFreeAsync(ptr, stream));
+    }
+  }
+}
+
+Tensor make_mxfp8_rowwise_from_columnwise(const Tensor& input, std::vector<void*>& buffers,
+                                          const bool output_swizzled, hipStream_t stream) {
+  NVTE_CHECK(input.has_columnwise_data(), "MXFP8 transpose-to-TN requires column-wise data.");
```
ipanfilo, line 276: Should it only be called when no rowwise data exist?

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1775,9 +2015,16 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
     handle = hipblaslt_handles[compute_stream_offset];
   }
 
-  hipblaslt_gemm(inputA, inputB, outputD, inputBias, outputPreGelu, m, n, k, lda, ldb, ldd, transa,
-                 transb, grad, workspace, workspaceSize, alpha, beta, use_split_accumulator,
-                 math_sm_count, use_service_stream ? ss_ctl.stream : stream, handle);
+  bool handled = try_mxfp8_non_tn_transpose_to_tn(inputA, inputB, outputD, inputBias,
```
ipanfilo, line 1979: Should hipblaslt_gemm be tried first and only if transposing combination is not supported and cannot be handled by Canonicalization, then try new w/a?

ipanfilo (changes requested):

transformer_engine/common/gemm/rocm_gemm.cu
matthiasdiener: Removed output_swizzled parameter here and from make_mxfp8_rowwise_from_columnwise in bd630823. Left it in launch_mxfp8_colwise_scale_to_rowwise generic  in case it's reused from other contexts in the future.

transformer_engine/common/gemm/rocm_gemm.cu
matthiasdiener: Added a comment in bd630823. Even if rowwise data exists, it can't substitute for the transpose. MXFP8 rowwise and columnwise scales cover different dimensions (K vs M), so the non-transposed operand must use columnwise data. The existing NVTE_CHECK(input.has_columnwise_data()) in make_mxfp8_rowwise_from_columnwise validates this at runtime.

transformer_engine/common/gemm/rocm_gemm.cu
matthiasdiener: Addressed in bd630823, see above response.

transformer_engine/common/gemm/rocm_gemm.cu
matthiasdiener: Not sure if it is worth it; there'd be a performance cost for every NN/NT call. I left it intact for now, but added a FIXME comment in bd630823

transformer_engine/common/gemm/rocm_gemm.cu
ipanfilo: Scales are nit interchangeable but isn't layout of rowwise_data the same? I.e. should only scales be redone if rowwise data is there?

matthiasdiener: From what I understand, we need to transpose both data and scales.
MXFP8 rowwise and columnwise data have different FP8 byte values because they use different block groupings (blocks of 32, but along different dimensions). The data and scales must be a matched pair. If we used existing rowwise_data with transformed columnwise scales, the scales wouldn't correspond to the actual FP8 values in the data. The function transposes both columnwise data and columnwise scales together to produce a consistent pair.

ipanfilo (approved):

PR #632, opened 2026-06-17
Introduce a fused padding + cast transpose kernel grouped linear

alextmagro:
Fuses the following 2 kernels:

Pad tensors in BF16
Cast from BF16 to FP8, transpose, and store

into a single kernel that:
 
Cast from BF16 to FP8, transpose, store with padding

2x speedup over unfused kernel, applicable to grouped linear.

### DIFF
--- a/tests/cpp/operator/test_multi_cast_transpose.cu
+++ b/tests/cpp/operator/test_multi_cast_transpose.cu
@@ -105,18 +105,12 @@ void performTest() {
   }
 
   // Transformer Engine implementation
-  auto make_nvte_vector = [](std::vector<Tensor>& tensor_list)
-    -> std::vector<NVTETensor> {
-    std::vector<NVTETensor> nvte_tensor_list;
-    for (auto& tensor : tensor_list) {
-      nvte_tensor_list.emplace_back(tensor.data());
-    }
-    return nvte_tensor_list;
-  };
-  nvte_multi_cast_transpose(num_tensors,
-                            make_nvte_vector(input_list).data(),
-                            make_nvte_vector(output_list).data(),
-                            0);
+  std::vector<NVTETensor> nvte_inputs(num_tensors), nvte_outputs(num_tensors);
+  for (size_t i = 0; i < num_tensors; i++) {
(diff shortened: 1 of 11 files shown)


tests/cpp/operator/test_multi_cast_transpose.cu
```diff
@@ -105,18 +105,12 @@ void performTest() {
...
-    for (auto& tensor : tensor_list) {
-      nvte_tensor_list.emplace_back(tensor.data());
-    }
-    return nvte_tensor_list;
-  };
-  nvte_multi_cast_transpose(num_tensors,
-                            make_nvte_vector(input_list).data(),
-                            make_nvte_vector(output_list).data(),
-                            0);
+  std::vector<NVTETensor> nvte_inputs(num_tensors), nvte_outputs(num_tensors);
```
ipanfilo: nit: guard it or commet it with ROCm: ..... so it will be clear on further IFU

alextmagro: Sorry, I actually mistakenly thought this was something that claude added. (it loves lambdas unless you keep discouraging it)

Since it is from upstream I will revert this change.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -237,6 +240,17 @@ def forward(
...
+
+        output_unpadded = False
+        if unpad_output and actual_m_splits is not None and IS_HIP_EXTENSION and actual_m_splits != m_splits:
+            out_unpadded = torch.empty(
+                [sum(actual_m_splits), out.shape[-1]],
+                dtype=out.dtype, device=out.device,
+            )
+            tex.fused_multi_row_unpadding(out, out_unpadded, m_splits, actual_m_splits)
+            out = out_unpadded
+            output_unpadded = True
```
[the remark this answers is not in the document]
alextmagro: Not valid concern as MoE inputs are always 2D

alextmagro (commented):
Addressed copilot comments

ipanfilo (approved):
Pending CI results

PR #634, opened 2026-06-18
[ROCm] Fix biased wgrad with fp32 gradient accumulation

XinyuJiangCMU:
## Problem

On ROCm, hipBLASLt cannot find a suitable algorithm for an fp32 weight gradient GEMM with fused bias gradient computation.

This causes training with --add-qkv-bias and --accumulate-allreduce-grads-in-fp32 to fail with:

```text
RuntimeError: Unable to find any suitable algorithms
```

## Fix

Run the weight gradient GEMM without the fused bias gradient and compute the bias gradient separately by summing grad_output.

The fix is implemented in general_gemm, covering delayed weight gradient execution and other callers using the same path. CUDA behavior is unchanged.

## Testing

Verified on MI350X:

- The isolated reproduction passes.
- Qwen2.5-0.5B GSM8K training passes the previously failing backward step.
- Re-enabled the ROCm wgrad numerics tests previously skipped by grouped GEMM change 434:
  - test_linear_accuracy_delay_wgrad_compute
  - test_layernorm_linear_accuracy_delay_wgrad_compute
  - test_layernorm_mlp_accuracy_delay_wgrad_compute

  All three use general_gemm.

Result:

```text
132 passed, 0 skipped, 0 failed
```

### DIFF
--- a/tests/pytorch/test_numerics.py
+++ b/tests/pytorch/test_numerics.py
@@ -1410,10 +1410,6 @@ def test_linear_accuracy_delay_wgrad_compute(dtype, bs, model, bias, fuse_wgrad_
 
     config = model_configs[model]
 
-    if IS_HIP_EXTENSION:
-        if dtype not in (torch.float32,) and fuse_wgrad_accumulation and bias:
-            pytest.skip(f"ROCm does not support fused wgrad accumulation for {dtype}.")
-
     te_linear_ref = Linear(
         config.hidden_size,
         4 * config.hidden_size,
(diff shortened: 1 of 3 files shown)


ipanfilo:
There are number of numerical errors on MI300: https://github.com/ROCm/TransformerEngine/actions/runs/27776108682/attempts/5. Rerun to see if they are persistent with those changes

ipanfilo (changes requested):
The change cause pytorch/test_numeric regressions

wenchenvincent:
@XinyuJiangCMU Could you also get the hipblasLt configs and file a feature request to support those configs?

PR #648, opened 2026-06-23
Add DeepSeek shapes to C++ benchmarks

alextmagro:
Adds DeepSeek V3, V4 shapes to benchmark scripts

Adds a EP=8 option to benchmarks to crudely simulate mGPU workloads

### DIFF
--- a/benchmarks/cpp/cast/bench_casttranspose.cpp
+++ b/benchmarks/cpp/cast/bench_casttranspose.cpp
@@ -44,7 +44,7 @@ using fp8_e4m3 = test::fp8e4m3;
   ->Args({496, 5760})              \
   ->Args({1792, 5760})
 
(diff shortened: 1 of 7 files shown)


benchmarks/cpp/cast/bench_casttranspose.cpp
```diff
@@ -44,7 +44,7 @@ using fp8_e4m3 = test::fp8e4m3;
   ->Args({496, 5760})              \
   ->Args({1792, 5760})
 
-// Tensor shapes from LLaMA (8B, 70B, 405B) and Qwen (7B, 72B)
+// Tensor shapes from LLaMA (8B, 70B, 405B), Qwen (7B, 72B), and DeepSeek-V3/V4
```
ipanfilo, line 47: nit: comment in the list which shapes block is LLama, which Qewn, which DS

ipanfilo (approved):

PR #652, opened 2026-06-26
hipblaslt Fallback hotfix

alextmagro:
Fixes a bug where HipKittens does not fall back to hipBLASlt for unaligned mxfp8 GEMMs.

### DIFF
--- a/transformer_engine/common/gemm/rocm_gemm.cu
+++ b/transformer_engine/common/gemm/rocm_gemm.cu
@@ -1985,7 +1985,7 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
 
 #ifdef USE_HIPKITTENS_GEMM
 
-  bool use_hipkittens = false, hipkittens_gemm_complete = false;
+  bool use_hipkittens = false;
   if (is_mxfp8) {
     bool is_gfx950 = (cuda::sm_arch() == 95);
     bool force_hipblaslt = false;
--- a/transformer_engine/common/gemm/rocm_gemm.cu
+++ b/transformer_engine/common/gemm/rocm_gemm.cu
@@ -1985,7 +1985,7 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
 
 #ifdef USE_HIPKITTENS_GEMM
 
-  bool use_hipkittens = false;
+  bool use_hipkittens = false, hipkittens_gemm_complete = false;
   if (is_mxfp8) {
     bool is_gfx950 = (cuda::sm_arch() == 95);
     bool force_hipblaslt = false;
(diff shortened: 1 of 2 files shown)


transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -2001,18 +2001,19 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
...
-                       static_cast<int>(param.Atype),
-                       static_cast<int>(param.Btype),
-                       inputBias->data.dptr,
-                       static_cast<int>(inputBias->data.dtype),
-                       outputPreGelu->data.dptr,
-                       static_cast<int>(outputD->data.dtype),
-                       static_cast<int>(outputPreGelu->data.dtype),
-                       workspace, workspaceSize, s);
-  } else {
+    hipkittens_gemm_complete = kittens_mxfp8_gemm(param.A, param.B, outputD->data.dptr,
```
wenchenvincent: So `kittens_mxfp8_gemm` will decide whether it would be able deal with this GEMM config and return False if it cannot?

alextmagro: That's right. This was the original intended behavior, but things got mangled right before I merged.

wenchenvincent (approved):

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -2001,18 +2001,19 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
...
+                                                  static_cast<int>(param.Atype),
+                                                  static_cast<int>(param.Btype),
+                                                  inputBias->data.dptr,
+                                                  static_cast<int>(inputBias->data.dtype),
+                                                  outputPreGelu->data.dptr,
+                                                  static_cast<int>(outputD->data.dtype),
+                                                  static_cast<int>(outputPreGelu->data.dtype),
+                                                  workspace, workspaceSize, s);
+  }
+  if (!use_hipkittens || !hipkittens_gemm_complete) {
```
ipanfilo: nit: no need to check use_hipkittens - hipkittens_gemm_complete can only be True if use_hipkittens is True. Only use_hipkittens can be sufficient (nohipkittens_gemm_complete) , if assign it at line 2004

ipanfilo (commented):
looks good, just minor note. it is already approved

transformer_engine/common/gemm/rocm_gemm.cu
alextmagro: I have simplified the check as you mentioned.

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1985,7 +1985,7 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
 
 #ifdef USE_HIPKITTENS_GEMM
 
-  bool use_hipkittens = false;
+  bool use_hipkittens = false, hipkittens_gemm_complete = false;
```
ipanfilo: This is unused variable

alextmagro: Sorry, missed that. Should be fixed now.

ipanfilo (approved):

PR #566, opened 2026-04-28
HipKittens MXFP8 GEMM Support

alextmagro:
Creates an MXFP8 GEMM with HipKittens that outperforms hipBLASlt, and offers additional epilogues such as BIAS and GELU AUX

Requires a workspace sized relative to the model. Often larger than hipBLASlt, but with significant performance improvements. Only builds for gfx950, and requires M / 256 and N / 256. 

Adds hipKittens header library as a submodule.

PR #658, opened 2026-07-06
blockwise fp8 gemm integration for gfx942 and gfx950

asdfvg123:
# Description

Integrating blockwise FP8 gemm for gfx942 and gfx950 using hipkittens. Using 2 submodules for now. The hipkittens submodule uses separate branch for CDNA3 and CDNA4.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
separate the kernel for gfx942 and gfx950. The host selects the kernel in runtime.
The kernel supports 1d2d, 1d1d GEMM with output bf16, fp32, fp16 / input fp8e5m2, fp8e4m3 / and different epilogues.

# Checklist:


asdfvg123:
1. check_fp8_block_scaling_quantization_support 
These were temporary APIs introduced in PR #609 to enable blockwise FP8 quantization only. Now that GEMM is also supported, they are folded into `is_fp8_block_scaling_available` and removed. This is an intentional public symbol removal with no known downstream users.

2. `constants.py` FP8 dtype lookup fix
`Custom_DType_Dict.__missing__` now compares the enum key with `==` instead of `is`. This is a bug fix on the rocm side. `tex.DType` values returned from the C++ bindings (i.e. `tensor._fp8_dtype`) are fresh pybind11 instances that compare equal to but are not identical to the module-level singletons, so the identity check could cache the `e4m3` key as `e5m2`. The dict is a process-global singleton, so this was a latent bug that surfaced when a C++-Enum lookup was the first `e4m3` lookup in the process (seen in `test_float8block_linear_fused_adam_master_weights`). The `Custom_DType_Dict` class is created in this fork (upstream does not have the class), so no upstream compatibility concern.

ipanfilo:
Is CDNA3 support expected to be ported to HK?

alextmagro:
> Is CDNA3 support expected to be ported to HK?

Currently HK supports CDNA3, CDNA4, and UDNA1. There is currently effort with the external HK team to move away from having separate git branches for each arch to having directory level separation.

tests/pytorch/test_float8_blockwise_gemm_exact.py
```diff
@@ -6,21 +8,45 @@
...
 
+fp8_e4m3_type = get_torch_float8_e4m3_type()
+fp8_e5m2_type = get_torch_float8_e5m2_type()
 
 def fp8_blockwise_gemm_supported() -> bool:
     supported = te.is_fp8_block_scaling_available()
     emulated = get_device_compute_capability() >= (10, 0)
     return supported and not emulated
 
+def rocm_blockwise_unsupported_reason(
```
alextmagro: Here generally I think we should return a tuple -- True/False, followed by an optional reason. Maybe rename to rocm_blockwise_is_supported too.

tests/pytorch/test_float8_blockwise_gemm_exact.py
```diff
@@ -228,6 +263,11 @@ def cublas_gemm_test_constraint_enforced(
 ):
     if not fp8_blockwise_gemm_supported():
         pytest.skip("CUDA version does not support blockwise FP8 gemm.")
+    if IS_HIP_EXTENSION:
```
alextmagro, line 269: We could also potentially use the rocm function above here too?

ipanfilo:
> > Is CDNA3 support expected to be ported to HK?
> 
> Currently HK supports CDNA3, CDNA4, and UDNA1. There is currently effort with the external HK team to move away from having separate git branches for each arch to having directory level separation.

In this case can we avoid having 2 copies of HK submodule?

alextmagro:
> > > Is CDNA3 support expected to be ported to HK?
> > 
> > 
> > Currently HK supports CDNA3, CDNA4, and UDNA1. There is currently effort with the external HK team to move away from having separate git branches for each arch to having directory level separation.
> 
> In this case can we avoid having 2 copies of HK submodule?

Yes, looks like they merged it in late last week. I think that will require some integration/cmake changes so that the right headers are imported for each kernel. We'll probably want to move our kittens kernels into cdna3/cdna4/udna1 folders as well, so might be better as a follow up ticket?

tests/pytorch/test_float8_blockwise_gemm_exact.py
```diff
@@ -728,6 +773,10 @@ def test_bgrad_not_supported(
     is_w_1d_scaled,
 ) -> None:
     # NOTE: BGRAD epilogue is not supported for fp8.
+    if IS_HIP_EXTENSION:
```
alextmagro, line 781: One of the great things about hipKittens is how easy it is to fuse epilogues into kernels. If you have time it might be worth looking into bgrad/accumulator support.

tests/pytorch/test_float8_blockwise_gemm_exact.py
```diff
@@ -839,6 +895,10 @@ def test_illegal_dtype_enforced(
     is_w_1d_scaled,
 ) -> None:
     # e5m2 by e5m2 not supported.
+    if IS_HIP_EXTENSION:
+        expected_err_msg = "does not support e5m2 by e5m2 inputs"
```
alextmagro: e5m2 should be supported by hipkittens, have a look at the mxfp8 kernels to see how we template for that

tests/pytorch/test_float8_blockwise_gemm_exact.py
```diff
@@ -952,7 +1022,7 @@ def test_unaligned_shapes(
             use_split_accumulator,
             is_x_1d_scaled,
             is_w_1d_scaled,
-            expected_err_msg="dimension requirement",
```
alextmagro, line 955: This change should be hip guarded

tests/pytorch/test_float8_blockwise_scaling_exact.py
```diff
@@ -38,10 +38,7 @@
 tensor_dump_dir_env = os.getenv("NVTE_TEST_BLOCK_CURRENT_SCALING_EXACT_TENSOR_DUMP_DIR")
 if tensor_dump_dir_env is not None:
     TENSOR_DUMP_DIR = pathlib.Path(tensor_dump_dir_env)
-recipe_available, reason_for_no_recipe = te.is_fp8_block_scaling_quantization_available(
```
alextmagro, line 41: Do we not require a check for quantization available? Ideally we would keep this source code and just always return true if it is always supported.

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
```diff
@@ -0,0 +1,693 @@
...
+#include <type_traits>
+#include "kittens.cuh"
+#include "blockwise_fp8_gemm.h"
+
+namespace blockwise_gfx942 {
+
+#include "blockwise_fp8_gemm_helper.cuh"
+
+constexpr int NUM_WARPS   = 8;
+constexpr int WARPS_ROW    = 2;
```
alextmagro: nit: alignment here

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
```diff
@@ -0,0 +1,693 @@
...
+    const void *gelu_aux;
+    int gelu_aux_dtype;
+    const OType *c_in;
+    float beta;
+    int M() const { return (int)c.rows(); }
+    int N() const { return (int)c.cols(); }
+    int K() const { return (int)a.cols(); }
+    dim3 grid()  { return dim3(((N() + BLOCK_N - 1) / BLOCK_N) * ((M() + BLOCK_M - 1) / BLOCK_M)); }
+    dim3 block() { return dim3(NUM_THREADS); }
+    size_t dynamic_shared_memory() { return 49152; }
```
alextmagro: Why do we require 48 MiB exactly? If possible I would rather use a macro for both cdna3 and cdna4 that is defined based on the needed value.

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
```diff
@@ -0,0 +1,693 @@
...
+    const float *sb_base = g.scale_b.raw_ptr + (sb_valid ? sb_block0 : 0) * num_k_steps;
+    int32x4_lds_t sb_srsrc = make_buf_res((const void*)sb_base, (uint32_t)num_k_steps * 4);
+    const int local_m0 = warp_row * REG_M;
+    const int local_m1 = (warp_row + WARPS_ROW) * REG_M;
+    const int tid = threadIdx.x;
+
+    const uint32_t sa_range = (uint32_t)((M - row * BLOCK_M) * 4);
+
+    const bool is_first_k_partial = is_k_partial && (num_k_steps == 1);
+    if (is_first_k_partial || is_last_m) load_tile_masked<NUM_THREADS>(As, g.a, row, 0, M, K);
```
alextmagro: Let's add brackets here so it is easier to read.

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
```diff
@@ -0,0 +1,693 @@
...
+        float sb_next;
+        kittens::load(at[0], kittens::subtile_inplace<REG_M, MFMA_K>(As, {warp_row, 0}));
+        kittens::load(at[1], kittens::subtile_inplace<REG_M, MFMA_K>(As, {warp_row + WARPS_ROW, 0}));
+        kittens::load(bt[0], kittens::subtile_inplace<REG_N, MFMA_K>(Bs, {warp_col, 0}));
+        __builtin_amdgcn_s_barrier();
+        __builtin_amdgcn_sched_barrier(0);
+
+        // Cluster 1
+        asm volatile("s_waitcnt lgkmcnt(0)");
+        __builtin_amdgcn_s_setprio(1);
```
alextmagro, line 172: We discussed this a bit on Friday but trying prio 2 might give a performance boost.

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
```diff
@@ -0,0 +1,693 @@
...
+        kittens::mma_ABt(partial[0], at[0], bt[0], partial[0]);
+        kittens::mma_ABt(partial[1], at[1], bt[0], partial[1]);
+        __builtin_amdgcn_s_setprio(0);
+        __builtin_amdgcn_s_barrier();
+        __builtin_amdgcn_sched_barrier(0);
+
+        // Cluster 6
+        asm volatile("s_waitcnt lgkmcnt(0)");
+        if (is_next_k_partial || is_last_m) load_tile_masked<NUM_THREADS>(As, g.a, row, k_step + 1, M, K);
```
alextmagro: brackets here too for readability

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
```diff
@@ -0,0 +1,693 @@
...
+        kittens::store(g.c, C_accum[0], {0, 0, row * 4 + warp_row,     col * 4 + warp_col});
+        kittens::store(g.c, C_accum[1], {0, 0, row * 4 + warp_row + WARPS_ROW, col * 4 + warp_col});
+    }
+}
+
+template <typename AType, typename BType, typename OType, bool IS_PARTIAL_M, bool IS_PARTIAL_N,
+          bool HAS_BIAS = false, bool HAS_GELU = false, bool HAS_BETA = false,
+          bool IS_PARTIAL_K = false>
+__global__ __launch_bounds__(NUM_THREADS, 2)
+void micro_tk_1d1d(const micro_globals<AType, BType, OType> g) {
```
alextmagro: Ideally we would template the 1d2d and 1d1d kernels together; if possible it would be great for future maintenance and additional formats such as 2d1d and 2d2d.

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
```diff
@@ -0,0 +1,693 @@
...
+    if      (!pm && !pn) launch(kernel.template operator()<false, false>());
+    else if ( pm && !pn) launch(kernel.template operator()<true,  false>());
+    else if (!pm &&  pn) launch(kernel.template operator()<false, true >());
+    else                 launch(kernel.template operator()<true,  true >());
+}
+
+template <bool IS_1D2D, typename AType, typename BType, typename OType, bool KP>
+static void dispatch_micro_k(micro_globals<AType, BType, OType> g,
+                             bool has_bias, bool has_gelu, bool has_beta) {
+    if (has_gelu) {
```
alextmagro: Generally we can use our BOOL_SWITCH to avoid these types of if elses for templated code. We should have a copy of that function already in our hipkittens headers.

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.h
```diff
@@ -0,0 +1,43 @@
...
+*************************************************************************/
+
+#pragma once
+
+#include <hip/hip_runtime.h>
+#include <cstddef>
+
+#ifndef KITTENS_DTYPE_ENUM_DEFINED
+#define KITTENS_DTYPE_ENUM_DEFINED
+enum KittensDType {
```
alextmagro: I'd rather not have to redefine these. I think we can have a parent folder header file that contains these enums and any other shared values.

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.h
```diff
@@ -0,0 +1,43 @@
...
+#ifndef KITTENS_SCALING_MODE_DEFINED
+#define KITTENS_SCALING_MODE_DEFINED
+enum KittensScalingMode {
+    KITTENS_BLOCK_SCALING_1D = 2,
+    KITTENS_BLOCK_SCALING_2D = 3,
+};
+#endif  // KITTENS_SCALING_MODE_DEFINED
+
+namespace blockwise_gfx942 {
+void kittens_blockwise_fp8_gemm_impl_cdna3(
```
alextmagro: Rather than a namespace for gfx942 and gfx950, etc, I think we would be better off with a purely virtual class at the parent folder level, with child class definitions for functions within each architectural folder. That way we can track more easily what is available and what is not for each architecture.

transformer_engine/common/gemm/kittens/cdna4/blockwise_fp8_gemm.cpp
```diff
@@ -0,0 +1,1459 @@
+/*************************************************************************
```
alextmagro, line 1: Most of my comments regarding the cdna3 kernel applies here as well -- otherwise the kernel looks good.

transformer_engine/common/gemm/kittens/CMakeLists.txt
```diff
@@ -3,50 +3,86 @@
...
-target_include_directories(kittens_gemm PRIVATE ${HIP_INCLUDE_DIRS})
-target_link_libraries(kittens_gemm PUBLIC hip::host hip::device)
-target_compile_options(kittens_gemm PRIVATE ${KITTENS_GEMM_COMPILE_OPTIONS})
-
-install(TARGETS kittens_gemm
-    DESTINATION ${CMAKE_INSTALL_PREFIX}/transformer_engine/lib)
+option(NVTE_KITTENS_USE_POWER_OF_2_SCALE
+    "Use HW E8M0 MFMA scaling for power-of-2 blockwise FP8 scales (gfx950)" ON)
+
+function(try_enable_hipkittens_gemm)
```
alextmagro: Not sure this needs to bea  function if we are only calling it once?

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1980,6 +2000,69 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
...
+          workspace, workspaceSize, s);
+
+      if (use_service_stream)
+      {
+        release_service_stream(stream, ss_ctl);
+      }
+      return;
+    }
+  }
+#endif
```
alextmagro, line 2067: if hipkittens is not being used, we should add an nvte_error here for blockwise usage

alextmagro (changes requested):

tests/pytorch/test_float8_blockwise_gemm_exact.py
asdfvg123: fixed

tests/pytorch/test_float8_blockwise_gemm_exact.py
asdfvg123: yes, now we are using the function

tests/pytorch/test_float8_blockwise_gemm_exact.py
asdfvg123: I will leave this as a future support. The current implementation supports the minimal cases that the test is expecting

tests/pytorch/test_float8_blockwise_gemm_exact.py
asdfvg123: Same here. The test expects e4m3 x e4m3 / e4m3 x e5m2 / e5m2 x e4m3 but not e5m2 x e5m2 and the current implementation also supports the former 3 but not the e5m2 x e5m2. While this is an easy support, I will leave it as a future PR

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.h
asdfvg123: fixed. Added kittens_gemm_enums.h

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.h
asdfvg123: fixed

transformer_engine/common/gemm/kittens/CMakeLists.txt
asdfvg123: fixed

transformer_engine/common/gemm/rocm_gemm.cu
asdfvg123: fixed

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
asdfvg123: fixed

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
asdfvg123: fixed

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
asdfvg123: fixed

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
asdfvg123: Yes, I will leave this as a future work for now

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
asdfvg123: fixed

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
asdfvg123: fixed.

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
asdfvg123: added

tests/pytorch/test_float8_blockwise_scaling_exact.py
asdfvg123: In upstream, quantization and GEMM share a single availability check. If blockwise scaling is available, the GEMM is assumed to be supported. In PR #609, blockwise GEMM was not yet implemented, so I added is_fp8_block_scaling_quantization_available as a temporary function to gate quantization independently of GEMM. Now that blockwise GEMM is fully supported, this separation is obsolete, so I removed the temporary function rather than keeping it as an always-true stub. This restores the code to upstream's single-check behavior. 
Happy to keep the function returning true instead if you'd prefer to preserve the API surface.

tests/pytorch/test_float8_blockwise_gemm_exact.py
asdfvg123: fixed. To keep the upstream lines untouched, I guarded early-return block, which makes the code a bit longer. Let me know if you have a cleaner approach.

tests/pytorch/test_float8_blockwise_gemm_exact.py
```diff
@@ -6,21 +8,45 @@
...
 
+fp8_e4m3_type = get_torch_float8_e4m3_type()
+fp8_e5m2_type = get_torch_float8_e5m2_type()
 
 def fp8_blockwise_gemm_supported() -> bool:
     supported = te.is_fp8_block_scaling_available()
     emulated = get_device_compute_capability() >= (10, 0)
     return supported and not emulated
 
+def rocm_blockwise_is_supported(
```
wangye805: Guard our rocm-specific functions via IS_HIP_EXTENSION

tests/pytorch/test_float8_blockwise_gemm_exact.py
```diff
@@ -45,12 +71,21 @@ def cublas_gemm_fp8_blockwise_case(
     atol: float = 0.0,
     rtol: float = 0.0
 ):
-    if x_dtype == torch.float8_e5m2 and w_dtype == torch.float8_e5m2:
+    if IS_HIP_EXTENSION:
+        atol = rtol = 8e-3
```
wangye805: Is NV upstream using a different atol/rtol? If so, why

tests/pytorch/test_float8_blockwise_gemm_exact.py
```diff
@@ -6,21 +8,45 @@
...
+    is_x_1d_scaled,
+    is_w_1d_scaled,
+    *,
+    x_columnwise: bool = False,
+    w_columnwise: bool = False,
+):
+    is_1d2d = is_x_1d_scaled and not is_w_1d_scaled
+    is_1d1d = is_x_1d_scaled and is_w_1d_scaled
+    if not (is_1d2d or is_1d1d):
+        return False, "Only 1D by 1D and 1D by 2D block scaling GEMM is supported"
```
wangye805: Actually, let's have a subsection in readme to state our support matrix as well, and how it is different from NV side?

tests/pytorch/test_float8_blockwise_gemm_exact.py
```diff
@@ -680,6 +722,10 @@ def test_split_accumulator_enforced(
     is_x_1d_scaled,
     is_w_1d_scaled,
 ) -> None:
+    if IS_HIP_EXTENSION:
+        expected_err_msg = "requires split accumulator"
```
wangye805, line 729: use_split_accumulator in line 703 is set to false

transformer_engine/common/gemm/kittens/cdna4/blockwise_fp8_gemm_helper.cuh
```diff
@@ -0,0 +1,428 @@
...
+    asm volatile("s_mov_b32 m0, %0" :: "s"(lds_base_warp));
+    raw_buffer_load_lds_(__builtin_bit_cast(i32x4_v_, srd), (as3_u32_ptr_)0, 4, voffset, 0, 0, 0);
+}
+
+__device__ inline float rtne_bias(float v) {
+    uint32_t bits = __builtin_bit_cast(uint32_t, v);
+    if ((bits & 0x7f800000u) == 0x7f800000u) return v;
+    bits += 0x7fffu + ((bits >> 16) & 1u);
+    return __builtin_bit_cast(float, bits);
+}
```
wangye805: I recall we have hardware rtne casting instructions in gfx950?

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -1980,6 +2000,74 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
     handle = hipblaslt_handles[compute_stream_offset];
   }
 
+#ifdef USE_HIPKITTENS_GEMM
```
wangye805, line 2003: Emm, any reason that we don't want to build and use hipkitten fp8 blockwise gemm?

wangye805 (changes requested):

tests/pytorch/test_float8_blockwise_gemm_exact.py
asdfvg123: fixed

tests/pytorch/test_float8_blockwise_gemm_exact.py
asdfvg123: This is the test intentionally set use_split_accumulator = False to test the kernel only takes use_split_accumulator = True

transformer_engine/common/gemm/kittens/cdna4/blockwise_fp8_gemm_helper.cuh
asdfvg123: fixed. Now gfx950 does not use sw rtne path.

transformer_engine/common/gemm/rocm_gemm.cu
asdfvg123: This gating originates from the HipKittens MXFP8 implementation (PR #566)

The reason why it is guarded by USE_HIPKITTENS_GEMM is that HipKittens requires C++20, whereas TE generally builds with C++17. The option is ON by default, but gemm/kittens/CMakeLists.txt automatically turns it OFF when C++20 isn't available, so a C++17-only build of TE doesn't break. When the backend is disabled and a blockwise GEMM is requested, the #else path raises a clear NVTE_ERROR telling the user to build with USE_HIPKITTENS_GEMM.

tests/pytorch/test_float8_blockwise_gemm_exact.py
asdfvg123: added on the readme.rst

tests/pytorch/test_float8_blockwise_gemm_exact.py
asdfvg123: NV upstrean uses atol = rtol = 0 since their reference reproduces the exact cuBLAS kernel.
Our kernel has algorithmic differences so I set atol, rtol to PyTorch's standard fp32 values (atol=1e-5, rtol=1.3e-6), and the tests still pass.

transformer_engine/common/gemm/rocm_gemm.cu
wangye805: I see. Thanks

tests/pytorch/test_float8_blockwise_gemm_exact.py
wangye805: I'm still confused. ROCm TE does not support use_split_accumulator=True (it will fail or error out)? And this testcase set it to false, so we should be fine? Why do we expect a message like "requires split accumulator"?

README.rst
```diff
@@ -281,6 +281,21 @@ To enable MXFP8 support, use NVTE_ROCM_ENABLE_MXFP8 environment variable which c
 * 1 - enable MXFP8 support in fp8;
 * 2 - make MXFP8 a default fp8 recipe.
 
+Blockwise FP8 GEMM support on ROCm (gfx94x and gfx95x)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Blockwise FP8 GEMM is supported on gfx94x and gfx95x GPUs through the HipKittens backend and is enabled by default.
```
wangye805: nit, gfx942 and gfx950, not gfx94x or gfx95x

tests/pytorch/test_float8_blockwise_gemm_exact.py
wangye805: Wait a second, what's NV side target and reference impl backends? Both are cublaslt kernels? If so, they were comparing to itself...

asdfvg123: To clarify, the reference is not cublaslt itself. It is tests/pytorch/references/blockwise_fp8_gemm_reference.py

The reference python code is written to be bit-exact with cuBLAS. Thus, they are using atol=rtol=0.
We are comparing with the same python reference, but because our algorithm is different, we are using non zero tolerance.

tests/pytorch/test_float8_blockwise_gemm_exact.py
asdfvg123: This is an existing negative test from NV upstream. I kept the same pass/fail test behavior when implementing the ROCm kernel, to match NVIDIA's contract exactly.

To clarify, ROCm blockwise FP8 GEMM requires use_split_accumulator=True. True is the correct/supported value and False is the invalid one. 

This test intentionally passes False and asserts the kernel rejects it via NVTE_CHECK(use_split_accumulator, "...requires split accumulator") (rocm_gemm.cu), so it passes only when that error is raised.

README.rst
asdfvg123: fixed

transformer_engine/common/gemm/kittens/cdna4/blockwise_fp8_gemm.cpp
asdfvg123: I have also applied a similar changes to to the cdna4 kernels as well

wangye805 (approved):

tests/pytorch/test_float8_blockwise_gemm_exact.py
alextmagro: If the kernel supports the config already it I think it would be better to just disable this test, and allow users to use it if they want.

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.h
alextmagro: Thanks, the code looks good. I think we can keep it header only though, and combine blockwise_fp8_gemm.cpp, blockwise_fp8_gemm.h, blockwise_fp8_gemm_backend.h, and kittens_gemm_enum.h into a single kittens_common.h file.

alextmagro: I think the kittens_blockwise_fp8_gemm function could exist in the class, or at least be a friend function though?

tests/pytorch/test_float8_blockwise_scaling_exact.py
alextmagro: Understood, we should keep in sync with upstream here then.

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
alextmagro: nit, but should we consider adding a static_assert at compile time to ensure we don't exceed LDS size with our SMEM_BYTES? So, 64 KiB/160KiB?

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
alextmagro: I'd prefer if you could just give it a try before we merge, so we know. If you don't have time, please add a TODO

alextmagro (changes requested):

.github/workflows/rocm-wheels-build.yml
```diff
@@ -89,6 +89,7 @@ jobs:
             3rdparty/ck_jit \
             3rdparty/hipify_torch \
             3rdparty/hipkittens \
+            3rdparty/hipkittens_cdna3 \
```
wenchenvincent: HipKittens recently merged cdna3 support into the main branch. Could you try to reorganize based on that and test it?

wenchenvincent (changes requested):

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
asdfvg123: Fixed, added a static_assert. CDNA4 uses explicit __shared__ declarations, so I only added the assert to the CDNA3 path.

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.cpp
asdfvg123: I added TODO for now.

transformer_engine/common/gemm/kittens/cdna3/blockwise_fp8_gemm.h
asdfvg123: Fixed.
Combined the 3 .h and 1 .cpp files into one header (kittens_common.h)

kittens_blockwise_fp8_gemm is now a friend of BlockwiseGemmBackend, and the arch moved into the class as private static members (get_cdna3()/get_cdna4()). The dispatcher picks one by arch and calls backend->run(args)

rocm_gemm.cu still call kittens_blockwise_fp8_gemm()

tests/pytorch/test_float8_blockwise_gemm_exact.py
asdfvg123: Fixed. I cleared the NVTE_CHECK in rocm_gemm and enabled e5m2 by e5m2, so the kernel now supports e5m2 x e5m2.

The negative test (checking that the kernel rejects the unsupported e5m2 by e5m2) is now skipped on the ROCm path, and the e5m2 by e5m2 case is enabled as a positive test for ROCm.

alextmagro (approved):

.github/workflows/rocm-wheels-build.yml
asdfvg123: merged

wenchenvincent (approved):

PR #413, opened 2026-01-15
Enhance GroupedLinear with integrating AITER triton kernels

sudhu2k:
# Description

This PR enhances the GroupedLinear module with Triton kernel support for grouped GEMM operations, providing optimized performances. The implementation includes a complete Triton-based grouped matrix multiplication (GMM) backend that can be enabled via environment variables, along with pre-tuned configurations for optimal performance.

- Added support for using Triton kernels in GroupedLinear, allowing for optimized performance based on environment variables.
- Updated the setup.py to include JSON configuration files for Triton kernels in the package data.
- Added a new test case for grouped GEMM functionality in the CI pipeline.
- Refactored the handling of input tensors and gradients to accommodate the new Triton kernel logic.

## Benchmark results:
https://github.com/ROCm/frameworks-internal/issues/13792#issuecomment-3739558113
https://github.com/ROCm/frameworks-internal/issues/13792#issuecomment-3746418683

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- **Added Triton kernel support for GroupedLinear**: Implemented a complete Triton-based grouped GEMM backend with support for dynamic kernel selection based on environment variables (`NVTE_USE_GROUPED_GEMM_TRITON`)
-   Added optional `m_splits_tensor` parameter to keep tensor data on GPU and avoid redundant CPU-GPU data transfers for improved performance
- **New GMM (Grouped Matrix Multiplication) module from AITER**: Added comprehensive Triton kernel implementation in `transformer_engine/pytorch/triton_kernels/gmm/` from AITER including:
  - `gmm_common.py`: Common utilities and helper functions
  - `gmm_kernels.py`: Core Triton kernel implementations for grouped GEMM operations
  - `gmm_wrapper.py`: High-level wrapper functions from AITER
  - `pid_preprocessing.py`: Process ID preprocessing for efficient kernel scheduling
- **Pre-tuned configurations**: Added JSON configuration files for AMD GPU architectures:
  - `gfx942-GMM.json`: pre-tuned configs for gfx942 arch
  - `gfx950-GMM.json`: pre-tuned configs for gfx950 arch
- **Updated setup.py**: Modified package data to include JSON configuration files for Triton kernels
[...]

PR #660, opened 2026-07-07
Ifu dev 20260706 v2.17

AllenFarcas:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


tests/jax/test_custom_call_compute.py
```diff
@@ -2108,6 +2108,7 @@ def f(x):
         assert_allclose(actual, expected, dtype=dtype)
 
 
+@pytest.mark.skipif(is_hip_extension(), reason="Standalone TopK (nvte_topk) is not supported on ROCm")
```
wenchenvincent, line 2111: Why is nvte_topk not supported on rocm? Is this a new kernel from upstream?

transformer_engine/pytorch/custom_recipes/quantization_ref_nvfp4.py
```diff
@@ -712,7 +712,9 @@ def _quantize_blockwise_reference(
             FLOAT8_E4M3_MAX = torch.tensor(240.0 if is_fp8_fnuz() else 448.0, device=x.device, dtype=torch.float32)
         else:
             FLOAT8_E4M3_MAX = torch.tensor(448.0, device=x.device, dtype=torch.float32)
-        global_scale_e4m3_max = float(nvfp4_e4m3_max if nvfp4_use_4over6 else 448)
+        # On fnuz archs the block scale is stored in fnuz e4m3 (max 240), so the global
+        # scale must use the same arch-aware max to match the kernel (see FLOAT8_E4M3_MAX).
+        global_scale_e4m3_max = float(nvfp4_e4m3_max) if nvfp4_use_4over6 else float(FLOAT8_E4M3_MAX)
```
wenchenvincent: I think this recipe is only for gfx1250 or later? On gfx1250, we don't have the fnuz variant.

wenchenvincent: Particularly, this is for the scale.

tests/pytorch/distributed/test_fusible_ops.py
```diff
@@ -1046,7 +1046,7 @@ def run_parallel_tests() -> None:
 @pytest.mark.parametrize("world_size", _world_sizes)
 def test_distributed_fuser_ops(world_size: int) -> None:
     """Launch parallel job that runs parallel tests"""
-    #TODO: find out why cannot align the following two lines with NV upstream
+    # TODO: find out why we cannot align the following two lines with NV upstream
```
wenchenvincent: What's not aligned with NVTE upstream?

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -140,7 +140,11 @@ void quantize_fwd_helper(const NVTETensor input, NVTETensor output,
                    "Row-scaled NVFP4 quantization does not support 2D quantization.");
         NVTE_CHECK(!output_tensor->has_columnwise_data(),
                    "Row-scaled NVFP4 quantization does not produce columnwise output.");
+#ifndef __HIP_PLATFORM_AMD__
         nvfp4::compute_rowwise_amax(*input_tensor, noop_tensor, output_tensor, stream);
+#else
+        NVTE_ERROR("Row-scaled NVFP4 quantization is not supported on ROCm.");
```
wenchenvincent: Did we not support rowwise nvfp4 quantization? @matthiasdiener

transformer_engine/common/fused_attn/flash_attn.cu
```diff
@@ -157,6 +159,7 @@ void prepare_flash_attn_bwd(Tensor q, Tensor k, Tensor v, Tensor qkv, cudaStream
 // multi_tensor_transpose_to_bhsd: BSHD/SBHD -> BHSD
 // ============================================================================
 
+#ifndef __HIP_PLATFORM_AMD__  // Disabled on ROCm
```
wenchenvincent, line 164: Is this a new kernel from upstream? Does it use TMA?

transformer_engine/common/fused_attn_rocm/fused_attn.cpp
```diff
@@ -348,12 +348,17 @@ void nvte_fused_attn_fwd(const NVTETensor Q, const NVTETensor K, const NVTETenso
...
-                         NVTE_QKV_Layout qkv_layout, NVTE_Bias_Type bias_type,
+                         NVTE_QKV_Layout qkv_layout, NVTE_QKV_Format o_format,
+                         NVTE_QKV_Format qkv_scale_inv_format, NVTE_Bias_Type bias_type,
                          NVTE_Mask_Type attn_mask_type, NVTE_Softmax_Type softmax_type,
                          int64_t window_size_left, int64_t window_size_right,
                          bool bottom_right_diagonal, NVTETensor workspace, cudaStream_t stream) {
   NVTE_API_CALL(nvte_flash_attn_fwd);
   using namespace transformer_engine;
+  // ROCm output is always in the Q format; FP8 scale-inv formats are unused.
+  NVTE_CHECK(o_format == nvte_get_q_format(qkv_layout),
```
wenchenvincent, line 359: Is this o_format added by usptream? If so, do you know why it is added?

wenchenvincent: Is this specific to fp8 attention?

transformer_engine/common/fused_router/utils.h
```diff
@@ -71,6 +71,17 @@ enum ReduceFuncType {
   MAX,
 };
 
+// Reduction helpers for masked_warp_reduce_on_shmem.
```
wenchenvincent, line 74: Why is this newly added?

transformer_engine/common/swizzle/swizzle.cu
```diff
@@ -2212,6 +2212,7 @@ void nvte_multi_tensor_unswizzle_scaling_factors(const NVTETensor* inputs, NVTET
 
 namespace transformer_engine {
 
+#ifndef __HIP_PLATFORM_AMD__  // Disabled on ROCm
```
wenchenvincent, line 2215: is this the multi-tensor swizzle kernel or a different one?

transformer_engine/common/include/transformer_engine/fused_attn.h
```diff
@@ -159,10 +159,7 @@ enum NVTE_Softmax_Type {
 enum NVTE_Fused_Attn_Backend {
   /*! No supported backend */
   NVTE_No_Backend = -1,
-  /*! cuDNN-based FP16/BF16 fused attention for max 512 sequence length.
-   *  Note: upstream v2.17 removed the max512 backend implementation/dispatch;
-   *  the ROCm fork retains the enumerator so its CUDA-path bindings/consumers
-   *  (pybind_helper.h, cpp_extensions/fused_attn.py) still compile. */
+  /*! cuDNN-based FP16/BF16 fused attention for <= 512 sequence length */
```
wenchenvincent: Is max512 still supported or not?

transformer_engine/pytorch/csrc/common.h
```diff
@@ -354,12 +354,14 @@ class MXFP4Quantizer : public Quantizer {
 
   void set_quantization_params(TensorWrapper* tensor) const override;
 
-  std::pair<TensorWrapper, py::object> create_tensor(const std::vector<size_t>& shape,
-                                                     DType dtype) const override;
+  std::pair<TensorWrapper, py::object> create_tensor(
+      const std::vector<size_t>& shape, DType dtype,
```
wenchenvincent, line 358: Is this change from upstream?

tests/cpp/operator/test_ck_grouped_mxfp8.cu
```diff
@@ -424,7 +424,7 @@ static ck_tile::HostTensor<ck_tile::bfloat16_t> run_ck_tile_reference_for_group(
   return c_ref;
 }
 
-static float max_abs_bf16_tensor(const Tensor& t, size_t rows, size_t cols) {
```
wenchenvincent, line 427: Why remove `const`?

transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh
```diff
@@ -72,8 +72,14 @@ __global__ void __launch_bounds__(512)
 #else
   float amax = (tensor_amax != nullptr) ? (ROW_SCALED_NVFP4 ? tensor_amax[y] : tensor_amax[0]) : 1.0f;
 #endif
+#if defined(__HIP_DEVICE_COMPILE__)
+  // ROCm fp8e4m3 max is arch-specific (240 fnuz / 448 OCP); match the encode scale.
+  constexpr float factor_inv =
```
wenchenvincent: We don't need to do that for fnuz

transformer_engine/pytorch/csrc/extensions/gemm.cpp
```diff
@@ -669,6 +669,7 @@ std::optional<std::vector<at::Tensor>> te_general_grouped_gemm(
   return bias;
 }
 
+#ifndef USE_ROCM  // CUDA-only path; ROCm uses te_general_grouped_gemm.
 py::object te_general_grouped_gemm_for_grouped_tensor(
     py::handle A, bool transa, py::handle B, bool transb, py::handle D, py::object bias,
```
wenchenvincent, line 676: Do we have this function `te_general_grouped_gemm_for_grouped_tensor` on ROCm? Or I think this is a new function newly added by upstream? Shall we make a similar function? What's the API difference?

transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -20,11 +20,6 @@
 import transformer_engine_torch as tex
 
 from transformer_engine.common.recipe import Recipe
-from transformer_engine.pytorch.tensor._quantization_helpers import (
```
wenchenvincent, line 21: Do we know why the circular import only exists on ROCm?

tests/pytorch/triton_kernels/test_cast.py
```diff
@@ -83,20 +83,31 @@ def test_quantize(scaling, shape, in_dtype, out_dtype):
...
-        use_torch_semantics=True
-    ),
-    te_compare_results(
-        quantized_out_triton._get_quantizer().amax,
-        quantized_out_tex._get_quantizer().amax,
-        atol=atol_scale, rtol=rtol_scale,
-        msg='AMAX results do not match!',
-        use_torch_semantics=True
-    )
+    if scaling == "delayed":
```
wenchenvincent, line 86: Is this change due to API change in upstream?

tests/pytorch/mxfp4/test_mxfp4_module_exact.py
```diff
@@ -49,21 +49,11 @@ def mxfp4_recipe_to_test(use_hadamard: bool = False):
...
-            # Output quantization not used
-            return None
-        if role == "linear_grad_output":
-            return MXFP4QuantizerRef(rowwise=True, columnwise=True, shuffle_rowwise_data=False, shuffle_columnwise_data=False, with_gemm_swizzled_scales=False, use_hadamard=use_hadamard, use_te_quantizer=True)
-        elif role == "linear_grad_input":
-            # Grad input quantization not used
-            return None
-        else:
-            return None
+    def factory(role):
```
wenchenvincent, line 52: Why do we change this? mxfp4 is not related to IFU, I suppose.

transformer_engine/pytorch/tensor/fsdp2_allgather_tensor.py
```diff
@@ -142,7 +142,7 @@ def fsdp_pre_all_gather(self, mesh):
 
             self._module.init_fp8_metadata(num_gemms=num_gemms)
         if not self._module.fp8:
-            return (self._data,), (self._data.requires_grad,)
+            return (self._data,), (self._data.requires_grad, None)
```
wenchenvincent, line 145: Is this a ROCm specific issue?

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -1653,8 +1655,14 @@ def forward(
 
         ctx.o_format = o_format
         ctx.qkv_scale_inv_format = qkv_scale_inv_format
-        # dqkv should have the same layout as the original qkv
-        ctx.dqkv_layout = original_qkv_layout
+        # dqkv should have the same layout as the original qkv. On ROCm the fused-attn
```
wenchenvincent, line 1658: Why does the dqkv layout come back as de-interleaved?

tests/pytorch/attention/test_attention.py
```diff
@@ -284,9 +284,13 @@ def test_dot_product_attention(
         flash_attn_supported, fused_attn_supported, unfused_attn_supported = available_backends
 
     # FlashAttention does not support pad_between_seqs, but _run_dot_product_attention
-    # mannually pads and unpads the input and output of FlashAttention for testing purposes
+    # mannually pads and unpads the input and output of FlashAttention for testing purposes.
```
wenchenvincent, line 287: @ipanfilo  Could you check if this is accurate?

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -3189,6 +3189,15 @@ def skip_correction(_output, _softmax_aux, output_per_step, softmax_aux_per_step
                     return output_per_step.astype(jnp.float32), softmax_aux_per_step
 
                 def correction(output, softmax_aux, output_per_step, softmax_aux_per_step):
+                    if is_hip_extension():
+                        # Fully-masked striped steps must contribute nothing; the CK kernel can
+                        # emit non-finite output/LSE for them, so neutralize to avoid 0 * NaN.
+                        softmax_aux_per_step = jnp.where(
```
wenchenvincent, line 3195: It is not clear to me why we need to do this. Are those failures due to upstream changes?

tests/pytorch/triton_kernels/test_grouped_gemm.py
```diff
@@ -352,6 +352,11 @@ def test_tgmm(
     trans_lhs = trans_lhs_from_str(trans_lhs_str)
     rng_seed = rng_seed_from_str(rng_seed_str)
 
+    # Skip very large M before allocating inputs; gen_tgmm_tensors allocates ~24 GiB and would
+    # OOM under memory pressure before the check ran.
+    if M > 1e6:
```
wenchenvincent, line 357: Is this related to IFU? This test is rocm specific, I suppose.

tests/jax/test_custom_call_compute.py
AllenFarcas: It is a new upstream kernel. Standalone `nvte_topk` was added in this commit from upstream (https://github.com/NVIDIA/TransformerEngine/commit/0c2e7b09c6d33109803ea089fbf80421a326e0a4). It is built on CUB which isn't ported to ROCm. This is distinct from the MoE fused-router top-k which does compile and run on ROCm, so only the standalone `nvte_topk` API/JAX FFI is unavailable.

transformer_engine/pytorch/custom_recipes/quantization_ref_nvfp4.py
AllenFarcas: NVFP4 execution is enabled on gfx94x and newer in transformer_engine/pytorch/quantization.py line 197 (https://github.com/ROCm/TransformerEngine/pull/660/changes/BASE..be3feea9474299e45c4c55fc1490fb554819a6a8#diff-7d9f4aea337f66e97045753428d52173c5f70d022f67f082c3c4ad63c2ade332L101). Since it includes gfx942, an fnuz arch (e4m3 max 240), the reference's global scale must use the same arch-aware max or it mismatches the kernel. 

The 680 NVFP4 exact tests were failing on gfx942 without this and pass with it. FLOAT8_E4M3_MAX is already defined arch-aware (240 if is_fp8_fnuz() else 448).

tests/pytorch/distributed/test_fusible_ops.py
AllenFarcas: This is a pre-existing ROCm workaround (commit ba59f81c , "resolve pytorch pytest conflicts"), not introduced by the IFU, we only touched the comment. Post-merge, v2.17 appears to use the same two lines, so the TODO may now be stale, worth re-verifying against upstream

transformer_engine/common/cast/dispatch/quantize.cuh
AllenFarcas: Rowwise (row-scaled) NVFP4 is supported on ROCm. Lines 143-147 aren't a rowwise gate; they were the 4over6 guards. I'll update this once I implement the 4over6 NVFP4.

transformer_engine/common/fused_attn/flash_attn.cu
AllenFarcas: Yes, this is a new upstream kernel, and it does use TMA.

transformer_engine/common/fused_attn_rocm/fused_attn.cpp
AllenFarcas: Added by upstream #2719 ("Add MXFP8 attention" - https://github.com/NVIDIA/TransformerEngine/commit/ee5dcec2258db9697d87c86c773a04d56b3023d7). 

The `o_format`/`qkv_scale_inv_format` params were introduced so FP8/MXFP8 attention can express output and scale-inv in a format distinct from the QKV layout. It's tied to FP8/MXFP8 attention. 

On ROCm the output is always in the Q format and the scale-inv formats are unused, so we assert `o_format == nvte_get_q_format(qkv_layout)` and ignore `qkv_scale_inv_format`.

transformer_engine/common/fused_router/utils.h
AllenFarcas: This isn't a new comment. We restored the sum()/max() template helpers dropped by upstream's v2.17 warp_reduce_on_shmem rewrite (#3012 - https://github.com/NVIDIA/TransformerEngine/commit/21ba49c59f35947a617874e5274f597d572755f9) 

`masked_warp_reduce_on_shmem` assigns them into a `T (*reduce_func)(T,T)` function pointer (`reduce_func = sum/max`), which requires real addressable functions. That template instantiates on ROCm, so without the helpers the build fails at the pointer assignment. 

The comment just documents the restored helpers.

transformer_engine/common/swizzle/swizzle.cu
AllenFarcas: This is the variable-shape grouped swizzle kernel (`grouped_swizzle_scaling_variable_shape_kernel`), not the general multi-tensor swizzle. The multi-tensor path works on ROCm; only the variable-shape grouped variant is guarded off. Uniform-shape grouped swizzle works.

transformer_engine/common/include/transformer_engine/fused_attn.h
AllenFarcas: Not supported as a backend. 

Upstream v2.17 removed the max512 backend implementation and dispatch. The `NVTE_F16_max512_seqlen = 0` enumerator is retained only in the CUDA-path enum so the ROCm fork's CUDA-path bindings/consumers (`pybind_helper.h`, `cpp_extensions/fused_attn.py`) still compile. It isn't dispatched to any kernel.

I will restore the note.

transformer_engine/pytorch/csrc/common.h
AllenFarcas: Yes, driven by upstream. v2.17 changed the base `Quantizer` virtuals. Our change just updates the ROCm `MXFP4Quantizer` override signatures to match the new base class so it compiles. No behavior change.

tests/cpp/operator/test_ck_grouped_mxfp8.cu
AllenFarcas: It's required by the v2.17 test-common API: `max_abs_bf16_tensor` reads host data via `Tensor::rowwise_cpu_dptr<bf16_t>()`, which is now a non-const accessor, so it can't be invoked on a `const Tensor&`. We dropped `const` from the helper and the two `compare_*` callers that feed it.

transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh
AllenFarcas: This path runs on gfx942 (fnuz), not just gfx1250. The arch-aware `factor_inv` is required so dequant matches the kernel's fnuz-stored (max-240) block scale; hardcoding 448 breaks the tests on gfx942.

transformer_engine/pytorch/csrc/extensions/gemm.cpp
AllenFarcas: This is a new upstream function. 

It wraps cuBLAS grouped GEMM taking a single packed `GroupedTensor` (with GEMM-swizzled scales). On ROCm we don't have that packed-grouped-tensor GEMM path, so we route grouped-tensor GEMM through the existing `te_general_grouped_gemm` (list-of-tensors API). 

API difference: upstream takes one `GroupedTensor`; the ROCm path takes a list of per-group tensors. A dedicated ROCm equivalent would only be worthwhile if the hipBLASLt grouped path adds packed-grouped-tensor support otherwise the list-based route covers the functionality.

transformer_engine/pytorch/quantized_tensor.py
AllenFarcas: ` _quantization_helpers.py` is new from upstream. Before that, those helpers lived inside `quantized_tensor.py`

The loop is import-order-sensitive: on CUDA the helper module is already loaded by the time that line runs, while on ROCm the entry order hits it while tensor is half-initialized. The fix defers the three imports to call time so they resolve after all packages finish loading.

tests/pytorch/triton_kernels/test_cast.py
AllenFarcas: Yes, current scaling keeps no scale/amax on the quantizer under the v2.17 quantizer API.

The computed scale lives on the output as `_scale_inv`. The test was updated to compare the delayed-scaling quantizer's `.scale` only where it exists and to compare current-scaling via the output's `_scale_inv`, matching the new API.

tests/pytorch/mxfp4/test_mxfp4_module_exact.py
AllenFarcas: Upstream changed the recipe quantizer "roles" API (a quantizer is created for every slot, including `role=None` for `output`/`grad_input`). 

Our `MXFP4BlockScalingRecipeState` change that adds roles mirrors that. The test factory was updated to return a quantizer for all slots so the MXFP4 reference matches the new native-recipe behavior, otherwise the reference and the module diverge.

transformer_engine/pytorch/tensor/fsdp2_allgather_tensor.py
AllenFarcas: The underlying bug is general, not ROCm-specific.

`Float8CurrentScalingQuantizer` has no `.scale` attribute, so `post_all_gather` crashed for current scaling. 

The fix threads the tensor's `scale_inv` through the all-gather metadata and is applied unconditionally (no IS_HIP gate). It surfaced on ROCm mGPU FSDP2 current-scaling tests.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
AllenFarcas: On the CPU-offload reload path, the reloaded QKV is reconstructed in a de-interleaved layout rather than the original interleaved packed layout, so `ctx.qkv_layout != original_qkv_layout`. 

On ROCm the fused-attn kernel uses a single `qkv_layout` for both the forward inputs and the backward gradients, so `dqkv` must follow that same de-interleaved layout, hence the `IS_HIP_EXTENSION && offload_reloaded` guard aligning `dqkv_layout`.

tests/pytorch/attention/test_attention.py
AllenFarcas: This comment is from upstream and my change only added a trailing period and appended the ROCm-specific note: FlashAttention is disabled for THD+padding on ROCm, so we don't force that leg; CK FusedAttention covers it.

transformer_engine/jax/cpp_extensions/attention.py
AllenFarcas: The failures aren't an upstream bug per se; they come from the interaction of an upstream feature with gfx942's CK kernel. 

SWA in striped ring context-parallel is a recent upstream feature. Under it, a striped ring step can be fully masked for a row (window excludes every key). On gfx942 the CK kernel returns non-finite output/LSE for such fully-masked tiles (gfx950 returns finite), and the online-softmax combine then computes `0 · NaN = NaN`. 

The guard neutralizes only non-finite per-step contributions (a fully-masked step must contribute zero, matching the reference which outputs 0 there), gated on `is_hip_extension()` and a no-op when values are finite, so CUDA/gfx950 numerics are untouched.

tests/pytorch/triton_kernels/test_grouped_gemm.py
AllenFarcas: This test is ROCm-specific (AITER triton grouped GEMM, from PR https://github.com/ROCm/TransformerEngine/pull/413), and the change isn't IFU-behavioral. 

The M > 1e6 skip was placed after `gen_tgmm_tensors` (which allocates ~24 GiB), so the large deepseekv2-16B shape would OOM under memory pressure before the skip ran instead of skipping. 

Moving the skip ahead of the allocation makes it skip deterministically regardless of free VRAM. Pre-existing latent bug I found during testing of the IFU and fixed.

transformer_engine/pytorch/csrc/extensions/gemm.cpp
wenchenvincent: Let's see whether NVTE uses single-kernel group gemm or multi-stream cublaslt gemm for this. And see if we can enable it on our side.

tests/pytorch/triton_kernels/test_grouped_gemm.py
wenchenvincent: @sudhu2k Could you check on this?

transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh
wenchenvincent: @matthiasdiener Did you enable the simulated nvfp4 on both gfx942 and gfx950? I think we should remove the gfx942 for the purpose of simplity.

tests/jax/test_custom_call_compute.py
wenchenvincent: Check hipCUB

tests/pytorch/distributed/test_fusible_ops.py
wenchenvincent: @wangye805 Is this TODO still applicable?

transformer_engine/common/swizzle/swizzle.cu
wenchenvincent: Double whether there is any test for it.

transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh
matthiasdiener: > @matthiasdiener Did you enable the simulated nvfp4 on both gfx942 and gfx950? I think we should remove the gfx942 for the purpose of simplity.

Yes, gfx942 support was done as part of the initial nvfp4 implementation. Feel free to remove it if it makes things easier.

build_tools/utils.py
```diff
@@ -343,6 +343,37 @@ def cuda_archs() -> str:
     return archs
 
 
+def nccl_ep_enabled(archs: str = None) -> bool:
```
ipanfilo, line 346: Better add early exit on ROCm

build_tools/VERSION.txt
```diff
@@ -1 +1 @@
-2.15.0.dev0
+2.17.0
```
ipanfilo, line 1: Looks like wrong upstream commit is used. Upstream dev branch always have .dev0 version

tests/cpp/operator/test_cast_nvfp4_transpose.cu
```diff
@@ -64,16 +72,18 @@ std::vector<InputType> create_transpose(const InputType* const input, const size
 }
 
 // Compute the global encode scale factor for a given global amax
-float compute_global_encode_scaling_factor_FP4(const float global_amax, const bool use_fast_math) {
+float compute_global_encode_scaling_factor_FP4(const float global_amax, const bool use_fast_math,
+                                               const int e4m3_max = 448) {
+  NVTE_CHECK(e4m3_max == 448 || e4m3_max == 256, "Unsupported NVFP4 E4M3 max.");
```
ipanfilo, line 77: If this method is called on gfx942, it should be called with e4m3_max == 224. Otherwise (likely because NVFP4 is not supported on gfx942) the following ifdef is not needed

tests/pytorch/distributed/test_comm_gemm_overlap.py
```diff
@@ -158,24 +190,26 @@ def _run_layer_with_overlap(
         raise AssertionError(result.stderr.decode())
 
 
+@pytest.mark.parametrize("use_cublasmp", (False, True))
```
ipanfilo: Should be only False for ROCm

tests/pytorch/distributed/test_fusible_ops.py
ipanfilo: TODO probably is not applicable  - we know why the code is different. So it should rather be other comment, similar to test_cast_master_weights_to_fp8() which later solved the same problem

tests/pytorch/nvfp4/test_nvfp4_gemm_exact.py
```diff
@@ -246,6 +468,10 @@ def check_nvfp4_gemm_versus_reference(
     ],
     ids=["rowxrow", "colxrow", "colxcol"],
 )
+@pytest.mark.parametrize("row_scaled_nvfp4", [False, True], ids=["nvfp4", "nvfp4_row_scaled"])
+@pytest.mark.parametrize("use_4over6", [False, True], ids=["default", "4over6"])
```
ipanfilo: Should be False for ROCm

tests/pytorch/nvfp4/test_nvfp4_gemm_exact.py
```diff
@@ -216,6 +258,186 @@ def check_nvfp4_gemm_versus_reference(
...
+    m_splits: list[int],
+    k: int,
+    n: int,
+    *,
+    use_bias: bool,
+    single_output: bool,
+    use_4over6: bool = False,
+    nvfp4_4over6_err_mode: str = "MAE",
+):
+    if IS_HIP_EXTENSION:
```
ipanfilo: It should rather be skipif() decorator for corresponding test

tests/pytorch/nvfp4/test_nvfp4_group_quantize.py
```diff
@@ -165,9 +189,47 @@ def test_rht_with_quantization_block_tiling_versus_reference(
...
+    # bulk_allocate_nvfp4_tensors in transformer_engine/pytorch/csrc). Without
+    # ``with_rht=True`` the flag is silently dropped, so the swizzled_sf row
+    # would just duplicate the compact_sf row — skip it instead of flooding the
+    # matrix with redundant cases.
+    if optimize_for_gemm and not with_rht:
+        pytest.skip("optimize_for_gemm requires with_rht=True (framework gate)")
+
+    # Swizzled SF is emitted only by the Blackwell RHT cast-fusion kernel; no fusion path on ROCm.
+    if IS_HIP_EXTENSION and optimize_for_gemm:
+        pytest.skip("RHT cast-fusion swizzled SF is Blackwell-only")
```
ipanfilo: The message should rather say it is not supported on ROCm

tests/pytorch/nvfp4/test_nvfp4_rht_quantize_swizzle_fusion.py
```diff
@@ -0,0 +1,297 @@
+# Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.
```
ipanfilo, line 2: Missing This file was modified for portability to AMDGPU

tests/pytorch/references/ref_per_tensor_cs.py
```diff
@@ -5,10 +5,9 @@
 # See LICENSE for license information.
 
 import torch
-import transformer_engine_torch as tex
 
 from transformer_engine.pytorch.constants import TE_DType_To_Torch
-from torch.utils.cpp_extension import IS_HIP_EXTENSION
```
ipanfilo, line 11: Is it still different from upstream?

tests/pytorch/test_numerics.py
```diff
@@ -2030,621 +2018,11 @@ def test_layernorm_mlp_accuracy_checkpoint(
...
-@pytest.mark.parametrize("num_gemms", [3, 6])
-@pytest.mark.parametrize("bs", batch_sizes)
-@pytest.mark.parametrize("model", ["126m"])
-@pytest.mark.parametrize("recipe", fp8_recipes + [None])
-@pytest.mark.parametrize("fp8_model_params", all_boolean)
-@pytest.mark.parametrize("fuse_wgrad_accumulation", all_boolean)
-@pytest.mark.parametrize("bias", all_boolean)
-@pytest.mark.parametrize("delay_wgrad_compute", all_boolean)
-@pytest.mark.parametrize("use_triton", all_boolean)
-def test_grouped_linear_accuracy(
```
ipanfilo, line 2116: It is probably moved to a new test_grouped_linear.py If so the latter should be updated with ROCm changes

tests/pytorch/test_recipe.py
```diff
@@ -510,8 +519,139 @@ def test_quantizer_update(self, module_class):
...
+        False,
+        True,
+        True,
+    ]
+
+
 @pytest.mark.skipif(not fp4_available, reason=reason_for_no_fp4)
 @pytest.mark.parametrize("dtype", [torch.float32, torch.bfloat16], ids=str)
+@pytest.mark.parametrize("row_scaled_nvfp4", [False, True], ids=["nvfp4", "nvfp4_row_scaled"])
+@pytest.mark.parametrize("use_4over6", [False, True], ids=["default", "4over6"])
```
ipanfilo, line 654: Only False for ROCm?

ipanfilo (changes requested):

build_tools/VERSION.txt
wenchenvincent: @ipanfilo Good catch!

@AllenFarcas The right commit should be this one: https://github.com/NVIDIA/TransformerEngine/commit/5d1dddfe23971a776d8cde9c1e942040b9171cb2

Did you pick the commit manually or let the skill do it?

tests/pytorch/triton_kernels/test_grouped_gemm.py
sudhu2k: There's a ticket about this issue, https://amd-hub.atlassian.net/browse/ROCM-24533
But I wasn't able to re-produce it. If @AllenFarcas was able to reproduce this, then this would be the proper fix. LGTM.

build_tools/jax.py
```diff
@@ -144,7 +144,7 @@ def setup_jax_extension(
     if bool(int(os.getenv("NVTE_WITH_CUBLASMP", 0))):
         cxx_flags.append("-DNVTE_WITH_CUBLASMP")
 
-    # NCCL EP is a CUDA/NCCL-only feature; never enabled on ROCm.
+    # Disabled on ROCm
```
wangye805, line 150: nit: usually for those nv side feature we didn't enabled during IFU, we put TODO so later grep can be easier. But if you keep a list of NV side feature then this is also fine

ci/pytorch.sh
```diff
@@ -64,7 +64,7 @@ run_test_config(){
     run 1 test_jit.py
     NVTE_ROCM_ENABLE_MXFP8=1 run_default_fa 1 test_multi_tensor.py
     run 1 test_numerics.py
-    NVTE_ROCM_ENABLE_MXFP8=1 run_default_fa_lbl "mxfp8" 1 test_numerics.py -k "recipe0 and 126m and not grouped"
+    NVTE_ROCM_ENABLE_MXFP8=1 run_default_fa_lbl "mxfp8" 1 test_numerics.py -k "MXFP8BlockScaling and 126m and not grouped"
```
wangye805: Why the model 126m cannot run mxfp8 recipe?

tests/cpp/operator/test_cast_mxfp4_transpose.cu
```diff
@@ -297,7 +297,7 @@ void print_detailed_tensor_comparison(const std::string& name,
     printf("==================================\n");
 }
 
-void compareResults_mxfp4(const Tensor &test,
+void compareResults_mxfp4(Tensor &test,
```
wangye805, line 300: Why previously we can use const Tensor& but now need Tensor&? do we edit some contents in test

tests/pytorch/distributed/fsdp2_tests/run_fsdp2_fused_adam.py
```diff
@@ -735,6 +735,19 @@ def test_dcp_output_parity(recipe_name, async_save):
             "requires using power of two scaling factors."
         )
 
+    if recipe_name == "NVFP4BlockScaling" and async_save and IS_HIP_EXTENSION:
```
wangye805: Is this a new test from upstream? If so, can you file a ticket and put a TODO comment with this link here

tests/pytorch/distributed/fsdp2_tests/run_fsdp2_mem_leak.py
```diff
@@ -272,6 +272,21 @@ def test_fp8_temp_accumulation_across_layers(recipe_name, quantized_model_init):
     """
     _maybe_skip(recipe_name, quantized_model_init)
 
+    if (
```
wangye805: Need similar todo to github ticket for this fsdp2 issue

tests/pytorch/distributed/test_fusible_ops.py
wangye805: Probably the todo here is stale. @AllenFarcas can you take a quick try? Thanks

transformer_engine/common/cast/nvfp4/rowwise_amax_nvfp4.cuh
wangye805, line 1: Why this file looks like a new file?

transformer_engine/jax/cpp_extensions/attention.py
wangye805: I think this issue is worth a new git ticket for fully understanding. 

If it's a CK issue in gfx942, we may need to report to CK team.

transformer_engine/jax/csrc/extensions/topk.cpp
```diff
@@ -12,9 +13,7 @@
 namespace transformer_engine {
 namespace jax {
 
-// nvte_topk (common/util/topk.cu) is CUDA-only: standalone_topk.cuh relies on
-// cub internals and cooperative_groups::reduce, which have no HIP equivalent, so
```
wangye805: Try hipcub

transformer_engine/pytorch/tensor/storage/nvfp4_tensor_storage.py
```diff
@@ -51,10 +52,18 @@ def forward(
...
 
-        # ``tex.dequantize`` requires CUDA-resident buffers. If the tensor has
+        # ``tex.dequantize`` requires CUDA-resident buffers. Move the raw buffers
+        # to CUDA via the metadata dict rather than the NVFP4 subclass's ``.to()``:
+        # the latter re-enters __torch_dispatch__ -> dequantize -> ``.to()`` and
+        # recurses infinitely when the tensor is CPU-resident (FSDP2 DCP staging).
         src_device = tensor.device
         if src_device.type != "cuda":
-            cuda_tensor = tensor.to(device=torch.device("cuda"))
+            cuda = torch.device("cuda")
```
wangye805, line 62: Is this a bug fix over NV side?

transformer_engine/pytorch/triton_kernels/cast.py
```diff
@@ -25,14 +25,14 @@ def _empty_tensor() -> torch.Tensor:
 def _setup_conditional_transpose_storage(
         tensor: QuantizedTensor,
     ) -> QuantizedTensor:
-        shape = tensor.shape
+        shape = tensor.size()
```
wangye805, line 28: Why the original .shape need to be changed to .size()?

wangye805 (changes requested):

tests/pytorch/distributed/test_fusible_ops.py
AllenFarcas: Confirmed stale. The upstream form works on ROCm, verified by `test_distributed_fuser_ops` passing and it matches `test_cast_master_weights_to_fp8`, which already aligned. Removed the TODO and aligned the two lines.

transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh
AllenFarcas: Removed NVFP4 from gfx942 for simplicity and tests are passing.

tests/cpp/operator/test_cast_nvfp4_transpose.cu
AllenFarcas: Removed it.

The reference uses the passed e4m3_max uniformly. Also gated this cast test off gfx942 so the 448 reference never runs against the fnuz kernel.

tests/pytorch/distributed/test_comm_gemm_overlap.py
AllenFarcas: Gated `use_cublasmp` to `(False,) if IS_HIP_EXTENSION else (False, True)` on all six comm/gemm-overlap test parametrizations. On ROCm only the `False` cases are generated; CUDA still gets `False+True`.

build_tools/utils.py
AllenFarcas: Added

tests/pytorch/nvfp4/test_nvfp4_gemm_exact.py
AllenFarcas: Adjusted accordingly

tests/pytorch/nvfp4/test_nvfp4_gemm_exact.py
AllenFarcas: Added the skipif() for the corresponding test `test_nvfp4_row_scaled_grouped_gemm_matches_per_gemm`

tests/pytorch/nvfp4/test_nvfp4_group_quantize.py
AllenFarcas: Corrected. Applied the same correction to the sibling module-level skip in `test_nvfp4_rht_quantize_swizzle_fusion.py` for consistency.

tests/pytorch/nvfp4/test_nvfp4_rht_quantize_swizzle_fusion.py
AllenFarcas: Added

tests/pytorch/references/ref_per_tensor_cs.py
AllenFarcas: Good catch, it wasn't actually different from upstream anymore. I removed the AMD header.

tests/pytorch/test_numerics.py
AllenFarcas: Restored the ROCm use_triton coverage that was dropped when this moved from test_numerics.py. Also added a skip for row-scaled NVFP4 grouped GEMM on ROCm (hipBLASLt has no algorithm for it) in skip_unsupported_backward_override.

tests/pytorch/test_recipe.py
AllenFarcas: This is a quantize/dequantize round-trip with no GEMM, and 4over6 dequantize is supported on ROCm. The `[False, True]` cases pass on gfx950 (gfx942 skips via fp4_available). The [False]-only restriction only applied to 4over6 GEMM in `test_nvfp4_gemm_exact.py`. No change needed.

AllenFarcas:
@ipanfilo @wangye805  related to your comment on `tests/pytorch/distributed/test_fusible_ops.py` line 1049 "TODO probably is not applicable - we know why the code is different. So it should rather be other comment, similar to `test_cast_master_weights_to_fp8()` which later solved the same problem"

I cannot comment there but ill add here my reply: I replaced the TODO with an explanatory comment in the `# ROCm:` style used in `test_cast_master_weights_to_fp8.py`: `# ROCm: fnuz FP8 (gfx942) caps E4M3 at 240; OCP (gfx950+/CUDA) caps at 448.`, so the reference max must branch on `is_fp8_fnuz()`. Also removed an unused `transformer_engine_torch` import left behind.

build_tools/jax.py
AllenFarcas: I enriched the jax.py comment to explain the gate, matching the fuller comment already in `build_tools/pytorch.py`, and kept `# Disabled on ROCm` as the grep marker.

ci/pytorch.sh
AllenFarcas: 126m does run mxfp8

The parametrize now uses `ids=recipe_id` which produces a readable name based id from the recipe class instead of `recipe0`, so the first recipe now shows up as `MXFP8BlockScaling`. I updated the filter from `recipe0` to `MXFP8BlockScaling` so it keeps selecting the same MXFP8 cases.

tests/cpp/operator/test_cast_mxfp4_transpose.cu
AllenFarcas: Upstream dropped the `const` qualifier from `to_cpu()` and the `*_cpu_dptr()` accessors in `tests/cpp/test_common.h`

`compareResults_mxfp4` calls those on test, so once they became non-const the `const Tensor&` parameter stopped compiling and had to become `Tensor&`.

transformer_engine/jax/csrc/extensions/topk.cpp
AllenFarcas: Will implement as a feature after the IFU is merged.

transformer_engine/common/cast/nvfp4/rowwise_amax_nvfp4.cuh
AllenFarcas: Row-scaled NVFP4 is new in upstream v2.17, so the IFU pulled it in and ROCm needed the row-wise amax to support it but upstream keeps that kernel inside the Blackwell-only `quantize_transpose_nvfp4.cuh`, which doesn't compile on ROCm, so instead of taking it as-is we extracted the portable amax into this new header.

transformer_engine/pytorch/tensor/storage/nvfp4_tensor_storage.py
AllenFarcas: Yes, bug fix over NV. Upstream's `tensor.to("cuda")` on the NVFP4 subclass re-enters `__torch_dispatch__` -> `dequantize` -> `.to()`, recursing infinitely when running the `test_dcp_output_parity[NVFP4BlockScaling-sync]` test. 

We move the raw buffers via the metadata `dict` and rebuild `NVFP4TensorStorage` directly to avoid the recursion. Worth upstreaming.

transformer_engine/pytorch/triton_kernels/cast.py
AllenFarcas: `tensor` is a quantized storage subclass (NVFP4/MXFP8) which overrides size() to infer the true logical shape from its raw packed buffers, while `.shape` isn't overridden, so it returns the raw packed storage dims and can be wrong or warn when only partial data is present. 

`_setup_conditional_transpose_storage` needs the logical shape to size the transpose buffer, so we switched to the storage-safe `.size()` (and `len(shape)` instead of `.ndim`). 

When running `test_quantize` in `tests/pytorch/triton_kernels/test_cast.py` I caught this.

tests/pytorch/distributed/fsdp2_tests/run_fsdp2_fused_adam.py
AllenFarcas: Working on it right now, will push an updated commit soon.

tests/pytorch/distributed/fsdp2_tests/run_fsdp2_mem_leak.py
AllenFarcas: Working on it right now, will push an updated commit soon.

ci/pytorch.sh
ipanfilo, line 1: Look at pytorch.sh changes in release_v2.15_rocm, you might need the same here

tests/pytorch/attention/test_attention_with_cp.py
```diff
@@ -372,25 +624,62 @@ def test_cp_with_fused_attention(
...
+        pytest.skip("Deterministic mode does not support post_scale_bias with requires_grad")
+    # Observed: cuDNN det THD backward asks for ~128 * bHSS bytes of workspace
+    # on sm90; at 1<<30 that's 128 GiB, won't fit on H100's 80 GB. Held exactly
+    # at b=2 + power-of-2 S in our sweep; for b>=3 the workspace was observed to
+    # grow super-linearly (b=4 took ~4x the b=2 amount, not 2x) — revisit if a
+    # config uses b>2.
+    SM90_DET_FUSED_THD_BWD_MAX_BHSS = 1 << 30
+    if (
+        _deterministic
+        and qkv_format == "thd"
```
ipanfilo, line 661: nit: add and not IS_HIP_EXTENSION

tests/pytorch/distributed/test_fusible_ops.py
```diff
@@ -1046,9 +1046,8 @@ def run_parallel_tests() -> None:
 @pytest.mark.parametrize("world_size", _world_sizes)
 def test_distributed_fuser_ops(world_size: int) -> None:
     """Launch parallel job that runs parallel tests"""
-    #TODO: find out why cannot align the following two lines with NV upstream
-    python_exe = sys.executable
-    current_file = os.path.abspath(__file__)
+    python_exe = pathlib.Path(sys.executable)
+    current_file = pathlib.Path(__file__).resolve()
```
ipanfilo, line 1050: If it is different from upstream, add comment saying why

tests/pytorch/test_grouped_linear.py
ipanfilo, line 1: AMD copyright is missing

ipanfilo, line 1: The test needs to be added to pytorch.sh

transformer_engine/common/cast/nvfp4/rowwise_amax_nvfp4.cuh
```diff
@@ -0,0 +1,119 @@
+/*************************************************************************
+ * Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.
```
ipanfilo, line 2: Missing header line about modificaton

wangye805 (approved):

ipanfilo (commented):
Number of files have AMD copyright header incorrect format

transformer_engine/pytorch/quantized_tensor.py
ipanfilo: So on ROCm _quantized_helpers import quntized_tensor? Why?

transformer_engine/pytorch/quantization.py
```diff
@@ -96,9 +192,9 @@ def _compute_nvfp4_support() -> Tuple[bool, str]:
     """Return if nvfp4 support is available"""
     if IS_HIP_EXTENSION:
         gpu_arch = get_device_compute_capability()
-        if gpu_arch in ((9, 4), (9, 5)):  # TODO: enabled for gfx1250 when ready
+        if gpu_arch in ((9, 5),):  # gfx950; TODO: enable for gfx1250 when ready
```
ipanfilo, line 194: While it is expected to remove NVFP4 from gfx942, is there a reason to combine this change with IFU?

AllenFarcas: gfx942 NVFP4 was simulation-only (fnuz E4M3 max is 240/256, not 448) and the v2.17 NVFP4 rework broke it. I tried fnuz workarounds to keep it alive, but for simplicity it is better to remove gfx942 NVFP4 here.

AllenFarcas:
@ipanfilo In response to your question about `_quantized_helpers`: the import was removed because it was the load-time dependency creating the quantized_tensor -> tensor/__init__ -> quantized_tensor cycle on ROCm. 

The cycle is ROCm-specific because `init_mxfp4_extension()` (guarded by `#ifdef USE_ROCM` in pybind) imports `mxfp4_tensor` -> `triton_kernels.cast` -> `quantized_tensor` at extension load, entering it before the tensor package; on CUDA that block is compiled out.

transformer_engine/common/cast/nvfp4/rowwise_amax_nvfp4.cuh
AllenFarcas: Added

tests/pytorch/distributed/test_fusible_ops.py
AllenFarcas: The previous comment was a TODO to find out why we cannot allow with upstream, but right now that is solved and the change reflects upstream exactly.

tests/pytorch/test_grouped_linear.py
AllenFarcas: Added

AllenFarcas: Added

tests/pytorch/attention/test_attention_with_cp.py
AllenFarcas: Added

AllenFarcas:
@ipanfilo I fixed all remaining AMD copyright and added any missing ones.

ci/pytorch.sh
AllenFarcas: release_v2.15_rocm sets `export NVTE_GROUPED_LINEAR_SINGLE_PARAM=1` in `run_test_config` while the v2.17 merge dropped it, so the single-param `GroupedLinear` cases were skipping on ROCm CI. 

Restored the export in the same place.

transformer_engine/common/cast/nvfp4/core_nvfp4.cuh
```diff
@@ -75,10 +77,14 @@ namespace core {
...
-// Compute the global encode scale factor for a given global amax
+// Compute the global encode scale factor for a given global amax.
+// NVFP4 uses the full E4M3 range by default. Some 4over6 tensors dispatch
+// E4M3_MAX=256 to leave room for map-to-4 scale expansion.
+template <int E4M3_MAX = 448>
 __device__ __forceinline__ float compute_global_encode_scaling_factor_FP4(const float global_amax) {
   using namespace detail;
-  constexpr float fp8_max = TypeExtrema<fp8e4m3>::max;  // 448.0f;
+  static_assert(E4M3_MAX == 448 || E4M3_MAX == 256, "Unsupported NVFP4 E4M3 max.");
+  constexpr float fp8_max = static_cast<float>(E4M3_MAX);
```
ipanfilo, line 85: Are any of those changes ROCm specific or all come from upstream?

ipanfilo:
> @ipanfilo I fixed all remaining AMD copyright and added any missing ones.
It doesn't seem so. There are still files with wrong copyright header format

transformer_engine/common/cast/nvfp4/core_nvfp4.cuh
AllenFarcas: I reverted the copyright since this file is all from upstream.

AllenFarcas:
@ipanfilo During my check of copyright I found a couple of files not touched by this IFU but which had AMD content but the copyright was missing: `test_cast_transpose_current_scaling.cu`, `tests/jax/test_layer.py`, `jax/csrc/extensions/quantization.cpp`,  `pytorch/csrc/pybind.h`

I checked again and now everything else should be clean with the copyrights.

transformer_engine/pytorch/quantized_tensor.py
AllenFarcas: No, `_quantization_helpers` doesn't import `quantized_tensor` at all, on ROCm or CUDA. The only reference is a TYPE_CHECKING type hint (helpers.py L15-16), which is identical on both platforms and never runs. So there's nothing ROCm-specific here. 

The circular import came from the reverse edge: `quantized_tensor.py`'s top-level import of `_quantization_helpers` pulled in tensor/__init__.py, which imports back from `quantized_tensor`. 

Moving those imports to function scope fixed it.

ipanfilo:
> > @ipanfilo I fixed all remaining AMD copyright and added any missing ones.
> > It doesn't seem so. There are still files with wrong copyright header format

By contract AMD modified files have language about it at copyright header. You modified files and added copyright but missed that language. 
Also comment about NV copyright in AMD originated file is not addressed

AllenFarcas:
@ipanfilo added the "This file was modified for portability to AMDGPU" line to all the modified-upstream files where it was missing. 

On `rowwise_amax_nvfp4.cuh`, as it's a separate file on our side soI left it with AMD only copyright.

ipanfilo (approved):

wenchenvincent (approved):

PR #661, opened 2026-07-07
Grouped MXFP8 GEMMs with HipKittens

alextmagro:
Modifies HipKittens MXFP8 GEMM kernels to handled single launch grouped GEMM. Usage is enabled through NVTE_USE_CUTLASS_GROUPED_GEMM, similar to CK FP8


transformer_engine/common/gemm/kittens/mxfp8_gemm.cpp
```diff
@@ -54,6 +54,22 @@ enum struct OutDtype {
...
 };
 
+constexpr int MAX_EXPERTS = 512;
+
+struct TileOffsets {
+    int data[MAX_EXPERTS + 1];
+};
+
+template <typename T>
+__device__ __forceinline__ int rocm_upper_bound(const T *arr, int n, T val) {
```
matthiasdiener: Isn't this function in rocm_device_utils.cuh?

matthiasdiener:
> Usage is enabled through NVTE_USE_CUTLASS_GROUPED_GEMM, similar to CK FP8

This probably needs an update in the Readme, hipkittens is essentially "overriding" the CK grouped Gemm (which is also used for bf16 I believe).

transformer_engine/common/gemm/kittens/mxfp8_gemm.cpp
```diff
@@ -113,22 +129,16 @@ __device__ __forceinline__ void gemm_epilogue(
...
-    const AuxGLType AuxGL,
-    const gl_scale_rt scale_A_gl,
-    const gl_scale_rt scale_B_gl,
-    const void *__restrict__ bias,
-    int bias_dtype,
-    int M, int N, int K) {
+void mxfp8_gemm_tn_kernel(const gl_fp8_rt A, const gl_fp8_rt B, const OutGL C, const AuxGLType AuxGL,
+    const gl_scale_rt scale_A_gl, const gl_scale_rt scale_B_gl,
+    [[maybe_unused]] const void *__restrict__ bias, [[maybe_unused]] int bias_dtype,
+    [[maybe_unused]] TileOffsets tile_offsets, [[maybe_unused]] int num_experts,
```
matthiasdiener: tile_offsets is passed by value into each launch, including the non-grouped launches via dummy.

Not sure if it is an issue, but could we maybe template this argument out, or pass a pointer instead?

transformer_engine/common/gemm/kittens/mxfp8_gemm.cpp
```diff
@@ -1101,4 +1142,87 @@ bool kittens_mxfp8_gemm(
...
+bool kittens_grouped_mxfp8_gemm(
+    const void *const *A_array, const void *const *B_array, void *const *C_array,
+    const void *const *scale_A_array, const void *const *scale_B_array,
+    const int *M_array, int N, int K, int num_experts,
+    bool transa, bool transb, int a_dtype, int b_dtype, int out_dtype,
+    void *workspace, size_t workspace_size, hipStream_t stream) {
+
+    if (transa && transb) return false;
+    if (N % BLOCK_COL != 0 || K % BLOCK_K != 0 || K < 256) return false;
+    if (num_experts <= 0 || num_experts > MAX_EXPERTS) return false;
```
aris134: Would it be useful to optionally emit warning messages on the reason for fallback, similarly to the CK setup?

transformer_engine/common/gemm/kittens/mxfp8_gemm.cpp
```diff
@@ -1101,4 +1142,87 @@ bool kittens_mxfp8_gemm(
...
+
+    int tiles_N = N / BLOCK_COL;
+    int k_iters = K / BLOCK_K;
+    int scale_K = K / 32;
+
+    TileOffsets tile_offsets;
+    int total_M = 0;
+    int total_m_tiles = 0;
+    for (int g = 0; g < num_experts; g++) {
+        if (M_array[g] % BLOCK_ROW != 0) return false;
```
aris134: Same comment here regarding warning emission. Given that the CK grouped GEMM has padding support: https://github.com/ROCm/TransformerEngine/blob/71eb59d645d81877f5e8fff608ee024bd40fa0ea/transformer_engine/common/gemm/ck_grouped_gemm/ck_grouped_gemm_fp16.cpp#L14-L41

would it make sense to have that for HK as well? Otherwise my question would be how frequently does HK grouped MXFP8 fallback for Qwen or DS runs?

transformer_engine/common/gemm/kittens/mxfp8_gemm.cpp
```diff
@@ -54,6 +54,22 @@ enum struct OutDtype {
     FP16,
 };
 
+constexpr int MAX_EXPERTS = 512;
+
+struct TileOffsets {
```
aris134: Would it make sense to carve tile_offsets out of the workspace?

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -2040,4 +2040,70 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
 
 #pragma GCC diagnostic pop
 
+#ifdef USE_HIPKITTENS_GEMM
+bool try_kittens_grouped_mxfp8_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor *D,
+    int num_gemms, bool transa, bool transb, NVTETensor *workspace,
+    bool accumulate, cudaStream_t stream) {
+    if (accumulate || num_gemms <= 1) return false;
```
aris134, line 2133: Again here, an optional warning message on unsupported accumulate option would be helpful

transformer_engine/common/gemm/kittens/cdna4/mxfp8_gemm.cpp
aris134, line 1: numerics tests for the HK grouped GEMM?

aris134, line 1: might also be a good idea to update the relevant README section https://github.com/ROCm/TransformerEngine/blob/dev/README.rst#grouped-gemm-using-ck_tile

aris134 (changes requested):

aris134:
Do we have any performance numbers? I know there isn't a CK MXFP8 group GEMM for gfx950, but do we have hipBlaslt or triton baselines?

transformer_engine/common/gemm/kittens/mxfp8_gemm.cpp
```diff
@@ -196,38 +218,38 @@ void mxfp8_gemm_tn_kernel(
...
-    G::load(As[tic][0], A, {0, 0, block_row * 2, 0}, sw_A, a_srd, a_base, a_lds[tic][0]);
-    G::load(Bs[tic][1], B, {0, 0, block_col * 2 + 1, 0}, sw_B, b_srd, b_base, b_lds[tic][1]);
-    G::load(As[tic][1], A, {0, 0, block_row * 2 + 1, 0}, sw_A, a_srd, a_base, a_lds[tic][1]);
+    G::load(Bs[tic][0], B, {0, 0, b_half0, 0}, sw_B, b_srd, b_base, b_lds[tic][0]);
+    G::load(As[tic][0], A, {0, 0, a_half0, 0}, sw_A, a_srd, a_base, a_lds[tic][0]);
+    G::load(Bs[tic][1], B, {0, 0, b_half1, 0}, sw_B, b_srd, b_base, b_lds[tic][1]);
+    G::load(As[tic][1], A, {0, 0, a_half1, 0}, sw_A, a_srd, a_base, a_lds[tic][1]);
 
     if (warp_m == 1) __builtin_amdgcn_s_barrier();
     asm volatile("s_waitcnt vmcnt(4)");
```
aris134: Consider replacing the magic vmcnt values with named constants. It's currently difficult to tell why the hard-coded values were chosen.

transformer_engine/common/gemm/kittens/mxfp8_gemm.cpp
```diff
@@ -264,7 +286,7 @@ void mxfp8_gemm_tn_kernel(
         __builtin_amdgcn_s_barrier();
         __builtin_amdgcn_sched_barrier(0);
 
-        G::load(Bs[tic][1], B, {0, 0, block_col * 2 + 1, k + 2}, sw_B, b_srd, b_base, b_lds[tic][1]);
+        G::load(Bs[tic][1], B, {0, 0, b_half1, k + 2}, sw_B, b_srd, b_base, b_lds[tic][1]);
         asm volatile("s_waitcnt vmcnt(6)");
```
aris134: For example here, I think the vmcnt(6) here means you're retiring k+1 scales and As[toc][1] k+1, but using named constants here could confirm

transformer_engine/pytorch/quantization.py
```diff
@@ -220,6 +220,8 @@ def get_default_recipe() -> Recipe:
 def get_align_size_for_quantization(recipe: Recipe) -> int:
     """Get the alignment size for quantization."""
     if recipe.mxfp8():
+        if IS_HIP_EXTENSION and get_device_compute_capability() == (9, 5) and os.environ.get("NVTE_USE_CUTLASS_GROUPED_GEMM", "0") == "1":
```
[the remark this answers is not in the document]
alextmagro: NVTE_USE_HIPKITTENS_GROUPED_GEMM added as an alternative. C++ side check added as well.

transformer_engine/common/gemm/kittens/mxfp8_gemm.cpp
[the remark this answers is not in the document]
alextmagro: static_asserts added

transformer_engine/common/gemm/kittens/cdna4/mxfp8_gemm.cpp
alextmagro: I have added the info to the README. Additionally, I have made it possible for users to use either CK or HIPKITTENS without enabling NVTE_USE_CUTLASS_GROUPED_GEMM.

alextmagro: Numeric tests are covered by test_grouped_linear_accuracy_cutlass and other grouped tests

transformer_engine/common/gemm/kittens/mxfp8_gemm.cpp
alextmagro: Good idea. I have done that and combined it into other variables passed to the kernel itself.

transformer_engine/common/gemm/kittens/mxfp8_gemm.cpp
alextmagro: Yes, however the kittens folder is built as a separate shared library that doesn't reference TransformerEngine, so we need a copy of this function here.

transformer_engine/common/gemm/kittens/mxfp8_gemm.cpp
alextmagro: Addressed this with the changes to how tileOffsets is used.

transformer_engine/common/gemm/kittens/mxfp8_gemm.cpp
alextmagro: I don't think this is the right approach, as the number is not so much a magic number, but depends on how many G::load calls are in flight at each point. Named constants would not hold if the kernel changes.

transformer_engine/common/gemm/kittens/mxfp8_gemm.cpp
alextmagro: I see you use the NVTE_CUTLASS_GROUPED_GEMM_WARN_FALLBACK flag to enable these warnings, so I have added them similarly.

transformer_engine/common/gemm/kittens/mxfp8_gemm.cpp
alextmagro: In practice, I don't think we would ever fall back for Qwen and DS as expert weights are all 256 aligned. Activations are padded already by Megatron, so no issue there either.

transformer_engine/common/gemm/kittens/cdna4/mxfp8_gemm.cpp
aris134: Can you double check test_grouped_linear_accuracy_cutlass? I ran it on gfx950 with:

NVTE_ROCM_ENABLE_MXFP8=1 NVTE_USE_HIPKITTENS_GROUPED_GEMM=1 NVTE_USE_CUTLASS_GROUPED_GEMM=0 NVTE_CUTLASS_GROUPED_GEMM_WARN_FALLBACK=1 pytest tests/pytorch/test_numerics.py::test_grouped_linear_accuracy_cutlass -vv -s

and am seeing assertion failures from the numerical correctness checks when comparing HK outputs against the reference implementation.

e.g.

Mismatched elements: 1559808 / 1572864 (99.2%)
Greatest absolute difference: 2.58984375 at index (1792, 0, 751) (up to 0.001 allowed)
Greatest relative difference: 9792.0 at index (0, 0, 148) (up to 0.008 allowed)
FAILED tests/pytorch/test_numerics.py::test_grouped_linear_accuracy_cutlass[hipkittens-True-True-recipe0-True-126m-1-6-torch.bfloat16] - AssertionError: Tensor-likes are not close!

transformer_engine/common/gemm/kittens/mxfp8_gemm.cpp
aris134: Right, but I don't think it's strictly the number of G::load calls themselves. The required vmcnt value is really tied to the underlying VMEM operations generated by those loads, which can vary depending on how the backend lowers them (e.g., vector load width per thread).

That gets back to my readability concern. I'm not necessarily advocating for named constants, but it might be helpful to add a brief comment describing what each non-zero vmcnt is waiting on at the pipeline level and why newer loads are allowed to remain in flight.

transformer_engine/common/gemm/kittens/mxfp8_gemm.cpp
aris134: or just describing in a comment as I said above

aris134 (changes requested):

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -2040,4 +2040,117 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
...
+
+        a_ptrs[i]  = a_data;
+        b_ptrs[i]  = b_data;
+        c_ptrs[i]  = tD->data.dptr;
+        sa_ptrs[i] = a_scale;
+        sb_ptrs[i] = b_scale;
+        n_arr[i]   = n_i;
+    }
+
+    // Activation data (B), activation scales (SB), and output (D) must be contiguous.
```
wangye805, line 2207: For main stream MOE models, permuted tokens will still be guaranteed to be continuous but usually the scales is hard to guarantee to be still continuous, especially after padding. 

So what's the motivation of this continuous assumption? Without it, the hipkitten kernel will drop a lot? Does our CK grouped gemm need this assumption?

README.rst
```diff
@@ -208,6 +208,22 @@ If you want to check that only previously tuned algorithms are used by your appl
   diff algo_tune.csv algo_tune_check.csv
 
 
+Grouped GEMM Backends on ROCm
```
wangye805: Technically multi-stream based hipblaslt/hipkitten is another backend. Maybe we should say explicitly "single-launch" grouped gemm in the section title

README.rst
```diff
@@ -208,6 +208,22 @@ If you want to check that only previously tuned algorithms are used by your appl
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+For Mixture of Experts workloads, ROCm TE supports single-launch grouped GEMMs. Two backends are available:
+
+* **HipKittens** -- MXFP8 grouped GEMM only, requires expert dimensions to be 256-aligned. Default when the grouped GEMM path is enabled on gfx950.
+* **CK** (Composable Kernel) -- Grouped GEMM used as a fallback when HipKittens constraints are not met, or when explicitly selected.
+
+The following environment variables control backend selection:
+
+* ``NVTE_USE_CUTLASS_GROUPED_GEMM=1`` -- Enable the grouped GEMM path. On gfx950, defaults to HipKittens with CK fallback.
```
wangye805: Is this env by default 1?

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -45,7 +45,7 @@ def get_cublas_workspace_size_bytes() -> None:
     if IS_HIP_EXTENSION:
         """Return 64 MiB for gfx50x, 32 MiB for all other architectures."""
         if get_device_compute_capability() == (9, 5):
-            return 67_108_864
+            return 134_217_728
```
wangye805: Emm, this workspace size was originally for hipblaslt

transformer_engine/common/gemm/rocm_gemm.cu
```diff
@@ -2040,4 +2041,216 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
...
+        sa_ptrs[i] = a_scale;
+        sb_ptrs[i] = b_scale;
+        m_arr[i]   = M_i;
+    }
+
+    if (ref_n < 0) return false;
+
+    auto *ws = convertNVTETensorCheck(workspace[0]);
+
+    return kittens_grouped_mxfp8_wgrad(
```
wangye805, line 2331: Just curious, does hipkitten wgrad need continuous b and b scale requirement?

transformer_engine/common/gemm/kittens/cdna4/mxfp8_gemm.cpp
```diff
@@ -1085,20 +1167,559 @@ bool kittens_mxfp8_gemm(
...
+                        workspace, workspace_size, stream);
+                } else {
+                    assert(0 && "kittens_mxfp8_gemm: TT layout is not supported");
+                }
+    ))) // NOLINT(*)
+    return result;
+}
+
+static size_t align256(size_t x) {
+    return (x + 255) & ~(size_t)255;
```
wangye805, line 1362: nit: since hipkittens kernels are a standalone lib.so, we can use its own env prefix, like HK_. Or still use NVTE_ but pass it from libte.so into libhipkitten.so

transformer_engine/common/gemm/kittens/cdna4/mxfp8_gemm.cpp
```diff
@@ -1085,20 +1167,559 @@ bool kittens_mxfp8_gemm(
...
+    const void *b_ptr;
+    void       *d_ptr;
+    int         k_iters;
+    int         sa_tile_offset;
+    int         sb_tile_offset;
+};
+
+template <bool ACCUMULATE, typename OutGL>
+__global__ __launch_bounds__(NUM_THREADS, 2)
+void mxfp8_wgrad_nt_kernel(
```
wangye805, line 1532: Emm, is this wgrad_nt_kernel based on/modified from regular nt_kernel? If so, can you have some comments on how the new one is derived. 

If this is a complete new one, maybe have some tiling + pipelining comments or a pointer to our internal doc?

wangye805 (changes requested):
in the PR description, add the ticket or confluence doc link for performance

transformer_engine/common/gemm/kittens/cdna4/mxfp8_gemm.cpp
alextmagro: There was a bug with non-contiguous inputs, that should be fixed now.

transformer_engine/common/gemm/kittens/mxfp8_gemm.cpp
alextmagro: I've added some comments, hopefully it clears up what is in flight

transformer_engine/common/gemm/kittens/cdna4/mxfp8_gemm.cpp
alextmagro: Sounds good, I have added hk prefix. I am hoping to avoid introducing a dependency on libte to keep cmake/builds clean

transformer_engine/common/gemm/kittens/cdna4/mxfp8_gemm.cpp
alextmagro: Done. This is similar to our regular nt kernel but the varying K dimension means that templating was getting messy so I separated it out. Also, a 1D grid performed better for the nature of expert wgrad where k is relatively small to m and n.

transformer_engine/common/gemm/rocm_gemm.cu
alextmagro: This was simply a simplification, but you are right that if data is not 256 aligned and scales are not 16 byte aligned then we are unable to use this method. Since we don't do padding for rocm though, generally I would expect us to be okay for mainstream MoE models like you mention.

A note on performance though, if we guarantee that all of output, activations, scales, and weights are contiguous we see about a 20-25% increase in performance for these grouped GEMM kernels. Currently coming from Megatron, only weights are non-contiguous after quantization. I have some ideas on how we can make that possible, or even better, that I am planning on working on in a followup ticket.

transformer_engine/pytorch/cpp_extensions/gemm.py
alextmagro: Sorry, I was doing some experiments and forgot to revert this. Should be back to 64 MB now.

transformer_engine/common/gemm/rocm_gemm.cu
alextmagro: No, as this kernel was designed fully per-expert due to the varying K dimension.

README.rst
alextmagro: I think technically speaking multistream is not really a grouped gemm, but a set of gemms... -- I see where you are coming from though and have added that clarification!

README.rst
alextmagro: No, this is unset by default. I have added a note at the top of this subsection.

transformer_engine/common/gemm/rocm_gemm.cu
wangye805: Two questions after your scale swizzling commit: 
1). Does scale swizzling recover the 20%-25% performance under the non-continuous scale scenario? Or they are orthogonal?
2). What's the performance comparison btw your hipkitten grouped gemm and ck grouped under the non-continuous token/scales scenario? If it's still better, then I think we can first enable it without such continuity check?

transformer_engine/pytorch/cpp_extensions/gemm.py
wangye805: In the PR it was not reverted :-)

wangye805 (approved):

aris134 (approved):

transformer_engine/common/gemm/rocm_gemm.cu
alextmagro: Good questions:

1) No, they are orthogonal. Scale swizzling helps out with VALU/register reduction, whereas the non-contiguous data issue is more of a cache hit rate issue I am guessing.
2) I don't believe we have performance results for CK MXFP8 grouped GEMM on MI350s, my understanding was that was not enabled yet. I could be wrong about that now though.

wangye805: Well, "CK mxfp8 grouped gemm not enabled" exactly strengthen the point that we should enable hipkitten grouped gemm for non-contiguous scenarios

alextmagro: That's true, I can work on that in a follow-up PR. I also want to explore the performance differences between having a non-contiguous kernel and a copy-to-contiguous + contiguous kernel as the latter might be more performant overall

PR #662, opened 2026-07-07
Added AITER V3 API check mechanism

Micky774:
# Description

Added mechanism to check for V3 API usage for future flexibility.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
+++ b/transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
@@ -441,20 +441,16 @@ void dump_bwd_timings(const char* dump_path, float average_runtime){
   file << average_runtime << "\n";
 }
 
-hipError_t ck_attn_bwd(const CkAttnBwdArgs& args, hipStream_t stream){
+// Populate the AITER mha_bwd_args from TE's CkAttnBwdArgs. Shared by ck_attn_bwd
+// (real launch) and ck_attn_bwd_uses_v3 (v3 availability probe) so the probe can
+// never disagree with the launch. v3_api_check is left false here; callers flip it.
+// The stream-dependent max_seqlen override (NVTE_CK_RUNTIME_MAX_SEQLEN) is applied
+// by ck_attn_bwd after this returns; it does not affect v3 kernel selection.
+static aiter::mha_bwd_args build_bwd_fmha_args(const CkAttnBwdArgs& args){
 
   bool has_dropout = (args.dropout_probability > 0.f);
   bool has_dbias = args.dbias_ptr != nullptr;
   bool is_mqa_gqa = (args.h > args.hg);
-
-  bool ck_log_config = false;
-  if (const char* env_p = std::getenv("CK_FUSED_ATTN_LOG_CONFIG") ) {
-    if (env_p != nullptr && std::string(env_p) == "1")
-      ck_log_config = true;
-  }
-  const char* dump_path = std::getenv("NVTE_DUMP_AITER_RT");
-  // print kernel name on verbose mode
-  ck_tile::stream_config stream_config{stream, dump_path!=nullptr, get_ck_log_stream() != nullptr};
 
   bias_enum bias_type = bias_enum::no_bias;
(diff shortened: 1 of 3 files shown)


wenchenvincent (changes requested):
Can we add a test?

Micky774:
> Can we add a test?

What kind of test would you want added? Since this is pretty internal to the CK FA API we'd need to either test directly in C w/ known kernel configs, or create a python binding to expose to pytest. I personally think it should be okay without a test (we had this mechanism before) and that it'll actually be _indirectly tested_ by other downstream mechanisms that rely on it (e.g. your PR).

wenchenvincent (approved):

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -441,21 +441,17 @@ void dump_bwd_timings(const char* dump_path, float average_runtime){
   file << average_runtime << "\n";
 }
 
-hipError_t ck_attn_bwd(const CkAttnBwdArgs& args, hipStream_t stream){
+// Populate the AITER mha_bwd_args from TE's CkAttnBwdArgs. Shared by ck_attn_bwd
+// (real launch) and ck_attn_bwd_uses_v3 (v3 availability probe) so the probe can
+// never disagree with the launch. v3_api_check is left false here; callers flip it.
+// The stream-dependent max_seqlen override (NVTE_CK_RUNTIME_MAX_SEQLEN) is applied
+// by ck_attn_bwd after this returns; it does not affect v3 kernel selection.
+static aiter::mha_bwd_args build_bwd_fmha_args(const CkAttnBwdArgs& args){
```
wangye805: nit: do we by default export all functions defined in libck_fused_attn.so? I was just curious about the static function usage.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -578,6 +574,42 @@ hipError_t ck_attn_bwd(const CkAttnBwdArgs& args, hipStream_t stream){
...
+hipError_t ck_attn_bwd(const CkAttnBwdArgs& args, hipStream_t stream){
+
+  // Recomputed here (also computed inside build_bwd_fmha_args) for the post-dispatch
+  // MQA/GQA reduction and dbias handling below.
+  bool is_mqa_gqa = (args.h > args.hg);
+  bool has_dbias = args.dbias_ptr != nullptr;
+  BiasShape bias_shape = BiasShape::k11SS;
+  if (!args.is_group_mode()) {
+    std::tie(std::ignore, bias_shape) = get_ck_bias_type_shape(&args);
+  }
```
wangye805: Since you have already reorganize or refactor build_bwd_fmha_args, why not put mqa/gqa, dbias, or bias shape into it?

wangye805 (changes requested):

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: Yes, by default we export all under the `ck_fused_attn` namespace. We don't really need the `static` here anyways, so I've removed it. If we want to be more careful about symbol visibility we should probably do so at the library build level.

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
Micky774: Originally it was just to keep the API symmetric for `build_{f,b}wd_fmha_args` but I'm fine with adding a struct to package the vars -- it's safer than recomputing them at least.

wangye805 (approved):

wenchenvincent:
@Micky774 Could you check if the CI failures are related?

Micky774:
> @Micky774 Could you check if the CI failures are related?

Failures seem unrelated. They're a combo of group GEMM failures, timeouts and resource errors.

PR #671, opened 2026-07-16
[ROCm] Add THD/ragged bf16 (atomic16) backward test coverage

wenchenvincent:
## Summary

On ROCm/gfx950 the CK fused-attention **backward** silently corrupts **dQ** for THD/ragged (packed / document-masked) attention on the default bf16 dq-accumulation path (`NVTE_CK_IS_V3_ATOMIC_FP32=0`). Forward, dK and dV are correct.

### Root cause
`build_bwd_fmha_args` allocates the `dq_acc` scratch as fp32-packed `(nsplits, H, total_q, d_qk)` with `batch_stride_dq_acc = 0`. That matches the fp32 `dq_convert` post-kernel, but the bf16 `dq_shuffle` post-kernel expects a **per-segment padded** layout — with `batch_stride=0` every ragged segment past `cu_seqlens` offset 0 is written to the offset-0 slot, so dQ is correct only for the first segment and garbage for the rest.

### Fix
When the v3 asm path actually runs — probed via **`ck_attn_bwd_uses_v3`** (added in #662) — switch `dq_acc` to AITER's per-segment bf16 layout `(nsplits, B, H, pad16(s_q), 128)` with matching `stride/nhead/batch/split` strides. The v2/fallback path and non-ragged/`atomic_fp32=1` keep the fp32-packed layout. `fused_attn_ck.cpp` sizes the scratch to hold **either** layout (the layout choice needs a GPU probe that must not run in the workspace-sizing pass).

### Tests
- Add an **equal-dim-128 bf16 GQA self-attn** config (BSHD + THD/RAGGED) to `test_fused_attn.py`. Previously there was no symmetric 128/128 THD backward coverage (only asymmetric 128/64) — exactly the llama3 shape.
- Add an **`NVTE_CK_IS_V3_ATOMIC_FP32=0`** (atomic16 / `dq_shuffle`) CI run of the THD backward tests in `ci/jax.sh`. The default fp32 path never exercised the bf16 `dq_acc` path.

### Verification (MI355X / gfx950)
TE THD `padding_causal` backward vs a pure-JAX block-diagonal reference:
- dQ cosine **0.0036 → 1.0000** with `atomic_fp32=0`; per-segment all correct across `[256]`, `[100,80,76]`, `[128,128]`, `[64×4]`, `[200,56]`, for **both** the v3 (`dq_shuffle`) and v2 paths.
- Still uses the fast `bwd_hd128_dq_shuffle_group` kernel (fix is on the fast path, not a fallback).
- `atomic_fp32=1` unchanged; end-to-end MaxText llama3.1-8B block-diagonal THD packing converges to eval-loss 3.3 (previously stalled ~6.1 and NaN'd).

Builds on the V3 API check mechanism from #662. Supersedes #657 (which targeted an earlier base before #662).


transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -605,6 +605,21 @@ hipError_t ck_attn_bwd(const CkAttnBwdArgs& args, hipStream_t stream){
...
+  // dq_acc scratch layout (mirrors AITER's asm_mha_varlen_bwd wiring). build_bwd_fmha_args sets the
+  // fp32-packed layout (nsplits, H, total_q, d_qk) with batch_stride_dq_acc=0 -- correct for the fp32
+  // dq_convert post-kernel. But the bf16 dq_shuffle post-kernel (is_v3_atomic_fp32=0) expects a
+  // per-segment padded layout (nsplits, B, H, pad16(s_q), 128); with batch_stride=0 it mis-indexes every
+  // ragged segment past cu_seqlens offset 0, corrupting dQ. Switch to the per-segment layout only when
+  // the v3 asm path actually runs (ck_attn_bwd_uses_v3 probe) -- v2/fallback keeps the fp32-packed layout.
+  // The dq_acc buffer is sized to hold either layout in fused_attn_ck.cpp.
+  if (args.is_group_mode() && !args.is_v3_atomic_fp32 && ck_attn_bwd_uses_v3(args)) {
+    const int64_t padded_sq = static_cast<int64_t>(((args.s_q + 15) / 16) * 16);
+    fmha_args.stride_dq_acc = 128;
```
wangye805: Is this only for hdim 128 cases?

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -605,6 +605,21 @@ hipError_t ck_attn_bwd(const CkAttnBwdArgs& args, hipStream_t stream){
...
 
+  // dq_acc scratch layout (mirrors AITER's asm_mha_varlen_bwd wiring). build_bwd_fmha_args sets the
+  // fp32-packed layout (nsplits, H, total_q, d_qk) with batch_stride_dq_acc=0 -- correct for the fp32
+  // dq_convert post-kernel. But the bf16 dq_shuffle post-kernel (is_v3_atomic_fp32=0) expects a
+  // per-segment padded layout (nsplits, B, H, pad16(s_q), 128); with batch_stride=0 it mis-indexes every
+  // ragged segment past cu_seqlens offset 0, corrupting dQ. Switch to the per-segment layout only when
+  // the v3 asm path actually runs (ck_attn_bwd_uses_v3 probe) -- v2/fallback keeps the fp32-packed layout.
+  // The dq_acc buffer is sized to hold either layout in fused_attn_ck.cpp.
+  if (args.is_group_mode() && !args.is_v3_atomic_fp32 && ck_attn_bwd_uses_v3(args)) {
+    const int64_t padded_sq = static_cast<int64_t>(((args.s_q + 15) / 16) * 16);
```
wangye805: 1). I presume this pad16(s_q) is coming from aiter source codes, better put a code pointer here for future reference
2). previously we never test cases with sq not multiple of 16. So if we wanted to put this in, we better add a testcase for non 16 multiple sq and also make sure ck flow does not break

transformer_engine/common/ck_fused_attn/src/ck_fused_attn_bwd.cpp
```diff
@@ -605,6 +605,21 @@ hipError_t ck_attn_bwd(const CkAttnBwdArgs& args, hipStream_t stream){
...
+    fmha_args.stride_dq_acc = 128;
+    fmha_args.nhead_stride_dq_acc = static_cast<int64_t>(padded_sq * 128);
+    fmha_args.batch_stride_dq_acc = static_cast<int64_t>(args.h * padded_sq * 128);
+    fmha_args.split_stride_dq_acc = static_cast<int>(args.b * args.h * padded_sq * 128);
```
wangye805: Actually, aiter do not have deterministic path so split stride should not affect the results. (their nsplits=1)

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -749,8 +749,18 @@ void fused_attn_ck_bwd_impl(
...
-  // CK requires dq_acc ptr; size depends on deterministic mode
-  void* dq_acc_ptr = planner.allocate(nsplits*h*max_tokens_q*d_qk*sizeof(float));
+  // CK requires a dq_acc scratch buffer. ck_attn_bwd picks the layout at launch time: fp32-packed
+  // (nsplits, H, total_q, d_qk), or -- when the v3 asm dq_shuffle path runs (is_v3_atomic_fp32=0, ragged)
+  // -- bf16 per-segment padded (nsplits, B, H, pad16(s_q), 128). The v3 choice needs a GPU probe that must
+  // not run in the workspace-sizing pass, so size the buffer to hold EITHER layout for the ragged/non-atomic
+  // case; ck_attn_bwd sets the strides. Non-ragged / atomic_fp32 only ever use the fp32-packed layout.
+  const size_t dq_acc_fp32_bytes = nsplits*h*max_tokens_q*d_qk*sizeof(float);
+  const size_t dq_acc_padded_sq = ((s_q + 15) / 16) * 16;
+  const size_t dq_acc_bf16_bytes = nsplits*b*h*dq_acc_padded_sq*128*nvte_dtype_size(dtype);
```
wangye805: This will over allocate for hdim 64 and under allocate for hdim 192 cases

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -749,8 +749,18 @@ void fused_attn_ck_bwd_impl(
...
+  const bool dq_acc_max_of_both = is_ragged && !nvte_ck_is_v3_atomic_fp32;
+  const size_t dq_acc_bytes = (dq_acc_max_of_both && dq_acc_bf16_bytes > dq_acc_fp32_bytes)
```
wangye805: I recall we have a is_v3_api_check which we can use now

wangye805 (changes requested):

wenchenvincent:
Thanks @wangye805 — addressed in 60f2d12:

- **"Only for hdim 128?"** Yes. The only ragged/group dq_shuffle kernel is `bwd_hd128_dq_shuffle_group`, so `ck_attn_bwd_uses_v3` can only return true here for `d_qk==128`. Made it explicit: named `kV3DqAccHeadDim=128` / `kV3DqAccSeqAlign=16` constants in the shared header (referenced at both the sizing and stride sites), and a hard `d_qk == kV3DqAccHeadDim` assert in the stride branch so a future hd64/hd192 dq_shuffle fails loudly instead of silently corrupting dQ.
- **pad16 code pointer + non-16 seqlen test.** Added the AITER `asm_mha_varlen_bwd.cu` reference (atomic16 `dq_accum` = `{1, batch, nhead, pad16(max_seqlen_q), 128}`) at both sites, and a non-16-multiple THD config (`s_q=2044`) to `test_fused_attn.py`. Verified locally with segments 103/77/61 (S=241 → pad16 256): dQ cos 1.0, finite — the CK flow handles it.
- **split_stride is a no-op (nsplits=1).** Correct — v3 has no deterministic path, so nsplits==1 and the split index is always 0. Noted it in a comment; set it consistently anyway.
- **Over/under allocation for hd64/hd192.** Fixed: the bf16 sizing is now gated on `d_qk == kV3DqAccHeadDim`, so hd64/hd192 use the fp32-packed size (no over-sizing). Also switched the element size to `sizeof(uint16_t)` with an `// atomic16` comment rather than `nvte_dtype_size(dtype)`.
- **"use is_v3_api_check in the sizing pass."** I looked at this — the catch is that `ck_attn_bwd_uses_v3` needs a `CkAttnBwdArgs`, but the dq_acc allocation happens in the workspace-**sizing** pass which early-returns *before* `ck_args` is built (in the exec pass). Building a second minimal args just for a sizing-pass probe would reintroduce exactly the probe-vs-launch divergence that `build_bwd_fmha_args` (#662) exists to prevent. So I instead gate the sizing on `d_qk == kV3DqAccHeadDim` — the exact and only condition under which the bf16 group dq_shuffle runs — which is conservative (never under-sizes) and needs no probe in the sizing pass. The strides still use the real `ck_attn_bwd_uses_v3` probe at launch. Happy to revisit if you'd prefer the reorder instead.

wenchenvincent:
**Rebased onto dev — this PR is now test-only.**

dev's #599 ("Update QoLA/AITER") landed the dQ fix independently and more cleanly than this PR's original approach: it routes `dq_acc` through AITER's `workspace_alloc` callback (AITER owns the allocation + striding, sized via `ck_attn_bwd_workspace_size()`/`v3_dq_acc_bytes()`), and bumps AITER to the commit with the CK-graph fix. That supersedes the manual `dq_acc` stride/sizing changes this PR carried, so I've dropped them and taken dev's version of all three source files.

What remains here is the **test coverage** that #599 did not add and that guards this class of bug:
- `tests/jax/test_fused_attn.py`: equal-dim **128/128 bf16 GQA self-attn** (BSHD + THD/RAGGED) — there was no symmetric 128/128 THD backward config before — plus a **non-16-multiple seqlen** THD case (`s_q=2044`) exercising the atomic16 `pad16(s_q)` rounding.
- `ci/jax.sh`: an **`NVTE_CK_IS_V3_ATOMIC_FP32=0`** (atomic16 / bf16 `dq_shuffle`) leg of the THD backward tests — the default fp32 path never exercised the bf16 `dq_acc` layout.

**Verification (MI355X / gfx950, built at the merged dev + #599 AITER):** with `NVTE_CK_IS_V3_ATOMIC_FP32=0` the THD `padding_causal` backward dQ cos is **1.0000** (was 0.0036 pre-#599), per-segment correct across `[256]`, `[100,80,76]`, `[128,128]`, `[64×4]`, `[200,56]`, and the non-16 case (segments 103/77/61, S=241→pad16 256) dQ cos 1.0. So dev #599's fix is confirmed, and these tests lock it in.

wenchenvincent:
@wangye805 Now the issue has been fixed in the AITER commit upgrade, this PR is to add a test only. Could you review again?

wangye805 (approved):

PR #677, opened 2026-07-23
microbenchmarks: add buffer rotation option

matthiasdiener:
# Description

Add new options, `--rotating [MByte]`, `--no-rotating`, to rotate input buffers (on by default).
Follows the discussions in:
- https://docs.nvidia.com/cutlass/latest/media/docs/cpp/gemm_performance_measurement_methodology_guidelines.html
- https://amd.atlassian.net/wiki/spaces/MLSE/pages/744192648/hipBLASLt+Standard+Benchmarking

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

# Checklist:


benchmarks/microbenchmarks/README.md
```diff
@@ -31,6 +31,24 @@ python benchmark_gemm.py --csv --csv-samples gemm_samples.csv
...
+Pass `--rotating-buffers` to instead cycle inputs through a ring of buffers
+sized to exceed the **last-level cache**, so each launch touches different
+memory (closer to a cold-cache, steady-state workload):
+
+```bash
+python benchmark_gemm.py --rotating-buffers        # auto-size the ring past the LLC
+python benchmark_casting.py --rotating-buffers 16  # or fix the ring size (16 buffers)
+```
+
+Passing `--rotating-buffers N` fixes the number of buffers; omitting `N`
```
alextmagro: Generally we would specify the # of bytes that the rotating buffer uses rather than the # of buffers -- I think that can be generally set to a safely large value, such as 4/8

benchmarks/microbenchmarks/utils.py
```diff
@@ -275,6 +367,16 @@ def make_parser(**kwargs):
             "--csv-samples is ignored in this mode."
         ),
     )
+    parser.add_argument(
+        "--rotating-buffers", nargs="?", type=int, const=0, default=None, metavar="N",
```
alextmagro: I would argue for this to be on by default, as a standard benchmarking practice

alextmagro (changes requested):

benchmarks/microbenchmarks/README.md
matthiasdiener: https://github.com/ROCm/TransformerEngine/pull/677/commits/2fc6bf032a9dee7c61d0559c859738d96443a236 changes it so it is a rotating memory budget, like you suggested. It now matches the semantics of `hipblaslt-bench --rotating`.

The number of buffers is derived internally as `ceil(budget / per-iteration input bytes)` (with a floor of 2 and a ceiling that mirrors hipblaslt-bench capping its block count at the iteration count, so a huge budget on a tiny input can't allocate an unbounded ring). Omitting the value auto-sizes the ring to 2x the last-level cache, like before.

benchmarks/microbenchmarks/utils.py
matthiasdiener: Done in https://github.com/ROCm/TransformerEngine/pull/677/commits/2fc6bf032a9dee7c61d0559c859738d96443a236,  rotation is now on by default, with `--no-rotating` to opt out.

benchmarks/microbenchmarks/benchmark_grouped_gemm.py
```diff
@@ -114,18 +115,21 @@ def bench_grouped_gemm(Case, B, M, N, K, dtype):
         params_dtype=dtype,
         device=device,
     )
-    x = torch.randn((sum_M, K), dtype=dtype, device=device, requires_grad=True)
+    # Rotate the activation buffer (off by default) so back-to-back grouped GEMMs
```
ipanfilo: It seems on by default

matthiasdiener: Good catch, thanks. Fixed in 8ad39eb2

alextmagro (approved):

ipanfilo (approved):

PR #678, opened 2026-07-24
[ROCm] Jax Add softmax sink (learnable off-by-one) support for the ROCm/CK fused attention backend

shurale-nkn:
# Description

Adds forward and backward support for softmax sink (NVTE_LEARNABLE_SOFTMAX) to the CK/AITER fused-attention backend on ROCm.

- Backward pass computes and returns d_sink (accumulated via atomics inside CK, one scalar per head); the gradient buffer is zero-initialized before launch.
- Since the AITER ASM v3 fwd/bwd kernels don't support sink, uses_fwd_v3/uses_bwd_v3 are forced off whenever has_sink is true, falling back to the (slower but correct) CK tile path.
- Fixes the JAX extension's aux-tensor indexing (_### PrepareFusedAttnBackwardAuxTensors_). Upstream mixes static (fixed, assume-always-present) and dynamic (config-dependent) positions for optional aux tensors: forward builds the pack dynamically based on the real bias_type, while backward forces a "dummy" always-bias-present bias_type/backend so nvte_fused_attn_bwd can pull from fixed, hardcoded slot indices internally. That static-slot assumption doesn't scale once a second independent optional tensor (sink) is added — bias and sink can each be present or absent independently, so a fixed index for either one is wrong in at least one of the four combinations. For ROCm, both forward and backward now build the pack the same way, with every optional tensor (bias, sink) appended sequentially and only when actually enabled — no dummy/static slot assumptions — and fused_attn.cpp/fused_attn_ck.cpp read the aux tensors back out using the same dynamic running index instead of hardcoded tensors[2]/tensors[3].
- Fixes a pre-existing positional-argument bug in the CUDA path(PrepareFusedAttnForwardAuxTensors): pre-PR the call was ..., softmax_aux, softmax_offset), which bound softmax_offset to the rng_state_buf parameter

Fixes # (issue)
CK code in QoLa
- fmha_bwd_kernel.hpp indexed sink_ptr
as i_batch * nhead + i_nhead, but the buffer holds one value per head (shape
[nhead]), matching how the forward kernels already read it, so i_batch must not
factor into the offset.
Affected all bwd tests
- CK fwd: K/V windows double-offset with sink + local mask. In BlockFmhaPipelineQRKSVS the K/V/bias windows are created at kv_load_start (already == seqlen_k_start), and the kHasSink block moved them again by seqlen_k_start - sink_seq_end. TE uses no StreamingLLM prefix (sink_size = 0 → num_sink_loop == 0), so the guard i_total_loops == 0 fired on the first iteration instead of never — live here because this pipeline increments in while(++i_total_loops ...), dead in the async one, which is why only bias configs (qr) broke and no-bias (qr_async) was fine. Any query tile whose window starts past column 0 read wrong/out-of-range K/V → garbage output
 ``18 tests fixed: POST_SCALE_BIAS-1HSS-{Mask,Seqlens,SegmentIDs}-SWA-DROP_0.0-<cfg>-LEARNABLE_SOFTMAX-<mask>``

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -104,7 +104,6 @@ void PrepareFusedAttnForwardAuxTensors(NVTETensorPack *tensor_pack, const size_t
       bias_aux_data.dtype = static_cast<NVTEDType>(dtype);
       nvte_set_tensor_param(&bias_aux, kNVTERowwiseData, &bias_aux_data);
     }
-#ifndef USE_ROCM
     // include softmax_offset if provided
     if (softmax_offset_buf != nullptr) {
```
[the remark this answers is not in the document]
shurale-nkn: Fixed — the writer and the readers are now on one predicate.

`PrepareFusedAttnForwardAuxTensors` takes an `NVTE_Softmax_Type softmax_type` (right after `bias_type`, so it reads symmetrically with the bias slot) and the sink slot is appended under `softmax_type != NVTE_Softmax_Type::NVTE_VANILLA_SOFTMAX` instead of `softmax_offset_buf != nullptr`. It is threaded through `PrepareFusedAttnBackwardAuxTensors` and forwarded on both branches of the `#ifndef USE_ROCM` split (the real `softmax_type` in both — only `bias_type`/`backend` are dummied on CUDA), plus both call sites (`FusedAttnForwardImpl`, `FusedAttnBackwardImpl`), where it was already in scope.

This keeps CUDA correct rather than just ROCm: `fused_attn_f16_arbitrary_seqlen.cu:1198` and `fused_attn/fused_attn.cpp:709` use the same predicate, so for `VANILLA_SOFTMAX` the writer no longer appends a slot the reader skips, and for a real offset the slot can no longer be dropped because of a null pointer.

This approach is also consistent with other code comments that suggest using aux as the preferred mechanism for retrieving tensor dimensions in the future.

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -77,14 +77,6 @@ bool is_ck_backend_supported(
...
 
-  // filter based on softmax type
-  if(softmax_type!=NVTE_VANILLA_SOFTMAX){
-    if(nvte_log_ck_config){
-      std::cout<<"AITER/CK fused attn does not support learnable sink yet"<<std::endl;
-    }
-    return false;
-  }
-
   // joint filter based on sliding window and attn_mask
```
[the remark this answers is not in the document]
shurale-nkn: Took option 2 — the dead parameter is gone (definition in `fused_attn_ck.cpp`, declaration in `fused_attn_ck.h`, argument at the `nvte_get_fused_attn_backend` call site in `fused_attn_rocm/fused_attn.cpp`).

On your first question: no narrower guard is needed, and the full matrix is the evidence. The complete `LEARNABLE_SOFTMAX` sweep in `tests/jax/test_fused_attn.py` — 486 tests spanning dropout, THD/group mode, MQA/GQA, ragged layouts and deterministic bwd, i.e. exactly the `atomicAdd`-per-head `d_sink` paths you flagged — passes with zero failures. So sink is correct across the surface CK accepts, and there is nothing left for a CK-side sink filter to reject.

`is_aotriton_backend_supported` keeps its `softmax_type` parameter: it still genuinely rejects sink (`fused_attn_aotriton.cpp:73`), so there is no signature-parity reason to touch it.

Micky774:
Hey there @shurale-nkn, thanks for the contribution! While we're looking into it, could you provide us some context regarding why this feature enablement is wanted/needed? What use case are you trying to enable? Thanks!

shurale-nkn:
> Hey there @shurale-nkn, thanks for the contribution! While we're looking into it, could you provide us some context regarding why this feature enablement is wanted/needed? What use case are you trying to enable? Thanks!

Hi @Micky774, I need sink support for train in MaxText. At the moment, TE is the only provider of fused attention for this framework, and without this PR, GPT-OSS will not work correctly.

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -474,9 +467,11 @@ void fused_attn_ck_fwd_impl(
   size_t *workspace_size,
   cudaStream_t stream){
 
+  const bool has_sink = softmax_type != NVTE_VANILLA_SOFTMAX;
```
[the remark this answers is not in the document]
ipanfilo: The concern looks valid

shurale-nkn:
Failed tests not related to PR. (timed out after 180 minutes)
Need review. Original kernel fix merged in CK
@ipanfilo @wenchenvincent @wangye805

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -474,9 +467,11 @@ void fused_attn_ck_fwd_impl(
...
   const bool nvte_log_ck_config = getenv<bool>("NVTE_LOG_CK_CONFIG");
 
-  bool nvte_ck_uses_fwd_v3 = getenv<int>("NVTE_CK_USES_FWD_V3", 1);
+  // ASM v3 does not support sink; force CK tile path
+  bool nvte_ck_uses_fwd_v3 = getenv<int>("NVTE_CK_USES_FWD_V3", 1) && !has_sink;
```
ipanfilo: nit: !has_sink can go first to skip env retrieving when not needed

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -487,6 +482,7 @@ void fused_attn_ck_fwd_impl(
   bool is_padding = is_padding_mask(mask_type);
   bool bshd_to_thd = is_BSHD && is_padding;
 
+
```
ipanfilo: nit: double empty line is not needed

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -657,9 +653,15 @@ void fused_attn_ck_fwd_impl(
...
     std::cout<<"nvte_ck_uses_fwd_v3: "<<nvte_ck_uses_fwd_v3<<", ";
-    std::cout<<"num_splits: "<<ck_args.num_splits<<std::endl;
+    std::cout<<"num_splits: "<<ck_args.num_splits<<", ";
+    std::cout<<"has_sink: "<<has_sink<<", ";
+    std::cout<<"sink_ptr: "<<(has_sink ? devPtrSoftmaxOffset : nullptr)<<std::endl;
   }
 
+  ck_args.has_sink = has_sink;
+  ck_args.sink_ptr = has_sink ? devPtrSoftmaxOffset : nullptr;
+  ck_args.uses_fwd_v3 = nvte_ck_uses_fwd_v3;
```
ipanfilo: uses_fwd_v3 has already be set in this method

transformer_engine/common/fused_attn_rocm/fused_attn.cpp
```diff
@@ -309,7 +309,6 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
         qkv_layout,
         bias_type,
         attn_mask_type,
-        softmax_type,
         dropout,
```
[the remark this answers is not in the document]
shurale-nkn: PyTorch attention tests passed

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -474,9 +467,11 @@ void fused_attn_ck_fwd_impl(
   size_t *workspace_size,
   cudaStream_t stream){
 
+  const bool has_sink = softmax_type != NVTE_VANILLA_SOFTMAX;
   const bool nvte_log_ck_config = getenv<bool>("NVTE_LOG_CK_CONFIG");
 
-  bool nvte_ck_uses_fwd_v3 = getenv<int>("NVTE_CK_USES_FWD_V3", 1);
+  // ASM v3 does not support sink; force CK tile path
+  bool nvte_ck_uses_fwd_v3 = !has_sink && getenv<int>("NVTE_CK_USES_FWD_V3", 1);
```
Micky774, line 474: Does AITER V3 not correctly fallback to V2 in this case? Why do we need a special exclusion?

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -741,10 +747,14 @@ void fused_attn_ck_bwd_impl(
...
   
+  const bool has_sink = softmax_type != NVTE_VANILLA_SOFTMAX;
   const bool nvte_log_ck_config = getenv<bool>("NVTE_LOG_CK_CONFIG");
   // bwd v3 is optional by enabling the following envs
   // default values follows the ck example setting
-  bool nvte_ck_uses_bwd_v3 = getenv<int>("NVTE_CK_USES_BWD_V3", 1);
+  // ASM v3 bwd does not compute the sink gradient; force the CK tile path so
+  // has_sink users still get a correct (if slower) d_sink, mirroring the fwd
+  // ASM v3 sink guard above.
+  bool nvte_ck_uses_bwd_v3 = getenv<int>("NVTE_CK_USES_BWD_V3", 1) && !has_sink;
```
Micky774, line 757: This sounds like an AITER bug that needs to be reported right?

Micky774 (commented):
A couple questions regarding the AITER V3 exclusion

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
shurale-nkn: The ASM path only validates the arguments that are part of fmha_fwd_v3_args. It does not perform any meaningful validation of the remaining parameters, so a fallback is not possible.
I also couldn't find any evidence of tests covering this structure or any usage of it outside of this single file.

In addition, the AITER v3 interfaces(fmha_v3_fwd) only expose a limited set of arguments to callers, and sink is not among them.

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
shurale-nkn: same as forward. In BWD, sink is not part of the v3 interface either

shurale-nkn:
@Micky774 @ipanfilo ping

Micky774:
Can you update to the latest QoLA main commit?

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: Sorry, do you mean AITER doesn't do any validation of the remaining parameters? Its dispatch searches for V3 kernel compatible kernels and dispatches to them if found, otherwise defaults to V2 kernels. I'm not certain I understand your point about them not doing any validation of the remaining parameters. Could you can help me understand this issue a bit better?

shurale-nkn: @Micky774  Yes, the framework does attempt to select a kernel automatically, but the set of configurations supported by V3 is extremely limited. The so-called V2 path supports a much broader range of configurations. In particular, some parameters are not even represented in the V3 interface.

As a result, it is more reliable to control this behavior from the application side, which is exactly what the NVTE_CK_USES_V3 flag provides.

Otherwise, if V3 determines that a configuration is not supported while an ASM kernel has been requested, it will emit a warning. Given that the set of supported configurations is significantly smaller, users would end up seeing warnings in many perfectly expected scenarios. I don't see much value in generating unnecessary log noise.

Also, as demonstrated by the earlier commit https://github.com/ROCm/QoLA/commit/100b315ab23e26673519ecbdd7c9d8564b9a367e implementing similar functionality, AITER currently does not expose the interface required to check sink support. In my case, ASM kernels are always unavailable, so adding additional logic to the checker would not provide any practical benefit and would only make the implementation more complex.

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
```diff
@@ -657,9 +655,14 @@ void fused_attn_ck_fwd_impl(
...
     std::cout<<"nvte_ck_uses_fwd_v3: "<<nvte_ck_uses_fwd_v3<<", ";
-    std::cout<<"num_splits: "<<ck_args.num_splits<<std::endl;
+    std::cout<<"num_splits: "<<ck_args.num_splits<<", ";
+    std::cout<<"has_sink: "<<has_sink<<", ";
+    std::cout<<"sink_ptr: "<<(has_sink ? devPtrSoftmaxOffset : nullptr)<<std::endl;
   }
 
+  ck_args.has_sink = has_sink;
+  ck_args.sink_ptr = has_sink ? devPtrSoftmaxOffset : nullptr;
+
```
Micky774, line 767: Can you move this to line 555 right below `ck_args.how_v3_bf16_cvt = nvte_ck_how_v3_bf16_cvt`

transformer_engine/common/fused_attn_rocm/fused_attn_ck.cpp
Micky774: The dispatcher here rejects sink w/ no fallback so let's cut the path explicitly w/ `ck_args.num_splits = has_sink ? -1 : ck_attn_fwd_num_splits(ck_args);`

Micky774 (changes requested):
Minor changes, otherwise looks good

Micky774 (approved):
pending CI

ipanfilo (approved):

PR #699, opened 2026-08-10
[Fix] add rocm10 te core package support

GeneDer:
# Description

Rocm10.0 release is coming up and the TE packages doesn't currently recognized the TE package built for rocm10. I have manually applied this patch in my docker build and seeing it working. Upstream this change to TE so we don't need to patch the packaged after all. 

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/build_tools/utils.py
+++ b/build_tools/utils.py
@@ -326,7 +326,7 @@ def get_cuda_include_dirs() -> Tuple[str, str]:
     if not force_wheels and cuda_toolkit_include_path() is not None:
         return [cuda_toolkit_include_path()]
 
-    # Use pip wheels to include all headers.        
+    # Use pip wheels to include all headers.
     try:
         import nvidia
     except ModuleNotFoundError as e:
@@ -556,7 +556,7 @@ def uninstall_te_wheel_packages():
             "transformer_engine_cu12",
             "transformer_engine_torch",
             "transformer_engine_jax",
-            "transformer_engine_rocm7",
+            "transformer_engine_rocm10",
             "transformer_engine_rocm_jax",
             "transformer_engine_rocm_torch",
         ]
--- a/build_tools/utils.py
+++ b/build_tools/utils.py
@@ -556,6 +556,7 @@ def uninstall_te_wheel_packages():
             "transformer_engine_cu12",
             "transformer_engine_torch",
             "transformer_engine_jax",
+            "transformer_engine_rocm7",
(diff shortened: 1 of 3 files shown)


ipanfilo (changes requested):
The fix does not seem completed. setup.py and build tools scripts also use rocm7. So rocm10 should be declared as rocm10 extra for meta package in setup.py and as to deinstallation list in build_tools/utils.py

GeneDer:
> The fix does not seem completed. setup.py and build tools scripts also use rocm7. So rocm10 should be declared as rocm10 extra for meta package in setup.py and as to deinstallation list in build_tools/utils.py

Good call thanks @ipanfilo! Can you PTAL?

build_tools/utils.py
```diff
@@ -556,7 +556,7 @@ def uninstall_te_wheel_packages():
             "transformer_engine_cu12",
             "transformer_engine_torch",
             "transformer_engine_jax",
-            "transformer_engine_rocm7",
+            "transformer_engine_rocm10",
```
ipanfilo, line 560: It should not replace but be added instead to keep compatibility with ROCm 7

setup.py
```diff
@@ -430,10 +430,10 @@ def git_check_submodules() -> None:
             "pytorch": [f"transformer_engine_torch=={__version__}"],
             "jax": [f"transformer_engine_jax=={__version__}"],
         } if not rocm_build() else {
-            "rocm": [f"transformer_engine_rocm7=={__version__}"],
-            "rocm7": [f"transformer_engine_rocm7=={__version__}"],
-            "rocm_pytorch": [f"transformer_engine_rocm7[pytorch]=={__version__}"],
-            "rocm_jax": [f"transformer_engine_rocm7[jax]=={__version__}"],
+            "rocm": [f"transformer_engine_rocm10=={__version__}"],
```
ipanfilo: Similar to utils.py - Default rocm should be rocm7 and have rocm10 extra added similar to rocm7. See how it is done for core/core_cu12/core_cu13 just above.
Not to duplicate rocm_pytorch and rocm_jax with rocm10_pytorch and rocm10_jax, corresponding dependencies can be replaced with f"transformer_engine_rocm_torch=={__version__}" / f"transformer_engine_rocm_jax=={__version__}". Also refer to the above

ipanfilo (approved):

PR #700, opened 2026-08-11
[gfx1250] detect when building on gfx1250, add to build archs

matthiasdiener:
# Description

So that NVTE_ROCM_ARCH does not need to be manually set when building on gfx1250 (both for main TE build and the tests).

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/build_tools/rocm_utils.cmake
+++ b/build_tools/rocm_utils.cmake
@@ -34,6 +34,22 @@ message(STATUS "ROCM_PATH: ${ROCM_PATH}")
 #Configure target GPU architectures
 if(NOT DEFINED ENV{NVTE_ROCM_ARCH})
     SET(CMAKE_HIP_ARCHITECTURES gfx942 gfx950)
+    # gfx1250 is not in the default set; when building natively on such a machine
+    # add it so it is targeted without having to set NVTE_ROCM_ARCH. Skipped
+    # silently if rocminfo or a GPU is unavailable.
+    find_program(NVTE_ROCMINFO_EXECUTABLE rocminfo HINTS "${ROCM_PATH}/bin")
+    if(NVTE_ROCMINFO_EXECUTABLE)
+        execute_process(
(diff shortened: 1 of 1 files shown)


ipanfilo (commented):
gfx1250 is eventually (after merging all gfx1250 changes) to be added to list of default arches, currently building for 1250 requires setting some extra variables anyway. So as a temporary change it is probably not worth to merge it

matthiasdiener:
> gfx1250 is eventually (after merging all gfx1250 changes) to be added to list of default arches, currently building for 1250 requires setting some extra variables anyway. So as a temporary change it is probably not worth to merge it

That's a fair point. I agree this is meant to be a temporary fix, but it is locally contained and easy to remove later.

Regarding "requires setting some extra variables anyway": the only other one I can think of is `export NVTE_FUSED_ATTN=0` (which is handled in #603).

One main motivation is the C++ tests build ergonomics. Because configure and build are separate CMake invocations, you currently have to repeat the arch on both:

```
NVTE_ROCM_ARCH=gfx1250 cmake -Bbuild
NVTE_ROCM_ARCH=gfx1250 cmake --build build
```

Otherwise, you (silently) get a gfx942/gfx950-only build that fails at runtime with cryptic errors like:

```
unknown file: Failure
C++ exception with description "/dockerx/TransformerEngine/tests/cpp/test_common.hip:1117 in function fillUniformLinearBufferDevice: CUDA Error: invalid device function" thrown in the test body.
```

Although this does not resolve all gfx1250 build differences to gfx942/gfx950 (again, the only one I can think of is the NVTE_FUSED_ATTN/aiter situation), in my view it does improve the out-of-the-box experience on gfx1250 noticeably.

ipanfilo:
IMO it would be more beneficial to make the changes generic and support native target to build n a system where it is going to run

PR #712, opened 2026-08-21
CI: upload Python coverage.json from pytest (non-blocking)

jiagaoxiang:
## Summary

- JUnit XML already records which **tests passed or failed**. This PR adds the missing layer: which **installed `transformer_engine` Python functions actually ran** during those tests.
- Existing `pytest_run` is wrapped with `coverage.py` only when `TE_COVERAGE=1` (set in ROCm CI). Local `ci/pytorch.sh` / `ci/jax.sh` runs are unchanged unless that variable is set.
- Each sGPU / mGPU job uploads `python-coverage-*/coverage.json` plus a small `coverage-meta.txt` with the git SHA. Export is best-effort: a missing or empty report is **not** uploaded and **does not fail** the job.

This is Python-only. It does not measure C++/HIP kernels (`.cu` / `.hip`); that would need a separate instrumented build and `llvm-cov`. `torchrun` / `mpirun` child processes are also not included in this first step.

## Test plan

- [ ] CI still fails only when a test suite fails (coverage export errors do not fail the job).
- [ ] After a green (or failed) sGPU/mGPU job, artifact `python-coverage-sgpu-<arch>` / `python-coverage-mgpu-<arch>-<framework>` is present **only if** `coverage.json` is non-empty and lists measured files.
- [ ] `coverage-meta.txt` records the commit SHA of that job.
- [ ] Local `ci/pytorch.sh` without `TE_COVERAGE=1` still invokes plain `python3 -m pytest`.

### DIFF
--- a/.github/workflows/rocm-ci.yml
+++ b/.github/workflows/rocm-ci.yml
@@ -179,6 +179,8 @@ jobs:
           pip install --upgrade hypothesis setuptools
           pip install --no-build-isolation --no-deps "$TE_TORCH_PKG"
           pip install --no-build-isolation --no-deps "$TE_JAX_PKG"
+          # Optional: wrap pytest with coverage.py. Export is non-blocking.
+          pip install 'coverage>=7'
           EOF
           )"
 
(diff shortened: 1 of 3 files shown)


ipanfilo:
This feature is  valuable for developers but might be overkill for automated CI run

PR #713, opened 2026-08-24
Bulk AG Overlap for bf16 on gfx950

alextmagro:
Enables fused wgrad AG + dgrad GEMM for BF16 + gfx950.

Adds additional test coverage and layer tests, and makes fused/bulk_fused the default overlap scheme for gfx950, with fallback to no overlap at the moment for unaligned shapes, dtypes, and reduce-scatter.

1.282x Geomean speedup for wgrad AG + dgrad GEMM step


tests/pytorch/distributed/run_gemm_with_overlap.py
```diff
@@ -187,7 +187,13 @@ def _parse_args(argv=None, namespace=None):
...
     if opts.bulk_overlap:
-        if opts.p2p:
+        if opts.fused and opts.comm_type != tex.CommOverlapType.AG:
+            warnings.warn("The fused bulk overlap is all-gather only.")
+            opts.fused = False
+        if opts.fused:
+            # `fused_overlap_bulk_ag` is a CommOverlapP2P entry point
+            opts.p2p = True
+        elif opts.p2p:
             warnings.warn("Point-2-point comms are not supported with bulk overlap.")
```
aris134, line 202: Consider updating the original warning message. Perhaps something like:

`Non-fused point-2-point comms are not supported with bulk overlap`

tests/pytorch/distributed/test_comm_gemm_overlap.py
```diff
@@ -539,3 +571,57 @@ def _hashes(out):
...
+
+@pytest.mark.skipif(not fused_available, reason=reason_for_no_fused)
+@pytest.mark.parametrize("nprocs", FUSED_PROC_COUNTS)
+def test_fused_layer_bulk_dgrad_bf16(nprocs):
+    """A column-parallel layer whose dgrad dimensions clear the fused contract."""
+    result = _run_fused_layer(nprocs, [f"--out-features={ELIGIBLE_OUT_FEATURES_PER_RANK * nprocs}"])
+    _assert_numerics_passed(result)
+    fused = _reported_names(result.stdout, "UB FUSED NAMES: ")
+    assert fused is not None, f"harness printed no fused name set\n{result.stdout.decode()}"
+    assert "qkv_dgrad" in fused, fused
```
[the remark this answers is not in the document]
alextmagro: Issue addressed, tests now properly test both sides

tests/pytorch/distributed/run_gemm_with_overlap.py
alextmagro: The original warning message is from upstream, so I want to leave as is. I have added the missing IS_HIP_EXTENSION guards throughout the file to make it more clear this is a difference from upstream.

transformer_engine/common/gemm/kittens/cdna4/fused_ag_gemm_nn.cuh
```diff
@@ -659,6 +678,40 @@ static persistent_fn_t get_persistent_fn(int M, int N, int K, int S) {
...
+
+static persistent_bulk_fn_t get_persistent_bulk_fn(int M, int N, int K, int S) {
+    (void)M; (void)N; (void)K;
+    if (S == 1) return launch_persistent_bulk<1>;
+    if (S == 2) return launch_persistent_bulk<2>;
+    if (S == 4) return launch_persistent_bulk<4>;
+    return nullptr;
+}
+
 // Split-K is gated on the workspace budget. 64 MiB cap is an intentional performance clamp, not a safety choice.
```
aris134, line 715: Sorry if this was asked in a previous PR, but I am wondering whether the 64 MiB cap could be raised to make S=8 affordable at larger tile counts? Was this already tried?

alextmagro: 64 MiB here is actually unrelated to the hipBLASLt workspace needed, and more of a happy coincedence. We have a split-K partial that needs to be stored, and that is a 256x256 fp32 tile, or 256 KiB. With 256 CUs, that means we need exactly (256 KiB * 256 CUs) = 64 MiB at most.

aris134 (approved):

tests/pytorch/distributed/test_comm_gemm_overlap.py
```diff
@@ -477,9 +474,9 @@ def _fused_launch_cmd(nprocs: int):
     return ["torchrun", f"--nproc_per_node={nprocs}"]
 
 
-def _run_fused_ag(quantization="none", nprocs=None):
+def _run_fused_ag(nprocs, bulk=False, quantization="none"):
```
ipanfilo, line 480: For my curiosity, why need making nproc positional arg?

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -711,6 +711,9 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
         "Bulk overlap All-Gather with a GEMM operation launched by another communicator",
         py::call_guard<py::gil_scoped_release>(), py::arg("allgather_communicator"),
         py::arg("send_stream"), py::arg("recv_stream"));
+#else
+  m.def("bulk_overlap_ag_with_external_gemm", &transformer_engine::pytorch::placeholder, "Dummy");
+#endif
   m.def(
```
ipanfilo, line 734: It still seems called from under IS_HIP_EXTENSION only. Why need to move it out of ifdef?

tests/pytorch/distributed/test_comm_gemm_overlap.py
alextmagro: Here I am adding TP=8 tests on top of TP=4 that upstream does, so made this a required arg rather than one with a default.

transformer_engine/pytorch/csrc/extensions/pybind.cpp
alextmagro: This is a bit of a git diff viewing issue, `bulk_overlap_ag_with_external_gemm` still remains in the else block, it is `reset_fused_ag_gemm_cache`  that has been moved.

ipanfilo: > This is a bit of a git diff viewing issue, `bulk_overlap_ag_with_external_gemm` still remains in the else block, it is `reset_fused_ag_gemm_cache` that has been moved.

Yes, I'm about reset_fused_ag_gemm_cache, it is called under IS_HIP_EXTENSION

alextmagro: Ah, understood. I have moved it back into the ifdef properly.

ipanfilo (approved):

PR #716, opened 2026-08-25
Add ROCm Triton blockwise FP8 grouped GEMM

sudhu2k:
# Description

Opt-in ROCm Triton path for blockwise FP8 grouped GEMM on PyTorch `GroupedLinear`, used for MoE expert GEMMs under `Float8BlockScaling`.

The default grouped GEMM path does not implement this blockwise layout (activation 1×128 along K, weights 128×128, columnwise segment-padded operand for variable-K wgrad). This PR adds Triton quantization + persistent grouped GEMM kernels and selects them from `GroupedLinear` when `NVTE_USE_BLOCKWISE_GMM_TRITON=1` and the call is otherwise compatible. Unsupported configs (bias, fused-pad / `unpad_output`, `save_original_input`, `backward_override`, cpu offloading, debug, non-128-aligned features, non-matching recipe dims, etc.) fall back to the existing path instead of raising.

Quantization is always from the original high-precision tensors (no double-quantization). When the caller passes a device `m_splits_tensor`, split lengths are taken from that tensor so the path does not issue a blocking H2D of pageable CPU `m_splits`.

Fixes # (issue)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue) — generic `single_grouped_weight` grad-state fix, see Changes.
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add `_forward_blockwise_fp8_triton` / `_backward_blockwise_fp8_triton` on `_GroupedLinear`, gated by `_is_blockwise_fp8_triton_grouped_gemm_supported()` and `NVTE_USE_BLOCKWISE_GMM_TRITON=1` under `Float8BlockScaling` (x 1D / w 2D / grad 1D).
- Add Triton kernels in `triton_kernels/blockwise_quantize.py` (1×128 activation / 128×128 weight / segment-padded columnwise) and `triton_kernels/blockwise_fp8_grouped_gemm.py` (persistent grouped GEMM and variable-K wgrad).
- Fuse the activation quantization: `quant_fp8_blockwise_grouped_kernel` produces both the rowwise (1×128 along K, forward/dgrad) and the segment-padded columnwise (1×128 along M, variable-K wgrad) operands from a single HBM read when both are needed (training), instead of two separate passes. A `quantize_fp8_blockwise_act_operands` dispatcher selects fused-both / rowwise-only / columnwise-only based on what the caller requests, and returns a single `Float8BlockwiseQTensor` carrying both operands (columnwise operand + its padded `vk_group_offs` ride in the QTensor's columnwise slots).
- Autotune the grouped GEMM kernels: `@triton.autotune` on the persistent forward/dgrad kernel (keyed on `G, N, K`) and the variable-K wgrad kernel (keyed on `G, OUT_M, OUT_N`) over a curated config set. The first call per key warms up on a balanced `group_offs` so the cached config is not locked to one uneven MoE routing.
- Fuse wgrad accumulation into the variable-K kernel: it accepts an existing output tensor and an `accumulate` flag, so wgrad adds directly into `main_grad` in-place instead of producing a separate gradient that is summed afterward (removes an extra tensor and add pass).
- Cache the quantized weight across microbatches on the blockwise path, mirroring the default grouped path's `is_first_microbatch` / `self._fp8_workspaces` reuse: the high-precision weight is quantized once (first microbatch) and the packed `[G*N, K]` `Float8BlockwiseQTensor` is reused for the rest, avoiding N× re-quantization during gradient accumulation. `is_first_microbatch=None` keeps the previous always-quantize behavior; the `fp8_model_params` case already skips re-quantization.
- Support a single contiguous `[G, N, K]` weight buffer (`single_grouped_weight`) to avoid per-forward concatenation of separate per-expert weight tensors. Upstream `GroupedLinear` should pass this through to use it.
- **Generic fix (not ROCm-specific, affects the shared CUDA path):** `GroupedLinear._get_weight_tensors()` / `_get_bias_tensors()` split the grouped Parameter into per-expert views (this fork's transitional pre-#3224 design; upstream returns `[self.weight]`). Those views did not carry the grad state the autograd Function reads off `weights[i]`, so `single_grouped_weight` training was broken on **every** backend: `requires_grad` was `False` (wgrad never ran) and `weights[i].main_grad` raised `AttributeError` under `fuse_wgrad_accumulation`. Now mirror `requires_grad`, per-expert aliasing views of `main_grad`, and `overwrite_main_grad` from the grouped Parameter. This is transitional and disappears if the fork adopts upstream's `[self.weight]` design (NVIDIA/TransformerEngine#3224). Note: `NVTE_GROUPED_LINEAR_SINGLE_PARAM` is off by default, so this path was previously untested here; a regression test is added (see below).
- Use caller-provided `m_splits_tensor` when present; otherwise copy CPU `m_splits` to device.
- Remove the device-to-host sync from FP8 padding: padding previously ran only on the CPU `m_splits` list, forcing a D2H sync. Extend `Fp8Padding.forward` with an optional `m_splits_tensor`; when provided, return the padded split sizes as an on-device tensor (rounded up on-device, no D2H/H2D) so callers keep split sizes on the GPU. Default list behavior is unchanged.

# Checklist:


ci/pytorch.sh
```diff
@@ -98,6 +98,7 @@ run_test_config(){
     run_default_fa 1 triton_kernels/test_cast_mxfp8.py
     run_default_fa 1 triton_kernels/test_cast_mxfp4.py
     run_default_fa 1 triton_kernels/test_grouped_gemm.py
+    run_default_fa 1 triton_kernels/test_blockwise_fp8.py
```
ipanfilo: Nit: keel alphabetical order, put it before triton_kernels/test_cast

transformer_engine/pytorch/triton_kernels/blockwise_fp8_grouped_gemm.py
```diff
@@ -0,0 +1,775 @@
...
+    global _KNOBS_SET
+    if _KNOBS_SET:
+        return
+    _KNOBS_SET = True
+    if hasattr(triton, "knobs") and hasattr(triton.knobs, "amd"):
+        triton.knobs.amd.use_async_copy = True
+        triton.knobs.amd.scalarize_packed_fops = True
+        triton.knobs.amd.use_block_pingpong = True
+    else:
+        os.environ.setdefault("TRITON_HIP_USE_ASYNC_COPY", "1")
```
ipanfilo: Why setdefault not set? With current implementation if branch forces settings while else branch only sets defaults

transformer_engine/pytorch/triton_kernels/blockwise_fp8_grouped_gemm.py
```diff
@@ -0,0 +1,775 @@
...
+        trans_b: If True, b[g] is [N, K] (transposed).
+        out_dtype: Output dtype (default bfloat16).
+
+    Returns:
+        [M_total, N] output in out_dtype.
+    """
+    if is_cdna4():
+        set_triton_knobs_gfx950()
+    else:
+        _set_amd_knobs(enable=True)
```
ipanfilo: From the comments, enable should be calculated form layout. Also, since knobs win/loss is gfx942 specific, it should be rather if is_cdna3() ... else

transformer_engine/pytorch/triton_kernels/blockwise_fp8_grouped_gemm.py
```diff
@@ -0,0 +1,775 @@
...
+    """Set AMD-specific Triton knobs (non-gfx950 fallback).
+    NOTE: use_async_copy and scalarize_packed_fops help NT/NN but regress
+    TN/wgrad ~5-8% on gfx942, so callers gate ``enable`` per layout.
+    """
+    if hasattr(triton, "knobs") and hasattr(triton.knobs, "amd"):
+        triton.knobs.amd.use_async_copy = enable
+        triton.knobs.amd.scalarize_packed_fops = enable
+
+
+NUM_XCDS = 8
```
ipanfilo, line 93: In MPX mode it will be different number. Why is it needed?

transformer_engine/pytorch/triton_kernels/blockwise_quantize.py
```diff
@@ -0,0 +1,328 @@
...
+@triton.jit
+def _floor_to_pow2(scale):
+    scale_bits = scale.to(tl.uint32, bitcast=True) & 0xFF800000
+    return scale_bits.to(tl.float32, bitcast=True)
+
+
+@triton.jit
+def compute_scale_and_quant(x_tile, x_tile_abs, axis, FP8_MAX, ROUND_POW2: tl.constexpr):
+    x_tile_max = tl.max(x_tile_abs, axis=axis, keep_dims=True)
+    x_tile_max = tl.maximum(x_tile_max, 1e-4)
```
ipanfilo, line 42: What is 1e-4?

ci/pytorch.sh
sudhu2k: Fixed

transformer_engine/pytorch/triton_kernels/blockwise_fp8_grouped_gemm.py
sudhu2k: Changed to force setting.

transformer_engine/pytorch/triton_kernels/blockwise_fp8_grouped_gemm.py
sudhu2k: This is used for swizzling of PIDs across dies for better L2 cache utilization.

https://github.com/ROCm/TransformerEngine/blob/237bc22266b09558b96625fb3d4da3d83f4a4317/transformer_engine/pytorch/triton_kernels/gmm/pid_preprocessing.py#L12-L27

Seems like we can't get this info programatically, so it is hardcoded.

https://github.com/ROCm/aiter/blob/7f184691e35627b3a672974687e617d057164836/aiter/ops/triton/utils/device_info.py#L25-L27

transformer_engine/pytorch/triton_kernels/blockwise_fp8_grouped_gemm.py
sudhu2k: Fixed this

transformer_engine/pytorch/triton_kernels/blockwise_quantize.py
sudhu2k: Scale is FP8_MAX / amax, this is to ensure that the minimum value of amax isn't 0 so we don't accidentally divide by 0. The value was chosen by primus turbo's implementation.

transformer_engine/pytorch/triton_kernels/blockwise_fp8_grouped_gemm.py
```diff
@@ -0,0 +1,780 @@
...
+            acc += partial * (a_s * b_s)[:, None]
+
+        # ── Store output ──
+        c = acc.to(C.type.element_ty)
+        rm_s = (pid_m * BLOCK_SIZE_M + tl.arange(0, BLOCK_SIZE_M)) % M_g
+        rn_s = (pid_n * BLOCK_SIZE_N + tl.arange(0, BLOCK_SIZE_N)) % N
+        rn_s = tl.max_contiguous(tl.multiple_of(rn_s, BLOCK_SIZE_N), BLOCK_SIZE_N)
+        c_mask = (rm_s[:, None] < M_g) & (rn_s[None, :] < N)
+        C_ = C + m_start_g * stride_cm + rm_s[:, None] * stride_cm + rn_s[None, :] * stride_cn
+        tl.store(C_, c, c_mask)
```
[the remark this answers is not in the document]
sudhu2k: Added in_features % 128 == 0 and out_features % 128 == 0 gate since the kernel assumes in and out features are a multiple of 128.

transformer_engine/pytorch/triton_kernels/blockwise_fp8_grouped_gemm.py
```diff
@@ -0,0 +1,780 @@
...
+        # ── Store output ──
+        rm_s = pid_m * BLOCK_SIZE_M + tl.arange(0, BLOCK_SIZE_M)
+        rn_s = pid_n * BLOCK_SIZE_N + tl.arange(0, BLOCK_SIZE_N)
+        rn_s = tl.max_contiguous(tl.multiple_of(rn_s % OUT_N, BLOCK_SIZE_N), BLOCK_SIZE_N)
+        c_mask = (rm_s[:, None] < OUT_M) & (rn_s[None, :] < OUT_N)
+        C_ = C + group_idx.to(tl.int64) * stride_cg + rm_s[:, None] * stride_cm + rn_s[None, :] * stride_cn
+        c = acc.to(C.type.element_ty)
+        if ACCUMULATE:
+            c += tl.load(C_, mask=c_mask, other=0)
+        tl.store(C_, c, c_mask)
```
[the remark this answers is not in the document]
sudhu2k: Removed % OUT_N and let the c_mask take care of bounds.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -399,6 +399,298 @@ def _forward_grouped_tensor(
...
+        split_src = m_splits_tensor if m_splits_tensor is not None else m_splits
+        group_lens = split_src.to(device=device, dtype=torch.int64)
+        group_offs = torch.zeros(num_gemms + 1, dtype=torch.int64, device=device)
+        group_offs[1:] = torch.cumsum(group_lens, 0)
+
+        w = _GroupedLinear._expert_weights_as_3d(weights, activation_dtype)
+
+        # Quantize: activation rowwise (1x128 along K), weights 128x128.
+        a_row, a_srow = quantize_fp8_blockwise(a, dt, axis=1, block_size=128)
+        b_fp8, b_scale = quantize_fp8_blockwise_weight(w, dt, block_size=128)
```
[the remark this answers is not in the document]
sudhu2k: Integrated Qparams values to enable or disable power of 2 scales

transformer_engine/pytorch/triton_kernels/blockwise_fp8_grouped_gemm.py
```diff
@@ -0,0 +1,780 @@
...
+        triton.knobs.amd.use_async_copy = enable
+        triton.knobs.amd.scalarize_packed_fops = enable
+
+
+def _apply_amd_compiler_knobs(*, is_tn: bool) -> None:
+    """gfx942: knobs from GEMM layout. Else (gfx950): always-on gfx950 knobs."""
+    if is_cdna3():
+        _set_triton_knobs_gfx942(enable=not is_tn)
+    else:
+        _set_triton_knobs_gfx950()
```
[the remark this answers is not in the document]
sudhu2k: Created a context helper which enables the knobs before kernel is run and disables it once it gets over.

transformer_engine/pytorch/triton_kernels/blockwise_fp8_grouped_gemm.py
```diff
@@ -0,0 +1,780 @@
...
+        as_ptrs_base = A_scales_ptr + (m_start_g + rm.to(tl.int64)) * stride_as_m
+
+        # B_scales pointer: B_scales[g, pn, ki] (2D block scaling)
+        bs_ptr_base = B_scales_ptr + group_idx.to(tl.int64) * stride_bs_g + pid_n * stride_bs_n
+
+        # ── K-loop with block-wise scaling (EVEN_K pattern) ──
+        loop_k = tl.cdiv(K, BLOCK_SIZE_K)
+        if not EVEN_K:
+            loop_k -= 1
+        tl.assume(loop_k > 1)
```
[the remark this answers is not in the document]
sudhu2k: Fixed it.

tests/pytorch/test_grouped_linear.py
```diff
@@ -398,8 +401,17 @@ def test_grouped_linear_accuracy(
...
+        # uses Triton quant + grouped GEMM. Budget two independent FP8 stacks.
+        atol, rtol = 0.25, 0.12
     for o, o_ref in zip(outputs, outputs_ref):
-        torch.testing.assert_close(o, o_ref, rtol=rtol, atol=atol)
+        if use_blockwise_triton:
+            mag = max(float(o.detach().abs().max()), float(o_ref.detach().abs().max()))
+            tensor_atol = max(atol, 0.05 * mag)
+            torch.testing.assert_close(o, o_ref, rtol=rtol, atol=tensor_atol)
+        else:
+            torch.testing.assert_close(o, o_ref, rtol=rtol, atol=atol)
```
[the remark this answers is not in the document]
sudhu2k: Reverted back to original tolerances used by `use_triton`

tests/pytorch/test_grouped_linear.py
```diff
@@ -754,8 +766,6 @@ def test_padding_grouped_linear_accuracy(
 ):
     if fp8_model_params and NVTE_TEST_NVINSPECT_ENABLED:
         pytest.skip("FP8 parameters are not supported in debug mode.")
-    if IS_HIP_EXTENSION and recipe is not None and recipe.float8_block_scaling():
-        pytest.skip("ROCm grouped GEMM does not yet support FP8 block scaling.")
     skip_unsupported_backward_override(
```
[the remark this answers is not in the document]
sudhu2k: the skip was stale. Float8BlockScaling passes on test_padding_grouped_linear_accuracy, test_padding_grouped_linear_accuracy_save_original_input, and test_grouped_linear_accuracy for both use_triton=True/False. So lifting it there is correct.
Added skip for blockwise scaling in test_grouped_linear_accuracy_rocm_backends.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -399,6 +399,298 @@ def _forward_grouped_tensor(
...
+        ``single_grouped_weight`` keeps experts as consecutive slices of one
+        ``GroupedTensor.rowwise_data`` buffer; those slices are a zero-copy view.
+        Discrete ``weight0..weightN`` still need a stack.
+        """
+        packed = _GroupedLinear._packed_3d_view(weights)
+        if packed is not None:
+            if packed.dtype != dtype:
+                packed = packed.to(dtype)
+            return packed if packed.is_contiguous() else packed.contiguous()
+        return torch.stack([wt.to(dtype).contiguous() for wt in weights], 0).contiguous()
```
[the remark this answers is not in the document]
sudhu2k: Excluded fp8 model parameters in _is_blockwise_fp8_grouped_gemm_supported since it's not supported currently. Default path should be good enough.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -1998,6 +2466,24 @@ def _get_weight_tensors(self) -> List[Union[torch.Tensor, QuantizedTensorStorage
...
+                per_expert_main_grad = main_grad.view(
+                    self.num_gemms, self.out_features, self.in_features
+                )
+            for i, w in enumerate(weight_tensors):
+                if w.requires_grad != want_grad:
+                    w.requires_grad_(want_grad)
+                if per_expert_main_grad is not None:
+                    w.main_grad = per_expert_main_grad[i]
+                if hasattr(grouped_weight, "overwrite_main_grad"):
+                    w.overwrite_main_grad = grouped_weight.overwrite_main_grad
```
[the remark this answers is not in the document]
sudhu2k: Added coverage for this issue, this is a bug which is present in ROCm and CUDA for single_grouped_weight. This is fixed in latest TE upstream using different technique. https://github.com/NVIDIA/TransformerEngine/commit/2d80391f77e542c06d0281688ddd17b6e34adec8#diff-e52c6ddc8c0f4d20cb5aa832e92dd3794687bac1f828e367b526f17f524238a5
Will wait for upstream integration, until this this fix is good enough.

tests/pytorch/test_grouped_linear.py
```diff
@@ -385,8 +386,10 @@ def test_grouped_linear_accuracy(
         delay_wgrad_compute,
     )
 
+    use_blockwise_triton = os.getenv("NVTE_USE_BLOCKWISE_GMM_TRITON", "0") == "1"
```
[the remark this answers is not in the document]
sudhu2k: Fixed.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -399,6 +399,404 @@ def _forward_grouped_tensor(
...
+        # ``single_grouped_weight``), no stack.
+        if isinstance(weights[0], Float8BlockwiseQTensor):
+            # fp8_model_params: reuse the stored rowwise data + scales directly (no
+            # re-quantization).
+            qw = _GroupedLinear._grouped_weight_blockwise_qtensor(
+                weights, activation_dtype, dt, pow2_w
+            )
+        else:
+            w = _GroupedLinear._expert_weights_as_3d(weights, activation_dtype)
+            qw = quantize_fp8_blockwise_grouped_weight_qtensor(w, dt, pow2=pow2_w)
```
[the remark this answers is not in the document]
sudhu2k: Added weight workspace caching feature.

tests/pytorch/test_grouped_linear.py
```diff
@@ -484,6 +516,12 @@ def test_grouped_linear_accuracy_rocm_backends(
     ):
         pytest.skip("CK MXFP8 grouped GEMM only supported on gfx1250.")
 
+    if recipe is not None and recipe.float8_block_scaling():
+        # The CUTLASS/HipKittens/CK grouped GEMM backends do not support FP8 block
```
ipanfilo: NV CUTLASS?

transformer_engine/pytorch/triton_kernels/blockwise_fp8_grouped_gemm.py
ipanfilo: Ok, since there is no readme for it, please comment here that it correctly works only for full config SPX mode

transformer_engine/pytorch/triton_kernels/blockwise_quantize.py
ipanfilo: Add comment for this

tests/pytorch/test_grouped_linear.py
sudhu2k: Removed CUTLASS reference.

transformer_engine/pytorch/triton_kernels/blockwise_fp8_grouped_gemm.py
sudhu2k: Done!

transformer_engine/pytorch/triton_kernels/blockwise_quantize.py
sudhu2k: Done!

ipanfilo (approved):

PR #725, opened 2026-08-31
RS + GEMM overlap for BF16, gfx950, w/ HipKittens + UserBuffers

alextmagro:
New fused reduce-scatter + GEMM forward kernel
New bulk reduce-scatter backward overlap kernel
HK overlap kernel tests moved to test_rocm_fused_overlap.py, replacing old userbuffer tests in CI
Extracted common AG and RS code to a shared header

For MI350:
fused RS + TN GEMM is 1.18x faster geomean, 1.24x faster ms weighted
bulk RS + NT GEMM is 1.41x faster geomean, 1.45x faster ms weighted.


transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -349,8 +372,220 @@ bool run_bulk_nn(const KittensFusedAgGemmArgs &args) {
...
+
+bool rs_guards_ok(const KittensRsGemmArgs &args) {
+    const int M       = args.n;
+    const int N_TOTAL = args.m;
+    const int K       = args.k;
+    const int tp_size = args.nranks;
+    const int bands   = (M / tp_size) / 256;
+
+    // Order matters here
+    return (tp_size == 4 || tp_size == 8) && tp_size <= args.peer_count &&
```
aris134: Just flagging this here because it seems implied that tp_size should be 8 https://github.com/ROCm/TransformerEngine/blob/392fd78363f7f3e4abd2dba0c941b0c3c4db8182/tests/pytorch/distributed/test_rocm_fused_overlap.py#L35-L37

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
```diff
@@ -0,0 +1,617 @@
...
+using namespace hk_overlap;
+
+struct TileDesc {
+    int chunk_id;
+    int tile_m;
+    int tile_n;
+};
+
+using namespace kittens;
+using G_group = kittens::group<NUM_WARPS>;
```
aris134: this line appears to be duplicated across fused_ag_gemm_nn.cuh, fused_ag_gemm_tn.cuh, and fused_rs_gemm_tn.cuh. Since NUM_WARPS is already specified in overlap_common.cuh, perhaps factor this line out into the common header as well

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
```diff
@@ -0,0 +1,617 @@
...
+    unsigned int *arrive[RS_MAX_TP];
+    bf16 *recv[RS_MAX_TP];
+    unsigned int *ready[RS_MAX_TP];
+    bf16 *stage[RS_MAX_TP];
+};
+
+
+#define RS_PUBLISH(p) __hip_atomic_fetch_add((p), 1u, __ATOMIC_RELEASE, __HIP_MEMORY_SCOPE_SYSTEM)
+#define RS_SPIN(p)    __hip_atomic_load((p), __ATOMIC_RELAXED, __HIP_MEMORY_SCOPE_SYSTEM)
+#define RS_ACQUIRE(p) ((void)__hip_atomic_load((p), __ATOMIC_ACQUIRE, __HIP_MEMORY_SCOPE_SYSTEM))
```
aris134: I don't think these macros are used anywhere

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
```diff
@@ -0,0 +1,617 @@
...
+                    if (buckets.cnt[bb] == 0) continue;
+                    if (__hip_atomic_load(&bucket_ctr[bb], __ATOMIC_RELAXED,
+                                          __HIP_MEMORY_SCOPE_AGENT) >= buckets.cnt[bb]) continue;
+                    const int idx = atomicAdd(&bucket_ctr[bb], 1);
+                    if (idx < buckets.cnt[bb]) {
+                        found = buckets.off[bb] + idx;
+                        break;
+                    }
+                }
+                s_tile_idx = (found < 0) ? num_tiles : found;
```
aris134: The work stealing code block appears identical across all headers, and could likely be factored out as a device helper in overlap_common.cuh

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
```diff
@@ -0,0 +1,617 @@
...
+                          const bf16 *__restrict__ local_stage, const RsPeers &peers,
+                          unsigned int *__restrict__ done, bf16 *__restrict__ out,
+                          size_t band_elems) {
+    const int w = (int)blockIdx.x;
+    const size_t lines = band_elems / 8;
+    const int peer = (w < TP - 1) ? (w + (w >= my_pe ? 1 : 0)) : -1;
+    typedef int v4i __attribute__((ext_vector_type(4)));
+
+    {
+        const int b0 = 0;
```
aris134: nit: b0 is a frozen 0. Do we need it in this block?

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -349,8 +372,220 @@ bool run_bulk_nn(const KittensFusedAgGemmArgs &args) {
...
+        (tokens == 16384 || tokens == 32768 || tokens == 65536)) {
+        return 4;
+    }
+    if (hidden == 8192 && k_local == 1024 && (tokens == 32768 || tokens == 65536)) {
+        return 12;
+    }
+    if (hidden == 4096 && k_local == 512) {
+        if (tokens == 65536) return 16;
+        if (tokens == 32768) return 12;
+    }
```
wangye805: nit: this hardcoding seemed to be applied onto specific customer requested configs. Okay for now. Is there a systematic way to tune comm_wg? Like with gemm bounded cases, decrease the comm_wg and for rs bounded cases, increase the comm_wg?

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
```diff
@@ -0,0 +1,617 @@
...
+
+template <int TP, bool NT>
+__device__ __forceinline__
+void pull_reduce_all_sent(int my_pe, int ncomm, int bands, int tiles_N,
+                          const bf16 *__restrict__ local_stage, const RsPeers &peers,
+                          unsigned int *__restrict__ done, bf16 *__restrict__ out,
+                          size_t band_elems) {
+    const int w = (int)blockIdx.x;
+    const size_t lines = band_elems / 8;
+    const int peer = (w < TP - 1) ? (w + (w >= my_pe ? 1 : 0)) : -1;
```
aris134: defined here but not used in this kernel

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
```diff
@@ -0,0 +1,617 @@
...
+constexpr int RS_MAX_TP = 8;
+
+#ifndef COMM_WG
+#define COMM_WG 8
+#endif
+
+// Sentinel selectors, declared here so the epilogue clamp below can see them. 
+// Poison and clamp are quiet negative NaNs so neither can collide with a hardware NaN.
+#define RS_SENT_BF16 0xFFAAu
+#define RS_SENT_DW ((unsigned int)RS_SENT_BF16 * 0x00010001u)
```
wangye805, line 37: Do you need a hipmemset write this 0xffaa into output buffer before each gemm call? I didn't see the writing function call

Also, this poison value usage is based on the assumption that user will not use produce -nan. What if user use this 0xffaa for the padding tokens?

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
```diff
@@ -0,0 +1,617 @@
...
+                    dst_ptr[(row + l * 2 + 1) * row_stride + col] =
+                        base_types::convertor<U, T>::convert(src.tiles[i][j].data[idx].y);
+                }
+            }
+        }
+    }
+}
+
+typedef int rs_v4i __attribute__((ext_vector_type(4)));
+typedef const volatile __attribute__((address_space(1))) rs_v4i *rs_gvol4;
```
wangye805, line 113: This is for poison value 0xffaa reading, right? It was written by other ranks and we are trying to read it in local rank. Without atomic writing, can you check if this read lower to cache-bypassing load (sc0/sc1/nt) on gfx950 in asm? 

If not, consumer may keep reading the stale l2 cache data...

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
```diff
@@ -0,0 +1,617 @@
...
+
+            G_group::load(Bs[1][1], B, {0, 0, block_col * 2 + 1, tile + 3}, sw_B, b_srd, b_base, b_lds_11);
+            asm volatile("s_waitcnt vmcnt(4)");
+            __builtin_amdgcn_s_barrier();
+
+            __builtin_amdgcn_s_setprio(1);
+            mma_ABt(c11, a_tile, b_tile_1, c11);
+            __builtin_amdgcn_s_setprio(0);
+            __builtin_amdgcn_s_barrier();
+        }
```
wangye805: nit: is this tn gemm tile same as that in ag+tn gemm? If so, can we consolidate them with a common macro?

wangye805 (changes requested):

aris134 (changes requested):

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
alextmagro: You're right, this isn't the best way to do this. I have replaced with a heuristic similar to AG.

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
alextmagro: You're right, test is older than this so I have added TP=4 to the test.

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
alextmagro: Okay, moved to header and shared across files

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
alextmagro: Good catch, cleaned up now.

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
alextmagro: I have checked, we do lower to a cache bypassing load. This was an intentional change to avoid an atomic write

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
alextmagro: This is an artifact, sorry. I have cleaned this up and some other dead code.

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
alextmagro: Probably not --cleaned up now

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
alextmagro: It is -- unfortunately we need to be very careful with using device helpers as the compiler optimizes things in a different order. Here, it was something like 10,000 line changes of assembly that hurt performance across the board. 

The solution to repeat code in situations like this is an .inc file, which I have now added, but something to keep in mind!

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
alextmagro: Yes, consolidated into an .inc file to keep ISA identical.

transformer_engine/common/gemm/kittens/cdna4/overlap/tn_mainloop.inc
aris134, line 1: copyright headers needed for *.inc files?

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
alextmagro: This is called in comm_gemm.cpp, but we don't need to write the poison into the output buffer, only the staging portion.

This poison is a signaling NaN, so not of the same class as an arithmetic NaN that the hardware would generate. I think if the user tried to use 0xffaa for padding it would corrupt the GEMM output, as presence of a NaN in a mfma instruction would output all NaNs

transformer_engine/common/gemm/kittens/cdna4/overlap/tn_mainloop.inc
alextmagro: Yes, they are needed and I have added. Thanks!

tests/pytorch/distributed/run_gemm_with_overlap.py
```diff
@@ -86,6 +89,12 @@ def _parse_args(argv=None, namespace=None):
         default=False,
         help="Select the fused AG+GEMM backend (gfx950 only).",
     )
+    parser.add_argument(
+        "--dgrad",
+        action="store_true",
+        default=False,
+        help="Shape the GEMM like the dgrad of a row-parallel layer (NN layout)"
+    )
```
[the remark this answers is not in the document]
alextmagro: Fixed

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
```diff
@@ -0,0 +1,368 @@
...
+                    acc = v;
+                } else {
+                    __hip_bfloat16 *a = reinterpret_cast<__hip_bfloat16 *>(&acc);
+                    const __hip_bfloat16 *x = reinterpret_cast<const __hip_bfloat16 *>(&v);
+#pragma unroll
+                    for (int j = 0; j < 8; j++) a[j] = bf16_add(a[j], x[j]);
+                }
+            }
+            if (pend != 0u) {
+                continue;
```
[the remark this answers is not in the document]
alextmagro: Changed to match ag_ready_kernel

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
```diff
@@ -0,0 +1,368 @@
...
+    const int bands   = (M / tp_size) / BLOCK_ROW;
+
+    gl<bf16, 1, 1, -1, -1> A_gl(d_a, nullptr, nullptr, (size_t)M,       (size_t)K);
+    gl<bf16, 1, 1, -1, -1> B_gl(d_b, nullptr, nullptr, (size_t)N_TOTAL, (size_t)K);
+
+    const int ncomm = (tp_size - 1) * cfg.comm_wg;
+
+    int grid = tiles_M * tiles_N + ncomm;
+    if (grid > GRID_CAP) grid = GRID_CAP;
+    if (grid < ncomm + 1) grid = ncomm + 1;
```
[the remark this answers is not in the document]
alextmagro: Good find, producer idxs moved to bottom of range.

aris134 (approved):

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh
wangye805: Maybe you can add a note in readme to warn users about this 0xffaa signaling NaN?

tests/pytorch/distributed/run_gemm_with_overlap.py
```diff
@@ -182,7 +188,7 @@ def _parse_args(argv=None, namespace=None):
         default=None,
         help=(
             "Override the absolute-error tolerance used in the numerical check. "
-            "When unset, defaults to 0.0625 for FP8/MXFP8 and 0.002 otherwise."
+            "When unset, defaults to 0.0625 for FP8/MXFP8, else 0.01 on ROCm / 0.00125 on CUDA."
         ),
```
wangye805, line 192: Regarding the tolerance relaxation, is it because of 1). we use different accumulation dtype vs NV upstream or 2) we added new tests that not in previous NV upstream tests? 

Also, I noticed in your reduce scatter kernel, bf16 accumulator is used: https://github.com/ROCm/TransformerEngine/blob/55957cde877c0c61ae9ea0c2651be76facd5bea0/transformer_engine/common/gemm/kittens/cdna4/overlap/fused_rs_gemm_tn.cuh#L164. If we change the intermediate buffer to fp32, will it improve the tolerance and degrade the performance too much?

wangye805 (changes requested):

tests/pytorch/distributed/run_gemm_with_overlap.py
alextmagro: These relaxations are actually from previous UB tests, but are reasonable here as we are comparing our HK gemms to  hipBLASLt, which is different that what upstream does, which is a cuBLASLt vs cuBLASLt comparison.

It will improve tolerance slightly, accumulation error was only around ~0.5% RMS. The more interesting finding is the fp32 intermediate reduce is actually faster for some reason, VALU count went down and geomean improved by 1% for the fused path, and 2% for bulk. 

I have made this change now.

wangye805 (approved):

PR #734, opened 2026-09-09
Persistent AG + GEMM MXFP8 Enablement

aris134:
# Description

Enables persistent MXFP8 AG + GEMM kernels for TN + NN layouts, and bulk overlap (wgrad AG + dgrad GEMM) with HipKittens and Userbuffers. Also swaps the NN operand slots in the baseline HipKittens MXFP8 GEMM so the transposing LDS read lands on the 2×-replicated warp axis instead of the 4× one — +6.9% geomean over 24 shapes, measured comm-free on MI350, and it lets the standalone and fused-AG NN kernels share one main loop.

Fixes https://github.com/AMD-ROCm-Internal/frameworks-internal/issues/339#issue-5246899585

## Performance (Note: baseline is sequential AG + HipKittens GEMM)

| Metric | Category | MI350X | MI355X |
|---|---|---|---|
| Geometric mean | TN | 1.274x | 1.368x |
| Median | TN | 1.298x | 1.329x |
| Arithmetic mean | TN | 1.287x | 1.377x |
| Min / Max | TN | 1.010x / 1.645x | 1.133x / 1.775x |
| Geometric mean | NN | 1.194x | 1.263x |
| Median | NN | 1.228x | 1.251x |
| Arithmetic mean | NN | 1.206x | 1.273x |
| Min / Max | NN | 0.949x / 1.462x | 1.033x / 1.555x |
| Geometric mean | Bulk AG | 1.339x | 1.407x |
| Median | Bulk AG | 1.366x | 1.421x |
| Arithmetic mean | Bulk AG | 1.344x | 1.414x |
| Min / Max | Bulk AG | 1.137x / 1.542x | 1.179x / 1.667x |

Sample Perfetto traces:

MXFP8 fused AG TN

32768 tokens × 2304 out_local × 16384 hidden

Baseline: RCCL AG + HK MXFP8 GEMM
Screenshot 2026-09-09 150725

Fused AG + MXFP8 GEMM
Screenshot 2026-09-09 150855

 speedup: 1.594x

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


ci/pytorch.sh
```diff
@@ -137,7 +137,8 @@ run_test_config_mgpu(){
     run_default_fa 1 test_gemm_sm_count.py
     run_default_fa 3 test_sanity_import.py
     run_default_fa 3 distributed/test_cast_master_weights_to_fp8.py
-    run_default_fa 3 distributed/test_rocm_fused_overlap.py
+    # NVTE_ROCM_ENABLE_MXFP8=1 opens up the fused AG+GEMM MXFP8 cases; the rest are unaffected
```
alextmagro: Comment not neded here

qa/L1_pytorch_distributed_unittest/test.sh
```diff
@@ -1,3 +1,4 @@
+# Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.
```
alextmagro: These are upstream qa files, should not be edited by us.

tests/pytorch/distributed/run_gemm_with_overlap.py
```diff
@@ -83,6 +83,13 @@ def _parse_args(argv=None, namespace=None):
     parser.add_argument(
         "--atomic", action="store_true", default=False, help="Test overlap with atomic GEMM."
     )
+    parser.add_argument(
```
alextmagro: I am concerned about some of the changes to this file. They don't seem to be hip guarded when we change upstream code, and additionally I am not quite sure why we can't go through the path that was defined for bf16, just with the --fp8 flag.

tests/pytorch/distributed/test_rocm_fused_overlap.py
```diff
@@ -165,6 +181,19 @@ def test_fused_bulk_ag_overlap_bf16(nprocs):
     _assert_numerics_passed(_run_fused_ag(nprocs, bulk=True))
 
 
+@pytest.mark.skipif(not fused_available, reason=reason_for_no_fused)
+@pytest.mark.parametrize("nprocs", FUSED_PROC_COUNTS)
+def test_fused_bulk_ag_overlap_mxfp8(nprocs):
+    """Bulk all-gather riding in an unrelated MXFP8 GEMM's grid.
```
alextmagro: docstrings should be one line. Ideally we would fuse this test into the bf16 test as well.

tests/pytorch/distributed/test_rocm_fused_overlap.py
```diff
@@ -241,6 +270,17 @@ def test_fused_rs_overlap_rejects_non_bf16(quantization, nprocs):
         pytest.skip(reason_for_no_fp8)
     if quantization == "mxfp8" and not mxfp8_available:
         pytest.skip(reason_for_no_mxfp8)
+    if quantization == "mxfp8":
```
alextmagro: This test should only be run for bf16 at the moment, so not quite sure what this addition is doing.

That being said, I think there is a bug at line 271 -- that should be an or condition not an and at the moment.

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -825,29 +825,43 @@ void CommOverlapP2PBase::initialize(const std::vector<size_t> &buffer_shape, DTy
 
   // Create workspace tensor with userbuffer
   NVTE_CHECK(buffer_shape.size() == 2, "Userbuffer shape must be 2-dimensional!");
-  size_t buffer_bytes = get_buffer_size_bytes(buffer_shape[0], buffer_shape[1], buffer_dtype);
-  int buffer_chunk_bytes = buffer_bytes / _tp_size;
+  size_t data_bytes = get_buffer_size_bytes(buffer_shape[0], buffer_shape[1], buffer_dtype);
+
+  // The E8M0 scales are carried inside the Userbuffers allocation unconditionally. Without it
```
alextmagro: This comment isn't relevant for the codebase, better left for the ticket. Also, this is editing upstream's buffer_bytes, which should be left alone. We can add to buffer_bytes if needed for scales

transformer_engine/common/comm_gemm_overlap/rocm_comm_gemm_overlap.cpp
```diff
@@ -251,16 +251,55 @@ static_assert(prev_is_inverse_of_next<7,8>(tp_next_8, tp_prev_8), "tp_prev_8 is
 
 #ifdef USE_HIPKITTENS_GEMM
 // Fused all-gather + GEMM, launched by fused_overlap_ag below.
-static bool hk_fused_ag_gemm(const TensorWrapper &A, bool transa, bool transb, TensorWrapper &D,
+static bool hk_fused_ag_gemm(const TensorWrapper &A, bool transa, const TensorWrapper &B, bool transb, TensorWrapper &D,
```
alextmagro: This doesn't look right, ubuf is the B tensor so we don't need to pass it. The following check is also not needed.

transformer_engine/common/comm_gemm_overlap/rocm_comm_gemm_overlap.cpp
```diff
@@ -251,16 +251,55 @@ static_assert(prev_is_inverse_of_next<7,8>(tp_next_8, tp_prev_8), "tp_prev_8 is
...
+                             size_t scale_chunk_bytes, cudaStream_t stream) {
   // TODO: Add bias support
+  NVTE_CHECK(B.dptr() == ubuf.dptr(),
+             "fused AG+GEMM reached with invalid B tensor!");
+
   NVTE_CHECK(!transb && !accumulate && bias.numel() == 0 && pre_gelu_out.numel() == 0 && B_copy.numel() == 0,
              "fused AG+GEMM reached with an unsupported epilogue");
-  NVTE_CHECK(A.dtype() == DType::kBFloat16 && ubuf.dtype() == DType::kBFloat16 && D.dtype() == DType::kBFloat16,
-             "fused AG+GEMM reached with a non-bf16 operand");
+  bool is_bf16 = A.dtype() == DType::kBFloat16 && ubuf.dtype() == DType::kBFloat16 && D.dtype() == DType::kBFloat16;
```
alextmagro: This can be cleaned up, the following lambda is unneeded, and dtype checks should be done elsewhere -- see functions that fall back to non-overlap when conditions aren't met in our python layers.

transformer_engine/common/comm_gemm_overlap/rocm_comm_gemm_overlap.cpp
```diff
@@ -270,15 +309,25 @@ static bool hk_fused_ag_gemm(const TensorWrapper &A, bool transa, bool transb, T
...
       static_cast<size_t>(GET_SEND_PTR_BY_INDEX(0, comm, reg, 0) - reinterpret_cast<char *>(comm->peer_ptr[0][0])),
       static_cast<size_t>(GET_RECV_PTR_BY_INDEX(1, comm, reg, 0) - GET_RECV_PTR_BY_INDEX(0, comm, reg, 0)),
       signal, static_cast<int>(m), static_cast<int>(n_chunk * tp_size), static_cast<int>(k), transa,
-      tp_id, tp_size, chunk.bytes(), workspace.dptr(), workspace.bytes(), stream};
+      tp_id, tp_size, chunk.bytes(), scale_base_offset, scale_chunk_bytes, workspace.dptr(), workspace.bytes(), stream};
+  if (A_tensor->scaling_mode == NVTE_MXFP8_1D_SCALING) {
+    // e4m3 -> 0, e5m2 -> 1, as in mxfp8_gemm.cpp. These name TE's operands, not the kernel's
+    // slots; comm_gemm.cpp binds them to CBSZ/BLGP for the layout it launches, so this side
+    // stays BLAS-canonical and never re-derives transa.
+    args.a_fp8_code = (A.dtype() == DType::kFloat8E5M2) ? 1 : 0;
```
alextmagro: fp8 codes should live in the kittens folder not TE

transformer_engine/common/comm_gemm_overlap/rocm_comm_gemm_overlap.cpp
```diff
@@ -343,16 +399,35 @@ static bool hk_bulk_ag_gemm(const TensorWrapper &A, bool transa, const TensorWra
...
+  if (bulk_fp8) {
+    // NN: A column-wise, B row-wise, matching the fused path's pairing.
+    scale_A = A_tensor->columnwise_scale_inv.dptr;
+    scale_B = B_tensor->scale_inv.dptr;
+  }
+
   const int rank_round_tp = comm->myrank - tp_id;
   KittensAgGemmArgs args{
-      A.dptr(), B.dptr(), D.dptr(),
+      // NN takes A column-wise, matching the scales chosen above.
```
alextmagro: There seem to be a lot of these short comments -- I think where the code is more or less self explanatory, we can remove these.

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -7,6 +7,9 @@
 #include "../../comm_gemm.h"
 #include "fused_ag_gemm_tn.cuh"
 #include "fused_ag_gemm_nn.cuh"
+#include "fused_ag_mxfp8_gemm_tn.cuh"
```
alextmagro: Can we combine these files with the bf16 files? Otherwise, we should rename the original files to be fp16 specific

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -7,6 +7,9 @@
...
+#include "fused_ag_mxfp8_gemm_nn.cuh"
+#include "../../kittens_kernel_common.cuh"
```
alextmagro: nit: this should go on line 8 after ../../comm_gemm.h

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -59,6 +62,50 @@ uint64_t ag_ready_warn_ticks() {
...
+    const size_t off = scale_base + static_cast<size_t>(peer) * chunk_bytes;
+    const v4i *src = reinterpret_cast<const v4i *>(peers.base[peer] + off);
+    v4i       *dst = reinterpret_cast<v4i *>(ub + off);
+
+    const size_t n4     = chunk_bytes / sizeof(v4i);
+    const size_t stride = static_cast<size_t>(gridDim.x) * blockDim.x;
+    const size_t step   = stride * U;
+    size_t i = static_cast<size_t>(blockIdx.x) * blockDim.x + threadIdx.x;
+
+    if (U > 1) {
```
alextmagro: This should be a constexpr

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -59,6 +62,50 @@ uint64_t ag_ready_warn_ticks() {
...
+            for (int u = 0; u < U; u++) v[u] = src[i + static_cast<size_t>(u) * stride];
+#pragma unroll
+            for (int u = 0; u < U; u++) {
+                if (NT) __builtin_nontemporal_store(v[u], &dst[i + static_cast<size_t>(u) * stride]);
+                else    dst[i + static_cast<size_t>(u) * stride] = v[u];
+            }
+        }
+    }
+    for (; i < n4; i += stride) {
+        if (NT) __builtin_nontemporal_store(src[i], &dst[i]);
```
alextmagro: Same, if (NT) should be constexpr

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -118,11 +165,51 @@ const std::vector<void *> *peer_bases(const void *peer_ub, int count) {
...
+    ln[(tile_id * NG + grp) * 64 + lane] =
+        kittens::pack_scales((const kittens::fp8e8m0 *)tile, grp * STEP);
+}
+
+// Gather+pack the activation scales inside the gatherer blocks instead of up front? Only pays at
+// large K: the gatherer reads one contiguous k-range per scale row, worth 8x the cache lines at
+// K=16384 but only 2x at K=4096, below which the per-tile pack inside the arrival gate costs more
+// than the removed prologue saves.
+int interleave_scales_enabled(int K) {
+    if (const char *e = getenv("HK_INTERLEAVE_SCALES")) return atoi(e) ? 1 : 0;
```
alextmagro: I am hesitant to add this env var if we have a heuristic that can replace it

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -118,11 +165,51 @@ const std::vector<void *> *peer_bases(const void *peer_ub, int count) {
...
+    return K >= 16384 ? 1 : 0;
+}
 
 // Peer-dedicated queues win when the queue is walked at most ~2 times and there are enough M-tiles
 // to fill the buckets. The 1024 is fixed, not derived from the grid cap.
 int auto_xcd_bucket(int num_tiles, int tiles_m) {
-    return (num_tiles <= 1024 && tiles_m >= 5) ? 1 : 0;
+    const int grid_cap = getenv("HK_GRID_CAP") ? atoi(getenv("HK_GRID_CAP")) : 256;
```
alextmagro: Comment is not longer in line with code here -- why do we need this code change?

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -145,6 +232,18 @@ int gath_wg_bulk(int k, int tp_size) {
     return GATH_WG;
 }
 
+// MXFP8 sibling of gath_wg_bulk; the identical thresholds are a measured result, not inheritance
+// -- the gathered tensor is MXFP8 too, so both sides of the balance halve together. The gate is k
+// (the dgrad reduction depth) alone: the gather moves m*N_TOTAL bytes against 2*m*N_TOTAL*k flops,
+// so flops/byte depends on neither m nor N_TOTAL. Kept separate so retuning bf16 cannot move this.
+int gath_wg_mxfp8_bulk(int k, int tp_size) {
```
alextmagro: If this is the same we can reuse gath_wg_bulk

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -241,10 +340,269 @@ bool run_tn(const KittensAgGemmArgs &args) {
     return hipGetLastError() == hipSuccess;
 }
 
+// Packs raw scales into the lane-native layout the mxfp8 GEMM reads: one fp8e8m0_4 per (group,
+// lane). A: STEP=64,NG=4 (256 words/tile); B: STEP=32,NG=8 (512 words, hi/lo tile pair).
+// COLWISE=false: raw uint8 [dim, K/32] row-major; COLWISE=true: [K/32, dim] col-major.
+template <bool COLWISE, int STEP, int NG>
+__global__ void pack_scales_kernel(const uint8_t *__restrict__ scales, uint32_t *__restrict__ ln,
```
alextmagro: Why do we need a new pack_scales_kernel? Why not pass just the local chunk to the previously available one?

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -241,10 +340,269 @@ bool run_tn(const KittensAgGemmArgs &args) {
...
+                        hipStream_t stream) {
+    const int tiles_per_col = dim / 256;
+    pack_scales_kernel<COLWISE, STEP, NG><<<k_iters * tiles_per_col, NG * 64, 0, stream>>>(
+        scales, ln, dim, scale_K, k_iters, tiles_per_col);
+}
+
+// Per-layout bindings for the shared run_mxfp8 host sequence.
+struct Mxfp8Tn {
+    using TileDesc = hk_mxfp8_ag_tn::TileDesc;
+    static constexpr int  PLAN_TAG        = 2;
```
alextmagro: nit: = signs don't get aligned for different var types

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -241,10 +340,269 @@ bool run_tn(const KittensAgGemmArgs &args) {
...
+    static std::vector<TileDesc> work_queue(int M, int N, int K, int tp, int pe) {
+        return hk_mxfp8_ag_tn::build_work_queue(M, N, K, tp, pe);
+    }
+    static hk_mxfp8_ag_tn::persistent_fn_t launch_fn(int M, int N, int K, int a_code, int b_code) {
+        return hk_mxfp8_ag_tn::get_persistent_fn(M, N, K, a_code, b_code);
+    }
+};
+
+// NN consumes TE's A operand (this kernel's B) column-wise.
+struct Mxfp8Nn {
```
alextmagro: I wonder if these structs could be combined and just initiated differently if really needed?

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -241,10 +340,269 @@ bool run_tn(const KittensAgGemmArgs &args) {
...
+            static_cast<void *const *>(const_cast<void *>(args.arrive_peers)), args.arrive_offset,
+            static_cast<const char *>(args.arrive_local), args.arrive_stride, args.arrive_value,
+            args.peer_first, args.peer_count, tp_size, ag_ready_warn_ticks());
+    }
+
+    // Activation scales live in the userbuffer, so they can only be packed after ag_ready_kernel
+    // has established that the peers finished writing their chunks. Row-wise in both layouts.
+    // scale_chunk_bytes != 0 means the peers' raw scales are reachable, which interleaving needs;
+    // it is always true for a fused MXFP8 AG buffer, so the guard is defensive.
+    const int interleave =
```
alextmagro: nit: no need for a line break here. If claude generated, Claude.md in TE repo specifies C++ styling info

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -382,6 +740,131 @@ bool run_bulk_nn(const KittensAgGemmArgs &args) {
...
+    const int tp_size = args.nranks;
+    const int m_local = M / tp_size;
+    const int tiles_m = M / BLOCK_ROW;
+    const int tiles_n = N_TOTAL / BLOCK_COL;
+    const int k_iters = K / BLOCK_K;
+    const int scale_K = K / 32;
+
+    // 16B alignment is what lets gather_scales run without a scalar tail. Here the region is the
+    // gathered tensor's shard, which is not a GEMM operand on this path.
+    if (args.scale_chunk_bytes &&
```
alextmagro: Are we not always 16B aligned with mxfp8?

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -581,16 +1064,18 @@ bool rs_guards_ok(const KittensRsGemmArgs &args) {
 
 // Shape and pointer requirements shared by ag entry points
 bool ag_guards_ok(const KittensAgGemmArgs &args) {
-    const int M       = args.n;
-    const int N_TOTAL = args.m;
+    // args.n is the gathered token axis, so that is what must split evenly across the TP group.
+    // The predicates are value-identical under the TN/NN naming swap, since BLOCK_ROW == BLOCK_COL.
+    const int M       = args.m;
```
alextmagro, line 1008: Let's be careful here since bf16 uses this code as well. If we aren't swapping for bf16 this would break that code

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -581,16 +1064,18 @@ bool rs_guards_ok(const KittensRsGemmArgs &args) {
...
+    const int N_TOTAL = args.n;
     const int K       = args.k;
     const int tp_size = args.nranks;
 
     // Order matters here
     return tp_size >= 1 && tp_size <= 8 && tp_size <= args.peer_count &&
            args.rank >= 0 && args.rank < tp_size &&
-           M % tp_size == 0 && M % 256 == 0 && N_TOTAL % 256 == 0 &&
-           K % 128 == 0 && K >= 256 && (M / tp_size) % 256 == 0 &&
+           N_TOTAL % tp_size == 0 && N_TOTAL % 256 == 0 && M % 256 == 0 &&
```
alextmagro, line 1016: M seems to have been dropped from this check. Is that correct?

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_ag_gemm_nn.cuh
```diff
@@ -220,14 +175,14 @@ void persistent_ag_bf16_gemm_body(
     };
     auto read_B = [&](auto &bt, int p, int j) {
         const uint32_t addr = (uint32_t)(uintptr_t)(&Bs[p][j][0]) + (uint32_t)(warp_n * 4096 + kittens::laneid() * 8);
-#define RDB(I, J, O0, O1)                                                             \
+        #define RDB(I, J, O0, O1)                                                     \
```
alextmagro: This alignment is incorrect, macros should be left aligned

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_ag_mxfp8_gemm_nn.cuh
```diff
@@ -0,0 +1,405 @@
...
+    const size_t doff       = (size_t)peer * chunk_bytes + (size_t)tn * tile_bytes;
+
+    size_t sub_bytes = (((tile_bytes + gath_wg - 1) / gath_wg) + 15) & ~size_t(15);
+    size_t o         = (size_t)sub * sub_bytes;
+    size_t l         = (o >= tile_bytes) ? 0 : ((o + sub_bytes <= tile_bytes) ? sub_bytes : tile_bytes - o);
+
+    if (l) gather_copy_wg<U, NT>(gather_dst + doff + o, (const char *)peers.base[peer] + doff + o, l);
+
+    __syncthreads();
+    if (NT) {
```
alextmagro: Missing constexprs here -- additionally, is this not repeat code from bf16?

transformer_engine/common/gemm/kittens/cdna4/mxfp8_gemm.cpp
```diff
@@ -646,173 +562,14 @@ __global__ __launch_bounds__(NUM_THREADS, 2) void mxfp8_gemm_nn_kernel(const gl_
...
-    }
+#include "mxfp8_nn_mainloop.inc"
 
     OutGL C_local(C);
     if constexpr (GROUPED) {
         C_local.raw_ptr = (typename OutGL::dtype *)c_expert_ptrs[expert_id];
     }
 
-    gemm_epilogue<EPILOGUE, ACCUMULATE, RT_C, RT_C_T>(cA, cB, cC, cD, C_local, AuxGL, bias, bias_dtype,
+    gemm_epilogue_nn<EPILOGUE, ACCUMULATE, RT_C>(cA, cB, cC, cD, C_local, AuxGL, bias, bias_dtype,
```
alextmagro: I am a bit confused by this change. It seems we had a templated gemm_epilogue and have untemplated it here.

transformer_engine/common/gemm/kittens/cdna4/mxfp8_gemm.cpp
```diff
@@ -1140,8 +897,17 @@ static void launch_gemm_typed(
                 nullptr, nullptr, nullptr,
                 N, K, tiles_M, tiles_N);
         } else if constexpr (!TRANSA && !TRANSB) {
+            // FLIPPED slots: the A slot takes B's tensor (the activation, K-contiguous [N,K]) and
```
alextmagro, line 822: We discussed this offline, but it might be worthwhile to experiment if flipping is always beneficial (certainly possible) or if it is something to be determined based on the size of A and B. If a simple heuristic says that transposing the smaller matrix is always best, then we should configure the dispatch to support that. Also, let's look into if this swap works for bf16 as well?

transformer_engine/common/gemm/kittens/comm_gemm.h
```diff
@@ -25,10 +27,17 @@ struct KittensAgGemmArgs {
...
+    size_t scale_base_offset;
+    size_t scale_chunk_bytes;
     void *workspace;
     size_t workspace_size;
     hipStream_t stream;
     void *gather_dst;     // Bulk all-gather only
+    // MFMA operand format codes, as mxfp8_gemm.cpp's fp8_code(): 0 = e4m3, 1 = e5m2. They select
+    // CBSZ/BLGP independently per operand, which HYBRID recipes need (E4M3 fwd, E5M2 bwd).
+    // A is the gathered activation operand (ub), B is the weight (args.A). Ignored by bf16 entries.
+    int a_fp8_code;
```
alextmagro: format codes should be available via the enum and templated out I would think.

transformer_engine/pytorch/module/base.py
```diff
@@ -915,16 +921,23 @@ def fill_userbuffers_buffer_for_all_gather(
...
-            [process_group_size * local_scale_inv_size[0]] + local_scale_inv_size[1:],
-            dtype=local_scale_inv.dtype,
-            device=local_scale_inv.device,
-        )
-        torch.distributed.all_gather_into_tensor(
-            global_scale_inv,
-            local_scale_inv,
-            group=process_group,
-        )
+        global_scale_inv_shape = [
```
alextmagro: This should be hip guarded

alextmagro (changes requested):

ci/pytorch.sh
aris134: Removed in 9723b4d

qa/L1_pytorch_distributed_unittest/test.sh
aris134: Thanks, corrected in 9723b4d

tests/pytorch/distributed/run_gemm_with_overlap.py
aris134: Thanks, went through and fixed in 9723b4d

tests/pytorch/distributed/test_rocm_fused_overlap.py
aris134: Resolved in 9723b4d

tests/pytorch/distributed/test_rocm_fused_overlap.py
aris134: Reverted the change in 9723b4d

transformer_engine/common/gemm/kittens/cdna4/overlap/fused_ag_mxfp8_gemm_nn.cuh
aris134: Refactored duplicated code in 4e20d56

transformer_engine/common/gemm/kittens/cdna4/mxfp8_gemm.cpp
aris134: reinstated the templated gemm_epilogue in 4e20d56

transformer_engine/common/gemm/kittens/comm_gemm.h
aris134: Thanks, I used the format codes already available in 4e20d56

transformer_engine/pytorch/module/base.py
aris134: Done in 705da39

transformer_engine/common/gemm/kittens/cdna4/mxfp8_gemm.cpp
aris134: The flip is bound to one matrix — the one whose contraction axis is its slowest dim. In dgrad W is `[N,K]` contracting along N, so W is the strided operand; dY is `[M,N]` with N contiguous. The transposed operand is therefore always W, independent of the relative sizes of W and dY.

What we do choose is which slot W occupies, and so how many warps redundantly load it: 4 on the M side, 2 on the N side. Moving W to the N side halves its transposed reads independent of shape.

The 24-shape MI350 sweep supports the point on shape independence. Flat against hidden/tokens across a 32× range: 18 shapes with W smaller give 1.067x, the 6 with W larger or equal give 1.076x. Not flat against the reduction dimension: 1.06–1.10x for n_iters ≥ 10, and mixed at n_iters = 6 — 1.067x, 1.024x, 0.995x, 0.904x, geomean 0.996. The shallow-end cost is the scalar store in the un-transposed epilogue, a fixed per-output-tile cost that longer reductions amortize. But I'm not sure this alone is enough to justify a swap heuristic, I would want to run more experiments on the shallower end.

On BF16, the swap is already baked into the implementation; enabling it for MXFP8 NN is what gave MXFP8 the boost. Happy to run some more experiments though.

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
aris134: Equivalent to before. Here we have the same predicate on args.m/args.n/args.k, just with the local names made consistent with the struct fields (M ↔ m, N_TOTAL ↔ n). bf16 is unaffected since callers pass raw args to the guard and alias locally themselves (kittens_fused_ag_gemm_bf16_cdna4 binds M = args.n; run_mxfp8 does the layout swap), so the guard's internal names aren't visible to them

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
aris134: Not dropped, just re-labeled. See response to last comment.

transformer_engine/common/gemm/kittens/cdna4/mxfp8_gemm.cpp
```diff
@@ -86,62 +92,116 @@ __device__ __forceinline__ void gemm_epilogue(
...
+            RT_C eA, eB, eC, eD;
+            kittens::load(eA, C, out_coord_A); kittens::load(eB, C, out_coord_B);
+            kittens::load(eC, C, out_coord_C); kittens::load(eD, C, out_coord_D);
+            kittens::add(cA, cA, eA); kittens::add(cB, cB, eB);
+            kittens::add(cC, cC, eC); kittens::add(cD, cD, eD);
+        }
+
+        kittens::store(C, cA, out_coord_A); kittens::store(C, cB, out_coord_B);
+        kittens::store(C, cC, out_coord_C); kittens::store(C, cD, out_coord_D);
+    }
```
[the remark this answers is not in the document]
aris134: 1. Yes, kittens::store has a dedicated col_layout overload
2. Two paths are being conflated. mxfp8_gemm.cpp is the standalone GEMM reached from rocm_gemm.cu, which passes is_transa/is_transb and the bias/pre-GELU pointers as independent arguments; layout and epilogue are orthogonal template parameters there. The fused overlap NN path is the one with no epilogue support — rocm_comm_gemm_overlap.cpp currently rejects bias and pre-GELU

transformer_engine/common/gemm/kittens/cdna4/mxfp8_gemm.cpp
```diff
@@ -1487,10 +1283,17 @@ static bool grouped_mxfp8_gemm(
                 (const void *const *)d_b_ptrs, (const void *const *)d_c_ptrs, d_sb_tile_offsets,
                 total_N, K, tiles_M, total_n_tiles);
         } else if (!transa && !transb) {
+            // FLIPPED slots: the A slot takes B's tensor (the activation) and the B slot A's (the
+            // weight); the expert pointer arrays swap with them. The activation view is sized by
+            // the widest expert so the buffer descriptor covers every per-expert base.
+            gl_fp8_rt gl_A_nn((kittens::fp8e4m3 *)B_array[0], nullptr, nullptr,
+                              (size_t)max_N, (size_t)K);
+            gl_fp8_rt gl_B_nn((kittens::fp8e4m3 *)A_array[0], nullptr, nullptr,
+                              (size_t)K, (size_t)M);
```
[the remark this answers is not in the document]
aris134: This follows what the file already does for grouped GEMMs — the other grouped path sizes its view by the largest expert and swaps the pointer per expert inside the kernel. The non-grouped path isn't quite the right comparison, since it doesn't re-base per expert and so can just use the real size.  For context on how the line got here: it used to be sized by the first expert, which is too small whenever a later expert is wider — those reads would have been clamped. The max is never smaller than that, so it's strictly better than what it replaced, even if it is looser than a true per-expert bound.

ci/pytorch.sh
```diff
@@ -137,7 +137,7 @@ run_test_config_mgpu(){
     run_default_fa 1 test_gemm_sm_count.py
     run_default_fa 3 test_sanity_import.py
     run_default_fa 3 distributed/test_cast_master_weights_to_fp8.py
-    run_default_fa 3 distributed/test_rocm_fused_overlap.py
+    NVTE_ROCM_ENABLE_MXFP8=1 run_default_fa 3 distributed/test_rocm_fused_overlap.py
```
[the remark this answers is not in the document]
aris134: Couldn't reproduce. Also follows pattern already present in the script.

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -825,29 +825,43 @@ void CommOverlapP2PBase::initialize(const std::vector<size_t> &buffer_shape, DTy
...
       num_ubuf_view_chunks = _num_ubuf_chunks;
     }
   }
 
+  size_t alloc_bytes = data_bytes;
+  if (mxfp8_ag_scales) {
+    _scale_chunk_bytes = buffer_chunk_bytes / 32;
+    _scale_base_offset = data_bytes;
+    alloc_bytes += _scale_chunk_bytes * _tp_size;
+  }
```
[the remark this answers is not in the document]
aris134: Done in 88844fbd5

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -241,10 +340,269 @@ bool run_tn(const KittensAgGemmArgs &args) {
...
+                    (const uint8_t *)args.scale_B + (size_t)args.rank * args.scale_chunk_bytes,
+                    packed_sa, args.rank * tiles_per_chunk, tiles_per_chunk, tiles_gath, scale_K,
+                    k_iters);
+        } else {
+            pack_local_scales_kernel<32, 8>
+                <<<k_iters * tiles_per_chunk, 8 * 64, 0, args.stream>>>(
+                    (const uint8_t *)args.scale_B + (size_t)args.rank * args.scale_chunk_bytes,
+                    packed_sb, args.rank * tiles_per_chunk, tiles_per_chunk, tiles_gath, scale_K,
+                    k_iters);
+        }
```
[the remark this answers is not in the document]
aris134: Addressed in 88844fbd5

transformer_engine/common/comm_gemm_overlap/rocm_comm_gemm_overlap.cpp
```diff
@@ -343,16 +399,35 @@ static bool hk_bulk_ag_gemm(const TensorWrapper &A, bool transa, const TensorWra
...
 
+  auto A_tensor = convertNVTETensorCheck(A.data());
+  auto B_tensor = convertNVTETensorCheck(B.data());
+  const void *scale_A = nullptr;
+  const void *scale_B = nullptr;
+  if (bulk_fp8) {
+    // NN: A column-wise, B row-wise, matching the fused path's pairing.
+    scale_A = A_tensor->columnwise_scale_inv.dptr;
+    scale_B = B_tensor->scale_inv.dptr;
+  }
```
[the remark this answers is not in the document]
aris134: Done in  11b588f

transformer_engine/common/gemm/kittens/cdna4/overlap/overlap_common.cuh
```diff
@@ -2,71 +2,314 @@
...
+constexpr int HALF_ROW     = BLOCK_ROW / 2;
+constexpr int HALF_COL     = BLOCK_COL / 2;
+constexpr int REG_M        = BLOCK_ROW / WARPS_ROW / 2;
+constexpr int REG_N        = BLOCK_COL / WARPS_COL / 2;
+constexpr int K_STEP       = 64;
+constexpr int NUM_THREADS  = NUM_WARPS * kittens::WARP_THREADS;
 
-constexpr int NUM_THREADS = NUM_WARPS * kittens::WARP_THREADS;
+// bf16 kernels only; the MXFP8 pair reads HK_GRID_CAP at launch instead.
+constexpr int GRID_CAP     = 256;
```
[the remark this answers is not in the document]
aris134: Resolved in 88844fbd5

transformer_engine/common/gemm/kittens/cdna4/overlap/overlap_common.cuh
```diff
@@ -2,71 +2,314 @@
...
+    if (l) gather_copy_wg<U, NT>(gather_dst + doff + o, (const char *)peers.base[peer] + doff + o, l);
+
+    __syncthreads();
+
+    // Ordering comes from ag_ready_kernel, same as the data copy above. Trip counts differ across
+    // blocks when gath_wg does not divide k_iters -- fine, __syncthreads() is per block.
+    {
+        const char *peer_scales = (const char *)peers.base[peer] + scale_base
+                                + (size_t)peer * scale_chunk_bytes
+                                + (size_t)tn * BLOCK_ROW * (size_t)scale_K;
```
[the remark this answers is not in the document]
aris134: Addressed in 88844fbd5

tests/pytorch/distributed/run_gemm_with_overlap.py
```diff
@@ -838,6 +847,30 @@ def _gemm():
 
     # Compare against standard GEMM
     numerics_failed = False
+
+    # The kernel populates the peers' rows of the Userbuffers scale region, which base.py hands
```
alextmagro: Can probably remove the second sentence of this comment.

tests/pytorch/distributed/test_rocm_fused_overlap.py
```diff
@@ -53,27 +57,48 @@ def _fused_launch_cmd(nprocs: int):
...
 ELIGIBLE_OUT_FEATURES_PER_RANK = 1536
 INELIGIBLE_OUT_FEATURES_PER_RANK = 1568
 UNALIGNED_SEQ_LENGTH = 1152
 
+# K = num_heads * head_dim, and the kernel only gathers scales in-flight at K >= 16384
+# (interleave_scales_enabled). ffn_hidden_size is pinned rather than left at 4x to keep the
+# weight at 33 MB per rank.
+INTERLEAVE_NUM_HEADS: int = 128
+INTERLEAVE_HEAD_DIM: int = 128
+INTERLEAVE_FFN_HIDDEN: int = 16384
```
alextmagro: Rather than hard coding this I wonder if we should have a MI350 specific var for K16384?

tests/pytorch/distributed/test_rocm_fused_overlap.py
```diff
@@ -124,24 +149,61 @@ def _assert_numerics_passed(result):
...
+            num_heads=INTERLEAVE_NUM_HEADS,
+            head_dim=INTERLEAVE_HEAD_DIM,
+            extra_args=[f"--ffn-hidden-size={INTERLEAVE_FFN_HIDDEN}"],
+        )
+    )
+
+
+@pytest.mark.skipif(not fused_available, reason=reason_for_no_fused)
+@pytest.mark.parametrize("nprocs", FUSED_PROC_COUNTS)
+def test_fused_ag_overlap_mxfp8_interleaved_ub_scales(nprocs):
```
alextmagro: How is this test different from the previous one?

tests/pytorch/distributed/test_rocm_fused_overlap.py
```diff
@@ -192,8 +257,13 @@ def test_fused_layer_bulk_wgrad_bf16(nprocs):
     assert "qkv_wgrad" in eligible, eligible
 
 
-def _run_fused_rs_layer(nprocs, extra_args, seq_length=SEQ_LENGTH):
-    """Run the layer harness on a row-parallel Linear, which is the fused reduce-scatter's path."""
+def _run_fused_row_parallel_layer(nprocs, extra_args, seq_length=SEQ_LENGTH):
+    """Run the layer harness on a row-parallel Linear.
```
alextmagro: Comment should be shortened to 1 line if possible

tests/pytorch/distributed/test_rocm_fused_overlap.py
```diff
@@ -232,6 +302,37 @@ def test_fused_rs_overlap_bf16(nprocs):
     assert "proj_fprop" not in disabled, disabled
 
 
+@pytest.mark.skipif(not fused_available, reason=reason_for_no_fused)
+@pytest.mark.parametrize("nprocs", FUSED_PROC_COUNTS)
+def test_fused_ag_overlap_row_parallel_mxfp8(nprocs):
+    """Row-parallel MXFP8 backward, where the grad output has to be gathered twice.
```
alextmagro: Same as above, 1-2 line comment if possible

transformer_engine/common/comm_gemm_overlap/rocm_comm_gemm_overlap.cpp
```diff
@@ -250,17 +250,66 @@ static_assert(prev_is_inverse_of_next<2,4>(tp_next_4, tp_prev_4), "tp_prev_4 is
...
+  uint64_t signal;
+};
+
 // Fused all-gather + GEMM, launched by fused_overlap_ag below.
 static bool hk_fused_ag_gemm(const TensorWrapper &A, bool transa, bool transb, TensorWrapper &D,
                              const TensorWrapper &bias, const TensorWrapper &pre_gelu_out,
                              const TensorWrapper &B_copy, TensorWrapper &workspace, bool accumulate,
-                             const TensorWrapper &ubuf, const TensorWrapper &chunk, communicator *comm,
-                             int reg, int tp_id, int tp_size, uint64_t signal, cudaStream_t stream) {
+                             const TensorWrapper &ubuf, DType ubuf_elt_dtype,
```
alextmagro: It looks like clang-format or something was used to limit chars to 100. I believe we have a slightly higher limit, 120 included in our claude.md

transformer_engine/common/comm_gemm_overlap/rocm_comm_gemm_overlap.cpp
```diff
@@ -250,17 +250,66 @@ static_assert(prev_is_inverse_of_next<2,4>(tp_next_4, tp_prev_4), "tp_prev_4 is
...
+               "GEMM would have no column-wise scale-inverses for the gathered tensor");
+    NVTE_CHECK(aux_ag->chunk_bytes == chunk.bytes() &&
+                   aux_ag->scale_chunk_bytes == scale_chunk_bytes,
+               "fused AG+GEMM: the auxiliary region shards ", aux_ag->chunk_bytes, " data and ",
+               aux_ag->scale_chunk_bytes, " scale bytes, but the gathered tensor shards ",
+               chunk.bytes(), " and ", scale_chunk_bytes,
+               "; both hold the same tensor and must agree.");
+  }
+
+  const void* scale_A = nullptr;
```
alextmagro: super nit: Convention in rocm files is to attach * to var name, C style

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -118,6 +167,44 @@ const std::vector<void *> *peer_bases(const void *peer_ub, int count) {
     return &g_peers.emplace(peer_ub, std::move(v)).first->second;
 }
 
+// Packs ONLY the local chunk's scale rows, for the interleaved path: every peer chunk is packed
+// in-kernel by the gatherer block that fetches it, but the local chunk has no gatherer.
+template <int STEP, int NG>
+__global__ void pack_local_scales_kernel(const uint8_t *__restrict__ scales,
```
alextmagro: Is this templatable into the current pack scales kernel used by non-overlapped GEMMs? If so that might be good, otherwise this can stay as is.

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
```diff
@@ -118,6 +167,44 @@ const std::vector<void *> *peer_bases(const void *peer_ub, int count) {
...
+    const size_t tile_id = (size_t)ki * tiles_per_col + (cblk_off + lblk);
+    ln[(tile_id * NG + grp) * 64 + lane] =
+        kittens::pack_scales((const kittens::fp8e8m0 *)tile, grp * STEP);
+}
+
+// Gather+pack the activation scales inside the gatherer blocks instead of up front? Only pays at
+// large K: the gatherer reads one contiguous k-range per scale row, worth 8x the cache lines at
+// K=16384 but only 2x at K=4096, below which the per-tile pack inside the arrival gate costs more
+// than the removed prologue saves.
+int interleave_scales_enabled(int K) {
```
alextmagro: This can probably just be inlined, with bool statements combined

alextmagro (changes requested):

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
aris134: Done in c787ece

transformer_engine/common/gemm/kittens/cdna4/overlap/comm_gemm.cpp
aris134: Thanks, this is a good suggestion. Done in c787ece

transformer_engine/common/comm_gemm_overlap/rocm_comm_gemm_overlap.cpp
aris134: Adapted to the convention in c787ece

transformer_engine/common/comm_gemm_overlap/rocm_comm_gemm_overlap.cpp
aris134: Redirected agent to use the higher limit from claude.md, and do a re-check across affected files in c787ece

tests/pytorch/distributed/test_rocm_fused_overlap.py
aris134: Done in c787ece

tests/pytorch/distributed/test_rocm_fused_overlap.py
aris134: Done in c787ece

tests/pytorch/distributed/test_rocm_fused_overlap.py
aris134: It was originally meant to isolate testing of all gathered scales with interleave, but it was similar enough to the previous one that I combined them in c787ece

tests/pytorch/distributed/test_rocm_fused_overlap.py
aris134: Already MI350 specific since under kittens/cdna4, but I did add some logic to ensure that if the K-threshold is changed beneath the test that it'll catch it here

tests/pytorch/distributed/run_gemm_with_overlap.py
aris134: Right, done in c787ece

PR #737, opened 2026-09-18
Fix blockwise numerical issues and add ROCm testing

alextmagro:
Fixes a numerical issue with blockwise scaling where a subnormal or zeroed scale would cause running results to hit infinity.

Both fixed through an epsilon addition to floor amax during quantization to avoid infinities, similar to triton and primus block quantizations.

ROCm specific GEMM tests introduced, testing both pow2 and non-pow2 scales, replacing upstream bitwise exact python tests only testing pow2.


transformer_engine/common/recipe/__init__.py
```diff
@@ -446,10 +446,12 @@ class Float8BlockScaling(Recipe):
 
     use_f32_scales: bool = os.getenv("NVTE_FP8_BLOCK_SCALING_FP32_SCALES", "0") == "1"
 
+    amax_epsilon = 1e-4
```
wenchenvincent: Modifying amax_epsilon here impacts other fp8 recipes. We want to limit the workaround only to blockwise fp8.

transformer_engine/common/common.h
```diff
@@ -655,7 +655,8 @@ struct GroupedTensor {
 
 struct QuantizationConfig {
   bool force_pow_2_scales = false;
-  float amax_epsilon = 0.0f;
+  // Floor the block/tensor amax so a degenerate input cannot send the scale to inf
+  float amax_epsilon = 1e-4f;
```
wenchenvincent: Modifying amax_epsilon here impacts other fp8 recipes. We want to limit the workaround only to blockwise fp8.

transformer_engine/common/gemm/kittens/CMakeLists.txt
```diff
@@ -79,14 +79,10 @@ else()
...
                      SOURCES cdna3/blockwise_fp8_gemm.cpp)
 
-    # -ffast-math folds fpext(fptrunc(x)) back to x, so any deliberate narrow-then-widen
-    # in blockwise_fp8_gemm.cpp has to be made opaque to the optimizer -- see
-    # round_to_out_dtype() in cdna4/blockwise_fp8_gemm_helper.cuh.
     kittens_add_arch(TAG cdna4 GFX gfx950
                      SOURCES cdna4/blockwise_fp8_gemm.cpp cdna4/mxfp8_gemm.cpp
                              cdna4/overlap/comm_gemm.cpp
-                     FLAGS -gline-tables-only
-                     SOURCE_FLAGS cdna4/blockwise_fp8_gemm.cpp "-ffast-math")
```
wenchenvincent, line 89: Did you check if there is performance impact by removing "-ffast-math"?

tests/cpp/operator/test_cublaslt_gemm.cu
```diff
@@ -1277,4 +1304,139 @@ TEST(InputGenTest, FillUniform_DoesNotGetOverwrittenByFromCpu) {
...
+
+struct BlockwiseParams {
+  size_t m, k, n;
+  bool a_2d_scaled;
+  bool fp32_scales;
+  float amax_epsilon;
+  bool degenerate;
+};
+
+// k-block 1 quantizes to a subnormal scale_inv and k-block 2 to ~22, so prev/curr spans >1e39.
```
wangye805, line 1328: Emm, scales coming from both A and B, so if both sa_prev = 1e4, sb_prev=1e4, even if sa_current and sb_current clamped to 1e-30, we can still easily get s_prev/s_curr > 1e39?

transformer_engine/common/gemm/kittens/cdna4/blockwise_fp8_gemm_helper.cuh
```diff
@@ -19,13 +19,17 @@ struct ColScale { float v[WIDTH]; };
 template <int HEIGHT>
 struct RowRatio { float v[HEIGHT][4]; };
 
+constexpr float kMinScaleInv = 1e-30f;
```
wangye805: Emm, the amax epsilon 1e-4 is doing the same side clamping (make sure scale_inv >= 1e-4/448), why do we still need a kMinScaleInv? To protect when upstream quantizer does not have this 1e-4 epsilon?

transformer_engine/common/gemm/kittens/cdna4/blockwise_fp8_gemm_helper.cuh
```diff
@@ -364,6 +362,8 @@ __device__ __forceinline__ void compute_a_ratios_and_promote(
         int e = g * (BLOCK_M / 4) + lt;
         float2 p = reinterpret_cast<const float2 *>(smem_sa_prev)[e];
         float2 c = reinterpret_cast<const float2 *>(smem_sa_curr)[e];
+        c.x = floor_scale_inv(c.x);
+        c.y = floor_scale_inv(c.y);
         float2 r = {p.x / c.x, p.y / c.y};
```
wangye805, line 367: Since the running scale is multiplying s_previous/s_current, should we also clamp the max scale inv?

transformer_engine/common/recipe/__init__.py
wangye805: This sounds like a big upstream change, need an ENV to guard. Since later we will use it in RL post training, this amax_epsilon may also need to align with rollout side

wangye805 (dismissed):

wangye805:
Do we have a scale then accumulate path?

transformer_engine/common/recipe/__init__.py
wenchenvincent: And 1e-4 is way too large.

wenchenvincent: Could you add a test case that fails with amax_epsilon=1e-4 and passes with a derived floor? The new degenerate
  cases all pass with 1e-4, so nothing currently pins the value.

  A tensor whose entire block-amax range sits below the epsilon does it. Real example — the layers.2.mlp.linear_fc1
  wgrad that motivated this fix: block amax min 2.76e-40, median 1.99e-9, max 4.10e-5. With 1e-4 all 393216 blocks get
  floored, 64% of elements quantize to zero, and relerr doubles (2.24e-2 → 4.20e-2). With fp8_max * FLT_MIN only 65
  blocks (0.017%) are floored.

wenchenvincent (changes requested):

transformer_engine/common/common.h
alextmagro: I misread what this struct does, I have fixed it to be blockwise specific.

transformer_engine/common/recipe/__init__.py
alextmagro: I had set this to match our triton blockwise quantization threshold, which quotes being derived from Primus.

I have removed this in favor of the derived threshold of 1e-35

alextmagro: Reverted, this was removed as part of the change to avoid hitting all fp8 paths.

transformer_engine/common/gemm/kittens/cdna4/blockwise_fp8_gemm_helper.cuh
alextmagro: Overflow is coming from the denominator being too small, a large numerator will stay in bounds I believe.

transformer_engine/common/gemm/kittens/cdna4/blockwise_fp8_gemm_helper.cuh
alextmagro: 1e-4 was dropped, this is the correct implementation. I have raised the value to 1e-13

tests/cpp/operator/test_cublaslt_gemm.cu
alextmagro: Good catch. The clamp has been increased to 1e-13 to account for the 2 scales.

I have also changed the test to have both scaleA and scaleb be degenerate

transformer_engine/common/gemm/kittens/CMakeLists.txt
alextmagro: Yes, it appears that removing --fast-math actually makes the fp32 scale kernel around 4% faster on cdna4, pow2 is unaffected.

transformer_engine/common/recipe/__init__.py
alextmagro: I have dropped the 1e-4 quantizer floor in favor of the running scale minimum that is kernel side, but I have added tests to ensure that the floor is not too high, which would hit the same issue.

alextmagro:
> Do we have a scale then accumulate path?

I believe this was phased out for performance / register reasons.

transformer_engine/common/gemm/kittens/cdna4/blockwise_fp8_gemm_helper.cuh
wangye805: I thought the overflow coming from a normal accumulator value times a huge s_previous/s_current. Denominator too small is definitely one factor of s_previous/s_current too large. 

We may need the exact reproducer from Wen's RL run to see whether only the large numerator is in bounds

wangye805 (commented):
Since we don't have the robust path (scale then accumulate with another accumulator vgpr/agpr), I think we can give this PR fix to Wen to test if this stabalize the training. Then we can decide whether to merge or pursue the robust path

In my opinion, the stability of running scale requires s_prev/s_current not overflow, which implies constraints on both max of s_prev and min of s_current. Your clamping min of scale_inv solves half the problem. 

In contrast, the conventional path (scale then accumulate) is also robust against max of s_scale is too large

alextmagro:
> Since we don't have the robust path (scale then accumulate with another accumulator vgpr/agpr), I think we can give this PR fix to Wen to test if this stabalize the training. Then we can decide whether to merge or pursue the robust path
> 
> In my opinion, the stability of running scale requires s_prev/s_current not overflow, which implies constraints on both max of s_prev and min of s_current. Your clamping min of scale_inv solves half the problem.
> 
> In contrast, the conventional path (scale then accumulate) is also robust against max of s_scale is too large

Correction, we do have a robust path, micro_tk_partial_k does scale then accumulate. It is gated on K < 2*BLOCK_K || K % BLOCK_K != 0, and due to a large amount of expected VGPR spills would require significant refactoring for performance.

You are right that a very large numerator could cause overflow too, but the impact of both sides is not the same. A tiny denominator comes from a block with negligible values, so clamping them has minimal impact on convergence or model quality. On the other hand, clamping large values on a dominant block would misscale the most important values in the tensor, which we want to avoid.

Regarding magnitude, the overflow value of the numerator depends on the denominator. For lower clamp 1e-13, this is K*amaxA*amaxB <= 3.4e12. For our largest sample K, 53248, that would amaxA*amaxB <= 6.4e7.  My understanding is the largest amax values we see in training are in the e3, low e4 range, so we have 3 magnitudes of scale overhead generally.

wangye805:
> > Since we don't have the robust path (scale then accumulate with another accumulator vgpr/agpr), I think we can give this PR fix to Wen to test if this stabalize the training. Then we can decide whether to merge or pursue the robust path
> > In my opinion, the stability of running scale requires s_prev/s_current not overflow, which implies constraints on both max of s_prev and min of s_current. Your clamping min of scale_inv solves half the problem.
> > In contrast, the conventional path (scale then accumulate) is also robust against max of s_scale is too large
> 
> Correction, we do have a robust path, micro_tk_partial_k does scale then accumulate. It is gated on K < 2*BLOCK_K || K % BLOCK_K != 0, and due to a large amount of expected VGPR spills would require significant refactoring for performance.
> 
> You are right that a very large numerator could cause overflow too, but the impact of both sides is not the same. A tiny denominator comes from a block with negligible values, so clamping them has minimal impact on convergence or model quality. On the other hand, clamping large values on a dominant block would misscale the most important values in the tensor, which we want to avoid.
> 
> Regarding magnitude, the overflow value of the numerator depends on the denominator. For lower clamp 1e-13, this is K_amaxA_amaxB <= 3.4e12. For our largest sample K, 53248, that would amaxA*amaxB <= 6.4e7. My understanding is the largest amax values we see in training are in the e3, low e4 range, so we have 3 magnitudes of scale overhead generally.

Oh, good to know that we have a robust path. Do you recall the performance? If it's not pretty bad, maybe we can give it an env first to let user choose which path to go (running scale or scale then accumulate)

Sorry if I make you mis-understand before. I didn't propose to clamp the max of scale_inv, it does not makes sense in math. I   generally don't know any good solution to this large scale_inv in running scaling flow besides your experience/observation "My understanding is the largest amax values we see in training are in the e3, low e4 range, so we have 3 magnitudes of scale overhead generally."

wenchenvincent (approved):

